PR #383, opened 2013-10-15
cc2420 hardware aes fix

renehummen:
ContikiMAC can preempt the standalone encryption process and access the SPI at the same time as the hardware encryption process. As a result, faulty AES results are returned.

In order to lock interactions with the radio, I moved the standalone encryption to cc2420.c. Alternatively, the lock could be made public. However, moving all cc2420-related functionality to one file was deemed the cleaner way.

Note, that I also added simple hardware aes test example for the 'sky' target. This example runs hardware aes several times in a row and evaluates the output of a hardware AES operation for a fixed key and input. If the outputs deviate, this may be an indication that other protothreads interfere with this protothread.


g-oikonomou:
@kkrentz I understand that #557 obsoletes this, right?

kkrentz:
yes

PR #397, opened 2013-10-29
I2C driver for cc2538 (re-pull requested)

areaarea:
Sorry, I was learning how to use git and messed this up completely (as well as my fork) the first time.  I *now* know how to use branches more effectively...

Anyway, here is the i2c driver that I had submitted in the past. It presents the 'normal' interfaces to read and write.  Not based on interrupts or the like, just straight function call. Should be pretty straight forward.


g-oikonomou:
#1005

Issue #384, opened 2013-10-15
Discussion on ADC interface abstraction


areaarea:
I have looked through the source for several targets and found no ADC support in any of them.  Lighting off the ADC portions of these targets is a key foundational block for making contiki a practical _sensor_ network solution.  I have been rolling around how to do this under the constraints of contiki. 

My proposal is that there be an init() function that does all the ADC management for the general interface and starts a tic timer in the ADC process but does not set up individual pins for use.  Instead a register() function is called per pin (for flexibility) that has a format of 

adc_service_register(PORT, PIN, samples_per_second,void * callback_function)

the samples per second field is limited in granularity and speed by CLOCK_SECONDS (in my case for the cc2538, 128 sps).  A simple data structure inside the process can be stored in an array (list?) set up that has the following fields:

struct adc_pin_instance {
uint8_t current_tics_to_fire
uint8_t reset_value
uint8_t port
uint8_t pin
void* callback_function
}

every time the tic timer fires in the process, all the current_tics_to_fire are decremented and the ones hitting 0 are serviced and then the tics to fire is reset.  

With an unregister() as well, this makes for a dynamically deployable ADC system inside contiki.  It has shortcomings; it is low speed, may have jitter, etc, but we need _something_ for this interface.  

Open to suggestions/comments/ridicule...

Adam

g-oikonomou:
Closing this due to inactivity

PR #398, opened 2013-10-30
ADC driver for cc2538 (re-pull request)

areaarea:
Here is an ADC driver for the cc2538.  It is a arch specific implementation right now, but I think the method is definitely scalable and abstractable.  General flow of the driver:

adc_init() does very little but set flags and the interrupt.  The real setup of each pin happens in the adc_register_pin() function.  It currently supports single ended reads on A0:7 or differential reads on A01:67.  The register function sets up an element that allows the pins to be tracked and serviced independently from each other.  A callback function is registered for the independent pin such that any number of modules can register the pin and the ADC driver needs no information on data handling itself. The upside is that you can fire all the pins on different time scales and from different modules.  The down side is there is a fundamental limit of 128sps and rounding make *exact* *jitter-free* measurements difficult (although power of 2 sps rates should have limited jitter.) 

 There is a process that runs for the driver that fires at 128Hz and counts down all the 'timers' of all  the active pins. When any (or all) pins need servicing, there is a helper function that handles the traversing of the structure and the isr() dispatches the data to the related cb function.

Was related to the discussion I tried to start in #384 but maybe an actual implementation will aid that discussion.


alignan:
Where are we on this long dweller?

simonduq:
No activity here? Will in one month if this stays idle.

simonduq:
closing after timeout

PR #468, opened 2013-11-26
update to fix webserver / atmega128rfa1 / avr raven

celsimor:
Several commits to reable web server example and default functionality for both avr-raven and avr-atmega128rfa1, for which i use a zigduino to test:

1) removed default target from webserver main makefile, and simplified it by removing unnecesary code. Default still set in makefile.webserver
2) added default make target to enable "make webserver6.eu;make webserver6.u" scenario to easy flash a sensor node with eeprom and hex
3) removed unnecessary files from 128rfa1. Zigduino does not have an lcd coprocessor nor any other 128rfa1 platfrom i am aware of.
4) Set the UIP_CONF buffer for bot avr and zigduino to 1280 bytes, to enable tcp ip communication. Needed to use the webserver, otherwise webserver fails opaquely
5) Updated the random number generator on 128rfa1 to now actually give random numbers with 8 bits of randomness. Consequetive calls will also keep random.
6) webserver now writes binary website file to example dir, so platform dir remains clean
7) Enabled serious out debugging on 128rfa1 / zigduino. Current code was legacy of raven debugging, which did not work and just gave garbage out.


simonduq:
Is there any interest? Will close in one month if not.

simonduq:
closing after timeout

PR #470, opened 2013-11-26
Refactoring Observers

darki2002:
Well I just wanted to implement the BaseRSSI support.
Therefore I needed an Observer from the AbstractRadioMedium so I was looking around.
First i remarked, that there is no setChanged() combined with notify() function, but you need to have both. So I added the ScnObservable class for the combined setChangedAndNotify() function.
Then I remarked that the so called RadioMediumObservable in AbstractRadioMedium in fact is a RadioTransmissionObservable so i refactored it. After that I added a RadioMediumObservable with right function and wrote some small documentation.
Then I found the settingsObservable in the MRM which does the same like the RadioMediumObservable, so I superseded it. It was even not used, but I added the functionality now.


simonduq:
Is there any interest? Will close in one month if not.

PR #495, opened 2013-12-07
core/sys/rtimer.h, rtimer.c: support multiple timers; allow cancellation

wpwrak:
This rewrite of rtimers introduces the following changes:
- more than one rtimer can be used concurrently
  
  Some operations are O(n) for the number of active rtimers, so
  their number should still not be excessive.
- when setting an rtimer, the handler is guaranteed to be called
  exactly once. Before, it could also happen that it was never
  called or that it was only called with a very large delay,
  depending on the underlying architecture support.
- rtimers can be cancelled with rtimer_cancel. Cancellation means
  that the handler function will not be invoked after rtimer_cancel
  returns. The rtimer structure may still be used internally and
  must not be overwritten or deallocated.

### DIFF
--- a/core/sys/rtimer.c
+++ b/core/sys/rtimer.c
@@ -113,7 +113,7 @@ set_locked(struct rtimer *rtimer, rtimer_clock_t time,
   }
   rtimer->time = time;
   rtimer->func = func;
-  rtimer->ptr = func;
+  rtimer->ptr = ptr;
   rtimer->cancel = 0;
 
   for(anchor = &next_rtimer; *anchor; anchor = &(*anchor)->next) {
--- a/core/sys/rtimer.c
+++ b/core/sys/rtimer.c
@@ -140,7 +140,9 @@ next_timer_locked(void)
   while(next_rtimer && !RTIMER_CLOCK_LT(now, next_rtimer->time)) {
     t = next_rtimer;   
     next_rtimer = t->next;
-    if(!t->cancel) {
+    if(t->cancel) {
+      need_sched = 1;
+    } else {
       t->func(t, t->ptr);
       need_sched |= t != next_rtimer;
     }
--- a/core/sys/rtimer.c
+++ b/core/sys/rtimer.c
@@ -45,6 +45,7 @@
  */
 
 #include <stdbool.h>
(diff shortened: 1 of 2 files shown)


simonduq:
This is such an important feature and sensitive piece of code that I'll avoid the "merge/close in one month" [1]. The features (multiple, cancellable rtimers) have been missing for too long in Contiki, so I'm a proponent of this PR. I have tested it with our TSCH implementation (heavily reliant on rtimer) and this seems to work. Also, we're shortly after a release, so IMO it's a good time to get important features (albeit perhaps slightly dangerous) in master.

One thing I'd like is to have the code a bit more documented, i.e. a block of comment before each function. Non-trivial things are going on and need explaining. For instance, what does the locking system exactly do, or what the purpose of the `more` field is and why it has a hardcoded size of `2`.

Also, a rebase is needed.

* Does anyone (maintainer or not) have experience with this patch (even in single rtimer context) or concerns about it?
* Does any maintainer have a strong objection to getting this merged?

[1] see [new PR merging policy](https://github.com/contiki-os/contiki/wiki/Code-Contributions#pull-request-merging-policy)

bkozak-scanimetrics:
Since I'm interested in having this feature too, I've had a look at this and I think I basically understand everything that's going on.

My only major concern with this is the complexity of the code. The author has really done quite the dance to get all of the locking working without having to block interrupts (he says as much above) and I'm concerned that it may end up being a little slow. I haven't been able to find any real problems with the logic though.

I could have a try at making those block comments if you like.

atiselsts:
I've been using this feature in heavy simulations tests: many Cooja scenarios with up to 100 nodes running up to 24 hours simulation time. Never seen any problems.

+1 for the randomized test approach. Haven't looked in the tests in depth, but the fact they exist gives additional evidence that rare corner cases do not cause problems. (But it's not a proof, of course...)

simonduq:
removing the timed out tag as the discussion is still open here

g-oikonomou:
Setting the time out flag here.

g-oikonomou:
#1290

simonduq:
closing after timeout

Issue #602, opened 2014-03-21
field client_process of struct simple_udp_connection is not initialized


xmxueping:
The field client_process is not initialized when function simple_udp_register() called,
but used by thead function simple_udp_process().

add below line in function simple_udp_register() ?
c->client_process = PROCESS_CURRENT();

Issue #601, opened 2014-03-21
different size of uip_conns[] between tcpip.c and uip6.c/uip.c


xmxueping:
There is a bug about UIP_CONNS in tcpip.c & uip6.c/uip.c.
Because the difference of include files,the value UIP_CONNS in tcpip.c is greater than UIP_CONNS in uip6.c/uip.c,
,this can cause overflow error when access uip_conns in tcpip.c.

//uipopt.h
//UIP_CONNS:
# ifndef UIP_CONF_MAX_CONNECTIONS
# define UIP_CONNS       10
# else /\* UIP_CONF_MAX_CONNECTIONS */
# define UIP_CONNS (UIP_CONF_MAX_CONNECTIONS)
# endif /\* UIP_CONF_MAX_CONNECTIONS */

//contiki-default-conf.h
//UIP_CONF_MAX_CONNECTIONS
/\* UIP_CONF_MAX_CONNECTIONS specifies the maximum number of
   simultaneous TCP connections. */
# ifndef UIP_CONF_MAX_CONNECTIONS
# define UIP_CONF_MAX_CONNECTIONS 8
# endif /\* UIP_CONF_MAX_CONNECTIONS */

//tcpip.c includes the file contiki-default-conf.h
tcpip.c <- contiki-net.h <- contiki.h <- contiki-conf.h+contiki-default-conf.h  <- project-conf.h

//uip6.c does not include the file contiki-default-conf.h
uip6.c <- uip.h <- uipopt.h <- contiki-conf.h <- project-conf.h

when we does not define PROJECT_CONF_H or does not define UIP_CONF_MAX_CONNECTIONS in project-conf.h file,
the value of UIP_CONNS in file tcpip.c is 10,but the value of UIP_CONNS in file uip6.c is 8.

To prevent this bug, we can change uip6.c/uip.c or uipopt.h
1)uip6.c/uip.c
add include file:
# include "contiki-net.h"

2)uipopt.h
change include file:
# include "contiki-conf.h"

to 
# include "contiki.h"

PR #604, opened 2014-03-22
patch for issue #601 & #602

xmxueping:
1)The field client_process is not initialized when function simple_udp_register() called,
but used by thead function simple_udp_process().

add below line in function simple_udp_register() 
c->client_process = PROCESS_CURRENT();

2)There is a bug about UIP_CONNS in tcpip.c & uip6.c/uip.c.
Because the difference of include files,the value UIP_CONNS in tcpip.c is greater than UIP_CONNS in uip6.c/uip.c,
,this can cause overflow error when access uip_conns in tcpip.c.

when we does not define PROJECT_CONF_H or does not define UIP_CONF_MAX_CONNECTIONS in project-conf.h file,
the value of UIP_CONNS in file tcpip.c is 10,but the value of UIP_CONNS in file uip6.c is 8.

### DIFF
--- a/core/net/ip/simple-udp.c
+++ b/core/net/ip/simple-udp.c
@@ -176,6 +176,7 @@ simple_udp_register(struct simple_udp_connection *c,
     uip_ipaddr_copy(&c->remote_addr, remote_addr);
   }
   c->receive_callback = receive_callback;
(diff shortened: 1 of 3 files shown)


g-oikonomou:
Closing due to long inactivity

PR #618, opened 2014-04-02
Changed AES-Interface and added AES-Hardware-Usage for platform econotag.

SmallLars:
Hope you like it and will merge it :)


simonduq:
OK, now Contiki has a generic API for both AES and AES-CCM*. This PR could get in if you first port your implementation to the new interface and rebase. Otherwise, will close in a month.

SmallLars:
In my view this API isn't suitable for AES hardware and AES in general.
Ill try to modify it this week to do a proposal for amend.

simonduq:
What is unsuitable?
Just to clarify; I'm talking about `struct aes_128_driver` and `struct ccm_star_driver`.

SmallLars:
Yes, we are talking about same driver.
I committed a proposal where I consider all the things I wrote down in my comment on 11 Jun 2014.
The test in examples/aes-test works for "native" and "econotag". In my view this is the best solution to support hardware and software in a same way.

Even if you like, DON'T MERGE at this time.
I still need to change CCM-Driver to work with new interface.

I would be happy about your comments or suggestions for improvement.

simonduq:
@kkrentz, you probably also want to take a look

SmallLars:
CCM* is using new AES driver now. All tests passing.
I also added a full CCM implementation for RFC 3610.

I still need to check the difference between both.
Maybe CCM* is a subset of CCM so we can us one implementation for both.
@kkrentz can help me there?

simonduq:
Looking only at `aes-128.h`, the code style is a bit off with capitalized function names, and the types are confusing: is `getData` for instance, why is `dst` as `uint8_t *` and `src` as `void *`. I also don't like all the `void *` pointers at the beginning of the new driver.

I general, is is also unclear why we would need things such as `ctr_result` etc. in a generic AES driver. This should be up to the CTR mode of operation, implemented separately (based on aes-128 or not depending on what the H/W offers).

simonduq:
My feeling from this PR is that you are trying to match the aes-128 interface to the specific needs of your platform. If this is not the case, please clearly explain what the limitations of the current interface are etc

SmallLars:
Its important for me to have an interface that support all the AES functions hardware provides.
With MC1322x there is a chip which provides CTR and CBC mode. I used the void pointers at the beginning to save register address for hardware and static memory address for software. Its like an enumeration for the different kinds of data. I used void pointers, because only the specific implementation is able to know what kind of pointer it is.

For Hardware not supporting CTR and CBC its very easy to implement this in software, with
a direct use of the hardware core.

I think an example for this is in dev/cc2420. It works in a similar way. There is also an initialisation and hardware register you need to fill with data. After calling a function you can read the result out of a register.

simonduq:
Regarding the `void *` pointers: shouldn't these rather be an enum, with possible values KEY, IV etc?

Exploiting the specifics of every H/W is wanted indeed, but I'm afraid every coprocessor comes with its own interface. Am still in favor of a simple and generic aes-128 driver, and if we need CBC, a new driver for CBC, which, on your platform, has an implementation that exploits all H/W features.

SmallLars:
I thought about an enum too, yes. I want to avoid an additional mapping from an enum value to a memory address and I am not sure how to use an enum where every driver can set its own values.
Ill try to figure out how that would work.

I already fixed get_data and set_data.

When I am thinking about a CTR and CBC driver, I haven't any idea how to realize.
When we want to do, we need a generic block-cipher driver and not a specific aes-128 driver?
So we are already on a wrong way. I am afraid we are blowing up the implementation for
an extreme generic way nobody need.

simonduq:
I see your point on the block-cipher driver.

As for the enum I'd simply do something similar to the radio interface, with setters and getters for a generic set of values. Implementations are free to support any subset.

SmallLars:
When there is an implementation which don't want to support any of the 6 void pointers, its possible to set the specific pointers to NULL and check in set_data and get_data. But I don't see any reason to do that. I also don't like setters and getters for constant values.

But I think you found an important point, when we talk about a subset implementation and look to encryption/decryption. It's different for AES and implementations should be able to support encryption only as a subset, because decryption isn't  needed for CCM.

I found PR #1164. That's why I added and additional parameter for encryption and decryption.
With the new return value of AES execution its possible to signal the calling function that decryption isn't supported.

simonduq:
removing the timed out tag as this is active again

kkrentz:
I just had a quick look at this so I may be on the wrong track: I was wondering whether you could also leverage the hardware-accelerated CTR and CBC of the econtag platform by implementing a special CCM* driver. That would keep things simple I suppose.

SmallLars:
The CCM mode of econotag, and I think of other hardware too, is CBC and CTR at the same time.
Its not the full function of CCM (RFC 3610) or CCM*. Thats why I use the AES interface in CCM, CCM* and CMAC implementation. So on a platform like econotag hardware-acceleration is used.

I am still not sure about the difference between CCM* and CCM.
In my view CCM* only is a "how to use CCM right". I added the 3 testvectors from IEEE 802.15.4-2006 to the CCM_128 test and it works fine. So I would prefer to remove the CCM* interface/implementation.

simonduq:
How would you exploit H/W able to perform CCM* directly then?

kkrentz:
I still don't see why the mode of operation should be part of the AES-128 driver. The difference between CCM and CCM* is that CCM* supersedes possible values for M (length of MICs) and L (length of the Length field). In CCM*, M may be either 0,4,8, or 16 and L is fixed to 2.

SmallLars:
Ok, so CCM* is a "subset of CCM" or "how to use CCM". When we add a return value to the configure function of CCM driver every implementation can realize a subset of the values for M and L. So its possible to create an implementation with CCM* hardware to. But i think its also ok to have both drivers to avoid confusion. Nevertheless we should think about a change of CCM* driver: There are 2 main functions. One for MAC/MIC and one for encryption. For me CCM(*) is design to calculate it in a parallel way, what is not possible at the moment. Do some of the steps twice costs time and energy.

To the operation mode question: I dont see any reason to support a basic AES encryption because its like insecure ECB mode. Nobody should use AES in a way like this. Its also useless without a decryption.

When we realize CTR and CBC mode outside of the driver: You want to remove the AES driver and create a block ciper driver? On this point ill ignore this things for econotag and implement a CCM driver with direct AES hardware usage, so all the stuff below is useless. At the moment it would be very easy for every platform to realize an own aes driver and use CCM(*) we have in the library.
Forcing platforms to create an own CCM(*) driver would cause code duplication. Its only wished and needed for platforms with hardware CCM(*) support.

I hope i was able to explain you my way of thinking.
Feel also free to tell me, when you think there is space to improve my behavior in a discussion.

kkrentz:
Thanks for your clarification. In #1146, I already combined MIC+CTR. I think it is a good idea to encapsulate the AES-128 implementation because many 802.15.4 transceivers have a hardware-accelerated AES-128 implementation. You can implement any modes of operation on top and do not have to care about the actual AES-128 implementation. However, there seems to be no use case for a generic block cipher interface at the moment. Platforms that want to make use of their hardware-accelerated CCM* implementation can implement their own CCM* driver already, as you proposed. We may also implement a CCM* driver that internally calls a CCM driver.

simonduq:
I fully agree with @kkrentz.
@SmallLars my suggestion for this PR is to provide CCM* for Econotag using the existing interface(s). As an ccm-star driver, and/or aes-128 (which can be called by the S/W implementation of ccm-star under `core/lib`).

simonduq:
(if you implement a ccm-star driver, I'd suggest using the future version from https://github.com/contiki-os/contiki/pull/1146)

SmallLars:
As you suggested I separated AES and its modes. But as I told you before, the old AES driver interface isn't suitable for hardware usage. That's why I changed it without adding the modes.

If you can accept I will update the ccm-star driver as in #1146 and implement it with using ccm-128, so that we finally can merge this PR.

When you cant accept I need to say: Sorry, I am not able to find a solution for implementing hardware usage and we need to close this PR.

simonduq:
I'm also not in favor of the new ccm-128 in core. It is too redundant with ccm-star, and used only by econotag anyway. I think the PR should narrow its focus down: no modification to core nor examples, only added implementation of the (new) ccm-star driver for Econotag. If you want to base this on your ccm-128 implementation, fine, but for now, I would also move this Econotag.

SmallLars:
There is no redundancy when I finish the ccm-star. Its only a simple function call to ccm-128.
Econotag has no hardware support for ccm or ccm-star. Its only aes-hardware for that I need a usable interface. So econotag will use ccm-128 and ccm-star implementation from lib.

simonduq:
Let me clarify once again that I'm not supporting this approach. aes-adv is very specific to the needs of your particular platform, and ccm-128 is heavily reliant on it. The option I would support is a ccm-star driver in your platform, that directly exploits the particular functions provided by your hardware (rather than use aes-adv and ccm-128). I don't think this PR should touch `core` at all.

SmallLars:
You are wrong when you say that "this PR is specific to the needs of my particular platform".
At the moment its a great stack fitting for the use of all other hardware.
When there is hardware with basic aes support implement aes-128.
When there is hardware with aes ctr, cbc and ccm support implement aes-adv.
When there is hardware with ccm(-star) support implement ccm-128 or ccm-star.
Its very easy because you only need to implement the lowest hardware level and get all the other algorithms as a present from lib. And aes_adv ist exactly what you want when you dont want a block_cipher driver and also don't want ctr and cbc included in aes driver. Its the glue between aes and ccm because you need aes-ctr and aes-cbc there. Only because of this ccm can be a small and compact implementation.

I already told you that the code for aes in dev/cc2420 works in a similar way. This would be a great example for implementing aes-128 in an easy way.

I would be happy when you can show me a platform where the different driver interfaces will not fit for. All the time when we discuss I try to give examples and implemented parts of you particular great suggestions. But I cant find any example or fact against my approach.

Even the new ccm-star implementation with aead function isn't a good way to implement ccm because the calculation of both parts (mic and data) isn't done in a parallel way. I updated the ccm-star implementation now with use of ccm-128. You can see there how easy encryption is in one step.

cabo:
I'm sorry, I'm late to this discussion, slowly acquainting myself to the issues.

I just had my first WTF moment:  aes_128_padded_encrypt -- what purpose does this serve?
(It's not in use in the master Contiki.)

alignan:
Closing after timeout

simonduq:
closing after timeout

PR #642, opened 2014-04-16
Support CSMA off

lebrush:
Clean packet buffer clearing when disabling the medium access. This is useful for sensors which during long periods sleeping or inactive and have no radio capabilities. 

For battery powered node, we need to keep the radio usage inside some boundaries. This caused the CSMA rtx queue to fill and loose control of the nodes. This solves the problem by clearing the queue whenever the MAC is turned off. In any case it's the correct thing to do since on the next turn on, the data is already too old.

### DIFF
--- a/core/net/mac/csma.c
+++ b/core/net/mac/csma.c
@@ -116,6 +116,7 @@ LIST(neighbor_list);
 
 static void packet_sent(void *ptr, int status, int num_transmissions);
 static void transmit_packet_list(void *ptr);
+static volatile uint8_t csma_on = 0;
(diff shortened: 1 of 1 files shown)


simonduq:
Interesting feature. No activity for a while though.
If this is still of interest, please rebase and let's see if/where Travis fails. If this is only sky code size problems we should change these particular tests (e.g. compile for the more roomy z1 or wismote).

If there is no activity within a month I'll have to close.

simonduq:
not rebased, closing

PR #643, opened 2014-04-16
RDNSS support

lebrush:
Implements RDNSS (RFC6106), this is sending NS information inside RA messages.


adamdunkels:
There is now an update version of this pull request in https://github.com/contiki-os/contiki/pull/999 so I'm closing this one.

PR #675, opened 2014-05-14
exit when RPL-Border router is started, so that tunslip6 can be invoked

mcr:
This lets the programming section exit so that a script can bring up tunslip6 immediately.


laurentderu:
I believe this PR is too specific, instead mc1322x-load.pl should be modified to exit when programming is done (and maybe return a proper result).

simonduq:
No activity after @laurentderu's comment, will close this PR in a month if nothing happens

malvira:
closing this. mc1322x-load.pl and .c both support the -e flag which will exit when done; if you don't want to drop to a terminal you should use the -e flag.

PR #676, opened 2014-05-14
some additions to tunslip6 to make it a bit easier to diagnose problems with it

mcr:
An additional important thing is to be able to reliably see console messages from contiki which may be interspersed into the code IP packets.

### DIFF
--- a/tools/tunslip6.c
+++ b/tools/tunslip6.c
@@ -224,7 +224,6 @@ serial_to_tun(FILE *inslip, int outfd)
           inet_pton(AF_INET6, ipaddr, &addr);
           if(timestamp) stamptime();
           fprintf(stderr,"*** Address:%s => %02x%02x:%02x%02x:%02x%02x:%02x%02x\n",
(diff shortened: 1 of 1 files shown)


laurentderu:
I see nothing wrong at first sight. @mcr are you still interested about this PR, if so could you rebase it ?

mcr:
I can rebase by Monday.

simonduq:
Nice, I'll merge in a month if nobody speaks against this PR

fvdnabee:
Just a quick question, why is the default behaviour to enable IPA? It seems to me that a more sensible default behaviour would be to disable IPA (and enabling it via a flag), as in the default rpl-border-router example the IPA message is not processed. With the default rpl-border-router, enabling IPA leads to "Got request message of type I" being echoed over and over again (see slip-bridge.c https://github.com/contiki-os/contiki/blob/master/examples/ipv6/rpl-border-router/slip-bridge.c#L74 ).

Unless I'm missing something, I would suggest changing the behaviour of the I flag to enable IPA rather than disabling it.

simonduq:
Yes, that's included in https://github.com/contiki-os/contiki/pull/1304

PR #690, opened 2014-05-30
Remove trailing whitespaces

sieben:
Simply removing trailing whitespaces.

How can we check code style systematically when making a new pull requests?


sieben:
@g-oikonomou I went with the solution presented here: https://help.github.com/articles/dealing-with-line-endings/ it seems to be the easiest way to deal with both line ending and trailing whitespaces.
As always you can use ?w=1 to the diff URL to only see the .gitattributes file.

PR #695, opened 2014-06-01
IPv6 prefix bug

sdefauw:
I found two bugs about the prefix diffusion in IPv6 Neighbor Discovery protocol.

It is about:
* the PIO is never sent in RA message
* problem about the return function when we add prefix in table (host version)

Sébastien

### DIFF
--- a/core/net/ipv6/uip-nd6.c
+++ b/core/net/ipv6/uip-nd6.c
@@ -121,6 +121,8 @@ static uint8_t *nd6_opt_llao;   /**  Pointer to llao option in uip_buf */
 #if !UIP_CONF_ROUTER            // TBD see if we move it to ra_input
 static uip_nd6_opt_prefix_info *nd6_opt_prefix_info; /**  Pointer to prefix information option in uip_buf */
 static uip_ipaddr_t ipaddr;
+#endif
+#if (!UIP_CONF_ROUTER || UIP_ND6_SEND_RA)
 static uip_ds6_prefix_t *prefix; /**  Pointer to a prefix list entry */
 #endif
 static uip_ds6_nbr_t *nbr; /**  Pointer to a nbr cache entry*/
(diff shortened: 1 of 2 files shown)


bradjc:
:+1:

alignan:
:+1:

core/net/ipv6/uip-nd6.c
```diff
@@ -682,7 +684,6 @@ uip_nd6_ra_output(uip_ipaddr_t * dest)
   nd6_opt_offset = UIP_ND6_RA_LEN;
 
 
-#if !UIP_CONF_ROUTER
```
simonduq: Don't we need `#if (!UIP_CONF_ROUTER || UIP_ND6_SEND_RA)` here?
The code uses `prefix`, which is conditionally declared

simonduq:
Good fix. If you address the comment I left in the code I'm willing to merge.

core/net/ipv6/uip-nd6.c
sdefauw: I think it will useless to add this macro. Indeed, at line 562 and 563, there are 2 macros saying that it needs to be a router and it sends RA. The code design is like this:
```
#if UIP_CONF_ROUTER
#if UIP_ND6_SEND_RA
static void
rs_input(void)
{
...
}
...
void
uip_nd6_ra_output(uip_ipaddr_t * dest)
{
...
}
#endif /* UIP_ND6_SEND_RA */
#endif /* UIP_CONF_ROUTER */
```

So the macro ``(!UIP_CONF_ROUTER || UIP_ND6_SEND_RA)`` will be always True because ``UIP_ND6_SEND_RA`` is set to True.

I think both macros are well written at line 562 and 563 because it needs to be a router to send a router advertisement (``UIP_CONF_ROUTER``). And all routers shouldn't send this type of message. We need to declare that this router must send RAs (``UIP_ND6_SEND_RA``).
 
About RA, we need to send Prefix Information Option (PIO) like described in the RFC4861. If you wouldn't like to send PIOs, you can set the flag ``advertise`` to 0 on the ``uip_ds6_prefix_list`` prefixes table.

simonduq:
Good point, I agree!

PR #704, opened 2014-06-09
CC2538: Creating a base CC2538 platform

bradjc:
The better the Contiki support for the CC2538 gets, the more CC2538 based hardware platforms that will be created. Right now, this is a bit tricky as one must copy the cc2538dk platform and make a few changes for the difference in hardware and then compile against the cpu/cc2538 folder. This works until the cpu/cc2538 changes at which point platform/cc2538dk gets updated but not the platforms sitting in other repositories and compilation fails.

Because the changes to the cc2538dk platform to create a new platform are usually small (typically initializing different hardware), we can create a "base" cc2538 platform and build specific hardware platforms off of that. This commit does just that. It spins most of the cc2538dk platform into just a "cc2538" platform and uses the cc2538dk platform for the /dev folder code and a couple of `init()` functions.

`contiki-main.c` now calls two new functions

- `platform_init()`
- `platform_process_init()`

in the `main()` function. The first `platform_init()` function calls any platform specific hardware functions that need to be initalized (like the smartrf06 buttons). The `platform_process_init()` is there to start and platform specific processes (like the sensors process). Both init functions are defined in `platform.c`.

This commit also adds a `platform-conf.h` file that can be used to override any of the defaults in `contiki-conf.h`.

Finally, this commit adds support for disabling the LEDs during boot.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -74,7 +74,7 @@ endif
 
 ifeq ($(CONTIKI_WITH_RIME),1)
   HAS_STACK = 1
-  CFLAGS += -DNETSTACK_CONF_WITH_RIME=1 
+  CFLAGS += -DNETSTACK_CONF_WITH_RIME=1
   MODULES += core/net/rime
 endif
 
@@ -177,8 +177,8 @@ CFLAGS += ${addprefix -D,${subst $(COMMA), ,$(DEFINES)}}
 
 ### Setup directory search path for source and header files
 
-CONTIKI_TARGET_DIRS_CONCAT = ${addprefix ${dir $(target_makefile)}, \
-                               $(CONTIKI_TARGET_DIRS)}
+CONTIKI_TARGET_DIRS_CONCAT ?= ${addprefix ${dir $(target_makefile)}, \
+                              $(CONTIKI_TARGET_DIRS)}
 CONTIKI_CPU_DIRS_CONCAT    = ${addprefix $(CONTIKI_CPU)/, \
                                $(CONTIKI_CPU_DIRS)}
 
--- a/platform/cc2538/Makefile.cc2538
+++ b/platform/cc2538/Makefile.cc2538
@@ -1,0 +1,33 @@
+# cc2538 base platform makefile
+
+ifndef CONTIKI
+  $(error CONTIKI not defined! You must specify where CONTIKI resides!)
+endif
+
+### These files must exist for all CC2538 platforms
+CONTIKI_TARGET_SOURCEFILES += leds.c leds-arch.c
+CONTIKI_TARGET_SOURCEFILES += contiki-main.c platform.c
+
+CONTIKI_SOURCEFILES += $(CONTIKI_TARGET_SOURCEFILES)
+
+### Unless the example dictates otherwise, build with code size optimisations
+ifndef SMALL
+  SMALL = 1
+endif
+
+### Define the CPU directory
+CONTIKI_CPU=$(CONTIKI)/cpu/cc2538
+include $(CONTIKI_CPU)/Makefile.cc2538
+
+MODULES += core/net core/net/mac \
+           core/net/mac/contikimac \
+		   core/net/llsec
+
+BSL = $(CONTIKI)/tools/cc2538-bsl/cc2538-bsl.py
+
+%.upload: %.bin
+ifeq ($(wildcard $(BSL)), )
+	@echo "ERROR: Could not find the cc2538-bsl script. Did you run 'git submodule update --init' ?"
+else
+	python $(BSL) -e -w -v $<
+endif
--- a/platform/cc2538/README.md
+++ b/platform/cc2538/README.md
@@ -1,0 +1,87 @@
+CC2538 Base Platform
+====================
+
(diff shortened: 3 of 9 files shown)


bradjc:
I think at this point either this should be pulled or closed. I think it would be useful to inherit CC2538 platforms, but if that's not in the cards there is no reason for this PR to be open.

Re travis: I don't think it is failing because of the changes in this commit.

ppannuto:
@g-oikonomou @adamdunkels bump?

g-oikonomou:
I still think we want a feature like that. Have a look and see what's been done with srf06-cc26xx, perhaps it would be suitable

adamdunkels:
I think having a base cc2538 platform is a great thing. There is significant overlap between all the existing cc2538 platforms already.

g-oikonomou:
@bradjc I am becoming _increasingly_ interested in this feature. Can we talk offline?

bradjc:
@g-oikonomou sure, you should be able to get my email from `git show af27d2d2`

I just rebased on to master, it looks like nothing major has changed so that went well.

bthebaudeau:
Sorry, rebase needed.

bradjc:
After looking at this again I have two notes:

1. Does it make sense to even have a `cc2538` platform? It doesn't really map to any particular hardware, and `make TARGET=cc2538` won't even compile. Should `contiki-main.c` and others just be moved to `cpu/cc2538` (as startup-gcc.c was)? The current problem is that there is a lot of code in `platform/cc2538dk` that has to be copied for any new platform, and that could be remedied by moving more to the cpu folder.

  I don't like the idea of cluttering the platform folder, but I am also sympathetic to any conceptual issues with moving more to the `cpu/cc2538` folder. But as I think about this more, a fair number of the function calls that `contiki-main.c` makes are to files in the cpu folder.

2. I would like to be able to support the notion of sub-sub-platforms. This would be useful for drop in modules with castellated headers or boards that support shields. Basically, the base platform would inherit from the cc2538 platform, and another platform could inherit from that new platform.

  I gave this a try, and ran into two main problems. First, include folders are specified as relative paths in the makefiles with the variable `CONTIKI_TARGET_DIRS`. Then in `Makefile.include` they are expanded using the platform makefile's path. Well in the case of nested platforms, there are two makefiles that both do this, and one ends up with the correct paths and the others are offset from the wrong base directory. To work around this I was able to change where `CONTIKI_TARGET_DIRS_CONCAT` gets set. Instead of in `Makefile.include` I put it at the end of the subplatform makefiles like this:

        mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
        CONTIKI_TARGET_DIRS_CONCAT += ${addprefix ${dir $(mkfile_path)}, $(CONTIKI_TARGET_DIRS)}
        include $(CONTIKI)/platform/cc2538/Makefile.cc2538

  It's a bit annoying to have to do this, but it does address the issue.

  Second, and more importantly, there is the issue of `board.h`. If there are multiple `board.h`s only one will get included, and if instead they are named something specific then I don't know where to include them. Inheriting `board.h` is critical so that sub-platforms don't have to redefine all of the pin mappings and whatnot of their parent platform, but I'm not sure a good way to handle this.

I did update the PR so that it should work with master if anyone wants to try it.

ppannuto:
@bradjc check out gcc's include_next, it may be the solution to
automatically concatenating board.h files

https://gcc.gnu.org/onlinedocs/gcc-3.0.2/cpp_2.html

g-oikonomou:
Moving main to the CPU dir has crossed my mind too, but I've always somehow ended up rejecting the idea.

As I hinted in one of my emails, I've been toying with the idea of a directory structure like this:

```
platform
  cc2538
    cc2538dk
    openmote
    zolertia
      remote
```

Then we could do `make TARGET=cc2538/zolertia/openmote` and we could probably nest as much as we like but I don't think we'd realistically need more than 3. Final platforms leaf nodes in this tree, so it might even be possible to avoid having to specify the entire 'path' but to simply do things the way we do today, e.g. `make TARGET=remote` and the build system should be able to pick this up. This would force unique directory names though: not a big deal

Just thinking aloud...

alignan:
Having unique names is not something too drastic and rather expected.  I prefer also to maintain platform-specific code as the cc2538 one in /platform and cc2538 specific in its current /cpu location.  For a platform I would basically would like to have in `platform.c` separate calls to initialize pins, peripherals, addressing and network stuff.

bthebaudeau:
I like @g-oikonomou's suggestion for the platform tree.

alignan:
Me too, but dislike the nested arguments for TARGET as commented before, I would prefer to stick to single TARGET=cc2538dk for example.

claudyus:
Hi all,
I'm next to provide some PRs that are using the CC2538 platform described here as base.
Here you can found this initial work: https://github.com/Unidata-SpA/contiki/commits/unidata

What do you suggest, should I rely on this base platform or there isn't interest in merging this feature soon or later?

g-oikonomou:
I'm now closing this. The `BOARD` business used by CC13xx/CC26xx platforms could be replicated here if there is interest in taking this forward.

PR #708, opened 2014-06-10
core/net/scan: New API for energy and beacon scans

darconeous:
This commit provides a mechanism for performing beacon
scans (or a "network scan") and energy scans.

Currently Contiki devices don't transmit beacons, but this
change is important step toward full network association
support. Beacons from other 802.15.4 devices will be
detectable with this API.

The beacon scan functionality is not 802.15.4 specific,
and can be used by any link layer that supports beacons.

Scanning support requires that the radio device implement
the new parameter API.

This API is documented in the associated headers. An example
is included demonstrating how the API should be used.


core/net/mac/csma.c
```diff
@@ -164,7 +164,7 @@ transmit_packet_list(void *ptr)
 }
 /*---------------------------------------------------------------------------*/
 static void
-free_packet(struct neighbor_queue *n, struct rdc_buf_list *p)
+csma_free_packet(struct neighbor_queue *n, struct rdc_buf_list *p)
```
simonduq: as per Contiki code style, static functions should not be prepended by the module name

simonduq:
any news, can you try a rebase?

simonduq:
closing after timeout

PR #712, opened 2014-06-11
Change the default IPv6 prefix from aaaa::0/? to fc00::0/7

adamdunkels:
Ever since 2008, the Contiki code has used an unauthorized IPv6 prefix `aaaa::0/?`, which at the time was picked for lack of a better prefix.

This pull request changes the default prefix to `fc00::0/7`, which is the unique local prefix as defined by RFC4193. Unique local addresses should not be globally routed, but may be used internally within a network or an organization, and makes a lot of sense to use as the default prefix in Contiki.

### DIFF
--- a/apps/rest-common/static-routing.c
+++ b/apps/rest-common/static-routing.c
@@ -30,7 +30,7 @@ void set_global_address(void)
 {
   uip_ipaddr_t ipaddr;
 
-  uip_ip6addr(&ipaddr, 0xaaaa, 0, 0, 0, 0, 0, 0, 0);
+  uip_ip6addr(&ipaddr, 0xfc00, 0, 0, 0, 0, 0, 0, 0);
   uip_ds6_set_addr_iid(&ipaddr, &uip_lladdr);
   uip_ds6_addr_add(&ipaddr, 0, ADDR_AUTOCONF);
 }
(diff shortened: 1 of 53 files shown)


g-oikonomou:
Still in favour of fd4d:4267:5f8c::

I'll either open pulls on top of this branch or close it deponding on reaction.

alignan:
:+1: for fd4d:4267:5f8c::/48, as it is registered already.

bradjc:
`aaaa::` and `fdfd::` are nice because they make it pretty clear that the derived addresses are not globally routed, but it's probably better to go with the address that is registered to contiki.

simonduq:
:+1 for the sixxs registered, standard-compliant one :)

joakimeriksson:
I vote for fd00::/64 or fdfd::/64 since they are very easy to type and indeed significantly better than the aaaa::/64 that is globally routable and might be used sooner or later by someone.

simonduq:
Looked at RFC 4193 again and it turns out the global ID must be generated randomly per network, not per implementation. The goal being to obtain globally unique IDs. Linux does not have one single hard-coded global ID for instance.

So, for the Contiki default prefix, used only when no PAN managed distributed a globally unique ID, I'm now in favor of the simpler `fd00::/64`. The longer one only adds confusion, as people might think they are compliant.

simonduq:
Just a heads up, there is a new PR addressing this: https://github.com/contiki-os/contiki/pull/1297

simonduq:
Closing now that https://github.com/contiki-os/contiki/pull/1297 is merged

PR #714, opened 2014-06-11
API proposal for AES and hardware random number generator

darconeous:
This isn't really a real pull request in the sense that I want it merged, I just wanted to share what I have with everyone to get some feedback on the appropriate way to abstract away things like this.

Ultimately we need a way to abstract away AES block operations (and potentially CBC-MAC) and random-number generation. However, probably want a AES-CCM+ implementation in the contiki tree (what would use the abstracted AES block operations). We also may want to have a contiki-specified whitening function for whitening raw entropy generated from hardware, because this is something that is often skipped and can significantly affect the resulting security of the device.

### DIFF
--- a/core/crypt/aes.h
+++ b/core/crypt/aes.h
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) 2014, Robert Quattlebaum
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


simonduq:
@darconeous, interesting discussion. Would be great if you could 1) update now that Contiki has an AES and AES-CCM* API, and 2) start the discussion as an issue.

simonduq:
closing after tmieout

PR #715, opened 2014-06-11
ENC28j60 Ethernet driver code

adamdunkels:
The ENC28j60 is a common Ethernet chip used by both the new OpenMote platform (http://www.openmote.com/), the do-it-yourself Thingsquare CC2538 setup (https://github.com/thingsquare/hardware/blob/master/ti/cc2538dk/assembly/README.md), and other upcoming platforms.

This pull request adds Thingsquare's ENC28j60 driver to the Contiki tree.

### DIFF
--- a/dev/enc28j60/enc28j60.c
+++ b/dev/enc28j60/enc28j60.c
@@ -1,0 +1,644 @@
+/*
+ * Copyright (c) 2012-2013, Thingsquare, http://www.thingsquare.com/.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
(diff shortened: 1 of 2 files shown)


markushx:
This pull request misses the arch specific files for e.g. eth2538, right? There's no implementation of enc28j60_arch_spi_write() enc28j60_arch_spi_read() currently in the tree. Shouldn't this pull request contain the arch specific drivers?

g-oikonomou:
@adamdunkels Adam, where are we on this one? There are some unanswered comments I think. Are they things that can be addressed or do we want this as-is?

adamdunkels:
This one is ready to be merged - we've been using this code for 2+ years now and while there certainly is room for improvement (e.g. some of the @bthebaudeau patches), I think it is more important that we get this first iteration into the code now than to wait even longer to patch it. We need this for a bunch of new platforms. We can always update via pull requests later.

nvt:
:+1:

PR #716, opened 2014-06-11
Make protothread semaphores interrupt-safe.

hexluthor:
Using PT_SEM_SIGNAL() within an interrupt service routine can cause a race condition if the counter decrement is not an atomic operation. Representing the semaphore count as a difference between two separate counters prevents such a race condition from occurring and makes PT_SEM_SIGNAL() interrupt-safe.


simonduq:
this looks legit, I'll merge this in a month if nobody has concerns

PR #717, opened 2014-06-11
core: Don't default UIP_CONF_BUFFER_SIZE to just 128 bytes.

darconeous:
Setting the uIP buffer size to 128 bytes is rather unreasonable,
especially for IPv6. uIP already manages default values for
macros like these. If a platform is really so constrained,
it can set this parameter itself in it's own `contiki-config.h`.

### DIFF
--- a/core/contiki-default-conf.h
+++ b/core/contiki-default-conf.h
@@ -117,13 +117,6 @@
 #define UIP_CONF_IPV6 0
 #endif /* UIP_CONF_IPV6 */
 
-/* UIP_CONF_BUFFER_SIZE specifies how much memory should be reserved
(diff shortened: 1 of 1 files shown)


simonduq:
I don't agree with removing `UIP_CONF_BUFFER_SIZE` from `contiki-default-conf.h` because this file is still Contiki's main documentation on basic configuration. Commenting it out is even worse.

If want a different default value, however, I don't see why anything smaller than 1280 would make sense. 1280 is the min MTU by standard. Anything smaller than that may result in a host not able to talk to the device. Of course setting it here to 1280 will take modification of many platform-specific config files...

I'm closing in a month if there is no update.

simonduq:
closing after timeout

PR #731, opened 2014-06-18
CC2538: Allow low-power modes to be disabled at run-time by calling lpm_set_max_pm(LPM_NONE).

hexluthor:
Currently the low-power modes can be disabled only at compile time by setting LPM_CONF_ENABLE = 0. This change allows the application to permanently or momentarily disable the low-power modes at run-time.


g-oikonomou:
CC2538 LPM will soon inherit from CC26xx LPM (which is a lot more sophisticated). Working with @alignan on this feature as we speak, so I'm closing this

PR #736, opened 2014-06-30
Added support for SPI slave and SPI-accel drivers

bmicho:
Added interupt support for slave SPI mode

Fixed minor bugs with PIN configuration:
When CS is configured with peripheral mode, SPI_WRITE will always terminate the transmission after each byte. This is expected since it waits for SPI not BSY.
for multiple byte transmissions use SPI_WRITE_FAST.

TI example software configures CS pins as software in accelerometer example. When i played with software controlled CS, SPI_WRITE_FAST() demonstrated a funny behaviour. In my oscilloscope the CS was changing values ... When CS is in software mode only use SPI_WRITE. SPI_WRITE works perfect in CS_Software mode.

Added check in SPI configuration of transmission size:
Now there is a max - min size check  in api driver (both init and configure function)

Removed wait time from SPI Flush, There is no point to use SPI_WAITFOREORx() before emptying the FIFO. 
it breaks when it is used since it waits for FIFO Not to be empty in order to empty it... if something a SSI_BSY
wait time should be added there to force the function to wait for end of transmission (if any) before erasing the FIFO

Added SPI master slave example. With 2 SMARTRF06 SPI master slave operation can be checked
Added smartrf06 accelerometer driver (not tested for calibrated)
added example with accelerometer

### DIFF
--- a/cpu/cc2538/dev/spi.c
+++ b/cpu/cc2538/dev/spi.c
@@ -32,6 +32,10 @@
  *
  * \file
  * Implementation of the cc2538 SPI peripheral
+ *
+ * \authors
+ *        Brad Campbell
+ *        Vasilis Michopoulos <basilismicho@gmail.com>
  */
 #include "contiki.h"
 #include "reg.h"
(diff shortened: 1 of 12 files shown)


simonduq:
OK, no update for so long, I'll close in a month if nothing happens

simonduq:
closing after timeout

PR #754, opened 2014-07-22
[CC2538] Fix SPI_FLUSH

wwhuang:
The current SPI_FLUSH macro first waits for the SPI RX FIFO to fill with something. So if an application makes a SPI_FLUSH call and the RX FIFO happens to already be empty, the application stalls.

### DIFF
--- a/cpu/cc2538/spi-arch.h
+++ b/cpu/cc2538/spi-arch.h
@@ -63,7 +63,6 @@
 #error "You must include spi-arch.h before spi.h for the CC2538."
 #endif
 #define SPI_FLUSH() do { \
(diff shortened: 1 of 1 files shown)


bradjc:
:+1:

bthebaudeau:
:+1:

Before merging, we just have to check that there aren't any in-tree users relying on the current wrong behavior of this macro on CC2538, and fix them in this PR if this is the case in order not to break Contiki.

bthebaudeau:
Checked: no in-tree CC2538 users of `SPI_FLUSH()`.

:+1:

PR #756, opened 2014-07-23
dos2unix fix

sieben:
I've found all the file with windows encoding by using the following
command :

    find . | xargs file | ack-grep "CRLF"

It's better to have only one encoding and since UNIX is the convention
used for core it should also be the one used everywhere else.


sieben:
@jimparis I've added a .gitattributes with the right argument in #690 I think it should be ok now.

jimparis:
Minor bug, you specified `tabwidth=4`, but it seems that existing files generally expect hard tabs to be 8 characters.

sieben:
What files are you referring to?

jimparis:
The .gitattributes that you added.

sieben:
I'm not sure to follow, my .gitattributes works plus hard tabs being 8
spaces seems very arbitrary. Do you mean that some files needs to have tabs
at 8 characters and if so which ones?

2015-02-25 6:36 GMT+01:00 jimparis <notifications@github.com>:

jimparis:
Hard tabs being 4 spaces also seems very arbitrary, which is why I was surprised to see you explicitly picked that value.  Why not just let it remain at the git default?  Most of the code currently uses 8, and the uncrusitfy.cfg specifies 2.  Adding a new file saying 4 seems extra confusing.

sieben:
Oh i see it now. Sorry, I will update this pr.

simonduq:
question: how does this relate to https://github.com/contiki-os/contiki/pull/690?

sieben:
I think we could make only one PR that will handle both issues of line
ending and trailing whitespaces.

Le ven. 28 août 2015 à 22:19, Simon Duquennoy <notifications@github.com> a
écrit :

simonduq:
After an off-line discussion with @sieben, we agreed to close https://github.com/contiki-os/contiki/pull/690 for now and focus on .gitattributes and line endings under this PR first.

sieben:
@simonduq if you have a windows machine can you test that it works as well?

    find . | xargs file | ack-grep "CRLF"

simonduq:
Nice tip, and as mentioned by @adamdunkels in #690, this also works: https://github.com/contiki-os/contiki/pull/756/files?w=1

:+1: for me, merging in a month if nobody opposes

simonduq:
merging after timeout

PR #771, opened 2014-08-08
Using gradle to build COOJA

sieben:
## Why using Gradle ?
- Can download dependencies dynamically (so we can delete lib that are bundled with contiki)
- Can perform check style (Android style guide)
- Can perform static analysis (FindBugs plugin)
- Cross platform
- Modern IDE (Eclipse, Idea IntelliJ,...) understands it
- Don't use XML as a build definition but a simple dynamic language
- Gradle is used to build up Android application (http://www.gradle.org)
- It's supported by travis-ci : http://docs.travis-ci.com/user/languages/java/

Let's use both ant and build.gradle. They can live together without any problems :D But gradle got much more features and I think in the long term should be preferred.


sieben:
I've rebased this PR but there is still errors :sweat_smile: Any help is appreciated to merge it :smile: 

@cmorty @fros4943 I've rebased but new errors appeared. I don't really know what went wrong wit : 
- https://travis-ci.org/contiki-os/contiki/jobs/68709436 any idea?

sieben:
@nvt @fros4943 The PR is now clean :) I'm at SICS now for a month

tools/cooja/apps/mspsim/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+            srcDir 'src'
+        }
+    }
+}
+
+task buildMspsimJar(type: Exec) {
+    environment System.getenv()
+    workingDir '../../../mspsim/'
+    commandLine 'sh', '-c', 'ant jar'
+}
```
cmorty: Is this tested for windows?

cmorty:
Why change the folder layout? You can configure input and  output paths. History will break, again, patches and links need adjustments and all that stuff. Changing Coojas namespace already was a PITA. It also makes the patch unnecessary large.

tools/cooja/apps/mspsim/build.gradle
sieben: I don't have windows for testing it. Do you have it?

2015-08-24 10:34 GMT+02:00 Moritz Strübe <notifications@github.com>:

sieben:
I think it's better to have a more standardized folder layout. The
changes are mostly simply renaming and simple reformatting code that an IDE
can do easily. No API are broken. @fros4943 comment from 25 Aug 2014 seems
to be ok with it.

2015-08-24 10:40 GMT+02:00 Moritz Strübe <notifications@github.com>:

tools/cooja/apps/mspsim/build.gradle
cmorty: Not really, especially there is a lack of time, but I think that using 'sh' is a very bad idea. Especially if there are gradle provides the functionality natively: https://docs.gradle.org/current/userguide/ant.html

cmorty:
Well, he is not maintaining out-of-tree plugins, patches and branches. All those will most certainly break. Thus I'm very in favor of just adding two instructions to the build tool and save others a lot of trouble.

sieben:
I see your point. @fros4943 what's your view on this?

fros4943:
I was originally mostly looking forward to the checkstyle integration, and also to replace Ant. I'm not convinced that this PR is ready to be merged yet however. I agree with @cmorty to not rename source files, so let's instead update the Gradle configuration.

cmorty:
@sieben Looking at the changelog I'm not sure why you don't start at the initial commit and than rebase/cherry-pick the other commits.

sieben:
@cmorty I'm trying to make the tests pass again. It was working before but now that I'm moving back to the old (and not standard) way there is lot of settings to finetune. I will rebase when it will be working again.

tools/cooja/apps/avrora/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+      srcDir 'src'
+    }
+  }
+
+}
+
+
+dependencies {
+    compile project(":")
+    compile files("lib/avrora-small.jar")
```
renatoathaydes: Depending on relative paths of local files for your Gradle build is really bad practice. You should instead install them in a local repository (I recommend using a local Maven repo - don't worry, nothing to do with Maven, just the structure of the files) if they are not part of the same build:

```groovy
repositores {
    mavenLocal()
    jcenter()
}
```

All jars will then be installed at `${user.home}/.m2/repository` (by default). You don't need to know that, though, as Gradle will make its magic from there.

Change all dependencies you have of this type to be regular dependencies of the form `compile "group:name:version"`. 

Of course, each project has to be given a group and version (name is the name of the directory unless you specify it).

Run the build with `gradle clean publishToMavenLocal`.

You can later even publish all your artifacts (jars) to bintray (JCenter) if you want to so people can depend on your "jars" in their projects too.

Regarding the dependency on the root project (`compile project(':')`), try replacing this line with the following

```groovy
compile rootProject
```

This tells Gradle your child project has a dependency on the root project.

But as I told you in the Java Meetup, think of the build as a tree of projects where a project cannot depend on another project which is higher in the same branch of the tree, they can depend only on "siblings".

Please have a look at some of my projects, I have a really interesting one with multiple projects built together [here](https://github.com/renatoathaydes/osgi-run/tree/master/osgi-run-test).

cmorty: I somewhat (strongly) disagree. We have a circular dependency. The plugins need Cooja to compile and we don't want Cooja without the plugins[^1]. And, as they are in the same build-tree, I think that, while having those artifacts in a maven-repo is the intended way to go, there are no real advantages in this case.
Especially as Cooja and its plugins never had, and very likely will never have, reasonable versioning. Without you'll come into the Maven-Versioning-Hell, because all builds and checkouts will share a single jar in the maven-repository and you'll spend hours to find out why things a not working as intended. This happened to me on another project where I was working on multiple checkouts at the same time.
To wrap it up: I'm very in favor of (local) maven-repositories, but I don't think they fit Cooja's development-model.

^1: Although I implemented plugin-autodiscovery a while ago, last I checked their paths were still hard-coded in the config.

renatoathaydes: You are mixing up a compile-time dependency with a distribution dependency. You don't have a circular dependency at compile time (it wouldn't compile if you did). What you're referring to is the fact the plugins are part of the same distribution as the core of Cooja, but you want them to be distributed together (but hopefully, the core Cooja does not need plugins to compile!).

You can achieve the same easily by keeping the jars the standard way (in a repo, Maven or not). You can even call the lib folder your repo (see Gradle's repositories guide) so you can keep it in the same location you have them today. The version for the plugins may be something just nominal, but you should definitely consider versioning them for real.

I am not sure what you're referring to with "all checkouts will share a single jar in the maven-repository". That's the opposite of what I am trying to say. You should have ONE jar for each Gradle module. These jars should be in a repository (even if it's just a local directory - the difference being just how you define that location, which is a big one as it allows you to later change the repository without touching any particular dependency). Distribution is something different altogether... You can have all your jars copied into  the distribution folder later (without explicitly referring to the location of those jars - if they are in a repo, they are just normal artifacts with a coordinate you can refer to). You can even use a Ant task from Gradle as Gradle has a Groovy's AntBuilder available as the variable `ant`.

Hope it makes sense. I'm claiming this model does fit the Cooja development-model, just like this model fits any other build I can imagine.

cmorty: Yea, I'm well aware that it does does not have a circular _build_ dependency. I just noticed that a "want _to run_ Cooja without the plugins." would have been a better wording.

But back to the topic at hand: To explain my issues in more detail:
You have two local git-checkouts of Cooja - Co1 and Co2. You compile and install the Cooja package in Co1. Now you compile a plugin in Co2. Due to the repo-magic it will use the Cooja-build from Co1 to build against as it as matching is done one name+version and both artifacts will have the same version, as Cooja has no versioning. Even if building does not fail this can cause problems when building Cooja in Co2 afterwards when the class-loader loads that plugin. This is for example the case when you add a function to one of Cooja's classes in Co2 that the is not used by that plugin, yet the class itself is referenced. 

So if you suggest manually setting the repo-path to the, again relative, path 'tools/.m2' or similar that is ok for me, yet I don't see any advantage over manually referencing those files, which is much easier to understand for people not familiar with the maven-repo thing.

renatoathaydes: I understand your concern better now. And I want to say that you would definitely benefit from the setup I suggest (even more than I thought before).

I will try to get some time on the weekend or something to study your build system in more detail. Hope I can help @sieben setup a really great and easy-to-understand Gradle build. Trust me, it can be really great!

Responding to some of your concerns:

* About not seeing the latest jars: you can tell Gradle to always reload the jars from the repo even if their versions did not change (so your build always see the latest jar). See this [discussion](https://discuss.gradle.org/t/how-to-get-gradle-to-download-newer-snapshots-to-gradle-cache-when-using-an-ivy-repository/7344).
* About the repo path: the important, in my view, is not where your repo is going to be located. The important is that projects refer to artifact coordinates only, not jars. Their exact locations do not matter for the projects that depend on it, why should they "know" it? They just tell Gradle, I need this artifact with this group:name:version (version is even optional). I don't care where you'll find. That's a different concern and its solution is specified in the repositories section - that's where you'll worry about where Gradle should look up things. And this is the greatest thing about Gradle. You can use JCenter, Maven Central, your local Maven repo, your file system, even a random url! Have a look at the dependency management/[repositories](https://docs.gradle.org/current/userguide/artifact_dependencies_tutorial.html#N10621) documentation. It's beautiful.

Believe me, there's a huge advantage in separating the concerns of specifying which artifacts you need and specifying where they might be physically located.

cmorty: > you can tell Gradle to always reload the jars from the repo even if their versions did not change (so your build always see the latest jar)

This will not fix the issue, as the problem is that the repo contains the wrong version.

> Their exact locations do not matter for the projects that depend on it, why should they "know" it?

You didn't get the point. I need to worry where Gradle gets that library, because I need that exact version, which however can not be distinguished from the wrong version by name and version, but _only_ by its location. Yet, if you don't believe me I might actually spend the time to break the approach once its implemented. I have spend enough time fighting this problem, because I had multiple checkouts of the same project and someone was of a the opinion that maven-artifacts are cool and created as many different artifacts as possible. It was a PITA to build two versions of the main application with a change somewhere down the dependency chain. I ended up giving them different version numbers which had to be propageted all the way up the dependency-chain.

Overall: Your solution is good for somewhat independent projects, like libraries providing functionality that might also be useful in another project. This is not the case here. What you should look into is https://docs.gradle.org/current/userguide/userguide_single.html#multi_project_builds . That approach described there might actually improve things and @sieben's current approach, but I'm not sure whether it is worth the trouble.

sieben:
I've added a script to test my work: https://gist.github.com/sieben/cec76a09f0a46e6b0ac9

sieben:
@renatoathaydes @cmorty @fros4943 Can you take a look? Regression tests about code quality is in the air ;)

tools/cooja/apps/mspsim/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+}
+
+task buildMspsimJar(type: GradleBuild) {
+    buildFile = 'importedmspsimBuild.gradle'
+    tasks = ['jar']
+}
+task buildCoffeeJar(type: GradleBuild) {
+    buildFile = '../../../coffee-manager/build.gradle'
```
cmorty: It might be nice to replace `../../../` with some variable like `contiki_tools_path`, but that's eye-candy

tools/cooja/apps/mspsim/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+repositories {
+   jcenter()
+   flatDir {
+       dirs 'lib'
+   }
+}
+
+project.gradle.buildFinished { buildResult ->
+  if (buildResult.getFailure() != null) {
+    println "Could not find the MSPSim build file. Did you run &quot;git submodule update --init&quot;?"
```
cmorty: As it seems that we expect too much from some people, maybe we should re-phase it to: `Please run: git submodule update --init` - though by now all those people automatically end up on my ignore-list. ;)

sieben: I like the Did you run. It sparks curiosity, which I think is a good thing
;-)

2015-09-29 9:52 GMT+02:00 Moritz Strübe <notifications@github.com>:

tools/cooja/build.gradle
```diff
@@ -0,0 +1,153 @@
...
+
+task run_bigmem(dependsOn: jar, type: JavaExec) {
+    description = "Run the simulation with a larger Java memory heap"
+    main = "org.contikios.cooja.Cooja"
+    workingDir = "build"
+    maxHeapSize = "1536m"
+}
+
+task run_nogui(dependsOn: jar, type: JavaExec) {
+    description = "Run the simulation without graphical interface"
```
cmorty: This isn't working, is it?

tools/cooja/build.gradle
```diff
@@ -0,0 +1,153 @@
...
+
+task runprof(dependsOn: jar, type: JavaExec) {
+    description = ""
+    main = "org.contikios.cooja.Cooja"
+    workingDir = "build"
+    jvmArgs "-agentlib:yjpagent"
+}
+
+task runfree(dependsOn: jar, type: JavaExec) {
+    description = ""
```
cmorty: What is this needed for?

tools/cooja/build.gradle
```diff
@@ -0,0 +1,153 @@
...
+task run_nogui(dependsOn: jar, type: JavaExec) {
+    description = "Run the simulation without graphical interface"
+    // <arg line="-nogui=${args}"/>
+    main = "org.contikios.cooja.Cooja"
+    workingDir = "build"
+    maxHeapSize = "512m"
+}
+
+task run_applet(dependsOn: jar, type: JavaExec) {
+    description = "Run the simulation as an applet"
```
cmorty: This isn't working either, right?

cmorty:
Didn't try it out, but it looks ok to me. Some notes: 
Why is `libsDirName` `../lib`and not `.` for the apps? It does make sense for Cooja as its sometimes run from the class-folder.
Does findbugs and checkstyle also scan the apps?
I think you should remove the targets that don't work. If they are needed someone can add them later.

tools/cooja/build.gradle
sieben: I've tried to mimick as much as possible the old ant command that were
limiting the HeapSize and other settings like this. I don't know if with
modern machines and recent development of Cooja they are still necessary.

2015-09-29 9:59 GMT+02:00 Moritz Strübe <notifications@github.com>:

sieben: For the lib I've tried also to imitate as much as possible legacy behavior.

2015-09-29 10:01 GMT+02:00 Remy 'Sieben' Leone <remy.leone@gmail.com>:

tools/cooja/apps/mspsim/build.gradle
cmorty: Me, too. But I'm always frustrated about how many fail reading and interpreting warnings.

tools/cooja/build.gradle
cmorty: I see that, but if it's not working I think it should be removed or implemented correctly.

tools/cooja/build.gradle
sieben: https://docs.gradle.org/current/dsl/org.gradle.api.tasks.JavaExec.html
maxHeapSize is working

cmorty: Yep, but for non-gui you need to pass `-nogui=...` which you commented out.

sieben:
@cmorty I've removed the target that don't work
@fros4943 What do you think of it now?

.gitignore
```diff
@@ -93,3 +93,12 @@ contiki-cc2530dk.lib
 # rl78 build artifacts
 *.eval-adf7xxxmb4z
 *.eval-adf7xxxmb4z.srec
+
+# Gradle
+
```
cmorty: In case you have the motivation: No newline over a heading.

tools/coffee-manager/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+  jcenter()
+}
+
+sourceSets {
+  main {
+    java {
+      srcDir 'org'
+    }
+  }
+
```
cmorty: Unnecesary newline.

tools/cooja/apps/collect-view/build.gradle
```diff
@@ -0,0 +1,18 @@
+apply plugin: "java"
+
+libsDirName = ".."
+
+repositories {
+    jcenter()
+}
```
cmorty: I have no idea whether I'm right, but aren't repositories also imported via the dependencies? And is it needed at all?

tools/cooja/build.gradle
```diff
@@ -0,0 +1,88 @@
...
+      jcenter()
+    }
+
+}
+
+apply plugin: 'findbugs'
+apply plugin: 'checkstyle'
+apply plugin: 'application'
+apply plugin: 'eclipse'
+apply plugin: 'idea'
```
cmorty: I'm not sure whether I've asked that before: Are the plugins checked, too?

tools/cooja/apps/collect-view/build.gradle
sieben: Indeed it's needed because you could have different repositories. The dependencies don't bring a specific repository with them.

tools/cooja/build.gradle
sieben: No they are downloaded on the fly.

cmorty: Ah sorry, that was badly phrased. I was wondering whether only Cooja is checker using checkstyle abd findbugs, or  the Cooja-Plugins, too.

sieben: They appear in the building log but they don't appear in the results logs. Anyhow, plenty of errors to fix in the main Cooja before even looking at the apps ;-)

cmorty: I'm less concerned about fixing, but adding new ones. Maybe a comment in the code would be nice.

sieben:
@fros4943 Are you still interested by this PR?

simonduq:
@fros4943, @cmorty, do you see any more reason not to have this merged?
I don't know the details here, but this is a new (optional) feature, with a great potential towards improved Cooja code quality, and the result of significant effort from several rounds of constructive comments over more than a year.

cmorty:
@simonduq  I think this might need some rework, but we'll only know once it's merged and people start complaining. ;) Thinking about it, it might be a good idea to add a dummy `build.xml` that always says something like: 
> We have now switched to gradle. Please run "gradle" instead of "ant". 
> For ubuntu/debian you can install grade running `apt-get install gradle`. For windows you'll find instructions at \<Link to a decent manual.>

The official instructions are shit. Most people will fail setting their environment variable. Possible manuals might be 
  * http://www.bryanlor.com/blog/gradle-tutorial-how-install-gradle-windows
  * http://bgasparotto.com/install-gradle-on-windows-and-ubuntu/

cmorty:
@simonduq  I think this might need some rework, but we'll only know once it's merged and people start complaining. ;) Thinking about it, it might be a good idea to add a dummy `build.xml` that always says something like: 
> We have now switched to gradle. Please run "gradle" instead of "ant". 
> For ubuntu/debian you can install grade running `apt-get install gradle`. For windows you'll find instructions at \<Link to a decent manual.>

The official instructions are shit. Most people will fail setting their environment variable. Possible manuals might be 
  * http://www.bryanlor.com/blog/gradle-tutorial-how-install-gradle-windows
  * http://bgasparotto.com/install-gradle-on-windows-and-ubuntu/

simonduq:
Ah ok I was assuming it was still possible to use ant. If not, I understand you/we need to be extra careful. Sorry for the spam.

cmorty:
@simonduq No, getting rid of ant was one of the goals. The only reasonable option would be to add an ant file that calls the gradle-wrapper. But that should print a warning that this is deprecated.
@renatoathaydes At least I wasn't. But it's a good idea, indeed.
@sieben I might be able to spare a few hours to get this done. Best contact me via IM.

If those two things are done I think it would be a good idea to merge and see what happens.

sieben:
I would like to have some feedback from @fros4943 before doing anymore
work. He got the :+1:

2015-11-18 8:44 GMT+01:00 Moritz Strübe <notifications@github.com>:

fros4943:
Hi @sieben 

Checked out the latest code and gave it a quick try on Windows/Cygwin. Here's some immediate feedback:
* `ant run` in `/tools/cooja` must give an informative error rather than just a bunch of compilation errors. Would be great with a pointer to how to install gradle on Windows/Instant Contiki, and that ant is no more to be used.

* My first attempt `gradle run` failed without any useful error message, whereas `gradle jar` instructed me to update the git submodules (great!), even though the output could use a minor update.
```Could not find the MSPSim build file. Did you run &quot;git submodule update --init&quot;?```

* Cooja complains about not finding `org.contikios.cooja.plugins.collectview.CollectView` as well as `quickhelp.txt`

Cooja does appear to run at this point.

I then tried to repeat the same steps in Instant Contiki, over a shared folder:
* `apt-get install gradle; gradle run` which fails with the following error
```
Build file '/mnt/hgfs/fros4943/contiki/tools/cooja/build.gradle' line: 5

* What went wrong:
A problem occurred evaluating root project 'cooja'.
> Could not find method jcenter() for arguments [] on repository container.
```

Ubuntu installed Gradle version 1.4, which does not yet contains `jcenter()`. Is this something that can be resolved using gradle-wrapper?

**Comment by renatoathaydes** [comment]
@fros4943 you should definitely use the Gradle Wrapper (I mentioned before about this, it allows you to build without downloading/installing Gradle as a bash/bat script is used to actually run the build - it downloads the Gradle version you configure the first time you use)...

I use Gradle for many projects and I don't even have it installed (every project can use whatever version it wants to). To start using it, all you have to do is type "gradle wrapper" (the first person to do it does need to have it installed, of  course)... this will create the files `gradlew` and `gradlew.bat` (for use in Unix-like systems and Windows, respectively) as well as a `gradle` directory containing a jar and a properties file... the properties file is where you set the Gradle version you want to use. Try to use version 2.9, the most current. The 1.4 version you mentioned is incredibly out-of-date.

Let me know if you want any more info.

**Comment by renatoathaydes** [comment]
One more note: you must add to source control all files created by Gradle wrapper for the build to be completely portable, including the bash/bat scripts and the jar(s).

**Comment by sieben** [comment]
@renatoathaydes Thank you very much for your insights :) I've added a gradle 2.9 wrapper just now. 
@fros4943 Thank you very much for your review. I will update both build.xml and build.gradle. That said I don't have a windows machine to test right now...

**Comment by sieben** [comment]
@fros4943 

Rapid heads up:
- I've added a wrapper and removed the dependencies in the .travis.yml for ubuntu-baked gradle
- Updated the whole chain to use the new gradle (including the regression tests)
- Clean format the warning for mspsim in case of unsuccessful build
- Fixed the collectview problem

I'm not sure to understand what the problem is with quickhelp.txt. It's probably a beginner mistake @renatoathaydes @cmorty do you have any clues ?

**Comment by renatoathaydes** [comment]
Where's the original `quickhelp.txt` file and where does Cooja expect it to be moved to?

**Comment by sieben** [comment]
The original is at the following path:
contiki/tools/cooja/config/quickhelp.txt
In the cooja.jar quickhelp.txt is at the root of the jar.
what really intrigues me is that the quickhelp.txt is at the root of the
jar in both cases.

2015-11-25 14:08 GMT+01:00 Renato Athaydes <notifications@github.com>:

**Comment by renatoathaydes** [comment]
In this line of the Gradle build file:

```groovy
resources {
    srcDir 'config'
}
```

You tell Gradle `config` is the directory which should be mapped to the root of the jar file, so all files in this directory will go into the root of the jar... if you want this file to be in, say, `help/files/quickhelp.txt`, inside the jar, just move it from `contiki/tools/cooja/config/quickhelp.txt` to `config/help/files/quickhelp.txt`.

cmorty:
@sieben So you finished your phd and are cleaning up?

sieben:
Indeed! I got a PhD and a post-doc and now I'm working at Scaleway. Because
I use github.com/pulls as a to-do list I wanted to clean up :)

PR #805, opened 2014-10-08
[cpu/cc253x] Makefile.customrules-cc253x: Disabled parallel makefile execution

comawill:
The makefile (especially the execution of bank-alloc.py) is not capable
of parallel execution. So I disabled it in general for this cpu.

This is the seperate PR as requested in #803


g-oikonomou:
The only solution I can think of is to copy the entire `obj_` directory for each different firmware that needs build. However, I really dislike this solution and it will only make a difference under very specific circumstances.

Thus, I'm closing this.

PR #809, opened 2014-10-09
[MSP430] Add gstabs debug information (dwarf is broken)

cmorty:
Currently there is an linker error when compiling with debug information.
This is only the case for dwarf (the default). Everything is fine with
stabs, thus allowing to debug and use all the other nice tools like
"objdump -S".


alignan:
:+1:

PR #832, opened 2014-10-23
change type of ascii2petscii and ptr to char

vac:
It will fix signedness warning.
It's related to: https://github.com/contiki-os/contiki/pull/831


simonduq:
The warning is now gone with https://github.com/contiki-os/contiki/pull/1051/files, I'm closing (happy to re-open if you still think the type of ascii2petscii still needs changing -- in that case please do the change, rebase on lastest master and squash all commits)

PR #837, opened 2014-10-27
nbr-table: add table purge method

lebrush:


simonduq:
Hi @lebrush,
No activity on this one for a while. Do you intend to address @nvt's and my comments? Will have to close otherwise.

PR #839, opened 2014-10-28
rf230bb: Handle all IRQ flags in one ISR call. (Q: Is the current ISR valid?)

gebart:
I think the if statements in the ISR in rf230bb look weird. If the first flag is matched then no other IRQ flags will be checked. According to the data sheet, the IRQ signal is deasserted after reading the IRQ status register, which, as I understand it, would cause the other unhandled IRQ flags to be lost forever. See AT86RF212 data sheet, chapter 8.1 "Frame Receive Procedure".

If my understanding of the data sheet is correct, then this behaviour can possibly lead to missed TRX_END interrupts, because RX_START is checked first. The reason that these two flags would be set at the same time could be that the handling of the interrupt from the radio has been delayed by other code either disabling interrupts, or another ISR with a higher priority was executing during the time between RX_START and TRX_END.

I have attached an untested example solution for the problem described above.

The reason I discovered this is that I am looking over the rf230bb code in order to try to find the source of an issue I am seeing with dropped packets from time to time. I have not yet had time to evaluate if this change fixes the problem.

### DIFF
--- a/cpu/avr/radio/rf230bb/halbb.c
+++ b/cpu/avr/radio/rf230bb/halbb.c
@@ -751,7 +751,8 @@ HAL_RF230_ISR()
 #endif
 #endif
 
-    } else if (interrupt_source & HAL_TRX_END_MASK){
+    }
+    if (interrupt_source & HAL_TRX_END_MASK){
 	   INTERRUPTDEBUG(11);	    	    
         
        state = hal_subregister_read(SR_TRX_STATUS);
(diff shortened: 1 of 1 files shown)


comawill:
Reduced some packet-loss by fixing the same thing, the same way (noticed this patch afterwards...).
Technically the fix is correct, the usage of INTERRUPTDEBUG might result in wrong states, if more than one interrupt flag is set in IRQ_STATUS. The easiest solution would be to remove the INTERRUPTDEBUG thing completely. If someone needs to debug the interrupt, it should be no problem for one to implement something practical on demand.

simonduq:
@malvira, do you have an opinion?
If no more activity in the coming month, I'll close.

adutze:
As the interrupt code looks now, only the RX_END bit is used with AUTO_ACK configured as default. The changes in the code are OK, I have been testing it with an RF233. It could improve things if the radio states transitions would be considered in the interrupt.

simonduq:
I see, this PR also makes sense to me now (although I known nothing about this particular platform...).
As we don't really get reviews anymore for this platform anymore, I'm of the opinion of merging this one if users give it good feedback.

Will merge in a month if nobody finds a flaw in this fix.

malvira:
merging this. I agree, those shouldn't be else ifs.

PR #849, opened 2014-11-05
Revert "CC2420: Added RSSI_OFFSET as described in the manual"

cmorty:
This reverts commit f513ef9ef0302f636e11b768de5eeffaaf750dd0. (#557)
Moving the RSSI to the right value will brake everything depending on
the wrong value. This includes the RSSI-Scanner and any research project
that relies on the wrong value.

### DIFF
--- a/dev/cc2420/cc2420.c
+++ b/dev/cc2420/cc2420.c
@@ -69,12 +69,8 @@
 #define FOOTER1_CRC_OK      0x80
 #define FOOTER1_CORRELATION 0x7f
 
-#ifdef CC2420_CONF_RSSI_OFFSET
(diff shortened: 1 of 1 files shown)


alignan:
Any updates on this?

simonduq:
how about, instead of reverting, setting the default `RSSI_OFFSET` to `0` as suggested by @kkrentz?

cmorty:
Thats fine for me. :)

cmorty:
Actually thinking about it. As it's broken/fixed for so long now, it might just do fixing the other tools....

simonduq:
OK @cmorty as you are saying this is settled, let's close (closed PR are still index and usable as reference).
Thanks!

PR #852, opened 2014-11-07
Assert that the ringbuf has the right size / Add compile time assert

cmorty:
This PR has to patches. The fist supports compile time asserts. The error itself is not really helpful, but as one still gets file name and line number its better than nothing.
The second uses the new compile-time assert-feature to ensure the the puffer passed the the ringbuf function fulfills the requirements.

### DIFF
--- a/core/dev/serial-line.c
+++ b/core/dev/serial-line.c
@@ -131,7 +131,7 @@ PROCESS_THREAD(serial_line_process, ev, data)
 void
 serial_line_init(void)
 {
-  ringbuf_init(&rxbuf, rxbuf_data, sizeof(rxbuf_data));
+  RINGBUF_INIT(&rxbuf, rxbuf_data);
   process_start(&serial_line_process, NULL);
 }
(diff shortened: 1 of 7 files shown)


simonduq:
I agree with @nvt that the macro on top of the init function makes the code more complex than it needs to. I would close this PR. Will give it another month in case someone wants to save it.

cmorty:
@simonduq Still didn't get any feedback from @nvt on 8634e7a

nvt:
I have a lukewarm feeling about this PR -- either merging or not merging is OK with me, but there are no clear supporters at the moment.

cmorty:
@nvt, @simonduq: I still think it's at least a nice example of best coding practice, as it's the only way of making decent compile time assertions in C - even adding a message. Doing so in runtime code IMO is not reasonable, because they add unnecessary overhead that nobody checks anyway. 
However, as learning a bit about C was my main gain out of this, I can sleep well if it is not merged.

simonduq:
not strong supporter, closing

PR #880, opened 2014-11-28
Streamline framer-802154.c

kkrentz:
Currently, `framer-802154.c` uses a temporary `struct` for creating and parsing frames. I modified `framer-802154.c` to write created frames directly to the packetbuf. Likewise parsed 802.15.4 headers are directly stored in packetbuf attributes. This saves RAM, ROM, as well as processing time.


simonduq:
Great initiative, looks excellent to me! I would need to look a bit deeper and try it out before fully supporting though.
Dare I ask for a rebase?

Some comments from a first look:
* why not removing `frame802154.h` altogether?
* alternative naming for `framer_802154_just_parse`: how about `framer_802154_peek_parse`?

Which platforms and configurations have you already tested this on? (with and/or without llsec, short/long address, RDCs and MACs)

kkrentz:
Hi there. Prior to this rebase, various combinations of RDCs and MACs worked fine for me. However, I only tested it on Sky motes.

`frame802154.h` is now gone. I moved the content of `llsec802154.h` and `frame802154.h` to [`ieee802154.h`](https://github.com/kkrentz/contiki/blob/shrink-framer/core/net/mac/ieee802154.h). `ieee802154.h` now contains very general definitions and settings - maybe too general to put them into framer802154.h.

simonduq:
Great, thanks. Will do some testing in the coming week(s).

simonduq:
OK, one of the tests I had in mind was to try this along with our IEEE 802.15.4e TSCH implementation. I'm now worried by one limitation of this PR: the assumption that the data is packetbuf means there is no way to parse frames from interrupts anymore. I believe this to be a real problem, in particular with the various MAC layers of 15.4e (soon to be a part of IEEE 802.15.4-2015). Enhanced ACKs and frame Information Elements often require some low-level parsing, from interrupt context, e.g. to decide whether to ACK or not and to build ACKs.

However, I really like many things in this PR, because it really cleans up the parser and gets rid of the `frame802154` - `framer-802154` confusion. Do you see a way to extend this PR with optional parsing from a specified buffer? Something like having `framer_802154_peek_parse` rely on a sub-function, similar to the former `frame802154_parse`, but where whenever `frame802154_t *pf` is `NULL`, will write directly to packetbuf addresses and attributes?

kkrentz:
We could, e.g., move all packetbuf attributes to a big struct like that:
```
struct packetbuf {
  uint8_t seqno;
  enum ieee802154_sec_lvl sec_lvl;
  ieee802154_frame_counter_t frame_counter;
...
}
```
and write such a function:
`framer_802154_peek_parse(struct packetbuf *p)`

However, I don't know whether this approach is viable for parsing information elements, too.

simonduq:
Anything should be possible to extend with information elements I'd say so we shouldn't worry too much; the main thing being to make sure parsing remains possible form interrupt.

One problem with your suggestion above, if I get it right, is that it will increase the number of attributes/fields, and as all of them get copied in queuebuf this feels like a waste. How about something like this:
* add a struct that contains all elements of a parsed frame, similar to `frame802154_t *pf`
* add one such struct to packetbuf, in addition to attributes
* have the parsing function work on this struct as you suggest
* some of the existing attributes might need to move to the struct, but we need to be extra careful when selecting them, because anything removed from attributes won't be saved in queuebuf

kkrentz:
Well, my idea is intrusive. That struct should contain all attributes, including addresses. That is, we won't have attributes any longer. This should actually save RAM because we can use specific types, as well as unions. What's more, accessing packetbuf atributes would become more readable and program memory efficient:
```
packetbuf_set_attr(PACKETBUF_ATTR_MAC_SEQNO, x);
packetbuf->seqno = x;
```
Of course, queuebuf should store the whole struct (instead of attributes and addresses).

simonduq:
Direct access to the fields instead of setters and getters sounds good to me here. But I'm a still worried with all the extra fields, needed only at parsing, that will get multiplied by the number of queuebufs in the system. Other opinions on this would be welcome :)

kkrentz:
Actually, I am also interested in parsing and creating frames within interrupts because I want to secure acknowledgement frames, too. Maybe, we don't need any additional attributes for that since we already have source, destination, sequence number, etc.. If we need any additional ones, we could add a second optional parameter to `framer802154_peek_parse`.

Another problem area I see is that the `queubuf` copies lots of data when moving a packet between the `queuebuf` and the `packetbuf`. According to my experiments, this results in a noticeable energy consumption, especially when using CSMA + ContikiMAC + phase optimization. It could be more lightweight if the packetbuf stored a pointer to the current packet. This pointer could then be set by `queuebuf` (instead of copying all data). 4f2fa1b52a1548df47a2a4a09b2846484625f9f4 already works towards this enhancement.

Unfortunately, the Rime stack currently fails, so I will try to figure out the cause.

simonduq:
This is getting so complex that it will be difficult to review. I did review carefully earlier versions but this one has changed massively, haven't yet gone through it. Questions:
* Does the current version support framing and parsing from interrupt?
* Is there any inter-dependency with https://github.com/contiki-os/contiki/pull/1146?
* Can you somehow split this PR in several smaller ones?

simonduq:
@kkrentz , ping?

kkrentz:
* Currently, there is no support for framing and parsing from interrupts. I started to work on this, but figured out that the rime stack does stuff that was incompatible with my design. So I have to start over.
* There is some interdependency with #1146.
* I can, e.g., cut the changes to llsec802154.h.

simonduq:
The problem is that there have two conflicting goals here: (1) parse/create directly to/from packetbuf (as a simplification and optimization) and (2) be still able to operate without packetbuf (for generality and in particular for frame parsing/creation from ISR). I'm very keen on the latter.

It seems to me that only a heavy redesign of packetbuf, with direct field access rather than attributes, can (maybe) achieve both. Is this what you're planning to do?

kkrentz:
My idea was to parse/create to/from a struct of packetbuf attributes. So, if you call create or parse from within an ISR, you can pass a pointer to a struct that does not interfere with the packetbuf (e.g. pointer to a local variable). As far as I remember, your objection to this was that we may have to add attributes and hence increase RAM usage, but I think we can even reduce RAM usage by using appropriate data types. A third goal may be to avoid copying within the queuebuf.

simonduq:
I agree (right, one problem with attributes is that we don't want to carry all framing information in every queuebuf)

kkrentz:
That is also true. We may want to divide the struct into two parts: things that have to be stored in queuebuf and things that are only needed when calling create/parse. However, what I actually meant was that the queubuf currently copies all packetbuf attributes one by one instead of simply adding a pointer to this data to some list.

simonduq:
I also agree with this.

kkrentz:
Finally, TSCH can also start streaming frames (at the cost of two additional packetbuf attributes ...). The way I did it was to enable ISR routines to change the location of packetbuf data. This is how it looks in action:
```c
struct packetbuf local_packetbuf;
struct packetbuf *actual_packetbuf;`

/* distort pointer to packetbuf data */
actual_packetbuf = packetbuf;
packetbuf = &local_packetbuf;

/* call packetbuf functions as usual */

/* restore pointer to packetbuf data */
packetbuf = actual_packetbuf;
```

Getting rid of `frame802154.c` completely is hard because it is being used by some old RDC Drivers and platforms, making it difficult to test modifications. Therefore, I suggest keeping it for the time being.

darbyShaw:
Anything I can contribute w.r.t #1904 ?

kkrentz:
We could add usage examples and adapt the cmd-broker to support beacon frames, too. Anyway, I would like to know if @simonduq agrees with these changes before putting more effort into this.

simonduq:
This PR contains many things and affects critical system functions.
If there a chance it can be broken up in several smaller, easier to handle PRs, with distinct functionality?

kkrentz:
To reduce the things that this PR affects, I suggest merging #1703 ;). Additionally, I could add my "streaming framer" as an additional framer, rather than overwriting the existing `framer_802154`. The deletion of the existing `framer_802154` and changes to TSCH can then become part of a new PR.

simonduq:
OK so let's wait for https://github.com/contiki-os/contiki/pull/1703 first.
I recommend against a separate "streaming framer", in particular if the end goal is to replace. Probably not worth the effort.

PR #884, opened 2014-12-01
Ensure that the data in ringbuff is accessed in the right order

cmorty:
This fixes a potential issue with concurrency and the compiler being allowed to reorder instructions in the ringbuff (The whole topic is quite interesting and caused quite some discussions in our working group). Unfortunately this causes msp430-gcc 4.7.0 to produce worse code (for some reason the optimizer does not detect, that a register isn't needed anymore and does not reuse it, therefore needing an additional instruction). Nonetheless the side effects caused by this in the future may become quite ugly.

### DIFF
--- a/core/lib/ringbuf.c
+++ b/core/lib/ringbuf.c
@@ -38,6 +38,7 @@
  */
 
 #include "lib/ringbuf.h"
+#include <sys/cc.h>
 /*---------------------------------------------------------------------------*/
 void
 ringbuf_init(struct ringbuf *r, uint8_t *dataptr, uint8_t size)
@@ -63,8 +64,15 @@ ringbuf_put(struct ringbuf *r, uint8_t c)
   if(((r->put_ptr - r->get_ptr) & r->mask) == r->mask) {
     return 0;
   }
-  r->data[r->put_ptr] = c;
-  r->put_ptr = (r->put_ptr + 1) & r->mask;
+  /*
+   * CC_ACCESS_NOW is used because the compiler is allowed to reorder
+   * the access to non-volatile variables.
+   * In this case a reader might read from the moved index/ptr before
+   * its value (c) is written. Reordering makes little sense, but
+   * better safe than sorry.
+   */
+  CC_ACCESS_NOW(uint8_t, r->data[r->put_ptr]) = c;
+  CC_ACCESS_NOW(uint8_t, r->put_ptr) = (r->put_ptr + 1) & r->mask;
   return 1;
 }
 /*---------------------------------------------------------------------------*/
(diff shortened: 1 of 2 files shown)


simonduq:
I would say that given the issue seems to affect only `ringbuf.c`, a strong solution (working with any compiler) should be preferred over a flexible and easy-to-use one (like the current `ACCESS_ONCE`). How about the following options:
* do not add anything to `cc.h`, but a simpler `ACCESS_BYTE_ONCE` defined in `ringbuf.c`
* keep the macro in `cc.h` but pass the type as parameter as suggested by @cmorty

simonduq:
@cmorty, thoughts?

cmorty:
@simonduq I'm still in favor of my solution: Add a ACCESS_NOW(type, variable) to cc.h, as it's a generic issue. Also, as in the Linux-Kernel, it allows to write more efficient code by removing unnecessary volatile declarations and only placing memory barriers where needed.

simonduq:
I'm still ok with that solution. If you update the PR (adding the type to the macro) I'll schedule this for merging in a month if nobody opposes.

nfi:
What do you think about the name CC_ACCESS_NOW as suggested earlier in the discussion? Most macros in sys/cc.h has CC as prefix (CC_INLINE, CC_ASSIGN_AGGREGATE, etc).

cmorty:
@nfi I really don't care. If you think that it's a good Idea I'll change it.

nvt:
Looks OK too me, but I would prefer <code>#ifndef ACCESS_NOW ... #endif</code> around the definition, which is what the other macro definitions have. For the naming, I think that either way is OK -- there is precedent for having no "CC_" around macros that can be seen to only have one possible definition, and this can be changed to "CC_" naming later if we want to.

core/lib/ringbuf.c
```diff
@@ -84,8 +92,17 @@ ringbuf_get(struct ringbuf *r)
...
+    /*
+     * ACCESS_NOW is used because the compiler is allowed to reorder 
+     * the access to non-volatile variables.
+     * In this case the memory might be freed and overwritten by 
+     * increasing get_prt before the value was copied to c.
+     * Opposed to the put-opertation this would even make sense, 
+     * because one the register used for mask can be reused to save c 
+     * (on some architectures).
+     */
+    c = ACCESS_NOW(uint8_t,r->data[r->get_ptr]);
```
nvt: There should be a space after the comma here.

core/lib/ringbuf.c
```diff
@@ -84,8 +92,17 @@ ringbuf_get(struct ringbuf *r)
      most platforms, but C does not guarantee this.
   */
   if(((r->put_ptr - r->get_ptr) & r->mask) > 0) {
-    c = r->data[r->get_ptr];
-    r->get_ptr = (r->get_ptr + 1) & r->mask;
+    /*
+     * ACCESS_NOW is used because the compiler is allowed to reorder 
+     * the access to non-volatile variables.
+     * In this case the memory might be freed and overwritten by 
+     * increasing get_prt before the value was copied to c.
```
nvt: get_prt -> get_ptr

core/lib/ringbuf.c
```diff
@@ -84,8 +92,17 @@ ringbuf_get(struct ringbuf *r)
...
+     * Opposed to the put-opertation this would even make sense,
```
nvt: opertation -> operation

core/lib/ringbuf.c
```diff
@@ -84,8 +92,17 @@ ringbuf_get(struct ringbuf *r)
...
   if(((r->put_ptr - r->get_ptr) & r->mask) > 0) {
-    c = r->data[r->get_ptr];
-    r->get_ptr = (r->get_ptr + 1) & r->mask;
+    /*
+     * ACCESS_NOW is used because the compiler is allowed to reorder 
+     * the access to non-volatile variables.
+     * In this case the memory might be freed and overwritten by 
+     * increasing get_prt before the value was copied to c.
+     * Opposed to the put-opertation this would even make sense, 
+     * because one the register used for mask can be reused to save c
```
nvt: "one" seems out of place here.

PR #890, opened 2014-12-02
Cleanup packetbuf: use rime attributes and addresses only when rime is e...

simonduq:
...nabled. Saves 32 byte per queuebuf in the non-rime case.


aguirrem:
I was just looking at drivers in the AVR platform but it appears I was looking at some old versions of the drivers.

Although maybe useless I am still curious how would one pass the number of re-tries a message took re-transmit w/o those buffer attributed?

simonduq:
This is done through the packet sent callbacks `mac_call_sent_callback`

PR #895, opened 2014-12-03
Do not use rpl_add_dag when RPL is used with a single dag per instance

simonduq:
No other purpose than save a little bit of space


nvt:
:+1:

PR #903, opened 2014-12-10
avoid updating routes that already are correct

joakimeriksson:
This PR make the routing table check that the nexthop we are trying to add is different than the nexthop already   in the routing table for the specific route entry. If that is the case it skips doing both a remove and an add.


nvt:
:+1:

PR #904, opened 2014-12-10
moved ETX storage variable from RPL to ds6-nbr

joakimeriksson:
This PR is for moving where the ETX is stored. Today it is in the RPL parents which works well until a global repair is made - then the ETX is gone - which make the global repair do a very bad job at repairing and finding the best parent. By moving to ds6-nbr which is then stored as long as the neighbor is there - independent of RPL state - this information will be kept during the global repair and RPL can pick the best parent with up-to-date information on ETX.


nvt:
:+1:

PR #906, opened 2014-12-10
fixed so that trickle timer and route lifetime rpl config is updated at global repair

joakimeriksson:
This PR makes sure that nodes in the RPL network pick up new configuration of trickle timers and route default lifetime at a global repair. This makes the whole network run the same settings as the RPL Root when it is changing the configuration (I have had problems getting all the nodes in a network to update timers and route-lifetime when updating the RPL Root with new config in a already deployed network).


nvt:
:+1:

PR #909, opened 2014-12-15
Updated cc2538-bsl submodule to version 1.1 (ff9182d).

JelmerT:
Bumping the cc2538-bsl submodule to the latest version.

I intended to include the `feature-readsize` branch (jelmert/cc2538-bsl@df1799a5ea6c9802345672a1b7a2e8e1486b1092) into this version. That branch includes a better recognition of the attached chip (exact model, die revision, sram and flash size) and a more correct (and safer) check if the backdoor enable bit is set. But I'm just not getting around to testing and finalizing it. I also only have the 512KB version of the chip, so I need some more people to try the code on other chip sizes.

But for now this version does include the following improvements:
- Support for Python3
- Fixed a nasty bug that would make the synch request fail the first time because of a buffer not being flushed
- Added support to use DTR and RTS pins to drive the bootloader_enable and reset pins on platforms that support it (standard enabled, no extra flags needed). This enables real 1 click uploading!
- Versioning system
- Possibility to set the secondary IEEE address
- General code clean-up
- Added error codes on exit

Thanks to @bradjc and @hexluthor for the commits.


JelmerT:
Is there anything holding back this PR? It's a fairly straight forward update of a submodule.

g-oikonomou:
I should have assigned this to myself long ago, not sure how I missed it. I am in fact playing around with CC2538 kit over the next few days, so I'll test and merge

JelmerT:
Thanks! Let me know if you run into trouble.

g-oikonomou:
:+1: This worked very nicely on OS X and I am being told by colleagues that it also works on Linux.

Any particular reason why we're not updating to JelmerT/cc2538-bsl@d6711e24 ?

JelmerT:
Nope no reason (I made this PR before pulling jelmert/cc2538-bsl@d6711e2).

Included the last commit into this PR.

PR #910, opened 2014-12-16
Allow disabling of DIO suppression as required by RFC

laurentderu:
The special value 0 of DIO Redundancy means that DIO suppression mechanism is disabled instead of triggered immediately as currently done.

<pre>
         In RPL, when k has the value
         of 0x00, this is to be treated as a redundancy constant of
         infinity in RPL, i.e., Trickle never suppresses messages.
</pre>


nvt:
:+1:

PR #911, opened 2014-12-16
Reset nopath_received flag when new DAO is received

laurentderu:
When a no_path DAO is received, an internal flag is set in the parent structure to not set again the DAO lifetime in case a redundant no-path DAO is received. However when a non no-path DAO is received the flag is not reset. This cause that all subsequent no-path DAO are ignored


nvt:
:+1:

PR #912, opened 2014-12-16
Add DODAG lifetime, update of parent rank when switching dag and multi dodag non-regression test

laurentderu:
Currently there is no automated way to remove obsolete DODAG in a node, if a node has seen MAX_DODAG_PER_INSTANCE DODAGS all new DODAG are ignored, the only way to remove the old DODAG is to reset the node.

This PR add a lifetime for each DODAG, the lifetime is reset each time a DIO advertising that DODAG is heard. When a DODAG is not updated since RPL_CONF_DAG_LIFETIME the DODAG is removed from the list. Note that the currently joined DODAG is immunized from this.

### DIFF
--- a/core/net/rpl/rpl-conf.h
+++ b/core/net/rpl/rpl-conf.h
@@ -122,6 +122,18 @@
 #endif /* RPL_CONF_DEFAULT_ROUTE_INFINITE_LIFETIME */
 
 /*
+ * Maximum lifetime of a DAG
+ * When a DODAG is not updated since RPL_CONF_DAG_LIFETIME times the DODAG
+ * maximum DIO interval the DODAG is removed from the list of DODAGS of the
+ * related instance, except if it is the currently joined DODAG.
+ */
+#ifdef RPL_CONF_DAG_LIFETIME
+#define RPL_DAG_LIFETIME                    RPL_CONF_DAG_LIFETIME
+#else
+#define RPL_DAG_LIFETIME                    3
+#endif /* RPL_CONF_DAG_LIFETIME */
+
+/*
  * 
  */
 #ifndef RPL_CONF_DAO_SPECIFY_DAG
--- a/core/net/rpl/rpl-dag.c
+++ b/core/net/rpl/rpl-dag.c
@@ -1060,6 +1060,7 @@ rpl_add_dag(uip_ipaddr_t *from, rpl_dio_t *dio)
       rpl_move_parent(previous_dag, dag, p);
     }
   }
+  p->rank = dio->rank;
 
   /* Determine the objective function by using the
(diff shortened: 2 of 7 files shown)


nvt:
I think that it is good to have a GC on unused DODAGs, but the comment describing this option should say why it exists, and in what unit the life time is. Is a uint32_t necessary to store a life time of this sort?

laurentderu:
Comments updated accordingly. The lifetime should be a uint32_t, uint16_t would give us only 65535 seconds (2/3rd of a day), which could be too short for DODAG with long DIO interval.

laurentderu:
Sky rom overflow is solved after rebase, travis is now happy.

simonduq:
Looks good but isn't the default lifetime a bit short? 20 min is close to the default longest DIO period, meaning a node with very few neighbors could easily end up freeing a DODAG that is still active, right?

laurentderu:
Indeed, a node could free a usable dodag because it has missed a few DIO. I could update the timeout to a larger value, let say one hour, or derive the timeout from the default max interval times a constant (maybe 3). What do you think ?

simonduq:
I'm equally fine with a longer time such as 1h or some multiplicator of the max interval of the instance (learnt from DIO rather than the compile-time default default)

simonduq:
@laurentderu, if you find the time to fix this (just set a longer lifetime) and rebase, I'll merge happily after a month!

laurentderu:
The longer time will not work if the DIO interval of the second DODAG is longer than it, so the only safe solution is to have a multiplicator of the max interval of the DODAG. I'm validating the modification and will update the PR.

simonduq:
sounds great!

simonduq:
any update @laurentderu?

laurentderu:
I updated the code, now the DAG lifetime is linked to the maximum DIO interval.

I added a non-regression test to validate multi-dodag support and dodag lifetime, it successively creates and removes 3 RPL Root and check that there is still connectivity afterwards.

However this non-regression test let me discover another issue related to multi-dodag : If a known parent switch from one dodag to another (new or existing), the rank of that parent is not updated and still the rank in its former dodag. This can lead to dead-lock in part of the tree as a parent with infinite rank in a previous dag can not become a parent in the new dag as the old rank is unacceptable.

simonduq:
OK, do you agree this is still good to merge anyway?
Maybe you should open an issue with the problem described here?

laurentderu:
I forgot that a change in the list of commits is not visible. I already added the fix and the non-regression test in this PR.

simonduq:
Looks good, waiting a little more before merging so people have time to comment on the additional fix

simonduq:
merging now

PR #914, opened 2014-12-16
Send no-path DAO only when the dag has a preferred parent

laurentderu:
In the current code, the No-Path DAO is send, even when the the node has no preferred parent (which happens when it first joins a DODAG), this cause a malformed DAO to be send.

Updated 3/3: The previous PR comment was completely wrong (but the title was right)


nvt:
:+1:

PR #915, opened 2014-12-16
Allow rpl default route lifetime to be infinite

laurentderu:
In high density network with DIO suppression activated it may happens that a node will never sent a DIO once the DIO interval becomes high as it has heard DIO from many neighbors already. As the default route to the preferred parent has a lifetime reset by receiving DIO from the parent, it means that the default route can be destroyed after a while.
This PR add the possibility to create default route with infinite lifetime, thus securing the upstream route. The default value is to not enable it though.
(It should be noted that TinyOS uses infinite lifetime for upstream routes in RPL)

### DIFF
--- a/core/net/rpl/rpl-conf.h
+++ b/core/net/rpl/rpl-conf.h
@@ -106,6 +106,22 @@
 #endif /* RPL_CONF_MAX_DAG_PER_INSTANCE */
 
 /*
+ * RPL Default route lifetime
+ * The RPL route lifetime is used for the downward routes and for the default
+ * route. In a high density network with DIO suppression activated it may happen
+ * that a node will never send a DIO once the DIO interval becomes high as it
+ * has heard DIO from many neighbors already. As the default route to the
+ * preferred parent has a lifetime reset by receiving DIO from the parent, it
+ * means that the default route can be destroyed after a while. Setting the
+ * default route with infinite lifetime secures the upstream route.
+ */
+#ifdef RPL_CONF_DEFAULT_ROUTE_INFINITE_LIFETIME
+#define RPL_DEFAULT_ROUTE_INFINITE_LIFETIME                    RPL_CONF_DEFAULT_ROUTE_INFINITE_LIFETIME
+#else
+#define RPL_DEFAULT_ROUTE_INFINITE_LIFETIME                    0
+#endif /* RPL_CONF_DEFAULT_ROUTE_INFINITE_LIFETIME */
+
+/*
  * 
  */
 #ifndef RPL_CONF_DAO_SPECIFY_DAG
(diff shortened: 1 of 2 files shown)


nvt:
Is this the solution of this PR preferable to having the user sets a default lifetime that is considered suitable for the network? The preprocessor conditionals are not that great to add for the purpose of having an infinite lifetime.

laurentderu:
The RPL route lifetime is used for the downward routes and for the default route, so if you increase the route lifetime to make the default route more resilient you may have obsolete downward routes in your tree for a long time. (And, nothing will guarantee you that your default route will be refreshed with a long lifetime if your network is stable and there always is DIO suppression).

Maybe the best solution is to always set an infinite lifetime for default routes created by RPL.

nvt:
I think that it is good to preserve the ability for the DODAG root to determine the lifetime, so it should not always be set to infinite lifetime.

I can accept this PR, but the implementation should preferably be without preprocessor conditionals; for example, 
<code> uip_ds6_defrt_add(from, RPL_DEFAULT_ROUTE_INFINITE_LIFETIME ? 0 :
RPL_LIFETIME(instance, instance->default_lifetime));</code>.

Furthermore, it would make it easier for the users to understand the option if you include the main points of this discussion in the comment above the option.

laurentderu:
Oups, I did the requested modification but I forgot to push it :) Here it is and rebased

simonduq:
Excellent :+1:

core/net/rpl/rpl-conf.h
```diff
@@ -106,6 +106,22 @@
 #endif /* RPL_CONF_MAX_DAG_PER_INSTANCE */
 
 /*
+ * RPL Default route lifetime
+ * The RPL route lifetime is used for the downward routes and for the default
+ * route. In high density network with DIO suppression activated it may happens
```
simonduq: "in high density network" -> "in a high density network"
happens -> happen (no "s")

core/net/rpl/rpl-conf.h
```diff
@@ -106,6 +106,22 @@
...
+ * that a node will never sent a DIO once the DIO interval becomes high as it
```
simonduq: sent -> send

core/net/rpl/rpl-conf.h
```diff
@@ -106,6 +106,22 @@
...
+ * has heard DIO from many neighbors already. As the default route to the
+ * preferred parent has a lifetime reset by receiving DIO from the parent, it
+ * means that the default route can be destroyed after a while. Setting the
+ * default route with infinite lifetime secure the upstream route.
```
simonduq: secure -> secures

simonduq:
a number of typos in the comment would need fixing though

nvt:
@laurentderu It looks like you have updated the PR, but there hasn't been any comment about this yet. Have you finished modifying it?

laurentderu:
Yes, it's done. I was waiting for Travis but then I forgot about it :)

nvt:
:+1:

PR #917, opened 2014-12-16
Allow user configuration of RPL hop-by-hop option

laurentderu:
Although Hop-by-hop option is extremely useful, it is not mandatory in RPL. This PR allows to deactivate properly RPL HBH


nvt:
This looks OK, but the comment for the option should preferably specify the behavior a little bit in more detail; e.g., say that we still handle incoming hop-by-hop options from other nodes.

laurentderu:
Comment updated as asked

nvt:
:+1:

PR #918, opened 2014-12-16
Multi DODAG support

laurentderu:
Currently Contiki only updates rank of parent belonging to the active DAG, parents belonging to other DAGs keep their initial rank even when receiving DIO from them.

This PR adds a mechanism to update the rank of these parents and also prevent the DODAG Root to join another DAG. With #912, a node can be part of MAX_DODAG_PER_INSTANCE DODAGs and drops dead ones.

### DIFF
--- a/core/net/rpl/rpl-dag.c
+++ b/core/net/rpl/rpl-dag.c
@@ -741,6 +741,9 @@ rpl_select_parent(rpl_dag_t *dag)
 
   if(best != NULL) {
     rpl_set_preferred_parent(dag, best);
+    dag->rank = dag->instance->of->calculate_rank(dag->preferred_parent, 0);
+  } else {
+    dag->rank = INFINITE_RANK;
   }
 
   return best;
--- a/core/net/rpl/rpl-dag.c
+++ b/core/net/rpl/rpl-dag.c
@@ -1285,6 +1285,9 @@ rpl_process_dio(uip_ipaddr_t *from, rpl_dio_t *dio)
     }
   }
 
+  /* Parent info has been updated, trigger rank recalculation */
+  p->flags |= RPL_PARENT_FLAG_UPDATED;
+
   PRINTF("RPL: preferred DAG ");
   PRINT6ADDR(&instance->current_dag->dag_id);
   PRINTF(", rank %u, min_rank %u, ",


core/net/rpl/rpl-dag.c
```diff
@@ -741,6 +741,7 @@ rpl_select_parent(rpl_dag_t *dag)
 
   if(best != NULL) {
     rpl_set_preferred_parent(dag, best);
+    dag->rank = dag->instance->of->calculate_rank(dag->preferred_parent, 0);
```
simonduq: Don't we also need to set the dag rank to infinite if there is no best parent?

core/net/rpl/rpl-dag.c
```diff
@@ -1023,8 +1024,9 @@ rpl_add_dag(uip_ipaddr_t *from, rpl_dio_t *dio)
   PRINTF("\n");
 
   ANNOTATE("#A join=%u\n", dag->dag_id.u8[sizeof(dag->dag_id) - 1]);
-
-  rpl_process_parent_event(instance, p);
+  if(dag->rank != ROOT_RANK(instance)) {
+    rpl_process_parent_event(instance, p);
+  }
```
simonduq: Why is this needed? A comment explaining this would be good!

core/net/rpl/rpl-dag.c
```diff
@@ -1282,6 +1302,9 @@ rpl_process_dio(uip_ipaddr_t *from, rpl_dio_t *dio)
     }
   }
 
+  //Parent info has been updated, trigger rank recalculation
```
simonduq: please use `/**/` comments instead

core/net/rpl/rpl-dag.c
```diff
@@ -1100,6 +1107,19 @@ rpl_recalculate_ranks(void)
     }
     p = nbr_table_next(rpl_parents, p);
   }
+  if (!updated) {
```
simonduq: the extra space after `if` should be removed

simonduq: I don't understand what this does. When no parent needed update, we update the preferred parent of every instance and dag? Clarification needed here

core/net/rpl/rpl-dag.c
laurentderu: Indeed, that is missing, I will add it

core/net/rpl/rpl-dag.c
laurentderu: rpl_process_parent_event() update the rank of the node, assuming it's a RPL router and not a Root. All the other calls to rpl_process_parent_event() are either protected the same way (rpl_process_dio() or global_repair()) or is never called (rpl_recalculate_ranks())

I will add a comment

core/net/rpl/rpl-dag.c
laurentderu: The PR and its commits are quite old, I believe this code was necessary when rpl_parents table did not exist and only the current dag parents were updated.

core/net/rpl/rpl-dag.c
laurentderu: Actually no longer needed since fd97136fee495ac0a261dbff642423a45db0ba0c

core/net/rpl/rpl-dag.c
laurentderu: I confirm, this is no longer needed, all the parents are now checked.

laurentderu:
PR updated and obsolete code removed. I let it run on during this week-end and network seems stable and multi-dodag still working.

simonduq:
:+1:

nvt:
:+1:

PR #920, opened 2014-12-16
Support more than one pending slip message

laurentderu:
Currently, the slip buffer can contain only one pending packet, which means that if there is a burst of packets coming from the BR all but the first are discarded. This PR transforms the slip buffer into a circular buffer able to store an unlimited number of packets (as long as they fit into the buffer)

### DIFF
--- a/core/dev/slip.c
+++ b/core/dev/slip.c
@@ -81,7 +81,7 @@ enum {
  */
 
 static uint8_t state = STATE_TWOPACKETS;
-static uint16_t begin, end;
+static uint16_t begin, next_free;
 static uint8_t rxbuf[RX_BUFSIZE];
 static uint16_t pkt_end;		/* SLIP_END tracker. */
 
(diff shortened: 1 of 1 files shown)


simonduq:
Excellent feature! How thoroughly has this been tested?

laurentderu:
It has been used in 6LBR since quite a long time, so I think it's quite stable and it increase the packet reception rate quite significantly.
However, looking again at the diff more than a year later make me think that some refactoring could make the code more readable.

simonduq:
OK in that case I think you should just go ahead and merge that one when you're done reworking it

simonduq:
@laurentderu shall we merge this one?

neophob:
:+1:

laurentderu:
The code is not really optimized but I guess it's better in than lingering out...

alignan:
:+1:

simonduq:
The only thing that stops me from adding :+1: now is that SLIP is not covered by regression tests, so I'd like @laurentderu to confirm the version in this PR is a well-tested one.

alignan:
I think he already confirmed that, as it has been used in 6lbr

simonduq:
Right, sorry for some reason I assumed this had been rebased in between but this is not the case indeed

simonduq:
:+1:

laurentderu:
I will try to improve the slip-radio non-regression tests and submit a new PR for them soon

simonduq:
sounds great!

PR #922, opened 2014-12-17
Atmel 230bb: always increase the rx frame ringbuffer cursor in rf230_read()

atiselsts:
Avoids packet reordering in cases when RF230_CONF_RX_BUFFERS > 1.

Until now, the code moved the cursor forward to the next slot only if the reception of a packet was successful. That lead to continuous packet reception reordering after once seeing a corrupt packet. The patch simply moves to the next packet even if the current packet cannot be successfully received.

It also looks like that the RS230bb driver enables automatic CRC,  but does not check is on reception. I got a lot of corrupt packets up to Rime level. Probably can be worked around by enabling the software checksum. I did not attempt to verify or patch that.

### DIFF
--- a/cpu/avr/radio/rf230bb/rf230bb.c
+++ b/cpu/avr/radio/rf230bb/rf230bb.c
@@ -552,7 +552,19 @@ rf230_is_ready_to_send() {
 static void
 flushrx(void)
 {
+  /* Clear the length field to allow buffering of the next packet */
   rxframe[rxframe_head].length=0;
+  rxframe_head++;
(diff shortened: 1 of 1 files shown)


simonduq:
@malvira, can you take a look?
This does not pass the tests and hasn't raised much attention. If no more activity in the coming month, I'll close.

herjulf:
Well the patch is interesting and does not seem to cause any problems in our setups.
thanks

malvira:
Has been rebased and tests pass. herjulf has tested it out so I'll count that as a  +1 from a domain expert and merge.

PR #927, opened 2014-12-19
Fix: set uip_ext_len to 0 after dropping packet

siskin:
###### Bug description:
If outgoing packet with hop-by-hop header is dropped in `tcpip_ipv6_output()` function, uip_ext_len is not set to 0. It causes next packet with no hop-by-hop header (for example ICMP) to become corrupt. Bug can be easily observed with ipv6/rpl-udp example client. If client looses connection with parrent, some outgoing DIO messages become corrupt (depending on the frequency of outgoing UDP messages). 

###### Fix description:
Issue can be solved by adding:
```
uip_ext_len = 0;
```
before every return statement in `tcpip_ipv6_output()`. 

After comment on mailing list from Laurent Deru, `uip_ext_len = 0` is also added in other places where packet is discarded.

### DIFF
--- a/core/net/ipv6/multicast/smrf.c
+++ b/core/net/ipv6/multicast/smrf.c
@@ -80,6 +80,9 @@ mcast_fwd(void *p)
   UIP_IP_BUF->ttl--;
   tcpip_output(NULL);
   uip_len = 0;
+#if NETSTACK_CONF_WITH_IPV6
+  uip_ext_len = 0;
+#endif /*NETSTACK_CONF_WITH_IPV6*/
 }
 /*---------------------------------------------------------------------------*/
 static uint8_t
--- a/core/net/ip/tcpip.c
+++ b/core/net/ip/tcpip.c
@@ -549,12 +549,14 @@ tcpip_ipv6_output(void)
   if(uip_len > UIP_LINK_MTU) {
     UIP_LOG("tcpip_ipv6_output: Packet to big");
     uip_len = 0;
+    uip_ext_len = 0;
     return;
   }
 
   if(uip_is_addr_unspecified(&UIP_IP_BUF->destipaddr)){
     UIP_LOG("tcpip_ipv6_output: Destination address unspecified");
     uip_len = 0;
+    uip_ext_len = 0;
     return;
   }
 
@@ -592,6 +594,7 @@ tcpip_ipv6_output(void)
           PRINTF("tcpip_ipv6_output: Destination off-link but no route\n");
 #endif /* !UIP_FALLBACK_INTERFACE */
           uip_len = 0;
+          uip_ext_len = 0;
           return;
         }
 
@@ -644,6 +647,7 @@ tcpip_ipv6_output(void)
 #if UIP_CONF_IPV6_RPL
     if(rpl_update_header_final(nexthop)) {
       uip_len = 0;
+      uip_ext_len = 0;
       return;
     }
 #endif /* UIP_CONF_IPV6_RPL */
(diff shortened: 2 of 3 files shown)


core/net/ipv6/multicast/smrf.c
```diff
@@ -80,6 +80,9 @@ mcast_fwd(void *p)
   UIP_IP_BUF->ttl--;
   tcpip_output(NULL);
   uip_len = 0;
+#if NETSTACK_CONF_WITH_IPV6
+  uip_ext_len = 0;
+#endif /*NETSTACK_CONF_WITH_IPV6*/
```
simonduq: I don't think we need the guards given the file is under `core/net/ipv6`

simonduq:
Looks good, but can you please point out where exactly the packet becomes corrupt? Wouldn't it be possible to, instead of this fix, set `uip_ext_len` before creating any IPv6 packet?

siskin:
Description:
1. Application running on node sends outgoing UDP packet (with `uip_ext_len > 0`).
2. There is no route to target, so packet is discarded by `tcpip_ipv6_output()`, but `uip_ext_len` is not set to 0.
3. Next outgoing ICMP packet is corrupt.

Easiest way to replicate bug: run ipv6/rpl-udp example client with RPL, let it connect to RPL Border Router and start sending datagrams. Then switch off the RPL border router. 

Setting `uip_ext_len = 0` before creating any IPv6 packet is certainly possible. I proposed solution above because of 2 reasons:
* `tcpip_ipv6_output()` already sets `uip_ext_len = 0` when packet is sent successfully. It should also set it to 0 when packet is discarded (consistent behavior).
* IPv6 packets in Contiki are created in many places: ICMP, RPL. It's much larger change.

simonduq:
I gave a closer look and I agree this is the way to go. However, as @laurentderu pointed out on the ML, there are many other places where `uip_len` is reset but not `uip_ext_len`.

My take on it is the following:
* There is no issue with input, because all incoming IP traffic goes through `tcpip_input`, which always resets `uip_ext_len` before returning;
* The issue is in output, where we need to make sure that `uip_ext_len` is always reset regardless of success or failure. As the flow is rather complex, we'd better reset too often than too rarely. I believe it should be reset every time `uip_len` is rest in all `core/net/ip` and `core/net/ipv6` files. If you agree, then a few more `uip_ext_len = 0;` should be added in `uip_icmp6.c` and `uip6.c`.

Please also take a look at the comment I left in the code (argue against or address).

nvt:
Rather than trying to reset both variables in conjunction every time, I think a small macro abstraction could decrease the risk of forgetting to reset uip_ext_len, while also making the code easier to read. E.g., to have something like uip_clear_buf() that resets both variables.

simonduq:
I fully agree with @nvt

simonduq:
The naming sounds good to me but this should be upper-cased right?

nvt:
That's usually the case, but uIP has most of it's function macros in lower case. Either way is fine with me.

core/net/ipv6/multicast/smrf.c
siskin: Good point. I agree.

siskin:
I agree with @nvt.

laurentderu:
+1 for uip_clear_buf(), that will help to always reset the variables.

laurentderu:
Everyone agree for uip_clear_buf() ? @siskin, will you do the modification ? if not, I'll do it.

joakimeriksson:
:+1: Yes - go for the uip_clear_buf() - much nicer!

siskin:
@laurentderu please do the modification if possible, I don't have the necessary environment set up.

simonduq:
https://github.com/contiki-os/contiki/pull/1116 is now merged, now closing this one

PR #928, opened 2014-12-20
Concept : cooperative mt scheduler with events

marcas756:
- Non preemtive mt-scheduler (context switch during irq not allowed for this implementation)
- sending events from a contiki-process to a smt-thread works by  using smt_post in the contiki-process
- sending events from a smt-thread to a smt-thread works by using smt_post in the source smt_thread
- sending events from a smt-thread to a contiki-process works by using process_post / process_post_synch in the smt-thread
- smt-thread can be sent to bed with smt_sleep. This does only block the smt-thread calling sleep, not the whole system.
- an smt-thread has to exit on its own. Killing the thread from outside is not possible
- A broadcast of an event to all smt-threads can be achieved by setting the target thread to SMT_BROADCAST when invoking smt_post.
- Don't forget to call mt_init() and smt_init() before using this lib.
- polling threads from irq, main thread or other threads context by using smt_poll()
- pausing mt_threads, similar to PROCESS_PAUSE
- does not break current mt_thread implementation

Changed stack size for native linux

This pull request replaces #923

### DIFF
--- a/core/sys/mt.c
+++ b/core/sys/mt.c
@@ -64,7 +64,6 @@ mt_remove(void)
 void
 mt_start(struct mt_thread *thread, void (* function)(void *), void *data)
 {
-  thread->needspoll = 0;
   /* Call the architecture dependant function to set up the processor
      stack with the correct parameters. */
   mtarch_start(&thread->thread, function, data);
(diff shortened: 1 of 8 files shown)


cmorty:
It is probably the most cited unsupported feature in survey-papers mentioning Contiki. :)

marcas756:
Hello, please refer to pull request #934, as this pull request got closed. #934 works even more efficient and does not require any change in existing contiki code, except for native stack size. Also it does not require separate post functions.

PR #929, opened 2014-12-21
Delete old CoAP implementation

bradjc:
It seems that this implementation of CoAP in Contiki is no longer maintained in favor of the `er-coap` implementation. This commit removes the code to prevent further bit-rot.

You may also read between the lines here: I spend a good deal of time trying to get this example to work on the CC2538. After adding a `sbrk()` function and editing the linker script to get it to compile, I dove into the RFC to figure out why the REST server wasn't working with Copper (Firefox CoAP plugin). Turns out this implementation supports some old version of the CoAP spec which is quite different from the RFC.


nvt:
It's a good idea to remove the old implementation, but please modify this PR so that the tests work again, as suggested by @ppannuto.

bradjc:
Ok finally got travis to like this PR. Hopefully it's good to go now.

ppannuto:
@nvt looks like @bradjc fixed everything that I saw.

simonduq:
The initial idea of having a common REST interface that can be instantiated as HTTP or CoAP transparently was appealing, but as the code is largely outdated, I'm also in favor of removing it now.
:+1:

alignan:
Some leftovers:

grep -lr "WITH_COAP" .
./platform/mbxxx/contiki-conf.h
./examples/seedeye/powerswitch/Makefile
./cpu/mc1322x/doc/rest-tutorial.md

grep -lr "WITH_HTTP" .
./examples/seedeye/powerswitch/Makefile

Besides that :+1:

bradjc:
Updated. I deleted `/cpu/mc1322x/doc/rest-tutorial.md` in a separate to commit to explain why it was removed, and I updated `./examples/seedeye/powerswitch` in a separate commit because it should compile now (I tested it with a compiler that I have, not the PIC compiler) after updating the code to the newer er-coap API.

alignan:
:+1:

Issue #930, opened 2014-12-21
examples/hello-world does not compile/run for target c64


marcas756:
The hello world example does not compile. The problem seems to be caused by the current work on IPv6 .

Regression tests do not cover the hello world example, thus travis does not complain about this issue.
I tried to compile the webserver and mail example (also done by travis)... both get compiled for taget c64 

I think that it would be a good idea to add  hello-world to regression test for all platforms, as
this example is the starting point for most people. I think it should work out of the box if the build/bin tools are correctly installed.

 make TARGET=c64
  CC        ../../cpu/6502/net/ethernet-drv.c
../../cpu/6502/net/ethernet-drv.c(64): Error: Undefined symbol: `IPBUF'
../../cpu/6502/net/ethernet-drv.c(64): Error: Struct pointer expected
../../cpu/6502/net/ethernet-drv.c(64): Error: Struct/union has no field named `srcipaddr'
../../cpu/6502/net/ethernet-drv.c(64): Error: Illegal address
../../cpu/6502/net/ethernet-drv.c(64): Error: Incompatible pointer types
../../cpu/6502/net/ethernet-drv.c(91): Error: Incompatible pointer types
make: *** [obj_c64/ethernet-drv.o] Fehler 1

see #931
cc65 issue?

marcas756:
Hello,

i was only able to get hello-world up and running with modifying Makefile (IPV4 = 1) and 6502 main loop. 

In the main loop I removed all network stuff to get it up and running. But I did not check it in as I am afraid to break something else with this.

The same I had to do for following pull request to work on 6502 (also did not check the mentioned changes in) :  #934

I am working with the current cc65 main branch.

Regards marcas

marcas756:
Hello,

in the meantime I got one issue solved. 

VICE complained that it could not open a file:

Fsdrive: error - cannot open file. status $02

As ethernet requires a disk image to load "contiki.cfg" I built contiki with 

export C1541=c1541
make disk TARGET=c64

and setting

CONTIKI_WITH_IPV4 = 1

in the hello-world example Makefile.

But with the same result, No "Hello world!"

```
load"$",8

searching for $
loading
ready.
list

0 "contiki         " 00 2a
68   "contiki"          prg
1    "contiki.cfg"      seq
3    "cs8900a.eth"      seq
3    "lan91c96.eth"     seq
589 blocks free.
ready.
load"*",8,1

searching for *
loading
ready.
run

ready.
```

[...]

PR #931, opened 2014-12-21
Add hello-world for target c64 to regression tests

marcas756:
see #930
Build failed as on my system


marcas756:
Created Makefile.c64.defines for the hello-world example due to discussion about it in #930.
Now it should compile and run, if vice is setup correctly for networking. 

helloworld

examples/hello-world/Makefile.c64.defines
```diff
@@ -0,0 +1,2 @@
+DEFINES = WITH_CLIENT,WITH_DNS,WITH_PFS,WITH_LOGGING
```
oliverschmidt: This line is supposed to be not necessary. Please try to remove it.

examples/hello-world/Makefile.c64.defines
```diff
@@ -0,0 +1,2 @@
...
+CONTIKI_WITH_IPV4 = 1
```
oliverschmidt: If you check out the recipe `savedefines` then you'll notice that Makefile.<target>.defines are not general containers for Makefile variable settings but are rather intended to contain exactly one line saying `DEFINES = ...`. From that perspective your approach to configure IPv4 doesn't seem valid to me.

As already said several times Contiki unfortunately presumes that IPv4 vs. IPv6 is purely a property of the example but but a property of the example-target tupel. I personally consider this a bug.

examples/hello-world/Makefile.c64.defines
marcas756: At least I would keep

DEFINES = WITH_LOGGING

as one can not know that hello-world (or any example) requires networking as provided by additional hardware or by vice hardware  emulation.

At least there would be an output on screen with the reason for the problem (not memory critical for hello world I think).

On the other hand one can not assume that the user has networking support (either by hardware or by emulation). 

So either the emulator or the original C64 has to provide network support ... or one has to write its own C64 platform for contiki.

For hello-world I think that this Info is very useful (also with the increased memory usage for hello-world) :

helloworlderrorinfo

Just my opinion, but also can be removed for this PR if requested.

oliverschmidt: > At least I would keep `DEFINES = WITH_LOGGING`

Yes, I agree with you.

examples/hello-world/Makefile.c64.defines
marcas756: I am honest ... just tired it :-)
I just assumed it to be an extention of the Makefile in the example directory.

marcas756: I had a look on savedefines plus README-BUILDING.md and it mentions:

``` 
Beside TARGET= there's DEFINES= which allows to set arbitrary variables for the
C preprocessor in form of a comma-separated list. Again it is possible to avoid
having to re-type i.e. DEFINES=MYTRACE,MYVALUE=4711 by running make TARGET=esb
DEFINES=MYTRACE,MYVALUE=4711 savedefines. A file called Makefile.esb.defines is
saved in the project's directory containing the currently saved defines for the
ESB platform.
```
The savedefines build receipe just saves DEFINE provided by the make invocation.

Thus 
```
make  TARGET=c64 DEFINES=WITH_LOGGING"\n"CONTIKI_WITH_IPV4=1  savedefines
```
would also deliver the above commented lines.

O.K. this is a little bit far-fetched, so you are right ... also doesn't seem valid to me now.

One problem I got with the current platform implementation for c64 is, that it assumes that the user wants to use it on a c64 with ethernet interface. The hello-world example does not require any ethernet support in my opinion.

Thus I recommend to add another define "WITH_ETH" or similar, to be able to enable the ethernet support explicitly.

I am quite  sure that this problem does not apply to platform c64  only ...

Regarding "CONTIKI_WITH_IPV4 = 1" : I will try to fix this one and hope that it does not break other platforms.

simonduq:
I have nothing against this, will merge in a month unless someone is against

simonduq:
(but please rebase to make sure the test still passes on the latest master!)

simonduq:
not rebased, closing atfer timeout

PR #934, opened 2014-12-29
Proposal: cooperative mt scheduler with events V2

marcas756:
# PR - Summary:

This pull request provides example code, located in examples/multi-threading/cmt-scheduler.

Tested for platforms:
- native linux            OK
- c64                          OK
- sky (cooja)              OK
- micaZ (cooja)          broken (also multi-threading example does not work properly) ... nice feature: mt_exit on return
- cooja (cooja)           broken (mtarch.c,h  is not  implemented properly) #1015

Intertarget portable part is located in cmt.c and cmt.h.

Some modifications were made to get it up and running for C64 as second test platform beside native linux (also see discussion in #930 and #931)
 
- Non preemtive mt-scheduler (context switch during irq not allowed for this implementation)
- In this implementation the cmt thread is derived from contiki process, thus a thread uses the contiki process scheduler directly  instead of an own mt scheduler as contiki process.
- As a cmt instance is also a process instance in this case, process post/post_sync can be used directly by processes and cmt_threads to post events to each other.
- This example also shows the close logical relationship between protothreads and usual cooperative threads. Think twice if you really need context switching or not!
-  For this special example it would be an advantage if process functions would use a void pointer for process passing, to avoid explicit casting on application level.
- It would be an advantage to be able to define thread specific stack sizes to avoid unnecessary RAM usage. Maybe by using cpu min stack size and multiplier? ... to be discussed

cmt_thread

replaces #928


marcas756:
Currently target c64 seems to be broken for main branch. For test purposes I modified contiki-main for c64 (removed network stuff), but did not check it in.

Maybe we are out of stack (> 256 Bytes)  for c64? ... Did not check (debug)  it in detail yet.
Also hello-world does not work for target c64.

Proof of concept for cmt_thread on C64:

c64_event_poll_test
c64_event_test
c64_event_pause_test

marcas756:
Gets it running for C64 as discussed in #931 and #930

So currently working targets are native linux and C64

cmt

At this point the PR also allows multiple "subprojects" for C64 disk image for the example projects.

Usually the "application" is called "contiki" in the C1541 disk file system.

The cmt-scheduler currently consists of three intependend tests that are all stored on the disk image now by name.

cmt-multiple

simonduq:
I can see a lot of effort went into this, thanks for submitting as a contribution here.
This hasn't raised any interest so far though, so I'll have to close in a month unless something happens

simonduq:
closing after timeout

rayui:
Dang! I need this! Can we resurrect?

simonduq:
Well the code is still in the original branch: `marcas756:PullReq-BASIC_MT_SCHEDULER_V2`
Feel free to open a new PR if you are willing to bring this back to life

marcas756:
I think that for this feature the current mt implementation needs to be streamlined for this (I am talking about mt_arch in this case).

Meanwhile I reduced it to native linux only ...(the contiki implementation with makecontext/getcontext/swapcontex). I had to touch  MTARCH_STACKSIZE for this ( 4096 -> 3*4096) .... can not imagine why ...

There were some issues with other architecures yet  ... see PR-Summary 

There is also still some work to do on contiki arch code to get it stable up and running for all architectures.

Shall I make a new PR with linux native only ? All code is located in examples/multi-threading/cmt-scheduler ... except the change in linux native (MTARCH_STACKSIZE  = 3*4096)

PR #945, opened 2015-01-17
Atmel AT86RF230 in avr-rcb: added missing macros and initialisations 

Tom435:
I added important modules in the Makefile.avr-rcb as well as some macros in contiki-conf.h and Makefile.avr-rcb to be able to compile applications for the avr-rcb.
To use the wireless interface it is necessary to initialize the NETSTACK_* protocols in /contiki-rcb-main.c.


platform/avr-rcb/contiki-rcb-main.c
```diff
@@ -109,21 +109,22 @@ init_lowlevel(void)
...
-  uint8_t error;
 
   PROCESS_BEGIN();
   
-  if((error = icmp6_new(NULL)) == 0) {
     while(1) {
       PROCESS_YIELD();
       
-#if NETSTACK_CONF_WITH_IPV6	  
+#if UIP_CONF_IPV6
```
nvt: NETSTACK_CONF_WITH_IPV6 is the normal way to check whether IPv6 is in use. Why was this changed?

Tom435:
You are right, that´s nonsense. I accidentally changed that macro because of a copy and paste mistake.

platform/avr-rcb/contiki-rcb-main.c
Tom435: You are right, that´s nonsense. I accidentally changed that macro because of a copy and paste mistake.

nvt:
:+1:

PR #946, opened 2015-01-21
CC2538 RF driver issues x3

e120guru:
When using the cc2538-rf driver without using radio duty cycling I discovered that if an error occurs, the radio is switched off, reinitialized, but never switched back on again, which killed all radio reception until the stack decided to switch the radio on again for any reason.

Also, the channel was reset to the default in this case, instead of what was last set. 

Thirdly, the radio was switched on whenever channel was changed, even if radio was currently off. This makes sleepy nodes switch on their radio of for instance using multichan.

* This patch stores the RF channel internally in the module and always reinitialize to the last set value.
* This patch only switch on radio when chaning channel iff the radio was on
* This patch switches radio back on after reinitializing

### DIFF
--- a/cpu/cc2538/dev/cc2538-rf.c
+++ b/cpu/cc2538/dev/cc2538-rf.c
@@ -964,17 +964,18 @@ PROCESS_THREAD(cc2538_rf_process, ev, data)
 
     /* If we were polled due to an RF error, reset the transceiver */
     if(rf_flags & RF_MUST_RESET) {
+      uint8_t was_on;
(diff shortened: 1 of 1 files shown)


g-oikonomou:
This is a good patch :+1: 

There is a minor style error in L978:
`if (was_on) {` (space before opening paren). Something you can fix?

e120guru:
Updated.... (waiting for Travis to finish)

g-oikonomou:
Thanks for addressing this so quickly :+1:

PR #950, opened 2015-01-29
Fix: Misconfigurations to compile for avr-zigbit platform

patoman007:
I had some reference errors for avr-zigbit platform.
I added to the makefile some necessary files and from contiki-avr-zigbit-main.c I commented the rime-stack initializations and modified the baud rate.


nvt:
:+1:

PR #951, opened 2015-02-01
Make the CC2538DK button sensor clearer to work with

hrzr:
It's now more convenient to add new buttons. If button is redirected to another port, this could doesn't need to be changed. The code itself now is more clear.


g-oikonomou:
Please, can you run this through uncrustify to fix code style and re-write it so the merge commit disappears? Thanks

hrzr:
I've uncrustified the code. Should I close this pull request?

hrzr:
Okay, I've got it. Thank you for the guidelines.

g-oikonomou:
I'm not convinced that the code is cleaner now, I think it was clear enough as it was.

Nevertheless, this pull does make the driver more generic, so I'll merge this :+1: Thanks

PR #953, opened 2015-02-03
added some documentation of native-border-router and slip-radio relationship

mcr:
This is just some documentation to clarify how these examples relate.

### DIFF
--- a/examples/ipv6/native-border-router/README.md
+++ b/examples/ipv6/native-border-router/README.md
@@ -1,21 +1,28 @@
-The native border router connects a TTY with an ethernet TUN device
-via the RPL protocol.  This works with the ../slip-radio example to permit
-a Linux host to have a 802.15.4 radio connected via a serial interface.
-What's on the SLIP interface is really not Serial Line IP, but SLIP framed
-15.4 packets.
+This code connects a 802.15.4 radio over TTY with the full uIPv6 stack of
+Contiki including 6LoWPAN and 802.15.4 framing / parsing. The native border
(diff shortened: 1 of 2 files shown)


joakimeriksson:
Great to see that we get documentation on the native border router and slip-radio. Some comments on the commands and on the text:

Commands from stdin to the border-router:
?M is used for requesting the MAC address from the radio in order to use it for uIP6 and its stateless address auto configuration of its IPv6 address. This will make the native border router have the address that correspond to the MAC address of the slip-radio. (response is !M<MAC> from the slip-radio)

?C is used for requesting the currently used channel for the slip-radio. The response is !C with a channel number (from the slip-radio).

!C is used for setting the channel of the slip-radio (useful if the motes are using another channel than the one used in the slip-radio).

Then I do think that it might make sense to expand the text on connecting TTY to a TUN interface. Maybe we could say something like "... connects a 802.15.4 radio over TTY with the full uIPv6 stack of Contiki including 6LoWPAN and 802.15.4 framing / parsing. The native border router also acts as a RPL Root and handles the routing and maintains the RPL network. Finally the native border router connects the full 6LoWPAN/RPL network to the host (linux/os-x) network stack making it possible for applications on the host to transparently reach all the nodes in the 6LoWPAN/RPL network."

BTW: we have several improvements on the native-border router that I am hoping to commit fairly soon.

mcr:
updated document with suggestions from joakimeriksson

nvt:
:+1:

PR #954, opened 2015-02-05
Fix: Cooja HC packet analyzer fix for SAC and DAC fields

TheGeorge:
This pull request fixes a wrong interpretation of some fields in the packet analyzer of Cooja, when dealing with a compressed header.
According to RFC 6282 section 3.2 (https://tools.ietf.org/html/rfc6282#section-3.2) a 0 in the address compression fields (SAC and DAC) indicates a stateless compression, not a stateful one.


nvt:
The modification looks reasonable, but the commit message says "SIC" and "DIC" fields. Can you fix this first?

By the way, it is unfortunate that several of the files in the same directory as the modified one are missing a copyright and license statement. But this is something that can be addressed only by the authors of those files in separate PRs.

TheGeorge:
I fixed the commit message.

nvt:
@TheGeorge Could you rebase so that the "Merge branch" commit is removed also?

TheGeorge:
I rebased the commit.

fros4943:
:+1:

PR #956, opened 2015-02-06
[z1] compilation error when z1 with specific reference number pattern attached

blemmens:
#820 introduced functionality that uses the 4 last digits of the built-in reference number as node_id. If however these 4 last digits start with a leading zero, the SERIALNUM macro will be interpreted as an octal number [here](https://github.com/contiki-os/contiki/blob/010a338630a78db15fbb269a199272180b7bacf8/platform/z1/contiki-z1-main.c#L227). If any of the other three last digits is a '9', a compilation error occurs.

This behaviour can be reproduced (without a z1 attached) in examples/hello-world with `make TARGET=z1 DEFINES=SERIALNUM=0789`.

This fix simply casts SERIALNUM to a hexadecimal value by prepending '0x'. As a side-effect, since we get rid of dec-hex base conversions, the last two bytes of node_mac will now match exactly with node_id when printed out.

### DIFF
--- a/platform/z1/Makefile.common
+++ b/platform/z1/Makefile.common
@@ -68,7 +68,7 @@ ifeq ($(HOST_OS),Darwin)
       ifneq (,$(REFNUM))
         # No device fo-und
         ifeq (,$(findstring und, $(REFNUM)))
-          CFLAGS += -DSERIALNUM=$(REFNUM)
+          CFLAGS += -DSERIALNUM=0x$(REFNUM)
         endif
       endif
   endif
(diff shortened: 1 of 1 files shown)


alignan:
Hello! a fix has also been proposed at https://github.com/contiki-os/contiki/pull/826, however it is based in removing leading zeroes only, it does not takes care of hex/dec conversions, is this something you prefer to have? if so I can close my PR.

blemmens:
@alignan I seem to have overlooked your PR on this issue, sorry about that. 
I don't have any real preference, since we always burn the node IDs in advance. I just wanted to get rid of the compilation error.

I do like the idea of generating the node_id from a hardware id though, but IMHO the current approach with REFNUM has some drawbacks: you can only have a single node attached, and you need to rebuild every time to pass on REFNUM via the SERIALNUM C macro. When you already have a %.z1 file, `make TARGET=z1 %.upload` will not recompile your sources, and you end up with a wrong REFNUM flashed to the node.

nvt:
The first proposed fix for this problem (PR #826) has been merged, so I'm closing this one. @blemmens points out some drawbacks with the REFNUM generation, which could give reason to implement an alternative solution in a future PR, if someone would volunteer.

PR #958, opened 2015-02-11
change value of RPL_DIO_MOP_MASK 

kshpylon:
There is a overlap between RPL_DIO_MOP_MASK and RPL_DIO_PREFRENCE_MASK.
according to rfc 6550, RPL_DIO_MOP_MASK should be 0x38


nvt:
:+1:

PR #961, opened 2015-02-15
Tidy-up multicast documentation

g-oikonomou:
This pull:

* Creates a doxygen group and adds multicast files to it, so that documentation appears within the module hierarchy, rather than simply under files
* Documents plenty of functions, macros, constants, data types which were previously undocumented
* Re-structures some doxygen comments to improve structure and readability


nvt:
:+1:

PR #965, opened 2015-02-17
Mqtt contrib

jonnteolsson:
This pull adds:

* An engine for MQTT 3.1 under the `apps` directory
* An example use of the engine for the CC2538DK platform
* A travis test for the engine + example

The code has been tested on Srf06+CC2538EM with mosquitto as well as with IBM's quickstart service through NAT64 with 6lbr.

The engine supports MQTT QoS levels 0 and 1.

### DIFF
--- a/apps/mqtt/Makefile.mqtt
+++ b/apps/mqtt/Makefile.mqtt
@@ -1,0 +1,1 @@
+mqtt_src = mqtt.c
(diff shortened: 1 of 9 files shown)


g-oikonomou:
I've spent a long part of the day testing this with mosquitto over a CC2538EM USB BR. Works a treat and reacts very nicely to various error conditions (e.g. BR dies, broker dies, something listening on 1883 without being a broker)

:+1:

PR #883, opened 2014-12-01
Move MAX, MIN and ABS to contiki-macros

cmorty:
Based on the discussion in #878 I started moving MIN, MAX and ABS macros to an extra file.

PR #966, opened 2015-02-18
Move cc2538 startup-gcc.c to the CPU dir

g-oikonomou:
This pull proposes moving CC2538's `startup-gcc.c` to the CPU dir. This file contains startup code and initialises the vector table.

@adamdunkels suggested this move a few months ago, but I had some concerns about the various `#if` statements used to manage some of the interrupt handlers. Looks like they were unfounded, but please take a look if you will.

In the longer term, I think we should use weak symbols here for handlers. It should help get rid of those `#if`s altogether.

This pull also sneaks in a minor code style fix.

### DIFF
--- a/cpu/cc2538/startup-gcc.c
+++ b/cpu/cc2538/startup-gcc.c
@@ -30,11 +30,11 @@
  *  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 /**
- * \addtogroup cc2538dk
+ * \addtogroup cc2538
  * @{
  *
  * \file
- * Startup code for the cc2538dk platform, to be used when building with gcc
+ * Startup code for the cc2538 chip, to be used when building with gcc
  */
 #include "contiki.h"
 #include "reg.h"
@@ -73,13 +73,13 @@ void udma_err_isr(void);
 #define FLASH_CCA_BOOTLDR_CFG_ACTIVE_LEVEL 0
 #endif
 
-#if ( (FLASH_CCA_CONF_BOOTLDR_BACKDOOR_PORT_A_PIN < 0) || (FLASH_CCA_CONF_BOOTLDR_BACKDOOR_PORT_A_PIN > 7) )
+#if ((FLASH_CCA_CONF_BOOTLDR_BACKDOOR_PORT_A_PIN < 0) || (FLASH_CCA_CONF_BOOTLDR_BACKDOOR_PORT_A_PIN > 7))
 #error Invalid boot loader backdoor pin. Please set FLASH_CCA_CONF_BOOTLDR_BACKDOOR_PORT_A_PIN between 0 and 7 (indicating PA0 - PA7).
 #endif
 
-#define FLASH_CCA_BOOTLDR_CFG ( FLASH_CCA_BOOTLDR_CFG_ENABLE \
+#define FLASH_CCA_BOOTLDR_CFG (FLASH_CCA_BOOTLDR_CFG_ENABLE \
   | FLASH_CCA_BOOTLDR_CFG_ACTIVE_LEVEL \
-  | (FLASH_CCA_CONF_BOOTLDR_BACKDOOR_PORT_A_PIN << FLASH_CCA_BOOTLDR_CFG_PORT_A_PIN_S) )
+  | (FLASH_CCA_CONF_BOOTLDR_BACKDOOR_PORT_A_PIN << FLASH_CCA_BOOTLDR_CFG_PORT_A_PIN_S))
 #else
 #define FLASH_CCA_BOOTLDR_CFG FLASH_CCA_BOOTLDR_CFG_DISABLE
 #endif
--- a/cpu/cc2538/startup-gcc.c
+++ b/cpu/cc2538/startup-gcc.c
@@ -40,11 +40,12 @@
 #include "reg.h"
 #include "flash-cca.h"
 #include "sys-ctrl.h"
-#include "uart.h"
 
 #include <stdint.h>
 /*---------------------------------------------------------------------------*/
 extern int main(void);
+/*---------------------------------------------------------------------------*/
+#define WEAK_ALIAS(x) __attribute__ ((weak, alias(#x)))
 /*---------------------------------------------------------------------------*/
 /* System handlers provided here */
 void reset_handler(void);
@@ -62,6 +63,9 @@ void cc2538_rf_rx_tx_isr(void);
 void cc2538_rf_err_isr(void);
 void udma_isr(void);
 void udma_err_isr(void);
+void usb_isr(void) WEAK_ALIAS(default_handler);
+void uart0_isr(void) WEAK_ALIAS(default_handler);
+void uart1_isr(void) WEAK_ALIAS(default_handler);
 
 /* Boot Loader Backdoor selection */
 #if FLASH_CCA_CONF_BOOTLDR_BACKDOOR
@@ -83,22 +87,6 @@ void udma_err_isr(void);
 #else
 #define FLASH_CCA_BOOTLDR_CFG FLASH_CCA_BOOTLDR_CFG_DISABLE
 #endif
-
-/* Link in the USB ISR only if USB is enabled */
-#if USB_SERIAL_CONF_ENABLE
-void usb_isr(void);
-#else
-#define usb_isr default_handler
-#endif
-
-/* Likewise for the UART[01] ISRs */
-#if UART_CONF_ENABLE
-void uart0_isr(void);
-void uart1_isr(void);
-#else /* UART_CONF_ENABLE */
-#define uart0_isr default_handler
-#define uart1_isr default_handler
-#endif /* UART_CONF_ENABLE */
 /*---------------------------------------------------------------------------*/
 /* Allocate stack space */
 static unsigned long stack[512];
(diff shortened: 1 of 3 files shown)


bradjc:
:+1: 

I have a few out-of-tree CC2538 platforms and I haven't had the need to change startup-gcc.c.

alignan:
:+1: @adamdunkels ?

g-oikonomou:
Community seems happy. @bthebaudeau and @adamdunkels , how do you feel about this?

g-oikonomou:
@bthebaudeau @uknoblic @areaarea @hexluthor

How do you feel about this? I want to get it merged and dusted.

areaarea:
Fine by me.

g-oikonomou:
>  Fine by me.

That was fast!

g-oikonomou:
I think this may need a lil bit of work after #1031

bthebaudeau:
It would be good to still be able to customize stuff in a platform other than cc2538dk after this PR, so keeping `TARGET_STARTFILES` alongside `CPU_STARTFILES` in `cpu/cc2538/Makefile.cc2538` would be good, and the definitions in `startup-gcc.c` should be made `weak`. Or, even easier, `cpu/cc2538/Makefile.cc2538` could define `CPU_STARTFILES` to `TARGET_STARTFILES` if the latter is defined (or non-empty), and use its current value in your code as a default otherwise.

> I think this may need a lil bit of work after #1031

I don't think so. The final and merged version of that PR did not rename the ISR functions.

adamdunkels:
:+1: from me!

g-oikonomou:
OK I'll update this based on the above suggestions over the next few days. Thanks for the input

hexluthor:
:+1: and I like @bthebaudeau's suggestions too.

cpu/cc2538/startup-gcc.c
```diff
@@ -30,22 +30,23 @@
...
 #include "reg.h"
 #include "flash-cca.h"
 #include "sys-ctrl.h"
-#include "uart.h"
 
 #include <stdint.h>
 /*---------------------------------------------------------------------------*/
 extern int main(void);
 /*---------------------------------------------------------------------------*/
+#define WEAK_ALIAS(x) __attribute__ ((weak, alias(#x)))
```
bthebaudeau: This definition is not platform-specific, so it should be moved to a common place like `cc.h`, all the more it may be useful elsewhere.

g-oikonomou: It is toolchain-specific though, so putting it in `cc.h` would mean wrapping it inside an `#if` config block, and it would have to default to empty. The platform would then have to `#define` it, which would defeat the purpose

bthebaudeau: Are there other compilers than GCC and IAR supported by Contiki? Such definitions could be made in a common place for all supported compilers (for all platforms), and a `#error` could be issued there if an unsupported compiler is detected in the `#else`. The issue that you describe could also be true for a single platform using several compilers.

g-oikonomou: SDCC, cc65.

I think the extent of what you are suggesting would need a separate pull request. In terms of this one here, I think we should be OK with the local definition, or we can simply just remove the macro and put `__attribute__ ((weak, alias(default_handler)))` directly in the prototype.

cmorty: I agree with @bthebaudeau. `cc.h` is where compiler/tool-chains specific stuff should go.

g-oikonomou: How about something like this?

```
#if CC_CONF_WEAK_ALIAS
#define CC_WEAK_ALIAS(f) __attribute__ ((weak, alias(#f)))
#else /* CC_CONF_WEAK_ALIAS */
#define CC_WEAK_ALIAS(f)
#endif /* CC_CONF_WEAK_ALIAS */
```

and then go to the platform's `contiki-conf.h` and add this:

```
#define CC_CONF_WEAK_ALIAS 1
```

bthebaudeau: > SDCC, cc65.

These are indeed very different and do not support this feature.

> I think the extent of what you are suggesting would need a separate pull request. In terms of this one here, I think we should be OK with the local definition, or we can simply just remove the macro and put __attribute__ ((weak, alias(default_handler))) directly in the prototype.

Maybe. No need to remove the macro.

Nothing should be required in the platform code, not even a `CC_CONF_WEAK_ALIAS`. The availability only depends on the toolchain and on the CPU (core) family. Many different MCUs / SoCs / platforms use the same toolchain and CPU, e.g. with GCC ARM. Everything should be auto-detected, and not configured.

How about this in `cc.h`?
```
#if __STDC_VERSION__ >= 199901L
#define PRAGMA(x) _Pragma(#x)
#endif

#if __GNUC__
#define WEAK_ALIAS(alias, ref) __attribute__((__weak__, __alias__(#ref)))
#elif __ICCARM__ || __ICCSTM8__
#define WEAK_ALIAS(alias, ref) PRAGMA(weak alias=ref)
#endif
```

In this way, if `WEAK_ALIAS()` is used where unsupported, the build will simply fail, and all cases supporting it are auto-detected and make it available.

These macros could even be built on the function prototypes if position matters, or to avoid duplicating the function name, but I think this would be too much before we find a case where it's strictly required.

g-oikonomou: > SDCC, cc65

win32 too by the way

> How about this in `cc.h`?

I cannot see anything wrong with this, other than that traditionally we have not been checking for compiler family / version etc in `cc.h`, see for example how `CC_CONF_INLINE` is being used. This could also have been done automatically, but hasn't. I'm not sure about the original reasons though.

Would `__IAR_SYSTEMS_ICC__` not be sufficient?

bthebaudeau: > I cannot see anything wrong with this, other than that traditionally we have not been checking for compiler family / version etc in `cc.h`, see for example how `CC_CONF_INLINE` is being used. This could also have been done automatically, but hasn't. I'm not sure about the original reasons though.

There's probably no strong reason for this.

Using the automatic preprocessor definitions generated by the toolchains is the standard way of doing such things.

> Would `__IAR_SYSTEMS_ICC__` not be sufficient?

No, because there are IAR toolchains for many other CPUs (all defining `__IAR_SYSTEMS_ICC__`), and only ARM and STM8 support this feature according to IAR documentation. I don't know what CPU families are used with IAR in Contiki, but it does not matter: we need to make things future-proof.

bthebaudeau: Also, if a given platform needs to support several incompatible toolchains, you cannot avoid detecting the toolchain to adjust the compiler-related macros accordingly, even if you choose to put the definitions in the platform code like now.

g-oikonomou: Fair point, I'm not at all familiar with IAR to be honest. Are there any strong reasons against prefixing `PRAGMA` and `WEAK_ALIAS` with `CC_`?

bthebaudeau: > Are there any strong reasons against prefixing `PRAGMA` and `WEAK_ALIAS` with `CC_`?

No, it's just a matter of simplification, because it's plain obvious that it's compiler-related, like in #883. It does not really matter for me if one or the other is chosen.

adamdunkels: I may have missed something, but is there a reason why this should go into `cc.h`?

How many other files in the tree besides this one file use this?

g-oikonomou: > How many other files in the tree besides this one file use this?

Many. And I just pushed a broken update. I'll revert

core/sys/cc.h
```diff
@@ -111,6 +111,19 @@
...
+ */
+#if __STDC_VERSION__ >= 199901L
+#define CC_PRAGMA(x) _Pragma(#x)
+#endif
+
+#if __GNUC__
+#define CC_WEAK_ALIAS(alias, ref) __attribute__((__weak__, __alias__(#ref)))
+#elif __ICCARM__ || __ICCSTM8__
+#define CC_WEAK_ALIAS(alias, ref) CC_PRAGMA(weak alias=ref)
+#endif
```
adamdunkels: Why is this in `cc.h`? Seems to only be used in `startup-gcc.c`.

g-oikonomou: This was a failed attempt to do what was discussed above. I've reverted to the previous version of the pull, the entire discussion should be visible now

cpu/cc2538/startup-gcc.c
adamdunkels: How many? And where in the tree are they located?

g-oikonomou: Sorry, not many.

* `cpu/arm/stm32f103/startup-STM32F10x.c` uses `#define UNHANDLED_ALIAS`
* `cpu/stm32w108/hal/micro/cortexm3/stm32w108/crt-stm32w108.c` and `cpu/arm/stm32f103/startup-STM32F10x.c` also use aliases but without defining any macros

cmorty: Independently of in the number of places it is used, IMO it should go into `cc.h` because is belongs there. Waiting until those things are defined in more of one spot almost always causes a duplication, because people are to lazy to refactor the code where it is already in use. Contiki has enaugh examples of this.

adamdunkels: Ok. By comparison the other macros in `cc.h` are used throughout the `core/` tree and by some 30+ files (by a cursory count). 

In any case, if someone really wants to discuss adding CPU-specific stuff to `cc.h`, let's make a separate pull request with those changes instead of slowing down this one (which shouldn't need to touch any core files).

g-oikonomou: In that case, I have nothing further to change and this is ready for review/merge from my perspective.

adamdunkels: Awesome - :+1: from me then.

PR #967, opened 2015-02-18
Use local variable to store RF on/off state

### LINKED ISSUE #823: Problem found in code for the CC2538 platform
Hello everyone,

I noticed a small problem while I'm playing with Contiki on the CC2538 chip.

I've found that the current consumption during an udp packet transmission lasts too long. I've created an stackoverflow question in the hope of receiving answers about that but I found the problem myself. I wished to share with you so that the problem may be fixed in future release. This is a small fix but is very helpful if you are limited in current. The post is here http://stackoverflow.com/questions/26179093/contiki-udp-packet-transmission-duration-with-cc2538/26184677. There are equally some graphs of the current consumption which help you understand the problem. 

What do you think about this problem?

g-oikonomou:
This improves the "if we were off, turn back off" logic in the RF driver.

This will prevent the RF from lingering in the RX state for longer than intended, and will thus reduce overall average energy consumption.

This is in response to the discussion with @darkopetrovic - I'd appreciate if you could have a look and verify this fix.

Fixes #823

### DIFF
--- a/cpu/cc2538/dev/cc2538-rf.c
+++ b/cpu/cc2538/dev/cc2538-rf.c
@@ -80,7 +80,6 @@
 /* Local RF Flags */
 #define RX_ACTIVE     0x80
 #define RF_MUST_RESET 0x40
-#define WAS_OFF       0x10
 #define RF_ON         0x01
 
 /* Bit Masks for the last byte in the RX FIFO */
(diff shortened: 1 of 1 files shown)


darkopetrovic:
Thanks for the fix.

If the message was for me, I'll check that of course, but actually I'm not ready to migrate to the last version to test that on the hardware. But anyway visually the algorithm seems ok for me.

darkopetrovic:
Sorry for the delay :p 

I moved to the last version and checked the multicast transmission. We just need to put the `was_off` variable as global in the file to resolve the problem. 

Before:


After:

PR #968, opened 2015-02-18
Move CC2538 VDD/3 and On-Chip Temp sensors to the CPU dir

g-oikonomou:
Currently, we have a single module ("ADC sensor") in the platform dir. This handles the VDD/3 sensor, the On-Chip Temp sensor as well as the Smart RF ALS.

With this pull, I am proposing:

* Separating the single driver into 3 individual ones.
* VDD/3 and Temp are built-in the CC2538, so this pull will place those drivers in the cpu dir
* Example projects adjusted accordingly

### DIFF
--- a/cpu/cc2538/Makefile.cc2538
+++ b/cpu/cc2538/Makefile.cc2538
@@ -52,7 +52,7 @@ CONTIKI_CPU_SOURCEFILES += nvic.c cpu.c sys-ctrl.c gpio.c ioc.c spi.c adc.c
 CONTIKI_CPU_SOURCEFILES += cc2538-rf.c udma.c lpm.c
 CONTIKI_CPU_SOURCEFILES += dbg.c ieee-addr.c
 CONTIKI_CPU_SOURCEFILES += slip-arch.c slip.c
-CONTIKI_CPU_SOURCEFILES += i2c.c
+CONTIKI_CPU_SOURCEFILES += i2c.c cc2538-temp-sensor.c vdd3-sensor.c
 
 DEBUG_IO_SOURCEFILES += dbg-printf.c dbg-snprintf.c dbg-sprintf.c strformat.c
 
--- a/cpu/cc2538/dev/cc2538-sensors.h
+++ b/cpu/cc2538/dev/cc2538-sensors.h
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2015, Zolertia - http://www.zolertia.com
+ * Copyright (c) 2015, University of Bristol - http://www.bristol.ac.uk
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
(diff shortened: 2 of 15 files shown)


alignan:
I have tested this for now a time now :thumbsup:

g-oikonomou:
This could no longer be merged after #1005 so I rebased it.

alignan:
:+1:

g-oikonomou:
@bthebaudeau @uknoblic @areaarea @hexluthor

How do you feel about this? I want to get it merged and dusted.

bthebaudeau:
The `SOC_ADC_*_BIT_RSHIFT` definitions are added but not used, and they're anyway just the complement to 16 bits, so just using `16 - enob` would be easier than using these definitions. There could be a macro to convert the decimation rate to the ENoB, and another one to convert the ENoB to the right shift. But anyway, `adc_get()` returns a signed 16-bit value independent of the decimation rate, so no shift is required, and the associated definitions could just be removed. There are just more or less useful LSBs in the returned value, which can be ignored rather than dealt with.

Apart from that, this PR looks very good.

hexluthor:
:+1:

areaarea:
@bthebaudeau shouldn't this in adc.c (lines 80ish) be used with div to limit the bits to ENoB's:

/* Read conversion result, reading SOC_ADC_ADCH last to clear
   * SOC_ADC_ADCCON1.EOC */
  res  = REG(SOC_ADC_ADCL) & 0xfc;
  res |= REG(SOC_ADC_ADCH) << 8;

Adam

bthebaudeau:
@areaarea No: as I explained, the extra lsbs are just "don't care" bits. They could be forced to 0 by this function before returning the value, but that would not be very useful, all the more it's possible that the hardware already does this (I've not checked this). The bits beyond the ENoB can just be considered as reading noise. It's like a measurement tool displaying a lot of decimal places, way beyond its accuracy.

hexluthor:
The two least-significant bits of `SOC_ADC_ADCL` always read as zero.

bthebaudeau:
@hexluthor Yes, but here we are talking about the next bits for a smaller ENoB. The largest ENoB is 12 bits for a decimation of 512, so this leaves 2 "don't care" bits in the lsbs, plus the 2 lsbs that you mention, which are not part of the result bit-field in the register description.

g-oikonomou:
OK thanks for the input folks, but I must say I am somewhat confused in terms of what changes need applied here. From what I can tell, we are just talking about nullifying the patch to `soc-adc.h`. Correct?

bthebaudeau:
@g-oikonomou Correct. If we wanted to force the unused lsbs to 0 (which would be for another PR), other definitions would be needed anyway, but I don't think we need this at all.

g-oikonomou:
... and done

bthebaudeau:
:+1:

areaarea:
👍

alignan:
:+1:

PR #971, opened 2015-02-22
Expand Macros when Building Docs

g-oikonomou:
TL;DR: This reduces the number of doxygen warnings by 77 and adjusts the travis job accordingly.

Doxygen gets confused by statements such as these below:

```
static int some_variable __attribute__((aligned(4)));
static const char hdr_name[] HTTPD_STRING_ATTR = "header";
```

The confusion is caused by `__attribute__()` stuff and by `HTTPD_STRING_ATTR`, resulting in "documented symbol was not declared or defined" warnings. Our current doxygen build is full of those.

We can try this:
```
#ifdef __GNUC__
#define CC_ALIGN_ATTR(n) __attribute__((aligned(n)))
#else
#define CC_ALIGN_ATTR(n)
#endif

static int some_variable CC_ALIGN_ATTR(4);
```
But this will still not work, because our doxygen configuration does _not_ expand macros. To fix warnings of this nature, we have to tell doxygen to expand macros by setting `MACRO_EXPANSION        = YES`

Then we have to select one of two possible approaches:
* Either we expand only macros defined in `Doxyfile`
```
EXPAND_ONLY_PREDEF     = YES
PREDEFINED` = ...
```
* or we expand all macros (`EXPAND_ONLY_PREDEF     = NO`)

The former is faster to build, but we will have to modify `Doxyfile` each time a pull request adds lines using declarations of this nature.

Thus, in this pull I am proposing the latter:
* Documentation builds will take somewhat longer, but
* If (current as well as new) sources correctly wrap those definitions with `#if` guards, then we won't have to touch `Doxyfile` that often (if at all).


nvt:
:+1:

PR #978, opened 2015-02-26
Implement set/get of the CCA threshold for CC2420

nvt:
Make it possible to set and get the CC2420's CCA threshold through the extended radio API.


g-oikonomou:
Looks good to me. I've restarted travis job 5.

PR #979, opened 2015-02-26
Time Drifting Motes

TheGeorge:
This pull requests implements support for motes that progress with different speed.

Currently the clock of all simulated motes is perfectly synchronized with the simulation time. In reality, the clocks of two nodes will drift apart (hardware imperfections/temperature differences/etc.). This PR enables to simulate and test systems where the clocks are not perfectly aligned. 

Each the mote clock interface has now the possibility to set a deviation factor between 0.0 and 1.0. A factor of 1.0 means that the clocks will be perfectly aligned with the simulation time. 0.5 means that the mote is running in half speed. 

Standard values are 1.0 and there is no gui to change the value, but it is accessible through the java interface for the motes.

### DIFF
--- a/tools/cooja/apps/avrora/src/org/contikios/cooja/avrmote/MicaZMote.java
+++ b/tools/cooja/apps/avrora/src/org/contikios/cooja/avrmote/MicaZMote.java
@@ -77,6 +77,10 @@ public class MicaZMote extends AbstractEmulatedMote implements Mote {
 
   private EEPROM eeprom = null;
   
+  private long executed = 0;
+  private long skipped = 0;
+  
+  
   /* Stack monitoring variables */
   private boolean stopNextInstruction = false;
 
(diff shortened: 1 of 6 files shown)


vsaw:
Will the rest of the code still work if the return value of `getDeviation` changes during a simulation?
This would allow to break the determinism of the simulation by returning carefully chosen random drifts (if wanted).

TheGeorge:
Changing the deviation factor during simulation should not break. The current implementation performs a time book keeping and skips the mote execution from time to time. If the deviation is changed, then the just rate of skipping should change.

TheGeorge:
There is some issue with the rpl tests in the travis test suite. See #980

TheGeorge:
I added a GUI (accessible via the Mote Interface Viewer) and saving/restoring the timing settings using the simulation file.

cmorty:
I  really like this one. A rebase would be nice though. :)

TheGeorge:
I think I messed that rebase up.

TheGeorge:
Is there anything else I can do to progress this pull request?

fros4943:
Great! :+1:
I know I will personally use this extensively :smiley_cat:

Merging!

PR #987, opened 2015-03-09
Send the first NS attempt inside tcpip_ipv6_output()

ioannisg:
This pull request enables tcpip_ipv6_output() to send the first of the (possibly multiple) NS messages for a new IPv6 neighbor immediately after this neighbor is added to the neighbor list.

Under the current implementation, tcpip_ipv6_output() prepares a new NS packet, places it in the uip buffer, but returns, before sending it, relying on the periodic calls of the IPv6 engine to send the pending frame. However, if a packet arrival from the underlying MAC layer (e.g. sicslowpan) occurs before the following call of uip_ds6_periodic, the pending NS frame is lost (sicslowpan overwrites the uip_buf and uip_len may be set to 0 after processing the incoming frame). This pull request will force the NS transmission before tcpip_ipv6_output() releases the control.

### DIFF
--- a/core/net/ip/tcpip.c
+++ b/core/net/ip/tcpip.c
@@ -675,7 +675,11 @@ tcpip_ipv6_output(void)
 
         stimer_set(&nbr->sendns, uip_ds6_if.retrans_timer / 1000);
         nbr->nscount = 1;
+        /* Send the first NS try from here (multicast destination IP address). */
       }
+#else /* UIP_ND6_SEND_NA */
+      uip_len = 0;
+      return;  
 #endif /* UIP_ND6_SEND_NA */
     } else {
 #if UIP_ND6_SEND_NA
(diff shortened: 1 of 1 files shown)


joakimeriksson:
We had a careful look at this and it seems to be a good bugfix which should in some cases avoid multiple packet drops (due to missing neighbor). :+1:

simonduq:
This looks ok to me, but I had a hard time understanding the fix. I have a suggestion to make the code clearer: how about, instead of removing a `return` statement, adding, right after `/* Send the first NS try from here (multicast destination IP address). */`, an explicit call to `tcpip_output` and then resetting `uip_len` and returning?

simonduq:
Ok I approve the fix and so does @joakimeriksson, so I will merge it in a month if nobody disapproves.

PR #988, opened 2015-03-10
Update CC2650 SensorTag SPI Flash Support

g-oikonomou:
This pull updates SPI clock IOID mapping and Device ID for the CC2650 SensorTag's external flash. It also fixes a doxygen comment.


g-oikonomou:
Obsoleted by #1047

PR #989, opened 2015-03-10
Added SHT25 sensor

alignan:
Included the Sensirion SHT25 Temperature and Humidity sensor driver, and a test example.


g-oikonomou:
Splendid!

kmartinez:
we meet again George! we have deployed some Z1's around the lab with these SHT25s after porting the driver too!

g-oikonomou:
We do indeed, Kirk!

PR #990, opened 2015-03-11
Removed duplicated condition in sht25 configuration

alignan:
Duplicated condition removed, the configuration should only be done on SENSORS_ACTIVE, to ble used from the SENSORS_ACTIVATE and SENSORS_DEACTIVATE macro.


nvt:
:+1:

PR #992, opened 2015-03-14
Added inital mtarch.c/h for platform/cooja

marcas756:
I am currently trying to get the multi-threading example up and running
for cooja motes.

I am not an expert in java and also not in getting native code running
in java.

I had a short look on the mtarch implementation in cooja. All stub
routines are empty at the moment.
cooja_mt_thread implements preemtive threading for rtimer and the
mainloop and may not be used for mt_threads I think.
I assume that this is used to simulate proper rtimer IRQ preemtion of
the main loop - right?

In a first step I just copied the mtarch.c(h) of cpu/native to get use
of fibers of windows and ucontext of posix.

I did not expected it to work from scratch, but I got this for a cooja
mote (one of the rare moments where an assumtion works out of the box):

...
22:24:40.118    ID:1    6543210
22:24:40.618    ID:1    GFEDCBA
22:24:41.118    ID:1    76543210
22:24:41.618    ID:1    HGFEDCBA
22:24:42.118    ID:1    876543210
22:24:42.618    ID:1    IHGFEDCBA
22:24:43.118    ID:1    9876543210
22:24:43.618    ID:1    JIHGFEDCBA
...

which in my opinion is the correct behaviour.

Is there another way planned for mtarch on cooja motes? Or is this
approach OK for a cooja pull request?

### DIFF
--- a/platform/cooja/Makefile.cooja
+++ b/platform/cooja/Makefile.cooja
@@ -47,7 +47,7 @@ CONTIKI_TARGET_DIRS = . dev lib sys cfs net
 # (COOJA_SOURCEDIRS contains additional sources dirs set from simulator)
 vpath %.c $(COOJA_SOURCEDIRS)
 
-COOJA_BASE	= simEnvChange.c cooja_mt.c cooja_mtarch.c rtimer-arch.c slip.c watchdog.c rimestats.c
+COOJA_BASE	= simEnvChange.c cooja_mt.c cooja_mtarch.c rtimer-arch.c slip.c watchdog.c rimestats.c mtarch.c
 
 COOJA_INTFS	= beep.c button-sensor.c ip.c leds-arch.c moteid.c \
 		    pir-sensor.c rs232.c vib-sensor.c \
--- a/platform/cooja/mtarch.c
+++ b/platform/cooja/mtarch.c
@@ -32,17 +32,59 @@
  *
  */
 
-#include "mtarch.h"
+#include "sys/mt.h"
+
+#ifndef MTARCH_STACKSIZE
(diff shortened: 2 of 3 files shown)


adamdunkels:
There is an mtarch implementation for the `cooja` platform, but it is has an unexpected name and is hiding in a somewhat unexpected place: https://github.com/contiki-os/contiki/blob/master/platform/cooja/sys/cooja_mtarch.c

marcas756:
I had a deeper look into cooja_mt and also cooja_mtarch ... the sp<->esb swapping is now clear for me and is a nice method.
But there is still a problem with cooja_mtarch_exec:

```
void
cooja_mtarch_exec(struct cooja_mtarch_thread *t)
{
  cooja_running_thread = t;
  cooja_sw();
  cooja_running_thread = NULL;
}
```


contiki-cooja-main.c :
```
JNIEXPORT void JNICALL
Java_org_contikios_cooja_corecomm_CLASSNAME_tick(JNIEnv *env, jobject obj)
{

.....

  if(simProcessRunValue == 0) {
    /* Rtimers done: Let Contiki handle a few events.
     * Sets simProcessRunValue */
    cooja_mt_exec(&process_run_thread); // runs process_run !!!
  }

.....
}
```
If I exec an mt_thread while another mt_thread is running, the cooja_running_thread gets overwritten.
And that is exactly what happens here in my case ... if you try to exec a thread in a contiki PROCESS, as done in the multi-threading example ... this can be a Problem.

Also the mapping to mtarch is missing (as told yet, mtarch stubs are currently empty)

If it does not matter, I will come up with another solution and will reopen this PR again as soon as I got one.

PR #993, opened 2015-03-15
Remove duplicate call to board_init

g-oikonomou:
The CC26xx startup sequence calls `board_init()` twice. This pull eliminates one of the calls.


g-oikonomou:
Obsoleted by #1047

PR #994, opened 2015-03-16
Examples: Change name of executable on d64 image to $(CONTIKI_PROJECT) for target c64

marcas756:
Currently only one executable is allowed for target c64 examples. Mentioned executable is stored on a d64 image with file name "contiki".

As it is also possible to define multiple subprojects (executables) in a CONTIKI_PROJECT, it  is obvious to store each subproject on disk by name.

As an example (derived from #934) :
```
cmt-scheduler/
├── cmt.c
├── cmt-event-test.c
├── cmt.h
├── cmt-join-test.c
├── cmt-poll-test.c
├── cmt-sleep-test.c
├── Makefile
└── Makefile.c64.defines
```
where the "test" source files are sources for stand alone executables.

Makefile would contain:
```
CONTIKI_PROJECT = cmt-sleep-test cmt-poll-test cmt-event-test cmt-join-test
```

Directory content after build:
```
cmt-scheduler/
├── cmt.c
├── cmt-event-test.c
├── cmt-event-test.c64
├── cmt.h
├── cmt-join-test.c
├── cmt-join-test.c64
├── cmt-poll-test.c
├── cmt-poll-test.c64
├── cmt-sleep-test.c
├── cmt-sleep-test.c64
├── contiki-c64.a
├── contiki-c64.map
├── contiki.d64
├── cs8900a.eth
├── cs8900a.eth.map
├── lan91c96.eth
├── lan91c96.eth.map
├── Makefile
├── Makefile.c64.defines
├── obj_c64
│   ├── ...
|   
├── w5100.eth
└── w5100.eth.map
```

The .c64 files are the executables that are build anyway. To store them on disk they would need unique names. This is provided by this PR.

On disk following would be stored:

subprojects

### DIFF
--- a/platform/c64/Makefile.c64
+++ b/platform/c64/Makefile.c64
@@ -46,7 +46,12 @@ endif
 
 disk: all
 	$(C1541) -format contiki,00 d64 contiki.d64
-	$(C1541) -attach contiki.d64 -write $(CONTIKI_PROJECT).$(TARGET)          contiki,p
+
+	# CONTIKI_PROJECT may contain more than one project
+	for SUBPROJECT in $(CONTIKI_PROJECT); do \
+	$(C1541) -attach contiki.d64 -write $$SUBPROJECT.$(TARGET)                $$SUBPROJECT,p ; \
+	done
+	
 	$(C1541) -attach contiki.d64 -write $(CONTIKI)/tools/$(TARGET)/sample.cfg contiki.cfg,s
 	$(C1541) -attach contiki.d64 -write cs8900a.eth                           cs8900a.eth,s
 	$(C1541) -attach contiki.d64 -write lan91c96.eth                          lan91c96.eth,s
--- a/platform/c64/Makefile.c64
+++ b/platform/c64/Makefile.c64
@@ -46,12 +46,7 @@ endif
 
 disk: all
 	$(C1541) -format contiki,00 d64 contiki.d64
-
-	# CONTIKI_PROJECT may contain more than one project
-	for SUBPROJECT in $(CONTIKI_PROJECT); do \
-	$(C1541) -attach contiki.d64 -write $$SUBPROJECT.$(TARGET)                $$SUBPROJECT,p ; \
-	done
-	
+	$(C1541) -attach contiki.d64 -write $(CONTIKI_PROJECT).$(TARGET)          $(CONTIKI_PROJECT),p
 	$(C1541) -attach contiki.d64 -write $(CONTIKI)/tools/$(TARGET)/sample.cfg contiki.cfg,s
 	$(C1541) -attach contiki.d64 -write cs8900a.eth                           cs8900a.eth,s
 	$(C1541) -attach contiki.d64 -write lan91c96.eth                          lan91c96.eth,s


oliverschmidt:
1. The same would be true for the other targets allowing to build disk images. And for consistency reasons those targets should behave equal.
2. I strongly disagree with polluting Makefiles with shell logic. So far the Contiki Makefiles are supposed to be independent from specific shells and/or shell tools.

marcas756:
@1 : Yes, I agree ... C128 would have been the next target ...

@2 : What would you propose in this case? 
The current build system allows to build multiple executables for examples. For me it was obvious to make it this way. A way would be to create a directory per subproject and to store 1 file per disk image (the contiki-executable) as it is done currently.

oliverschmidt:
> The current build system allows to build multiple executables for examples.

Sorry for the question, but...do you actually have a true use case failing because of the Makefile not supporting multiple executables - or is this "just because it should work"?

The reason for asking is that whatever solution is brought up will complicate the Makefile - and I don't want to complicate it "just for fun"...

marcas756:
You may ask ... no problem.

Maybe it is a good idea to have a usecase (requirements) management tool for contiki.

Usecase or userstory: Transfer one disk image with some testcases as fast as possible to following device (emulation not allowed).

myc64

oliverschmidt:
>  ... some testcases ...

So the question would be if those testcases are really part of a single project. Or do you "just define" them to be part of a single project as means to have them together on a single disk image?

I don't want to state that it solves your issue but you should anyway be aware of https://github.com/contiki-os/contiki/blob/master/tools/6502/Makefile

marcas756:
I define an example (in contiki/examples) to be a summary of executables (1 ... n), where n has to be selected in a way that the example fits on any targets single disk image.

Your link to a makefile  tries to build multiple disk images with a lot of apps to build a userfriendly disk set ... right? 

You also refer to executables in the example dir in your Makefile ... e.g.:
../../examples/ftp/ftp-client.apple2enh
Which is build anyway (also if you got subprojects in CONTIKI_PROJECT).
So I do not see any conflict here (except the usage of shell script code in my PR).

CONTIKI_PROJECT is not  contained in your Makefile.

Now I would like to ask (if it is not a problem) : what is the meaning of CONTIKI_PROJECT?

oliverschmidt:
The only other non-cc65 target Makefile I found that makse use of CONTIKI_PROJECT is https://github.com/contiki-os/contiki/blob/master/platform/eval-adf7xxxmb4z/Makefile.eval-adf7xxxmb4z and that seems to presume too that CONTIKI_PROJECT consists of one word.

The 'disk' goal in the Makfile(s) in question is just a convenience goal to allow for a quick turnaround during development of a program. That's by the way the reason why it's not  (part of) the default goal.

The tools makefile wasn't mentioned to show a conflict. It was mentioned to show that the 'disk' goal isn't meant as primary way to create a disk image.

After all I don't see me merging a change to extend the 'disk' goal to meet the expectations presented in this pr.

marcas756:
I reverted subprojects, but I would at least change the name from contiki to $(CONTIKI_PROJECT):

c64

So one single file per contiki example disk image with the name of the project (executable).

OK (Y/N)?

oliverschmidt:
> ... I would at least change the name from contiki to $(CONTIKI_PROJECT)

If you check out tools/6502/Makefile you'll see that both the Apple2 and the Atari limit the base name of binaries 8 chars. That means that either

- there would be need for some name truncation logic

or

- the naming logic is different between the CBM and non-CBM cc65-based targets.

Both alternatives pose downsides which don't seem worth the benefit of more meaningful names.

marcas756:
> there would be need for some name truncation logic

I checked it yet, and c1541 seems to properly truncate file names:

```
c1541 -attach contiki.d64 -write cmt-sleep-testkdjfhskdjfhskjdfhsfjh.c64          cmt-sleep-testkdjfhskdjfhskjdfhsfjh,p
Writing file `CMT-SLEEP-TESTKDJFHSKDJFHSKJDFHSFJH.C64' as `CMT-SLEEP-TESTKDJFHSKDJFHSKJDFHSFJH,P' to unit 8.
```

results in 

c64

> the naming logic is different between the CBM and non-CBM cc65-based targets.

At this point the PR refers to platform/c64/Makefile.c64 and no other platform. Can also be extended to platform C128.

oliverschmidt:
I'm well aware of both aspects you mention. However I want all cc65-based targets to be as similiar as possible. And because the non-CBM targets don't handle "too long" names nicely I want to avoid them altogether in the first place.

PR #995, opened 2015-03-18
CC2538: added support for SSI1

uknoblic:
I am preparing a new platform using the cc2538 + cc1200 + enc28j60. The cc1200 and the enj28j60 use different SPI moduls, so i have added support for the SSI1. I did the following changes:
- renaming the "old" functions spi_foo() to spi0_foo()
- adding support for SSI1 by duplicating the code spi0_foo() -> spi1_foo() and changing the registers accordingly
- adding macros for the old spi_foo() functions + defines so the still work
When using SSI0, the macros have to be defined

*     SPI0_CLK_PORT               SPI0_CLK_PIN
*    SPI0_MOSI_PORT              SPI0_MOSI_PIN
*    SPI0_MISO_PORT              SPI0_MISO_PIN

When using SSI1, the macros have to be defined

*    SPI1_CLK_PORT               SPI1_CLK_PIN
*    SPI1_MOSI_PORT              SPI1_MOSI_PIN
*    SPI1_MISO_PORT              SPI1_MISO_PIN

The old macros 

*   SPI_CLK_PORT               SPI_CLK_PIN
*    SPI_MOSI_PORT              SPI_MOSI_PIN
*    SPI_MISO_PORT              SPI_MISO_PIN

still work and enable the use of SSI0

### DIFF
--- a/core/sys/cc.h
+++ b/core/sys/cc.h
@@ -129,12 +129,18 @@
 
 #define CC_CONCAT2(s1, s2) s1##s2
 /**
- * A C preprocessing macro for concatenating to
- * strings.
+ * A C preprocessing macro for concatenating two preprocessor tokens.
  *
  * We need use two macros (CC_CONCAT and CC_CONCAT2) in order to allow
  * concatenation of two \#defined macros.
  */
 #define CC_CONCAT(s1, s2) CC_CONCAT2(s1, s2)
+#define CC_CONCAT_EXT_2(s1, s2) CC_CONCAT2(s1, s2)
+
+/**
+ * A C preprocessing macro for concatenating three preprocessor tokens.
+ */
+#define CC_CONCAT3(s1, s2, s3) s1##s2##s3
+#define CC_CONCAT_EXT_3(s1, s2, s3) CC_CONCAT3(s1, s2, s3)
 
 #endif /* CC_H_ */
--- a/cpu/cc2538/dev/spi.c
+++ b/cpu/cc2538/dev/spi.c
@@ -41,268 +41,308 @@
 #include "dev/spi.h"
 #include "dev/ssi.h"
 #include "dev/gpio.h"
-
-#if defined(SPI_CLK_PORT) && defined(SPI_CLK_PIN) && \
-	defined(SPI_MISO_PORT) && defined(SPI_MISO_PIN) && \
-	defined(SPI_MOSI_PORT) && defined(SPI_MOSI_PIN)
-
-/* For compatibility reasons */
-#define spi_init()						spi0_init()	
-#define spi_cs_init(port, pin)			spi0_cs_init(port, pin)	
-#define spi_enable()					spi0_enable()	
-#define spi_disable()					spi0_disable()	
-#define spi_set_mode(fr, po, ph, sz)	spi0_set_mode(fr,po,ph,sz)	
-
-#define USE_SPI0						1
-
-#define SPI0_CLK_PORT					SPI_CLK_PORT	
-#define SPI0_CLK_PIN					SPI_CLK_PIN	
-#define SPI0_MOSI_PORT					SPI_MOSI_PORT
-#define SPI0_MOSI_PIN					SPI_MOSI_PIN
-#define SPI0_MISO_PORT					SPI_MISO_PORT
-#define SPI0_MISO_PIN					SPI_MISO_PIN
-
-#elif defined(SPI0_CLK_PORT) && defined(SPI0_CLK_PIN) && \
-	defined(SPI0_MISO_PORT) && defined(SPI0_MISO_PIN) && \
-	defined(SPI0_MOSI_PORT) && 	defined(SPI0_MOSI_PIN)
-
-#define USE_SPI0						1
-
-#endif	/* #if SPI0 port / pins defined */
-
-#if USE_SPI0
-
-#define SPI0_CLK_PORT_BASE				GPIO_PORT_TO_BASE(SPI0_CLK_PORT)
-#define SPI0_CLK_PIN_MASK				GPIO_PIN_MASK(SPI0_CLK_PIN)
-#define SPI0_MOSI_PORT_BASE				GPIO_PORT_TO_BASE(SPI0_MOSI_PORT)
-#define SPI0_MOSI_PIN_MASK				GPIO_PIN_MASK(SPI0_MOSI_PIN)
-#define SPI0_MISO_PORT_BASE				GPIO_PORT_TO_BASE(SPI0_MISO_PORT)
-#define SPI0_MISO_PIN_MASK				GPIO_PIN_MASK(SPI0_MISO_PIN)
-
-#endif	/* #if USE_SPI0 */
-
-#if defined(SPI1_CLK_PORT) && defined(SPI1_CLK_PIN) && \
-	defined(SPI1_MISO_PORT) && defined(SPI1_MISO_PIN) && \
-	defined(SPI1_MOSI_PORT) && 	defined(SPI1_MOSI_PIN) 
-
-#define USE_SPI1						1
-#define SPI1_CLK_PORT_BASE				GPIO_PORT_TO_BASE(SPI1_CLK_PORT)
-#define SPI1_CLK_PIN_MASK				GPIO_PIN_MASK(SPI1_CLK_PIN)
-#define SPI1_MOSI_PORT_BASE				GPIO_PORT_TO_BASE(SPI1_MOSI_PORT)
-#define SPI1_MOSI_PIN_MASK				GPIO_PIN_MASK(SPI1_MOSI_PIN)
-#define SPI1_MISO_PORT_BASE				GPIO_PORT_TO_BASE(SPI1_MISO_PORT)
-#define SPI1_MISO_PIN_MASK				GPIO_PIN_MASK(SPI1_MISO_PIN)
-
-#endif	/* #if SPI1 port / pins defined */
-
 /*---------------------------------------------------------------------------*/
-/**
- * \brief Initialize the SPI bus for SSI0
- *
- * This SPI init() function uses the following defines to set the pins:
- *    SPI0_CLK_PORT               SPI0_CLK_PIN
- *    SPI0_MOSI_PORT              SPI0_MOSI_PIN
- *    SPI0_MISO_PORT              SPI0_MISO_PIN
- *
- * This sets the mode to Motorola SPI with the following format options:
- *    Clock phase:               1; data captured on second (rising) edge
- *    Clock polarity:            1; clock is high when idle
- *    Data size:                 8 bits
- */
-#if USE_SPI0
-void
-spi0_init(void)
+/* Map the old port / pin names to the new ones for compatibility reasons */
+#if (SPI_DEFAULT_INSTANCE==0)
+#ifdef SPI_CLK_PORT
+#define SPI0_CLK_PORT					SPI_CLK_PORT
+#endif
+#ifdef SPI_CLK_PIN
+#define SPI0_CLK_PIN					SPI_CLK_PIN
+#endif
+#ifdef SPI_MOSI_PORT
+#define SPI0_TX_PORT					SPI_MOSI_PORT
+#endif
+#ifdef SPI_MOSI_PIN
+#define SPI0_TX_PIN						SPI_MOSI_PIN
+#endif
+#ifdef SPI_MISO_PORT
+#define SPI0_RX_PORT					SPI_MISO_PORT
+#endif
+#ifdef SPI_MISO_PIN
+#define SPI0_RX_PIN						SPI_MISO_PIN
+#endif
+#elif (SPI_DEFAULT_INSTANCE==1)
+#ifdef SPI_CLK_PORT
+#define SPI1_CLK_PORT					SPI_CLK_PORT
+#endif
+#ifdef SPI_CLK_PIN
+#define SPI1_CLK_PIN					SPI_CLK_PIN
+#endif
+#ifdef SPI_MOSI_PORT
+#define SPI1_TX_PORT					SPI_MOSI_PORT
+#endif
+#ifdef SPI_MOSI_PIN
+#define SPI1_TX_PIN						SPI_MOSI_PIN
+#endif
+#ifdef SPI_MISO_PORT
+#define SPI1_RX_PORT					SPI_MISO_PORT
+#endif
+#ifdef SPI_MISO_PIN
+#define SPI1_RX_PIN						SPI_MISO_PIN
+#endif
+#else
+#error "Invalid SPI instance. Valid values are 0 or 1"
+#endif
+/*---------------------------------------------------------------------------*/
+/* Check port / pin settings for SPI0 and provide default values for spi_cfg */
+#ifndef SPI0_CLK_PORT
+#define SPI0_CLK_PORT					(-1)
+#endif
+#ifndef SPI0_CLK_PIN
+#define SPI0_CLK_PIN					(-1)
+#endif
+#if SPI0_CLK_PORT >= 0 && SPI0_CLK_PIN < 0 || \
+    SPI0_CLK_PORT < 0 && SPI0_CLK_PIN >= 0
+#error "Both SPI0_CLK_PORT and SPI0_CLK_PIN must be valid or invalid"
+#endif
+
+#ifndef SPI0_TX_PORT
+#define SPI0_TX_PORT					(-1)
+#endif
+#ifndef SPI0_TX_PIN
+#define SPI0_TX_PIN						(-1)
+#endif
+#if SPI0_TX_PORT >= 0 && SPI0_TX_PIN < 0 || \
+    SPI0_TX_PORT < 0 && SPI0_TX_PIN >= 0
+#error "Both SPI0_TX_PORT and SPI0_TX_PIN must be valid or invalid"
+#endif
+
+#ifndef SPI0_RX_PORT
+#define SPI0_RX_PORT					(-1)
+#endif
+#ifndef SPI0_RX_PIN
+#define SPI0_RX_PIN						(-1)
+#endif
+#if SPI0_RX_PORT >= 0 && SPI0_RX_PIN < 0 || \
+    SPI0_RX_PORT < 0 && SPI0_RX_PIN >= 0
+#error "Both SPI0_RX_PORT and SPI0_RX_PIN must be valid or invalid"
+#endif
+
+/* Here we check that either all or none of the ports are defined. As
+   we did already check that both ports + pins are either defined or
+   not for every pin, this means that we can check for an incomplete 
+   configuration by only looking at the port defines */
+#if (SPI0_CLK_PORT < 0 && SPI0_TX_PORT < 0 && SPI0_RX_PORT >= 0) ||  \
+	(SPI0_CLK_PORT < 0 && SPI0_TX_PORT >= 0 && SPI0_RX_PORT < 0) ||  \
+	(SPI0_CLK_PORT < 0 && SPI0_TX_PORT >= 0 && SPI0_RX_PORT >= 0) || \
+	(SPI0_CLK_PORT >= 0 && SPI0_TX_PORT < 0 && SPI0_RX_PORT < 0) ||  \
+	(SPI0_CLK_PORT >= 0 && SPI0_TX_PORT < 0 && SPI0_RX_PORT >= 0) || \
+	(SPI0_CLK_PORT >= 0 && SPI0_TX_PORT >= 0 && SPI0_RX_PORT < 0) ||
+#error "SPI0 port / pin definition incomplete"
+#endif
+/*---------------------------------------------------------------------------*/
+/* Check port / pin settings for SPI1 and provide default values for spi_cfg */
+#ifndef SPI1_CLK_PORT
+#define SPI1_CLK_PORT					(-1)
+#endif
+#ifndef SPI1_CLK_PIN
+#define SPI1_CLK_PIN					(-1)
+#endif
+#if SPI1_CLK_PORT >= 0 && SPI1_CLK_PIN < 0 || \
+    SPI1_CLK_PORT < 0 && SPI1_CLK_PIN >= 0
+#error "Both SPI1_CLK_PORT and SPI1_CLK_PIN must be valid or invalid"
+#endif
+
+#ifndef SPI1_TX_PORT
+#define SPI1_TX_PORT					(-1)
+#endif
+#ifndef SPI1_TX_PIN
+#define SPI1_TX_PIN					   	(-1)
+#endif
+#if SPI1_TX_PORT >= 0 && SPI1_TX_PIN < 0 || \
+    SPI1_TX_PORT < 0 && SPI1_TX_PIN >= 0
+#error "Both SPI1_TX_PORT and SPI1_TX_PIN must be valid or invalid"
+#endif
+
+#ifndef SPI1_RX_PORT
+#define SPI1_RX_PORT				   	(-1)
+#endif
+#ifndef SPI1_RX_PIN
+#define SPI1_RX_PIN					   	(-1)
+#endif
+#if SPI1_RX_PORT >= 0 && SPI1_RX_PIN < 0 || \
+    SPI1_RX_PORT < 0 && SPI1_RX_PIN >= 0
+#error "Both SPI1_RX_PORT and SPI1_RX_PIN must be valid or invalid"
+#endif
+
+/* Here we check that either all or none of the ports are defined. As
+   we did already check that both ports + pins are either defined or
+   not for every pin, this means that we can check for an incomplete 
+   configuration by only looking at the port defines */
+#if (SPI1_CLK_PORT < 0 && SPI1_TX_PORT < 0 && SPI1_RX_PORT >= 0) ||  \
+	(SPI1_CLK_PORT < 0 && SPI1_TX_PORT >= 0 && SPI1_RX_PORT < 0) ||  \
+	(SPI1_CLK_PORT < 0 && SPI1_TX_PORT >= 0 && SPI1_RX_PORT >= 0) || \
+	(SPI1_CLK_PORT >= 0 && SPI1_TX_PORT < 0 && SPI1_RX_PORT < 0) ||  \
+	(SPI1_CLK_PORT >= 0 && SPI1_TX_PORT < 0 && SPI1_RX_PORT >= 0) || \
+	(SPI1_CLK_PORT >= 0 && SPI1_TX_PORT >= 0 && SPI1_RX_PORT < 0) ||
+#error "SPI1 port / pin definition incomplete"
+#endif
+/*---------------------------------------------------------------------------*/
+typedef struct
 {
-  spi0_enable();
-
-  /* Start by disabling the peripheral before configuring it */
-  REG(SSI0_BASE + SSI_CR1) = 0;
-
-  /* Set the IO clock as the SSI clock */
-  REG(SSI0_BASE + SSI_CC) = 1;
-
-  /* Set the mux correctly to connect the SSI pins to the correct GPIO pins */
-  ioc_set_sel(SPI0_CLK_PORT, SPI0_CLK_PIN, IOC_PXX_SEL_SSI0_CLKOUT);
-  ioc_set_sel(SPI0_MOSI_PORT, SPI0_MOSI_PIN, IOC_PXX_SEL_SSI0_TXD);
-  REG(IOC_SSIRXD_SSI0) = (SPI0_MISO_PORT * 8) + SPI0_MISO_PIN;
-
-  /* Put all the SSI gpios into peripheral mode */
-  GPIO_PERIPHERAL_CONTROL(SPI0_CLK_PORT_BASE, SPI0_CLK_PIN_MASK);
-  GPIO_PERIPHERAL_CONTROL(SPI0_MOSI_PORT_BASE, SPI0_MOSI_PIN_MASK);
-  GPIO_PERIPHERAL_CONTROL(SPI0_MISO_PORT_BASE, SPI0_MISO_PIN_MASK);
-
-  /* Disable any pull ups or the like */
-  ioc_set_over(SPI0_CLK_PORT, SPI0_CLK_PIN, IOC_OVERRIDE_DIS);
-  ioc_set_over(SPI0_MOSI_PORT, SPI0_MOSI_PIN, IOC_OVERRIDE_DIS);
-  ioc_set_over(SPI0_MISO_PORT, SPI0_MISO_PIN, IOC_OVERRIDE_DIS);
-
-  /* Configure the clock */
-  REG(SSI0_BASE + SSI_CPSR) = 2;
-
-  /* Configure the default SPI options.
-   *   mode:  Motorola frame format
-   *   clock: High when idle
-   *   data:  Valid on rising edges of the clock
-   *   bits:  8 byte data
-   */
-  REG(SSI0_BASE + SSI_CR0) = SSI_CR0_SPH | SSI_CR0_SPO | (0x07);
-
-  /* Enable the SSI */
-  REG(SSI0_BASE + SSI_CR1) |= SSI_CR1_SSE;
-}
-#endif	/* #if USE_SPI0 */
+	int8_t port_num;
+	int8_t pin_num;
+} spi_pad_t;
+typedef struct {
+	uint32_t base;
+	uint32_t ioc_ssirxd_ssi;
+	uint32_t ioc_pxx_sel_ssi_clkout;
+	uint32_t ioc_pxx_sel_ssi_txd;
+	spi_pad_t clk;
+	spi_pad_t txd;
+	spi_pad_t rxd;
+} spi_cfg_t;
 /*---------------------------------------------------------------------------*/
-/**
- * \brief Initialize the SPI bus for SSI1
- *
- * This SPI init() function uses the following defines to set the pins:
- *    SPI1_CLK_PORT               SPI1_CLK_PIN
- *    SPI1_MOSI_PORT              SPI1_MOSI_PIN
- *    SPI1_MISO_PORT              SPI1_MISO_PIN
- *
- * This sets the mode to Motorola SPI with the following format options:
- *    Clock phase:               1; data captured on second (rising) edge
- *    Clock polarity:            1; clock is high when idle
- *    Data size:                 8 bits
- */
-#if USE_SPI1
-void
-spi1_init(void)
+static const spi_cfg_t spi_regs[SSI_INSTANCE_COUNT] = {
+	{
+		.base = SSI0_BASE,
+		.ioc_ssirxd_ssi = IOC_SSIRXD_SSI0,
+		.ioc_pxx_sel_ssi_clkout = IOC_PXX_SEL_SSI0_CLKOUT,
+		.ioc_pxx_sel_ssi_txd = IOC_PXX_SEL_SSI0_TXD,
+		.clk = {SPI0_CLK_PORT, SPI0_CLK_PIN},
+		.tx = {SPI0_TX_PORT, SPI0_TX_PIN},
+		.rx = {SPI0_RX_PORT, SPI0_RX_PIN}
+	}, {
+		.base = SSI1_BASE,
+		.ioc_ssirxd_ssi = IOC_SSIRXD_SSI1,
+		.ioc_pxx_sel_ssi_clkout = IOC_PXX_SEL_SSI1_CLKOUT,
+		.ioc_pxx_sel_ssi_txd = IOC_PXX_SEL_SSI1_TXD,
+		.clk = {SPI1_CLK_PORT, SPI1_CLK_PIN},
+		.tx = {SPI1_TX_PORT, SPI1_TX_PIN},
+		.rx = {SPI1_RX_PORT, SPI1_RX_PIN}
+	}
+};
+/*---------------------------------------------------------------------------*/
+void spix_init(uint8_t instance)
 {
-  spi1_enable();
+	const spi_cfg_t *cfg; 
+	
+	if (instance >= SSI_INSTANCE_COUNT) {
+		return;
+	}
 
-  /* Start by disabling the peripheral before configuring it */
-  REG(SSI1_BASE + SSI_CR1) = 0;
+	spix_enable(instance);
 
-  /* Set the IO clock as the SSI clock */
-  REG(SSI1_BASE + SSI_CC) = 1;
+	cfg = &spi_cfg[instance];
 
-  /* Set the mux correctly to connect the SSI pins to the correct GPIO pins */
-  ioc_set_sel(SPI1_CLK_PORT, SPI1_CLK_PIN, IOC_PXX_SEL_SSI1_CLKOUT);
-  ioc_set_sel(SPI1_MOSI_PORT, SPI1_MOSI_PIN, IOC_PXX_SEL_SSI1_TXD);
-  REG(IOC_SSIRXD_SSI1) = (SPI1_MISO_PORT * 8) + SPI1_MISO_PIN;
+	if (cfg->clk.port < 0)
+	{ 
+		/* Port / pin configuration invalid. We checked for completeness
+		   above. If clk.port is < 0, this means that all other defines are
+		   < 0 as well */
+		return;
+	}
 
-  /* Put all the SSI gpios into peripheral mode */
-  GPIO_PERIPHERAL_CONTROL(SPI1_CLK_PORT_BASE, SPI1_CLK_PIN_MASK);
-  GPIO_PERIPHERAL_CONTROL(SPI1_MOSI_PORT_BASE, SPI1_MOSI_PIN_MASK);
-  GPIO_PERIPHERAL_CONTROL(SPI1_MISO_PORT_BASE, SPI1_MISO_PIN_MASK);
+	/* Start by disabling the peripheral before configuring it */
+	REG(cfg->base + SSI_CR1) = 0;
+	
+	/* Set the IO clock as the SSI clock */
+	REG(cfg->base + SSI_CC) = 1;
 
-  /* Disable any pull ups or the like */
-  ioc_set_over(SPI1_CLK_PORT, SPI1_CLK_PIN, IOC_OVERRIDE_DIS);
-  ioc_set_over(SPI1_MOSI_PORT, SPI1_MOSI_PIN, IOC_OVERRIDE_DIS);
-  ioc_set_over(SPI1_MISO_PORT, SPI1_MISO_PIN, IOC_OVERRIDE_DIS);
+	/* Set the mux correctly to connect the SSI pins to the correct GPIO pins */
+	ioc_set_sel(cfg->clk.port, 
+				cfg->clk.pin,
+				cfg->ioc_pxx_sel_ssi_clkout);
+	ioc_set_sel(cfg->tx.port, 
+				cfg->tx.pin,
+				cfg->ioc_pxx_sel_ssi_txd);
+	REG(cfg->ioc_ssirxd_ssi) = (cfg->rx.port * 8) + cfg->rx.pin;
+	
+	/* Put all the SSI gpios into peripheral mode */
+	GPIO_PERIPHERAL_CONTROL(GPIO_PORT_TO_BASE(cfg->clk.port),
+							GPIO_PIN_MASK(cfg->clk.pin));
+	GPIO_PERIPHERAL_CONTROL(GPIO_PORT_TO_BASE(cfg->tx.port),
+							GPIO_PIN_MASK(cfg->tx.pin));
+	GPIO_PERIPHERAL_CONTROL(GPIO_PORT_TO_BASE(cfg->rx.port),
+							GPIO_PIN_MASK(cfg->rx.pin));
+	
+	/* Disable any pull ups or the like */
+	ioc_set_over(cfg->clk.port, cfg->clk.pin, IOC_OVERRIDE_DIS);
+	ioc_set_over(cfg->tx.port, cfg->tx.pin, IOC_OVERRIDE_DIS);
+	ioc_set_over(cfg->rx.port, cfg->rx.pin, IOC_OVERRIDE_DIS);
 
-  /* Configure the clock */
-  REG(SSI1_BASE + SSI_CPSR) = 2;
-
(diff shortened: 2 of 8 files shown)


bthebaudeau:
Useful feature, but duplicating code is bad:
 - it multiplies the work to maintain the code,
 - it makes it more likely to have diverging copies, which can introduce hard-to-trace bugs,
 - it increases the binary size.

It would be great if you could do something similar to what has been done for the UART in commit d93d129da6200d9872f9906d2d55832e2f84fd09.

uknoblic:
Hi, i can onyl partialy agree with the duplicate code argument in this case. Its only a few lines of code which are duplicated, and 
- i need this few lines anyway (surrounded by an if / else if put in one function)
- It will lead to MORE code (checking an additional parameter), doing if / else stuff and so on
- If done similar to the UART module (having one additional parameter): the API changes
- My code does NOT increase the binary size because the corresponding function is only compiled if the SII module is used (the port / pin macros are set)
i personally would still prepare having two seperate functions spi0_foo() + spi1_foo() against having
one function spi_foo(0|1,...).

bthebaudeau:
> Hi, i can onyl partialy agree with the duplicate code argument in this case. Its only a few lines of code which are duplicated,

spix_init/cs_init/enable/disable/set_mod(): 5 functions, about 29 lines of code.

> and
>
> - i need this few lines anyway (surrounded by an if / else if put in one function)

No, you don't: with a structure similar to `uart_regs_t` used to initialize a `const` array like `uart_regs[]`, you just have to pass an index to the functions: `uart_init(uint8_t uart)`, then `const uart_regs_t *regs = &uart_regs[uart];`, then you can simply use `regs` throughout the function. No if-s required.

Also, using `REG(regs->base + OFFSET)` instead of the current `REG(regs->base | OFFSET)` found in the current CC2538 code base (unrelated to your PR) should generate more optimal assembly code (@g-oikonomou). `|` is broken by design if the base is not aligned enough for the offset, and CPU instructions support addressing modes using additive offsets only (not OR-ed offsets), so computing a OR for each address used in the IP address range must be more expensive than the standard base + offset (because the compiler does not know the alignment of the variable base, so it can not optimize the `|` into a `+`), except if base is a constant known at compile time. This should be changed in the UART code (and everywhere else it appears), and the SPI code is OK here.

> - It will lead to MORE code (checking an additional parameter), doing if / else stuff and so on

Not necessarily. The only overhead would be the initialization of a pointer at the beginning of each function, and the `const` array. Depending on the optimizations of the compiler, this may be balanced by the explicit variable base + constant offset addressing instead of having multiple constant addresses. The resulting overhead, if any, would be small, and it would only exist if the duplicated code variant is only instantiated for a single SPI.

Note that you could also "macro-ize" the functions to duplicate only the binary, and not the source code, but this would be ugly.

> - If done similar to the UART module (having one additional parameter): the API changes

Yes, but that's not a big problem. The API change would be light. In-tree users would be updated. Out-of-tree users could easily adapt their code. Moreover, legacy macros like the ones you added to keep compatibility could be added in this case too.

> - My code does NOT increase the binary size because the corresponding function is only compiled if the SII module is used (the port / pin macros are set) i personally would still prepare having two seperate functions spi0_foo() + spi1_foo() against having one function spi_foo(0|1,...).

Your code does not make USE_SPI0/1 mutually exclusive, so it will increase the binary image size for some use cases (if both SPIs are used at the same time), even if this is not your use case.

uknoblic:
Ok, i will try. But how do i handle the fact that the spi prototype (spi_init()) is already defined in /core/dev/spi.h? Do you have a proposal how to overwrite this?

bthebaudeau:
We have the following options:
 - Update the SPI API for the whole Contiki. That might mean quite some work.
 - Define a new API (e.g. `void spix_init(unsigned int spi)`) alongside the existing API, either in `/cpu/cc2538/dev/ssi.h`, or in a new `/cpu/cc2538/dev/spi.h`, or in `/core/dev/spi.h`. With the latter, it would be possible to progressively convert Contiki to the new API, and to deprecate the original API after some time. Once the conversion is complete, an SPI instance ID should be passed to the drivers under `/dev/`. In the meantime, the old API should be implemented using the new API on platforms supporting it, thanks to the definition of a default SPI instance in the board header file. This is what I would do.

So:

`/core/dev/spi.h`:
```
void spi_init(void);
void spix_init(unsigned int spi);
```

`/platform/cc2538dk/dev/board.h`:
```
#define SPI_CONF_DEFAULT_SPI 0
```

Your custom `board.h` for your new platform:
```
#define SPI_CONF_DEFAULT_SPI 0
#define CC1200_CONF_SPI 0
#define ENC28J60_CONF_SPI 1
```

`/cpu/cc2538/dev/spi.c`:
```
void
spi_init(void)
{
  spix_init(SPI_CONF_DEFAULT_SPI);
}

void
spix_init(unsigned int spi)
{
  ...
```

Your custom `cc1200.c` for your new platform:
```
spix_do_something(CC1200_CONF_SPI);
```

Your custom `enc28j60.c` for your new platform:
```
spix_do_something(ENC28J60_CONF_SPI);
```

uknoblic:
Ok, i did the changes, made tests with my platform (also testing compatibility with the "old " API) and pushed new files. I would be happy if you could comment on them!

alignan:
Hello!

Thanks for the contribution!

Just one thing, if you have both Ethernet/CC1200 in your board.h file, then:

#if (SPI_DEFAULT_INSTANCE == 0)
#ifdef SPI_CLK_PORT
#define SPI0_CLK_PORT         SPI_CLK_PORT
#endif
#elif (SPI_DEFAULT_INSTANCE == 1)
#ifdef SPI_CLK_PORT
#define SPI1_CLK_PORT         SPI_CLK_PORT
#endif
#endif

Doesn't enable to easily define your modules but rather override the old SPI_CLK_PORT per module, instead why don't use:

#ifdef SPI0_CLK_PORT_CONF
#define SPI0_CLK_PORT
#endif

So in board.h you can have:

#define SPI0_CLK_PORT_CONF CC1200_CLK_PORT
#define SPI1_CLK_PORT_CONF ENCJ60_CLK_PORT

Cheers!

--Antonio

uknoblic:
Hi, 
i kept the old names SPI_CLK_PORT, SPI_CLK_PIN, SPI_MISO_PORT, SPI_MISO_PIN, SPI_MOSI_PORT, SPI_MOSI_PIN for compatibility reasons (only one instance supported), If using 2 instances, the new macros SPIx_CLK_PORT ... SPIx_RX_PIN have to be used. This is the way i did it in my board file:
/**
 * \name CC1200 configuration
 *
 * These macros configure the interface to the CC1200 including SPI
 * @{
 */
/* Which SSI instance to use? Valid values are 0 and 1 */
#define CC1200_SPI_INSTANCE				0
/* PA1 */
#define CC1200_SCLK_PORT_NUM			GPIO_A_NUM
#define CC1200_SCLK_PIN_NUM				1
/* PB1 */
#define CC1200_MOSI_PORT_NUM			GPIO_B_NUM
#define CC1200_MOSI_PIN_NUM				1
/* PB0 */
#define CC1200_MISO_PORT_NUM			GPIO_B_NUM
#define CC1200_MISO_PIN_NUM             0
/* PA0 */
#define CC1200_CS_N_PORT_NUM			GPIO_A_NUM	
#define CC1200_CS_N_PIN_NUM				0	
/* PC7 */
#define CC1200_RESET_N_PORT_NUM			GPIO_C_NUM	
#define CC1200_RESET_N_PIN_NUM			7	
/* PC4 */
#define CC1200_GPIO0_PORT_NUM			GPIO_C_NUM	
#define CC1200_GPIO0_PIN_NUM			4	
/* PC5 */
#define CC1200_GPIO2_PORT_NUM			GPIO_C_NUM	
#define CC1200_GPIO2_PIN_NUM			5	
/* PC6 */
#define CC1200_GPIO3_PORT_NUM			GPIO_C_NUM	
#define CC1200_GPIO3_PIN_NUM			6
/* No changes needed below.
   See ssi.h / spi-arch.c for the corresponding macros */
#ifndef CC1200_SPI_INSTANCE
/* Check if old SPI API is still working */
#define SPI_CONF_DEFAULT_INSTANCE		0
#define SPI_CLK_PORT					CC1200_SCLK_PORT_NUM
#define SPI_CLK_PIN						CC1200_SCLK_PIN_NUM
#define SPI_MOSI_PORT					CC1200_MOSI_PORT_NUM
#define SPI_MOSI_PIN					CC1200_MOSI_PIN_NUM
#define SPI_MISO_PORT					CC1200_MISO_PORT_NUM
#define SPI_MISO_PIN					CC1200_MISO_PIN_NUM
#define CC1200_SPI_TXBUF				SPI_TXBUF	
#define CC1200_SPI_RXBUF				SPI_RXBUF	
#define CC1200_SPI_WAITFORTxREADY()		SPI_WAITFORTxREADY()	
#define CC1200_SPI_WAITFOREORx()		SPI_WAITFOREORx()
#define CC1200_SPI_FLUSH				SPI_FLUSH()
#else
/* New SPI API */
#if (CC1200_SPI_INSTANCE==0)
#define CC1200_SPI_TXBUF				SPI0_TXBUF
#define CC1200_SPI_RXBUF				SPI0_RXBUF
#define CC1200_SPI_WAITFORTxREADY()		SPI0_WAITFORTxREADY()
#define CC1200_SPI_WAITFOREORx()		SPI0_WAITFOREORx()
#define CC1200_SPI_FLUSH				SPI0_FLUSH()
/* The following macros are needed by spi.c */
#define SPI0_CLK_PORT					CC1200_SCLK_PORT_NUM
#define SPI0_CLK_PIN					CC1200_SCLK_PIN_NUM
#define SPI0_TX_PORT					CC1200_MOSI_PORT_NUM
#define SPI0_TX_PIN						CC1200_MOSI_PIN_NUM
#define SPI0_RX_PORT					CC1200_MISO_PORT_NUM
#define SPI0_RX_PIN						CC1200_MISO_PIN_NUM
#elif (CC1200_SPI_INSTANCE==1)
#define CC1200_SPI_TXBUF				SPI1_TXBUF
#define CC1200_SPI_RXBUF				SPI1_RXBUF
#define CC1200_SPI_WAITFORTxREADY()		SPI1_WAITFORTxREADY()
#define CC1200_SPI_WAITFOREORx()		SPI1_WAITFOREORx()
#define CC1200_SPI_FLUSH				SPI1_FLUSH()
/* The following macros are needed by spi.c */
#define SPI1_CLK_PORT					CC1200_SCLK_PORT_NUM
#define SPI1_CLK_PIN					CC1200_SCLK_PIN_NUM
#define SPI1_TX_PORT					CC1200_MOSI_PORT_NUM
#define SPI1_TX_PIN						CC1200_MOSI_PIN_NUM
#define SPI1_RX_PORT					CC1200_MISO_PORT_NUM
#define SPI1_RX_PIN						CC1200_MISO_PIN_NUM
#endif
#endif

/** @} */
/**
 * \name ENC28J60 configuration
 *
 * These macross configure the interface to the ENC28J60 including SPI.
 * We use a different SPI module for the CC1200 and the ENC28J60
 * @{
 */
/* Which SSI module to use? Valid values are 0 and 1 */
#define ENC28J60_SPI_INSTANCE			1
/* PA3 */
#define ENC28J60_SCLK_PORT_NUM			GPIO_A_NUM
#define ENC28J60_SCLK_PIN_NUM			3
/* PA5 */
#define ENC28J60_MOSI_PORT_NUM			GPIO_A_NUM
#define ENC28J60_MOSI_PIN_NUM			5
/* PA4 */
#define ENC28J60_MISO_PORT_NUM			GPIO_A_NUM
#define ENC28J60_MISO_PIN_NUM           4
/* PA2 */
#define ENC28J60_CS_N_PORT_NUM			GPIO_A_NUM	
#define ENC28J60_CS_N_PIN_NUM			2
/* PA6 */
#define ENC28J60_INT_N_PORT_NUM			GPIO_A_NUM	
#define ENC28J60_INT_N_PIN_NUM			6	
/* PA7 */
#define ENC28J60_RESET_N_PORT_NUM		GPIO_A_NUM	
#define ENC28J60_RESET_N_PIN_NUM		7	
/* No changes needed below. 
   See ssi.h / spi-arch.c for the corresponding macros */
#if (ENC28J60_SPI_INSTANCE==0)
#define ENC28J60_SPI_TXBUF				SPI0_TXBUF
#define ENC28J60_SPI_RXBUF				SPI0_RXBUF
#define ENC28J60_SPI_WAITFORTxREADY()	SPI0_WAITFORTxREADY()
#define ENC28J60_SPI_WAITFOREORx()		SPI0_WAITFOREORx()
#define ENC28J60_SPI_FLUSH				SPI0_FLUSH()
/* The following macros are needed by spi.c */
#define SPI0_CLK_PORT					ENC28J60_SCLK_PORT_NUM
#define SPI0_CLK_PIN					ENC28J60_SCLK_PIN_NUM
#define SPI0_TX_PORT					ENC28J60_MOSI_PORT_NUM
#define SPI0_TX_PIN						ENC28J60_MOSI_PIN_NUM
#define SPI0_RX_PORT					ENC28J60_MISO_PORT_NUM
#define SPI0_RX_PIN						ENC28J60_MISO_PIN_NUM
#elif (ENC28J60_SPI_INSTANCE==1)
#define ENC28J60_SPI_TXBUF				SPI1_TXBUF
#define ENC28J60_SPI_RXBUF				SPI1_RXBUF
#define ENC28J60_SPI_WAITFORTxREADY()	SPI1_WAITFORTxREADY()
#define ENC28J60_SPI_WAITFOREORx()		SPI1_WAITFOREORx()
#define ENC28J60_SPI_FLUSH				SPI1_FLUSH()
/* The following macros are needed by spi.c */
#define SPI1_CLK_PORT					ENC28J60_SCLK_PORT_NUM
#define SPI1_CLK_PIN					ENC28J60_SCLK_PIN_NUM
#define SPI1_TX_PORT					ENC28J60_MOSI_PORT_NUM
#define SPI1_TX_PIN						ENC28J60_MOSI_PIN_NUM
#define SPI1_RX_PORT					ENC28J60_MISO_PORT_NUM
#define SPI1_RX_PIN						ENC28J60_MISO_PIN_NUM
#endif
/** @} */





maybe this is what i already did. My board file looks like this:

uknoblic:
Sorry, pasting from my board file didn't work as expected. How can paste my code?

uknoblic:
And i would be very happy if somebody could explain me why the travis build fails ;-)

g-oikonomou:
> Sorry, pasting from my board file didn't work as expected. How can paste my code?

https://help.github.com/articles/github-flavored-markdown/#fenced-code-blocks

> And i would be very happy if somebody could explain me why the travis build fails ;-)

Now _that_ sounds like an excellent question!

g-oikonomou:
And while travis is running, here are a few thoughts:

* This is a useful feature. We want this! Definitely for the CC2538, but my personal opinion is we want a new SPI API in core anyway, so this could be a move forward
* What Benoit has done with UART0/1 is what I'd recommend doing for SSI0/1 as per his comments
* Lasty, please make sure to read the notes about code style in the wiki page as well as in `CONTRIBUTING.md`

uknoblic:
Thanks George. But may i ask if you response to my travis related question is meant ironic?
This is my board file:
````
/**
 * \name CC1200 configuration
 *
 * These macros configure the interface to the CC1200 including SPI
 * @{
 */
/* Which SSI instance to use? Valid values are 0 and 1 */
#define CC1200_SPI_INSTANCE				0
/* PA1 */
#define CC1200_SCLK_PORT_NUM			GPIO_A_NUM
#define CC1200_SCLK_PIN_NUM				1
/* PB1 */
#define CC1200_MOSI_PORT_NUM			GPIO_B_NUM
#define CC1200_MOSI_PIN_NUM				1
/* PB0 */
#define CC1200_MISO_PORT_NUM			GPIO_B_NUM
#define CC1200_MISO_PIN_NUM             0
/* PA0 */
#define CC1200_CS_N_PORT_NUM			GPIO_A_NUM	
#define CC1200_CS_N_PIN_NUM				0	
/* PC7 */
#define CC1200_RESET_N_PORT_NUM			GPIO_C_NUM	
#define CC1200_RESET_N_PIN_NUM			7	
/* PC4 */
#define CC1200_GPIO0_PORT_NUM			GPIO_C_NUM	
#define CC1200_GPIO0_PIN_NUM			4	
/* PC5 */
#define CC1200_GPIO2_PORT_NUM			GPIO_C_NUM	
#define CC1200_GPIO2_PIN_NUM			5	
/* PC6 */
#define CC1200_GPIO3_PORT_NUM			GPIO_C_NUM	
#define CC1200_GPIO3_PIN_NUM			6
/* No changes needed below.
   See ssi.h / spi-arch.c for the corresponding macros */
#ifndef CC1200_SPI_INSTANCE
/* Check if old SPI API is still working */
#define SPI_CONF_DEFAULT_INSTANCE		0
#define SPI_CLK_PORT					CC1200_SCLK_PORT_NUM
#define SPI_CLK_PIN						CC1200_SCLK_PIN_NUM
#define SPI_MOSI_PORT					CC1200_MOSI_PORT_NUM
#define SPI_MOSI_PIN					CC1200_MOSI_PIN_NUM
#define SPI_MISO_PORT					CC1200_MISO_PORT_NUM
#define SPI_MISO_PIN					CC1200_MISO_PIN_NUM
#define CC1200_SPI_TXBUF				SPI_TXBUF	
#define CC1200_SPI_RXBUF				SPI_RXBUF	
#define CC1200_SPI_WAITFORTxREADY()		SPI_WAITFORTxREADY()	
#define CC1200_SPI_WAITFOREORx()		SPI_WAITFOREORx()
#define CC1200_SPI_FLUSH				SPI_FLUSH()
#else
/* New SPI API */
#if (CC1200_SPI_INSTANCE==0)
#define CC1200_SPI_TXBUF				SPI0_TXBUF
#define CC1200_SPI_RXBUF				SPI0_RXBUF
#define CC1200_SPI_WAITFORTxREADY()		SPI0_WAITFORTxREADY()
#define CC1200_SPI_WAITFOREORx()		SPI0_WAITFOREORx()
#define CC1200_SPI_FLUSH				SPI0_FLUSH()
/* The following macros are needed by spi.c */
#define SPI0_CLK_PORT					CC1200_SCLK_PORT_NUM
#define SPI0_CLK_PIN					CC1200_SCLK_PIN_NUM
#define SPI0_TX_PORT					CC1200_MOSI_PORT_NUM
#define SPI0_TX_PIN						CC1200_MOSI_PIN_NUM
#define SPI0_RX_PORT					CC1200_MISO_PORT_NUM
#define SPI0_RX_PIN						CC1200_MISO_PIN_NUM
#elif (CC1200_SPI_INSTANCE==1)
#define CC1200_SPI_TXBUF				SPI1_TXBUF
#define CC1200_SPI_RXBUF				SPI1_RXBUF
#define CC1200_SPI_WAITFORTxREADY()		SPI1_WAITFORTxREADY()
#define CC1200_SPI_WAITFOREORx()		SPI1_WAITFOREORx()
#define CC1200_SPI_FLUSH				SPI1_FLUSH()
/* The following macros are needed by spi.c */
#define SPI1_CLK_PORT					CC1200_SCLK_PORT_NUM
#define SPI1_CLK_PIN					CC1200_SCLK_PIN_NUM
#define SPI1_TX_PORT					CC1200_MOSI_PORT_NUM
#define SPI1_TX_PIN						CC1200_MOSI_PIN_NUM
#define SPI1_RX_PORT					CC1200_MISO_PORT_NUM
#define SPI1_RX_PIN						CC1200_MISO_PIN_NUM
#endif
#endif
/** @} */
/**
 * \name ENC28J60 configuration
 *
 * These macross configure the interface to the ENC28J60 including SPI.
 * We use a different SPI module for the CC1200 and the ENC28J60
 * @{
 */
/* Which SSI module to use? Valid values are 0 and 1 */
#define ENC28J60_SPI_INSTANCE			1
/* PA3 */
#define ENC28J60_SCLK_PORT_NUM			GPIO_A_NUM
#define ENC28J60_SCLK_PIN_NUM			3
/* PA5 */
#define ENC28J60_MOSI_PORT_NUM			GPIO_A_NUM
#define ENC28J60_MOSI_PIN_NUM			5
/* PA4 */
#define ENC28J60_MISO_PORT_NUM			GPIO_A_NUM
#define ENC28J60_MISO_PIN_NUM           4
/* PA2 */
#define ENC28J60_CS_N_PORT_NUM			GPIO_A_NUM	
#define ENC28J60_CS_N_PIN_NUM			2
/* PA6 */
#define ENC28J60_INT_N_PORT_NUM			GPIO_A_NUM	
#define ENC28J60_INT_N_PIN_NUM			6	
/* PA7 */
#define ENC28J60_RESET_N_PORT_NUM		GPIO_A_NUM	
#define ENC28J60_RESET_N_PIN_NUM		7	
/* No changes needed below. 
   See ssi.h / spi-arch.c for the corresponding macros */
#if (ENC28J60_SPI_INSTANCE==0)
#define ENC28J60_SPI_TXBUF				SPI0_TXBUF
#define ENC28J60_SPI_RXBUF				SPI0_RXBUF
#define ENC28J60_SPI_WAITFORTxREADY()	SPI0_WAITFORTxREADY()
#define ENC28J60_SPI_WAITFOREORx()		SPI0_WAITFOREORx()
#define ENC28J60_SPI_FLUSH				SPI0_FLUSH()
/* The following macros are needed by spi.c */
#define SPI0_CLK_PORT					ENC28J60_SCLK_PORT_NUM
#define SPI0_CLK_PIN					ENC28J60_SCLK_PIN_NUM
#define SPI0_TX_PORT					ENC28J60_MOSI_PORT_NUM
#define SPI0_TX_PIN						ENC28J60_MOSI_PIN_NUM
#define SPI0_RX_PORT					ENC28J60_MISO_PORT_NUM
#define SPI0_RX_PIN						ENC28J60_MISO_PIN_NUM
#elif (ENC28J60_SPI_INSTANCE==1)
#define ENC28J60_SPI_TXBUF				SPI1_TXBUF
#define ENC28J60_SPI_RXBUF				SPI1_RXBUF
#define ENC28J60_SPI_WAITFORTxREADY()	SPI1_WAITFORTxREADY()
#define ENC28J60_SPI_WAITFOREORx()		SPI1_WAITFOREORx()
#define ENC28J60_SPI_FLUSH				SPI1_FLUSH()
/* The following macros are needed by spi.c */
#define SPI1_CLK_PORT					ENC28J60_SCLK_PORT_NUM
#define SPI1_CLK_PIN					ENC28J60_SCLK_PIN_NUM
#define SPI1_TX_PORT					ENC28J60_MOSI_PORT_NUM
#define SPI1_TX_PIN						ENC28J60_MOSI_PIN_NUM
#define SPI1_RX_PORT					ENC28J60_MISO_PORT_NUM
#define SPI1_RX_PIN						ENC28J60_MISO_PIN_NUM
#endif
/** @} */
````

**Comment by g-oikonomou** [comment]
No! Definitely not ironic, I just posted while at the same time scratching my head... Sorry if it came across the wrong way!

Job 5 fails cause... it's job 5: see #980. Job 1 failed cause travis failed to `apt-get update` and therefore tried to build docs with an old version of doxygen. This should be transient. No idea about Job 2.

**Comment by uknoblic** [comment]
George,
i think my latests commits reflect what Benoit proposes. The remaining question is wether we can somehow come to a common understanding with regard to the naming convention. According to Benoit's proposal, the new funtions and macros are all prefixed with either "spix" or "SPIX" - we were more or less forced to choose new names due to the fact that spi_init() was already defined in /core/dev/spi.h and i didn't dare to change this because it's a contiki-wide prototype. But i am not happy with this fact because we now have at least two modules supporting multiple instances with different names.

**Comment by g-oikonomou** [comment]
Thinking aloud here: I think the current SPI API is a little tentative (so is the sensors API IMHO). Various people have had to use various workarounds to achieve various goals and this is far from ideal.

What I think we need is a good SPI (and I2C for that matter) abstraction in core. I don't have a good suggestion though, and I am certainly not suggesting that this pull should aim to come up with one.

On the other hand, is adding yet another workaround what we really want?

I can see the reasoning behind `spix_`, but it's likely to raise some eyebrows in terms of naming convention use and abuse (devil's advocate speaking).

**Comment by alignan** [comment]
Hello again!

Thanks for the clarification.  Do you have an instruction on how to compile? I'm compiling the vanilla cc2538dk example and fails to compile due to:

````
In file included from ../../cpu/cc2538/dev/spi.c:41:0:
../../core/./dev/spi.h:58:19: error: macro "spi_init" passed 1 arguments, but takes just 0
 void spi_init(void);
````
From:
````
commit 22d2bb49d77e233e67eab84caa0a9a099601a151
Author: Ulf Knoblich <ulf.knoblich@weptech.de>
Date:   Mon Mar 23 09:58:16 2015 +0100
    Proposal for a new SPI API supporting multiple instances
````

**Comment by uknoblic** [comment]
Hi, this should be fixed with the latest commit. It was caused by the macro spi_init() defined in spi-arch.h. I now moved it to a function call which then maps to spix_init()

File: cpu/cc2538/dev/spi.c [hunk 1]
```diff
@@ -41,102 +41,337 @@
...
+#endif
+#ifdef SPI_MISO_PORT
+#define SPI1_RX_PORT          SPI_MISO_PORT
+#endif
+#ifdef SPI_MISO_PIN
+#define SPI1_RX_PIN           SPI_MISO_PIN
+#endif
+#else
+#error "Invalid SPI instance. Valid values are 0 or 1"
+#endif
```
bthebaudeau: The legacy SPI_* definitions above would only come from board header files. They are not part of the core SPI API. Hence, we don't care about them. They can just be removed, and out-of-tree users should just slightly upgrade their code. Then, board.h files should directly define SPI pins as SPI0_* or SPI1_*.

cpu/cc2538/dev/spi.c
```diff
@@ -41,102 +41,337 @@
...
+/* Check port / pin settings for SPI0 and provide default values for spi_cfg */
+#ifndef SPI0_CLK_PORT
+#define SPI0_CLK_PORT         (-1)
+#endif
+#ifndef SPI0_CLK_PIN
+#define SPI0_CLK_PIN          (-1)
+#endif
+#if SPI0_CLK_PORT >= 0 && SPI0_CLK_PIN < 0 || \
+  SPI0_CLK_PORT < 0 && SPI0_CLK_PIN >= 0
+#error "Both SPI0_CLK_PORT and SPI0_CLK_PIN must be valid or invalid"
```
bthebaudeau: No quotation marks in `#error` directives. Everything following a `#error` is printed, including the quotation marks if any, so the whole error message must not be quoted (this is not a C string).

cpu/cc2538/dev/spi.c
```diff
@@ -41,102 +41,337 @@
...
+#ifndef SPI0_TX_PORT
+#define SPI0_TX_PORT          (-1)
+#endif
+#ifndef SPI0_TX_PIN
+#define SPI0_TX_PIN           (-1)
+#endif
+#if SPI0_TX_PORT >= 0 && SPI0_TX_PIN < 0 || \
+  SPI0_TX_PORT < 0 && SPI0_TX_PIN >= 0
+#error "Both SPI0_TX_PORT and SPI0_TX_PIN must be valid or invalid"
+#endif
```
bthebaudeau: Please use the standard MOSI / MISO naming convention for SPI, instead of TX / RX, which is unusual, unclear and confusing.

cpu/cc2538/dev/spi.c
```diff
@@ -41,102 +41,337 @@
...
+   not for every pin, this means that we can check for an incomplete
+   configuration by only looking at the port defines */
+#if (SPI0_CLK_PORT < 0 && SPI0_TX_PORT < 0 && SPI0_RX_PORT >= 0) || \
+  (SPI0_CLK_PORT < 0 && SPI0_TX_PORT >= 0 && SPI0_RX_PORT < 0) || \
+  (SPI0_CLK_PORT < 0 && SPI0_TX_PORT >= 0 && SPI0_RX_PORT >= 0) || \
+  (SPI0_CLK_PORT >= 0 && SPI0_TX_PORT < 0 && SPI0_RX_PORT < 0) || \
+  (SPI0_CLK_PORT >= 0 && SPI0_TX_PORT < 0 && SPI0_RX_PORT >= 0) || \
+  (SPI0_CLK_PORT >= 0 && SPI0_TX_PORT >= 0 && SPI0_RX_PORT < 0)
+#error "SPI0 port / pin definition incomplete"
+#endif
```
bthebaudeau: This could be expressed more efficiently:
```
/* If some SPI0 pads are valid, */
#if SPI0_CLK_PORT >= 0 || SPI0_MOSI_PORT >= 0 || SPI0_MISO_PORT >= 0
/* but not all */
#if SPI0_CLK_PORT < 0 || SPI0_MOSI_PORT < 0 || SPI0_MISO_PORT < 0
#error Some SPI0 pad definitions are invalid
#endif
#define SPI0_PADS_VALID
#endif
```

And same for SPI1, then you can add this coherency check at the end:
```
#ifdef SPI_DEFAULT_INSTANCE
#if SPI_DEFAULT_INSTANCE == 0
#ifndef SPI0_PADS_VALID
#error SPI_DEFAULT_INSTANCE is set to SPI0, but its pads are not valid
#endif
#elif SPI_DEFAULT_INSTANCE == 1
#ifndef SPI1_PADS_VALID
#error SPI_DEFAULT_INSTANCE is set to SPI1, but its pads are not valid
#endif
#else
#error Invalid SPI_DEFAULT_INSTANCE: valid values are 0 and 1
#endif
#endif
```

cpu/cc2538/dev/spi.c
```diff
@@ -41,102 +41,337 @@
...
+  int8_t pin;
+} spi_pad_t;
+typedef struct {
+  uint32_t base;
+  uint32_t ioc_ssirxd_ssi;
+  uint32_t ioc_pxx_sel_ssi_clkout;
+  uint32_t ioc_pxx_sel_ssi_txd;
+  spi_pad_t clk;
+  spi_pad_t tx;
+  spi_pad_t rx;
```
bthebaudeau: Ditto for mosi / miso.

cpu/cc2538/dev/spi.c
```diff
@@ -41,102 +41,337 @@
...
+} spi_cfg_t;
```
bthebaudeau: Please name this `spi_regs_t` for consistency with the UART driver. This is mostly the definition of the registers to use for one SPI instance, even if for the pads this comes from a configuration.

cpu/cc2538/dev/spi.c
```diff
@@ -41,102 +41,337 @@
...
+/*---------------------------------------------------------------------------*/
+static const spi_cfg_t spi_cfg[SSI_INSTANCE_COUNT] = {
```
bthebaudeau: Ditto: `spi_regs_t spi_regs`.

cpu/cc2538/dev/spi.c
```diff
@@ -41,102 +41,337 @@
...
+             uint32_t clock_phase, uint32_t data_size)
+{
+  spix_set_mode(SPI_DEFAULT_INSTANCE, frame_format, clock_polarity,
+                clock_phase, data_size);
+}
+void
+spi_cs_init(uint8_t port, uint8_t pin)
+{
+  spix_cs_init(port, pin);
+}
```
bthebaudeau: Only `spi_init()` above is defined by the core SPI API. The other SPI functions above are legacy CC2538 functions that can be removed.

cpu/cc2538/dev/spi.c
```diff
@@ -41,102 +41,337 @@
...
+/*---------------------------------------------------------------------------*/
+void
+spix_init(uint8_t instance)
```
bthebaudeau: Use `spi` rather than `instance`. It's shorter and more consistent with the UART driver.

cpu/cc2538/dev/spi.c
```diff
@@ -41,102 +41,337 @@
...
+  spix_enable(instance);
+
+  cfg = &spi_cfg[instance];
+
+  if(cfg->clk.port < 0) {
+    /* Port / pin configuration invalid. We checked for completeness
+       above. If clk.port is < 0, this means that all other defines are
+       < 0 as well */
+    return;
+  }
```
bthebaudeau: Preliminary validity checks must be performed before actually doing anything with the hardware, i.e. before calling `spix_enable()`.

cpu/cc2538/dev/spi.c
```diff
@@ -41,102 +41,337 @@
...
+
   /* Enable the clock for the SSI peripheral */
-  REG(SYS_CTRL_RCGCSSI) |= 1;
+  if(instance == 0) {
+    /* Enable the clock for the SSI peripheral */
+    REG(SYS_CTRL_RCGCSSI) |= 1;
+  } else {
+    REG(SYS_CTRL_RCGCSSI) |= 2;
+  }
```
bthebaudeau: Don't make the code overcomplicated. Just do this:
```
REG(SYS_CTRL_RCGCSSI) |= 1 << spi;
```

cpu/cc2538/dev/spi.c
```diff
@@ -41,102 +41,337 @@
...
+    return;
+  }
+
   /* Gate the clock for the SSI peripheral */
-  REG(SYS_CTRL_RCGCSSI) &= ~1;
+  if(instance == 0) {
+    REG(SYS_CTRL_RCGCSSI) &= ~1;
+  } else {
+    REG(SYS_CTRL_RCGCSSI) &= ~2;
+  }
```
bthebaudeau:
```
REG(SYS_CTRL_RCGCSSI) &= ~(1 << spi);
```

cpu/cc2538/spi-arch.h
```diff
@@ -34,63 +34,173 @@
...
+#else
+#define SPI_DEFAULT_INSTANCE      0
+#endif
+/*---------------------------------------------------------------------------*/
+/* We define a new set of "SPIX" macros which will then be mapped to the
+   corresponding macros defined later on */
+#define SPIX_WAITFORTxREADY(x)      SPI##x##_WAITFORTxREADY()
+#define SPIX_WAITFOREOTx(x)       SPI##x##_WAIT_FOREOTx()
+#define SPIX_WAITFOREORx(x)       SPI##x##_WAITFOREORx()
+#define SPIX_FLUSH(x)         SPI##x##_FLUSH()
```
bthebaudeau: You should add a macro level here. With your implementation, the following code:
```
#define ENC28J60_CONF_SPI 0
SPIX_WAITFORTxREADY(ENC28J60_CONF_SPI)
```

would be expanded as `SPIENC28J60_CONF_SPI_WAITFORTxREADY()` instead of as the expected `SPI0_CONF_SPI_WAITFORTxREADY()`.

To fix this:
```
#define TPASTE3(a, b, c) a##b##c
#define SPIX_WAITFORTxREADY(x) TPASTE3(SPI, x, _WAITFORTxREADY)()
```

A set of `TPASTEn()` macros (`n` from 2 to at least 3 for now) would be added to a core header file (one to put common useful macros and definitions, e.g. also MIN / MAX, etc.).

cpu/cc2538/spi-arch.h
```diff
@@ -34,63 +34,173 @@
...
 
-#define SPI_WAITFORTxREADY() do { \
-  while(!(REG(SSI0_BASE + SSI_SR) & SSI_SR_TNF)); \
+#define SPI_CS_CLR(port, pin)     SPIX_CS_CLR(port, pin)
+#define SPI_CS_SET(port, pin)     SPIX_CS_SET(port, pin)
+/*---------------------------------------------------------------------------*/
+/* New API macros */
+#define SPI0_WAITFORTxREADY() do { \
+    while(!(REG(SSI0_BASE + SSI_SR) & SSI_SR_TNF)) ; \
 } while(0)
```
bthebaudeau: You don't have to duplicate the code of these macros. You can just directly:
```
#define SPIX_WAITFORTxREADY(spi) do { \
  while(!(REG(TPASTE3(SSI, spi, _BASE) + SSI_SR) & SSI_SR_TNF)) ; \
} while(0)
```

And remove `SPI0_WAITFORTxREADY()` and `SPI1_WAITFORTxREADY()`.

cpu/cc2538/spi-arch.h
```diff
@@ -34,63 +34,173 @@
...
-
-#define SPI_TXBUF REG(SSI0_BASE + SSI_DR)
-
-#define SPI_RXBUF REG(SSI0_BASE + SSI_DR)
-
-#define SPI_WAITFOREOTx() do { \
-  while(REG(SSI0_BASE + SSI_SR) & SSI_SR_BSY); \
+#define SPI0_TXBUF REG(SSI0_BASE + SSI_DR)
+#define SPI0_RXBUF REG(SSI0_BASE + SSI_DR)
```
bthebaudeau: It's the same register, and these definitions are internal to the driver (they're exposed, but not supposed to be used elsewhere), so merge them like this:
```
#define SPIX_BUF(spi) REG(TPASTE3(SSI, spi, _BASE) + SSI_DR)
```

bthebaudeau:
I agree with @g-oikonomou: the legacy core SPI transfer API is ugly:
 - it's made from macros made from macros: not good for binary size and type checking,
 - it's byte-oriented and polling-driven: slow, wasting CPU time and power, binary even bigger.

For the new core SPI driver API, I would see the following specifications:
 - only functions (no macros),
 - API at buffer level (not at byte level),
 - single transfer function replacing read and write (the SPI hardware enforces simultaneous read and write, even if the software then decides to discard read or written data), with a NULL pointer if the software wants to only read or write,
 - choice between polling-, interrupt- and DMA-based implementations depending on hardware capabilities, i.e. blocking and non-blocking APIs, the latter with a function returning the status, or a callback function, or an OS event.

cpu/cc2538/dev/spi.c
uknoblic: Hi, what about SPI slave mode which might come at some day? RX / TX is unambiguous even in this case!

adamdunkels:
I really like the idea of updating the SPI API in the core code (which is pretty bad as it stands).

Agree with all points made by @bthebaudeau, but don't want the API to be really complicated so I'd advise against allowing too different implementations with a choice of OS events and/or callbacks. I'd stick with one API to make things simpler.

uknoblic:
One additional question with regard to the copyright header - would you propose to change the headers after having all these changes?

cpu/cc2538/dev/spi.c
bthebaudeau: > Hi, what about SPI slave mode which might come at some day? RX / TX is unambiguous even in this case! And even in the users guide + data sheet the pins are name Tx and Rx...

OK, that makes sense. If part of this code is reused one day for slave SPI, using TX / RX would indeed avoid switching the pad names in the initialization. They use TX / RX in the user's guide because this SSI IP is not SPI-only. So keeping your TX / RX naming is justified.

bthebaudeau:
@adamdunkels

> I'd advise against allowing too different implementations with a choice of OS events and/or callbacks. I'd stick with one API to make things simpler.

Yes, we have to make a choice between OS events and callbacks for a non-blocking API.

A weak blocking API could come on top of the non-blocking API as a convenience. Platforms not supporting interrupts or DMA for SPI could still overload it.

Actually, the blocking and non-blocking APIs could even be the same function. The distinction would be made when calling the SPI initialization function, telling the driver that it has to work in the chosen preferred mode if supported.

bthebaudeau:
@uknoblic

> One additional question with regard to the copyright header - would you propose to change the headers after having all these changes?

The changes are significant, so it would probably be fair to add your copyright to the header if you want to, but the original copyright should not be removed. @g-oikonomou: What do you think?

g-oikonomou:
>The changes are significant, so it would probably be fair to add your copyright to the header if you want to, but the original copyright should not be removed. @g-oikonomou: What do you think?

For files with extensive changes, yes I'd say add your own copyright line under the existing one.

@bthebaudeau #883

adamdunkels:
@bthebaudeau the blocking-API-with-a-non-blocking-implementation sounds great to me. I'd prefer callbacks to OS evens too.

uknoblic:
Hi, maybe out of topic (no idea where to discuss it) - but with regard to "blocking" and "callbacks": the same counts for the transmit function of the radio. How to handle slow data rates or long wake-up preambles? I thing we need a solution there, too.

cpu/cc2538/spi-arch.h
```diff
@@ -34,66 +38,131 @@
...
-
-#define SPI_WAITFORTxREADY() do { \
-  while(!(REG(SSI0_BASE + SSI_SR) & SSI_SR_TNF)); \
+/*---------------------------------------------------------------------------*/
+/* The default SPI instance to use. You can configure either instance 0 or 1 */
+#ifdef SPI_CONF_DEFAULT_INSTANCE
+#define SPI_DEFAULT_INSTANCE			SPI_CONF_DEFAULT_INSTANCE
+#else
+#define SPI_DEFAULT_INSTANCE			0
+#endif
```
bthebaudeau: The dependency on `board.h` might be an issue here, because the definitions would change depending on whether or not it has been included before `spi-arch.h`. It might be good to `#include "contiki.h"` above.

cpu/cc2538/spi-arch.h
```diff
@@ -34,66 +38,131 @@
...
+/*---------------------------------------------------------------------------*/
+#if SPI_DEFAULT_INSTANCE > (SSI_INSTANCE_COUNT - 1)
+#error Invalid SPI_DEFAULT_INSTANCE: valid values are 0 and 1
+#endif
+/*---------------------------------------------------------------------------*/
+/* New API macros */
+#define TPASTE3(a, b, c)				a##b##c
```
bthebaudeau: This macro could be moved to `/core/sys/cc.h` as previously said.

uknoblic:
Some of the cc2538 related travis builds fail because the corresponding macros (SPI1_TX_PORT and so on) are missing. How do i have to handle this? Do i have to modify these examples? I provided the coressponding macros in spi-arch.h but removed them as you can remember.

bthebaudeau:
You're talking about the job #2264.14, right?

The root cause of the failure seems to be that `SPI_DEFAULT_INSTANCE` is always defined, even if the SPI is unused:
```
../../../cpu/cc2538/dev/spi.c:142:2: error: #error SPI_DEFAULT_INSTANCE is set to SPI0, but its pads are not valid
```

So you could just define `SPI_DEFAULT_INSTANCE` to `SPI_CONF_DEFAULT_INSTANCE` if the latter is defined (no change), but do not define it otherwise (instead of defining it as `0` in this case).

If the SPI macros are not used, everything should build. If they are used, then the build will fail, but `SPI_CONF_DEFAULT_INSTANCE` would be needed in this case anyway. The only issue is that the error message would not be straightforward in this case.

cpu/cc2538/spi-arch.h
uknoblic: And renamed to CC_CONCAT3?

core/sys/cc.h
```diff
@@ -129,12 +129,16 @@
...
 #define CC_CONCAT2(s1, s2) s1##s2
 /**
- * A C preprocessing macro for concatenating to
- * strings.
+ * A C preprocessing macro for concatenating two strings.
  *
  * We need use two macros (CC_CONCAT and CC_CONCAT2) in order to allow
  * concatenation of two \#defined macros.
  */
 #define CC_CONCAT(s1, s2) CC_CONCAT2(s1, s2)
```
bthebaudeau: The name of this macro is not well chosen since the same kind could be added for the 3-token version. This macro just makes it possible to use `CC_CONCAT2()` from anywhere, even outside the context of another macro having `s1` and `s2` as parameters.

Maybe it could be renamed to something like `CC_CONCAT_EXT_2`.

core/sys/cc.h
```diff
@@ -129,12 +129,16 @@
...
 
+/**
+ * A C preprocessing macro for concatenating three strings.
```
bthebaudeau: Actually, it's not about strings. `##` is the C preprocessor token-pasting operator. So this macro and those above concatenate preprocessing tokens.

cpu/cc2538/spi-arch.h
```diff
@@ -34,66 +38,130 @@
...
+    GPIO_CLR_PIN(GPIO_PORT_TO_BASE(port), GPIO_PIN_MASK(pin)); \
 } while(0)
-
-#define SPI_CS_SET(port, pin) do { \
-  GPIO_SET_PIN(GPIO_PORT_TO_BASE(port), GPIO_PIN_MASK(pin)); \
+#define SPIX_CS_SET(port, pin) do { \
+    GPIO_SET_PIN(GPIO_PORT_TO_BASE(port), GPIO_PIN_MASK(pin)); \
 } while(0)
 /*---------------------------------------------------------------------------*/
+/* Deprecated macros provided for compatibility reasons */
```
bthebaudeau:
```
+#ifdef SPI_DEFAULT_INSTANCE
```

cpu/cc2538/spi-arch.h
```diff
@@ -34,66 +38,130 @@
...
+#define SPI_RXBUF						SPI_TXBUF
+#define SPI_WAITFOREOTx()				SPIX_WAITFOREOTx(SPI_DEFAULT_INSTANCE)
+#define SPI_WAITFOREORx()				SPIX_WAITFOREORx(SPI_DEFAULT_INSTANCE)
+#ifdef SPI_FLUSH
+#error You must include spi-arch.h before spi.h for the CC2538
+#else
+#define SPI_FLUSH()						SPIX_FLUSH(SPI_DEFAULT_INSTANCE)
+#endif
+#define SPI_CS_CLR(port, pin)			SPIX_CS_CLR(port, pin)
+#define SPI_CS_SET(port, pin)			SPIX_CS_SET(port, pin)
```
bthebaudeau:
```
+#endif
```

uknoblic:
Hopefully everthing is ok now, thank you for your patience! I did not modify CC_CONCAT() as it it used already but added CC_CONCAT_EXT2 + CC_CONCAT_ETX3. Now that the cc2538dk demos compile, i would be happy if someone could help me to find out why the travis build fails...

uknoblic:
Now everything compiles well but the travis build for the exp5438 platform. Could somebody help me here please? I get the same error when i try to compile this example locally at my pc using the official contikit clone...

adamdunkels:
The problem is that the `example/webserver-ipv6` example has gotten too large for the `exp5438` platform. This one has been problematic before and it is running dangerously close to the memory limit every time. I think we should disable it, at least for now. I created a pull request against your branch: https://github.com/uknoblic/contiki/pull/1 If you merge it, the commit will automatically make its way into this pull request.

bthebaudeau:
:+1:

PR #999, opened 2015-03-24
RDNSS support, again

adamdunkels:
This is an update of https://github.com/contiki-os/contiki/pull/643, which introduces a nice new API for the DNS resolver library and adds support for RDNSS. Unfortunately the original pull request failed on some platforms as the code was not optimized for the case when there was only one DNS server configured.

This updated pull request adds a special-case optimization for when we only have one DNS server which makes the code smaller in that situation.

Ping @lebrush and @oliverschmidt.

### DIFF
--- a/apps/dhcp/dhcp.c
+++ b/apps/dhcp/dhcp.c
@@ -88,7 +88,7 @@ makestrings(void)
   uip_getdraddr(&addr);
   makeaddr(&addr, gateway);
 
-  addrptr = resolv_getserver();
+  addrptr = uip_nameserver_get(0);
   if(addrptr != NULL) {
     makeaddr(addrptr, dnsserver);
   }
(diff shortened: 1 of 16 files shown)


oliverschmidt:
Hi,

I see two perspectives here:

1. More or less every byte added to the Contiki core requires adjustments in the Contiki apps for the retro targets - either by reducing buffer sizes or removing functionality. Having the Travis builds fail only means that the issue is a major one. The apps _all_ use heap memory so only actually running the apps shows if there's still enough memory available.

2. Designing Contiki development around the retro targets for sure means to wag the dog. In fact I wasn't the driver behind adding retro builds to Travis in the first place.

Regards,
Oliver

adamdunkels:
I was the one who wanted to have the `cc65` builds in travis. I think it is very useful to have them there to avoid the system becoming too large. It isn't just the 6502 platforms where we need to fight against tight memory constraints, even in 2015. This specific pull request actually shows how useful it is to have them there: without `cc65`, the memory optimization in this pull request wouldn't have happened which would have caused memory problems on other platforms.

oliverschmidt:
Hi Adam,

Nice that you explicitly state it this way :-) Sometimes the GH threads feel a bit like it would be just for my personal benefit to keep memory consumption down...

Regards,
Oliver

adamdunkels:
This morning we had a build fail with an overflow of 3 bytes of RAM - on a modern state-of-the-art wireless system-on-a-chip. So keeping the memory consumption low still is pretty useful.

lebrush:
Hey! Cool for the memory fixtures for smallest platforms. I also think is a good idea to keep memory footprint to a minimum. However, sometimes there significant trade-off between code complexity (#if #else everywhere) and memory footprint. In any case, look good to me :+1:

nvt:
:+1:

PR #1001, opened 2015-03-24
Cleanup: remove the obsolete reference packetbuf

adamdunkels:
The `packetbuf` module contains a minor optimization that was added a long time ago with the hope that it would be helpful but never turned out to be very helpful: reference packetbufs. The idea was that the application code could optimize memory access by providing a reference to the data in the packetbuf, instead of copying the data into the buffer. This introduced additional complexity.

The only part of the code that ever used this feature was the Rime module called `rudolph0` which was designed to do single-hop dissemination of a binary image. This module was in itself never used much.

This pull request removes the reference packetbuf to clean up the code and reduce complexity.


simonduq:
excellent, I agree this was unneeded complexity :+1:

nvt:
The statement <code>RIMESTATS_ADD(queuebuf_outofbuf);</code> fails to compile in Travis.

PR #1003, opened 2015-03-24
New HTTP socket, DNS64, IP64 Cooja tests

adamdunkels:
This is a pretty large, but important, pull request. With this pull request, Contiki devices can now (finally) access the IPv4 Internet just like they would access nodes on their IPv6 mesh which is a pretty big step for Contiki.

The pull request includes:

* A new `http-socket` module that is able to do both GET and POST. Thanks to @bthebaudeau there is also support for resuming a download.
* A bunch of bugfixes to the `tcp-socket` module, which is used by the `http-socket` module
* A new DNS64 module for the existing IP64 module. This makes it possible to do IPv6 DNS requests in the IPv6 network that are translated into IPv4 DNS requests before they are sent out onto the Internet. The DNS replies are then translated into IPv6 replies before it hits the IPv6 network. The IPv6 mesh thinks it talks to IPv6 hosts, but in reality they talk to the Internet.
* A new `cooja-ip64` Cooja platform that acts as an IP64 router node, so that we can do automated tests with IPv6-mesh-to-Internet communication.
* A new regression test that uses the `http-socket` module, the IP64 and DNS64 modules, and the new `cooja-ip64` Cooja platform to ensure that the entire Contiki system works: from the IPv6 RPL mesh and all the way onto the Internet.
* A new RPL module, `rpl-dag-root`, that makes it really easy to start a DAG root, without having to worry about existing networks in the air. By default, the module waits for 20 seconds before becoming a RPL root, so that any existing RPL networks will be joined before becoming root.


g-oikonomou:
These are really cool features!

I've been developing an engine for unidots and http sockets will be extra handy. I also use NAT64 a lot so have an interest in DNS64, means I'll test it first chance.

Not sure I'll be testing the other features, so if nobody else does in the meantime we can always split the pull into two I guess

alignan:
Cool patch! I will also test the DNS64 right away, I have also a couple of
examples which gladly will make use of this :thumbsup:

On Tue, Mar 24, 2015 at 5:50 PM, George Oikonomou <notifications@github.com>
wrote:

> These are really cool features!
>
> I've been developing an engine for unidots and http sockets will be extra
> handy. I also use NAT64 a lot so have an interest in DNS64, means I'll test
> it first chance.
>
> Not sure I'll be testing the other features, so if nobody else does in the
> meantime we can always split the pull into two I guess
>
> —
> <https://github.com/contiki-os/contiki/pull/1003#issuecomment-85594933>.
>



-- 

  Antonio Liñan
  [image: http://]about.me/antoniolignan
  <http://about.me/antoniolignan?promo=email_sig>

adamdunkels:
Great that you're testing things right away! In terms of testing, the code itself is pretty well-tested (the `http-socket` code used in at least one commercial product on the market today and on its way out into several others and the DNS64 code is used in the upcoming Thingsquare platform, also used in several upcoming products) but the integration with the current contiki-os tree is new and needs a bit more work (e.g. travis fails at the moment).

@g-oikonomou what do you mean with splitting the pull request into two?

g-oikonomou:
I mean that if I test N of M features, if they are in a separate pull I can merge them instead of block waiting for the remaining M-N

adamdunkels:
Ah ok. But everything in this pull request pretty much depends on each other, so splitting into separate pull requests would be tricky.

g-oikonomou:
Ok well I need to look at the code, but surely HTTP and TCP sock changes should be very easy to factor out. No?

adamdunkels:
The regression test for the `http-socket` module needs the DNS64 code and the Cooja ip64 platform and everything in between to run. 

In theory would be possible to break this up into several, individually blocking, pull requests:
* HTTP socket, Cooja IP64 platform, and DNS64 as three separate pull requests first, wait for them to hopefully be merged
* IP64 example code with RPL DAG patch as two separate pull requests second, wait for them to hopefully be merged
* Then, finally, the regression test as the final pull request.

But this would take months, which wouldn't exactly be ideal.

adamdunkels:
Btw @g-oikonomou, what is unidots? Tried googling, didn't find much that seems relevant :)

g-oikonomou:
Yea I saw the typo after posting, but couldn't edit from mobile.

It's ubidots! http://ubidots.com

g-oikonomou:
First version on g-oikonomou/contiki btw

adamdunkels:
Ah, makes more sense! :)

examples/http-socket/http-example.c
```diff
@@ -0,0 +1,64 @@
...
+  } else if(e == HTTP_SOCKET_ABORTED) {
+    printf("HTTP socket error: aborted\n");
+  } else if(e == HTTP_SOCKET_HOSTNAME_NOT_FOUND) {
+    printf("HTTP socket error: hostname not found\n");
+  } else if(e == HTTP_SOCKET_CLOSED) {
+    printf("HTTP socket closed, %d bytes received\n", bytes_received);
+  } else if(e == HTTP_SOCKET_DATA) {
+    bytes_received += datalen;
+    printf("HTTP socket closed\n");
+  } else if(e == HTTP_SOCKET_DATA) {
```
g-oikonomou: There are two `} else if(e == HTTP_SOCKET_DATA) {` here

examples/ip64-router/ip64-router.c
```diff
@@ -0,0 +1,27 @@
+#include "contiki.h"
+#include "contiki-net.h"
+#include "ip64.h"
```
alignan: Should we use ```#include "net/ip64/ip64.h"``` instead of ```#include "ip64.h"```? fails to compile otherwise

examples/http-socket/http-example.c
adamdunkels: Ouch, nice catch! I'll push an update.

examples/ip64-router/ip64-router.c
adamdunkels: The platform should have a `MODULE += core/net/ip64` and with that, it *should* be enough to include only `"ip64.h"`. What platform are you compiling for?

adutze:
Hello.
Very nice addition to Contiki, thank you.
I have some comments on the http-socket module:

I think that calling the http_callback functions before removing the socket from the list is a bit dangerous when, for example the application is feeding http_sockets from a memory pool and it requires some clean up after closing the TCP connection:
+    call_callback(s, HTTP_SOCKET_ABORTED, NULL, 0);
+    removesocket(s);
+    printf("Aborted\n");

If the allocated memory for the socket is free() in the http calback, then removesocket(s) will use a NULL pointer as parameter and removesocket() is only deleting it from the internal list of http_sockets. I propose to reverse the order of the removesocket() and call_callback() functions to avoid this problem.

Another useful thing would be to add a callback also for HTTP_SOCKET_OK or add a new event like HTTP_SOCKET_CONNECTED with a callback when the TCP connection is connected. With this mechanism, for example, the radio could be kept ON until the http socket is closed/aborted/...

examples/ip64-router/ip64-router.c
alignan: Oops I forgot to throw in the module, I'm testing for the Z1 and Re-Mote platforms.

adamdunkels:
Changing the order of the `removesocket()` and `call_callback()` calls is probably a good idea, good input!

adamdunkels:
Does anyone know what is up with the Doxygen travis build? It shows hundreds of warnings that are unrelated to this pull request, then fails. I'm pretty clueless as to how fix it.

g-oikonomou:
Hundreds is not the problem (it's always been hundreds, albeit fewer hundreds nowadays!). The problem is 4 new warnings compared to the last known count.

Is the pull adding doxygen stuff?

g-oikonomou:
These doxygen output lines are you culprit

```
/home/travis/build/contiki-os/contiki/platform/cooja-ip64/contiki-cooja-ip64-main.c:301: warning: argument 'start' of command @param is not found in the argument list of Java_org_contikios_cooja_corecomm_CLASSNAME_getMemory(JNIEnv *env, jobject obj, jint rel_addr, jint length, jbyteArray mem_arr)
/home/travis/build/contiki-os/contiki/platform/cooja-ip64/contiki-cooja-ip64-main.c:301: warning: The following parameters of Java_org_contikios_cooja_corecomm_CLASSNAME_getMemory(JNIEnv *env, jobject obj, jint rel_addr, jint length, jbyteArray mem_arr) are not documented:
  parameter 'env'
  parameter 'obj'
  parameter 'rel_addr'
  parameter 'mem_arr'
/home/travis/build/contiki-os/contiki/platform/cooja-ip64/contiki-cooja-ip64-main.c:327: warning: argument 'start' of command @param is not found in the argument list of Java_org_contikios_cooja_corecomm_CLASSNAME_setMemory(JNIEnv *env, jobject obj, jint rel_addr, jint length, jbyteArray mem_arr)
/home/travis/build/contiki-os/contiki/platform/cooja-ip64/contiki-cooja-ip64-main.c:327: warning: The following parameters of Java_org_contikios_cooja_corecomm_CLASSNAME_setMemory(JNIEnv *env, jobject obj, jint rel_addr, jint length, jbyteArray mem_arr) are not documented:
  parameter 'env'
  parameter 'obj'
  parameter 'rel_addr'
```

adamdunkels:
Ah, I need to update the `doxygen.cnt` file to fix it - thanks!

g-oikonomou:
Or to fix the actual warning :)

adamdunkels:
In principle, I definitely agree that it is better to fix the warning by documenting stuff, but in this case the offending function belongs to the Dark Secrets of Cooja and is probably better left untouched :scream_cat:

adamdunkels:
I rebased this, squashed a few intermediate bug commits, and travis is green now!

nvt:
:+1:

Issue #998, opened 2015-03-23
Use Additive Offsets for CC2538 Register Access


g-oikonomou:
Fork from #995 : 
> Also, using `REG(regs->base + OFFSET)` instead of the current `REG(regs->base | OFFSET)` found in the current CC2538 code base (unrelated to your PR) should generate more optimal assembly code. `|` is broken by design if the base is not aligned enough for the offset, and CPU instructions support addressing modes using additive offsets only (not OR-ed offsets), so computing a OR for each address used in the IP address range must be more expensive than the standard base `+` offset (because the compiler does not know the alignment of the variable base, so it can not optimize the | into a `+`), except if base is a constant known at compile time. This should be changed in the UART code (and everywhere else it appears), and the SPI code is OK here.

http://infocenter.arm.com/help/topic/com.arm.doc.dui0552a/BABJGHFJ.html

So the UART code is the obvious candidate. Where else do we need this?

bthebaudeau:
Yes, UART obviously, then this can be extended to the whole CC2538 code base as a coding rule, even if most other instances (if any) are probably compile-time constants, in which case the generated binary will be unchanged.

g-oikonomou:
#1009

PR #1009, opened 2015-03-28
Use additive offsets

bthebaudeau:
OR-ing an offset to a base address instead of adding it is dangerous
because it can only work if the base address is aligned enough for the
offset.

Moreover, if the base address or the offset has a value unknown at
compile time, then the assembly instructions dedicated to 'base +
offset' addressing on most CPUs can't be emitted by the compiler because
this would require the alignment of the base address against the offset
to be known in order to optimize 'base | offset' into 'base + offset'.
In that case, the compiler has to emit more instructions in order to
compute 'base | offset' on most CPUs, e.g. on ARM, which means larger
binary size and slower execution.

Hence, replace all occurrences of 'base | offset' with 'base + offset'.
This must become a coding rule.

Here are the results for the cc2538-demo example:
 - Compilation of uart_init():
    * before:
        REG(regs->base | UART_CC) = 0;
        200b78:	f446 637c 	orr.w	r3, r6, #4032	; 0xfc0
        200b7c:	f043 0308 	orr.w	r3, r3, #8
        200b80:	2200      	movs	r2, #0
        200b82:	601a      	str	r2, [r3, #0]

    * now:
        REG(regs->base + UART_CC) = 0;
        200b7a:	2300      	movs	r3, #0
        200b7c:	f8c4 3fc8 	str.w	r3, [r4, #4040]	; 0xfc8

 - Size of the .text section:
    * before:	0x4c7c
    * now:	0x4c28
    * saved:	84 bytes

This fixes issue #998.


g-oikonomou:
Splendid!

PR #1011, opened 2015-03-29
Update #if guard to match header filename

g-oikonomou:
Very simple change to include guard so that it matches the filename it appears in.


adamdunkels:
:+1:

PR #1013, opened 2015-03-30
Remove unnecessary #if guards

### LINKED ISSUE #935: Code / Comment conflict for UIP_CONF_IPV6
The code at line 78 in uip.c does not match the comment that goes with it.  The compiler directive simply checks if the parameter UIP_CONF_IPV6 is true or false.  The comment, however, says that it is determining if the parameter is defined or not.  Since I do not yet have a working version of the code, I cannot check if this works by accident in some cases and doesn't in others, or if this is an acceptable way of describing the code.  To a noob like me, this looks wrong.

g-oikonomou:
`net/ipv4/uip.c` will not even be included in the list of sources in the first place if we are building with NETSTACK_CONF_WITH_IPV6 (CONTIKI_WITH_IPV6). Thus, there is no longer a need to wrap the entire source code in an #if guard.

Closes #935

### DIFF
--- a/core/net/ipv4/uip.c
+++ b/core/net/ipv4/uip.c
@@ -75,18 +75,9 @@
 #include "net/ipv4/uip_arp.h"
 #include "net/ip/uip_arch.h"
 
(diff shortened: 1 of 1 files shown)


nvt:
:+1:

PR #1016, opened 2015-04-03
Remove all the DODAG from the instance when the node is set as Root

laurentderu:
When a node becomes a DODAG root within a given instance, it must leave any DODAG it has joined previously, not only the current DODAG and remove any references to the known DODAGs.


nvt:
:+1:

PR #1017, opened 2015-04-03
Do not delete instance when a new DODAG can not be created

laurentderu:
When the maximum number of DODAG for a given instance is reached, the current code delete that instance.


nvt:
:+1:

Issue #921, opened 2014-12-17
CC2538 - GPIO Power-up interrupt doesn't execute the callback function


darkopetrovic:
I've disabled ContikiMAC with `NETSTACK_RDC.off(0)`. Then there is no more automatic wake-up from the SM Timer. I've a button connected to a pin on the port D to wake-up the SoC when it is pressed. I added in the configuration of button:

```cpp
GPIO_POWER_UP_ON_FALLING(BUTTON_USER_PORT, BUTTON_USER_PIN_MASK);
GPIO_ENABLE_POWER_UP_INTERRUPT(BUTTON_USER_PORT, BUTTON_USER_PIN_MASK);
``` 
The SoC now wakes-up correctly from LPM2. 

But when the ISR of the port `gpio_port_d_isr` is triggered, this latter doesn't execute the callback function associated with the pin because the GPIO_MIS register is not updated at this time (since the source of the interrupt is coming from the power-up interrupt and not from the general interrupt: they are separated). 

Then in order that this interrupt handler can see the source of the interrupt is coming from the power-up interrupt, we need to look at the GPIO_IRQ_DETECT_ACK register too, thus I replaced:

```
notify(REG(GPIO_D_BASE | GPIO_MIS), GPIO_D_NUM);
```

by 

```
notify(REG(GPIO_D_BASE | GPIO_MIS) | REG(GPIO_D_BASE | GPIO_IRQ_DETECT_ACK)>>24 , GPIO_D_NUM);
```

There is no such problem when the RDC is activated because the SoC wakes-up continuously and can see the general interrupt the next time it wakes-up. 

I know that it's not in the philosophy of Contiki to disable the RDC but in the case one would save power from the battery, we should be able to deactivate the RDC when it is not needed and reactivate it for example by pressing an external button. Therefore, in order to do that properly, we need to check equally the `GPIO_IRQ_DETECT_ACK` register in the GPIO interrupt handler.  

Reference: http://e2e.ti.com/support/wireless_connectivity/f/158/p/320838/1376612

g-oikonomou:
This should be fixed now (#1031)

Issue #1019, opened 2015-04-07
CC2538 GPIO interrupt handling


uknoblic:
I am working on a driver for the CC1200 supporting more than 127 bytes payload (802.15.4g). I therefore have to use two GPIO interrupts on seperate pins. I now face the problem that one interrupt is fired, and within the interrupt service routine the next edge on the other GPIO occurs. Due to the handling in the GPIO module this interrupt is lost (ALL pending interrupts are cleared inside the port interrupt handler). Any proposal how to solve this? The best way would be just to reset the flags for the interrupts served?

PR #1031, opened 2015-04-20
cc2538: gpio: Improve and fix IRQ management

bthebaudeau:
This PR factors out duplicated ISR code, and it fixes several cases of missed interrupts.

This fixes issues #921 and #1019.

@g-oikonomou: This has been build-tested, but *not* runtime-tested.
@darkopetrovic: Please test that this works for your use case.


g-oikonomou:
Tested successfully with Srf+CC2538EM. I have no way of testing whether #921 is indeed fixed, but I'll merge this anyway since this is an improvement regardless.

:+1:

PR #1039, opened 2015-04-28
CC2538 I2C bug in clock computation

uknoblic:
There is a bug in the sys clock computation of the I2C module. Please check.


bthebaudeau:
We're only interested in your last commit for this PR. It is completely unrelated to PR #995. You should just keep the last commit based on contiki-os:master in your uknoblic:cc2538_i2c branch. You can use a forced push to completely recreate your branch without closing the PR. That would make the PR clearer.

Apart from that: :+1:

alignan:
:+1: for the I2C part.

PR #1040, opened 2015-04-28
Improve multichannel support in Cooja

atiselsts:
* Fix CCA detection in Cooja in the case when the receiver switches on the right channel during an ongoing transmission. Always add a connection on transmission, even when the channel is not correct. Initially the connection is in a dormant state; this mimics what Cooja is doing when the receiver radio is turned off; when the receiver is turned on and switched to the right channel, `updateSignalStrengths()` is called, and the connection starts to receive PHY-level traffic.

* Add "channel" property for DGRM edges.

* Avoid cross-channel interference on DGRM and MRM radio mediums

Test applications are available here: https://github.com/atiselsts/contiki/tree/master/examples/multichannel-test

### DIFF
--- a/tools/cooja/java/org/contikios/cooja/radiomediums/AbstractRadioMedium.java
+++ b/tools/cooja/java/org/contikios/cooja/radiomediums/AbstractRadioMedium.java
@@ -236,8 +236,9 @@ public abstract class AbstractRadioMedium extends RadioMedium {
 				case RECEPTION_STARTED:
 				case RECEPTION_INTERFERED:
 				case RECEPTION_FINISHED:
+					break;
+
 				case UNKNOWN:
-					return;
 				case HW_ON: {
 					/* Update signal strengths */
 					updateSignalStrengths();
@@ -292,7 +293,7 @@ public abstract class AbstractRadioMedium extends RadioMedium {
 				break;
 				case TRANSMISSION_FINISHED: {
 					/* Remove radio connection */
-					
+
 					/* Connection */
 					RadioConnection connection = getActiveConnectionFrom(radio);
 					if (connection == null) {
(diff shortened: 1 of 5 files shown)


tools/cooja/java/org/contikios/cooja/radiomediums/AbstractRadioMedium.java
```diff
@@ -292,7 +293,7 @@ public void execute(long t) {
 				break;
 				case TRANSMISSION_FINISHED: {
 					/* Remove radio connection */
-					
+
```
cmorty: Reformatting empty lines should be avoided.

tools/cooja/java/org/contikios/cooja/radiomediums/AbstractRadioMedium.java
```diff
@@ -236,8 +236,9 @@ public void update(Observable obs, Object obj) {
 				case RECEPTION_STARTED:
 				case RECEPTION_INTERFERED:
 				case RECEPTION_FINISHED:
+					break;
+
 				case UNKNOWN:
-					return;
```
cmorty: Why remove the return?

cmorty:
I like this one. Could you also add your examples as test-cases (in an extra patch).

simonduq:
I can't tell about the code but the problem addressed in this PR is one I've also seen before -- a fix is very welcome.

tools/cooja/java/org/contikios/cooja/radiomediums/AbstractRadioMedium.java
beshrns: I believe that mspsim fires an `UNKOWN` event when channel is changed in cc2420
https://github.com/contiki-os/contiki/blob/master/tools/cooja/apps/mspsim/src/org/contikios/cooja/mspmote/interfaces/Msp802154Radio.java#L159

beshrns:
This PR makes a lot of sense IMO and it eliminates the issue to a great level. 
I believe, however, that there still could be incorrect inter-channel interference or missed packets due to race conditions inherent in the Observer model and the asynchronous nature of radio events.

fros4943:
Very nice! The underlying reason for these issues is that the majority of the radio medium code was written pre radio channel support. Ideally, the AbstractRadioMedium should be rewritten with multiple radio channels, background noise interference, and capture effect in mind. Nevertheless, this PR goes a long way to fix a set of annoying bugs.

:+1: Merging

PR #1042, opened 2015-04-29
CC2538: added support for SSI1

uknoblic:
Added support for two SPI ports for cc2538 similar to uart modul.
This is a re-pull request for the already approved pull request based on a new branch in my own fork. Hopefully i don't create to much confusion...

### DIFF
--- a/core/dev/spi.h
+++ b/core/dev/spi.h
@@ -57,36 +57,36 @@ extern unsigned char spi_busy;
 void spi_init(void);
 
 /* Write one character to SPI */
-#define SPI_WRITE(data)                         \
-  do {                                          \
-    SPI_WAITFORTx_BEFORE();                     \
-    SPI_TXBUF = data;                           \
-    SPI_WAITFOREOTx();                          \
+#define SPI_WRITE(data) \
+  do { \
+    SPI_WAITFORTx_BEFORE(); \
+    SPI_TXBUF = data; \
+    SPI_WAITFOREOTx(); \
   } while(0)
 
 /* Write one character to SPI - will not wait for end
    useful for multiple writes with wait after final */
-#define SPI_WRITE_FAST(data)                    \
-  do {                                          \
-    SPI_WAITFORTx_BEFORE();                     \
-    SPI_TXBUF = data;                           \
-    SPI_WAITFORTx_AFTER();                      \
+#define SPI_WRITE_FAST(data) \
+  do { \
+    SPI_WAITFORTx_BEFORE(); \
+    SPI_TXBUF = data; \
+    SPI_WAITFORTx_AFTER(); \
   } while(0)
 
 /* Read one character from SPI */
-#define SPI_READ(data)   \
-  do {                   \
-    SPI_TXBUF = 0;       \
-    SPI_WAITFOREORx();   \
-    data = SPI_RXBUF;    \
+#define SPI_READ(data) \
+  do { \
+    SPI_TXBUF = 0; \
+    SPI_WAITFOREORx(); \
+    data = SPI_RXBUF; \
   } while(0)
 
 /* Flush the SPI read register */
 #ifndef SPI_FLUSH
 #define SPI_FLUSH() \
-  do {              \
-    SPI_RXBUF;      \
-  } while(0);
+  do { \
+    SPI_RXBUF; \
+  } while(0)
 #endif
 
 #endif /* SPI_H_ */
--- a/core/sys/cc.h
+++ b/core/sys/cc.h
@@ -129,12 +129,18 @@
 
 #define CC_CONCAT2(s1, s2) s1##s2
 /**
- * A C preprocessing macro for concatenating to
- * strings.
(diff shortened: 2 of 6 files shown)


bthebaudeau:
See conversation in #995.

I see that, since #995, you have added the `ssi_cprs_cpsdvsr` configuration. That's good, except that:
 - the indentation is now broken on the two lines where you initialize `.ssi_cprs_cpsdvsr`,
 - where you test that this configuration is valid, a negative value could be passed, so you should change these tests to:
```
#if (SPI0_CPRS_CPSDVSR & 1) || SPI0_CPRS_CPSDVSR < 2 || SPI0_CPRS_CPSDVSR > 254
```
and
```
#if (SPI1_CPRS_CPSDVSR & 1) || SPI1_CPRS_CPSDVSR < 2 || SPI1_CPRS_CPSDVSR > 254
```

You have also removed `Germany` from your copyright in `spi-arch.h` for consistency, which is OK.

bthebaudeau:
:+1:

g-oikonomou:
Turns out I still have a couple of questions about this one after all

core/dev/spi.h
```diff
@@ -54,39 +54,46 @@
 
 extern unsigned char spi_busy;
 
+/* Deprecated */
 void spi_init(void);
+/*---------------------------------------------------------------------------*/
+/* New SPI API supporting mutliple instances.
+   See /cpu/cc2538/spi-arch.h + cpu/cc2538/dev/spi.c for an example on
+   how this new API could be implemented. */
+void spix_init(uint8_t instance);
```
g-oikonomou: Why are we declaring this function's prototype here as well as in spi-arch.h?

g-oikonomou:
The discussion in #995 was in part about a new SPI API to replace the one in core.

This pull here is not proposing a new API, it merely extends CC2538 to support multiple instances. This is a very welcome but entirely platform-specific addition, and I don't see the reason to change anything in the core spi.h. The addition of the `spix_init` prototype is not strictly required and all other changes seem to be merely the result of an uncrustify run.

I think spi.h should be left alone until we have come up with a proposal for a new API. The platform code looks good to me!

g-oikonomou:
And by the way, I can't think how this pull could have anything to do with the webserver example breaking on travis.

core/dev/spi.h
bthebaudeau: Good point. Only the CPU-specific declarations and definitions should be in `spi-arch.h`. The official SPI API should be in `spi.h`. The `spix` API is only a temporary thing before we come up with a better SPI API, so maybe it could be kept in `spi-arch.h`.

bthebaudeau:
> The discussion in #995 was in part about a new SPI API to replace the one in core.
> 
> This pull here is not proposing a new API, it merely extends CC2538 to support multiple instances. This is a very welcome but entirely platform-specific addition, and I don't see the reason to change anything in the core spi.h. The addition of the spix_init prototype is not strictly required and all other changes seem to be merely the result of an uncrustify run.
> 
> I think spi.h should be left alone until we have come up with a proposal for a new API. The platform code looks good to me!

I think that @uknoblic's need was to use these multiple SPI instances from device drivers. Until these drivers are mainlined, I agree that changes to `spi.h` could be postponed and eventually discarded at the benefit of a new SPI API.

> And by the way, I can't think how this pull could have anything to do with the webserver example breaking on travis.

Travis looks good. Are you speaking about an issue that got fixed after restarting Travis?

g-oikonomou:
> Travis looks good. Are you speaking about an issue that got fixed after restarting Travis?

The `webserver-ipv6/exp5438` test started breaking at some point close to the time when this pull was opened, and this pull's history contains a commit which disables the test in question. I'm not opposed to disabling the test here, I'm just scratching my head a bit as to exactly what happened.

bthebaudeau:
:+1:

uknoblic:
Maybe somebody could check why the travis build failed?

bthebaudeau:
All your commits have the same purpose, so please squash them into a single one with a proper commit message. Just keep Adam's commit untouched and separate at the end since it's unrelated.

uknoblic:
:sweat_smile:

uknoblic:
Pushed the wrong button...

bthebaudeau:
Good for the rebase + squash.

I had previously missed tabs in `board.h`. Please replace them with spaces (keeping the commit log as it is now, i.e. replace the first commit, do not add a new one for that).

alignan:
:1: for me, as soon as @bthebaudeau comments about the tabs in `board.h` are addressed I'm in favor of finally merging, thanks for the long work on this one!

bthebaudeau:
Rebase now needed too because of a small change in `spi-arch.h` (`SPI_FLUSH()`) merged into master.

bthebaudeau:
:+1:

You've just lost Adam's commit in the process, but don't worry, I'll create a new PR for it.

bthebaudeau:
I see that Adam's patch had actually been merged through another PR, so you did good.

PR #1044, opened 2015-04-30
cc2538: allow for configuration of processor clock speed

uknoblic:
In order to speed up SPI + process clock in general for the CC2538, i would like to be able to configure the clock divisors. Could you please comment on this? The following modules are also affected of this modification: clock.c. uart.c (something else?), i will commit them once we aggree on the procedure.


g-oikonomou:
From my PoV, that would be a welcome change

bthebaudeau:
That would be useful. A few remarks:
 - The indentation of `val =` is now broken.
 - The two `>> 8` are useless.
 - Add the following to `sys-ctrl.h`, after your definitions of `SYS_CTRL_SYS_DIV` and `SYS_CTRL_IO_DIV`:
```
#define SYS_CTRL_SYS_CLOCK (SYS_CTRL_32MHZ >> SYS_CTRL_SYS_DIV)
#define SYS_CTRL_IO_CLOCK (SYS_CTRL_32MHZ >> (8 + SYS_CTRL_IO_DIV))
```
 - `get_sys_clock()` should be moved from `i2c.c` to `sys-ctrl.c`, and use `SYS_CTRL_32MHZ >> (REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SYS_DIV)` instead of `SYS_CTRL_32MHZ / (1 << (REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SYS_DIV))` (could be faster and smaller).
 - Add a `get_io_clock()` to `sys-ctrl.c` (could be used for SSI and UART).
 - `spi.c`: it would be better to deduce CPSR from the I/O clock and a configured target SPI clock, rather than directly configuring a blind value for CPSR.
 - `uart.c` should define `UART_CLOCK_RATE` to `SYS_CTRL_IO_CLOCK`. It could also compute the baud rate register values at runtime from `get_io_clock()` like `i2c.c` with the system clock, but that would be mostly useless until `sys-ctrl.c` provides a means of dynamically changing the system and I/O clocks, which means that `i2c.c` could also use `SYS_CTRL_SYS_CLOCK` instead of `get_sys_clock()` for now, but this is not required.
 - `clock.c`:
    * Define `RELOAD_VALUE` to `(SYS_CTRL_SYS_CLOCK / CLOCK_CONF_SECOND - 1)`. That's fine for a system clock down to 2 MHz, but then this will be less accurate, in which case a `#warning` or a `#error` must be issued by `clock.c`.
    * Set `TAPR` to `SYS_CTRL_SYS_CLOCK / 1000000 - 1`. The system clock has to be >= 1 MHz for this to work, so `clock.c` must issue a `#error` if this is not the case.
    * Remove the references to 16 MHz in the comments about the system clock.
 - `lpm.c`: You have to take care of the erratum "1.5 Possible Incorrect Value of Clock Dividers after PM2 and PM3" (see http://www.ti.com/lit/er/swrz045a/swrz045a.pdf) if the  system or I/O clock is set to 32 MHz.

uknoblic:
Sorry Benoit, just pushed my new modifications without having read your comments. Will update them as soon as possible. Some of your comments are already considered.

uknoblic:
Maybe you could especially have a look at the lpm module. I decided to use defines to do all of the stuff as up to now we don't change the speed dynamically. I would also propose to use this approach for the uart, spi + i2c module (no need for sys_ctrl_get_sys_clock() + sys_ctrl_get_io_clock()). I don't have my hardware avialable to check, but maybe somebody could tell me the difference between the IO and the SYS clock for UART + SPI module? To my understanding both modules use the sys clock (as far as i can remember i also check this in hardware...)

bthebaudeau:
@uknoblic

I'll review your changes later.

> I don't have my hardware avialable to check, but maybe somebody could tell me the difference between the IO and the SYS clock for UART + SPI module? To my understanding both modules use the sys clock (as far as i can remember i also check this in hardware...)

According to §7.3 in the User's Guide (http://www.ti.com/lit/ug/swru319c/swru319c.pdf), the system clock is simply not applicable to the SSI and UART modules for baud rate. It's the I/O clock in this case:
```
SYS_DIV is the frequency setting for the system clock and all module core clocks not
supported by IO_DIV. IO_DIV is the frequency setting for baud rate clocks for the SSI and
UART modules. There is only one IO_DIV in the system, which is used by all SSI and UART
instances.
```

Though, the chapters 19 (SSI) and 20 (UART) mention the system clock, not the I/O clock. But 'system clock' could be used as a generic expression here, rather than meaning the system clock against the I/O clock. I'd rather trust the chapter 7 here, knowing how such manuals are made. In any case, it would be worth checking this on hardware.

@g-oikonomou: Do you have hardware to check this? With the current mainline master branch, you could just change IO_DIV in sys-ctrl.c (e.g. to 8 MHz), and see if the text printed on UART becomes garbage, which would mean that chapter 7 is correct. If not, then this should happen if SYS_DIV is slowed down, which would mean that chapter 7 is wrong.

g-oikonomou:
Regarding the LPM module, I've got a big pull to open over the next few days for the CC26xx and the respective LPM module uses some good ideas that we can copy over at a later stage.

I'll test the UART thing, yes. Hopefully over the weekend

cpu/cc2538/clock.c
```diff
@@ -62,7 +62,21 @@
 #include <stdint.h>
 /*---------------------------------------------------------------------------*/
 #define RTIMER_CLOCK_TICK_RATIO (RTIMER_SECOND / CLOCK_SECOND)
-#define RELOAD_VALUE (125000 - 1)  /** Fire 128 times / sec */
+
+/* Prescaler for GPT0:Timer A used for clock_delay_usec(). */
+#if SYS_CTRL_SYS_CLOCK < SYS_CTRL_1MHZ
+#error System clock speeds below 1MHz are not supported
+#else
+#define PRESCALER_VALUE     (SYS_CTRL_SYS_CLOCK / 1000000 - 1)
```
bthebaudeau: You could even use `SYS_CTRL_1MHZ` instead of `1000000` here to better match the test above.

cpu/cc2538/clock.c
```diff
@@ -62,7 +62,21 @@
...
+#endif
```
bthebaudeau: Move this `#endif` to replace the `#else` above (this line can't be reached if the `#error` is triggered).

cpu/cc2538/clock.c
```diff
@@ -62,7 +62,21 @@
...
+
+/* Reload value for SysTick counter */
+#if SYS_CTRL_SYS_CLOCK < SYS_CTRL_2MHZ
+/* Clock speeds below 2 MHz will lead to reduced accurracy */
+#error System clock speeds below 2MHz are not supported
```
bthebaudeau: Actually, it's 2 MHz now only because `CLOCK_SECOND` is 128, but `CLOCK_CONF_SECOND` could be changed, which would make this test wrong. So the generic test would be:
```
#if SYS_CTRL_SYS_CLOCK % CLOCK_SECOND
#error System clock speed too slow for CLOCK_SECOND, leading to reduced accuracy
```

cpu/cc2538/clock.c
```diff
@@ -62,7 +62,21 @@
...
+#else
+#define RELOAD_VALUE      (SYS_CTRL_SYS_CLOCK / CLOCK_SECOND - 1)
```
bthebaudeau: Keep this definition and the previous one aligned.

cpu/cc2538/clock.c
```diff
@@ -62,7 +62,21 @@
...
+#endif
```
bthebaudeau: Move this `#endif` to replace the `#else` above (this line can't be reached if the `#error` is triggered).

cpu/cc2538/clock.c
```diff
@@ -98,15 +112,14 @@ clock_init(void)
...
-
   /* 16-bit */
   REG(GPT_0_BASE + GPTIMER_CFG) = 0x04;
 
   /* One-Shot, Count Down, No Interrupts */
   REG(GPT_0_BASE + GPTIMER_TAMR) = GPTIMER_TAMR_TAMR_ONE_SHOT;
 
-  /* Prescale by 16 (thus, value 15 in TAPR) */
-  REG(GPT_0_BASE + GPTIMER_TAPR) = 0x0F;
+  /* Prescale depending on sys clock used  */
```
bthebaudeau: - `sys` -> `system`.
 - Remove added extra space after `used`.

cpu/cc2538/clock.c
```diff
@@ -133,7 +146,7 @@ clock_wait(clock_time_t i)
   clock_time_t start;
 
   start = clock_time();
-  while(clock_time() - start < (clock_time_t)i);
+  while(clock_time() - start < (clock_time_t)i) ;
```
bthebaudeau: Unwanted cosmetic change: no space before semicolon.

cpu/cc2538/clock.c
```diff
@@ -148,7 +161,7 @@ clock_delay_usec(uint16_t dt)
   REG(GPT_0_BASE + GPTIMER_CTL) |= GPTIMER_CTL_TAEN;
 
   /* One-Shot mode: TAEN will be cleared when the timer reaches 0 */
-  while(REG(GPT_0_BASE + GPTIMER_CTL) & GPTIMER_CTL_TAEN);
+  while(REG(GPT_0_BASE + GPTIMER_CTL) & GPTIMER_CTL_TAEN) ;
```
bthebaudeau: Unwanted cosmetic change: no space before semicolon.

cpu/cc2538/dev/sys-ctrl.c
```diff
@@ -70,22 +70,22 @@ sys_ctrl_init()
...
    */
 
   val = SYS_CTRL_OSCS | SYS_CTRL_CLOCK_CTRL_OSC_PD
-      | SYS_CTRL_CLOCK_CTRL_IO_DIV_16MHZ | SYS_CTRL_CLOCK_CTRL_SYS_DIV_16MHZ;
+    | SYS_CTRL_IO_DIV | SYS_CTRL_SYS_DIV;
   REG(SYS_CTRL_CLOCK_CTRL) = val;
 
   while((REG(SYS_CTRL_CLOCK_STA) & (SYS_CTRL_CLOCK_STA_OSC32K |
-        SYS_CTRL_CLOCK_STA_OSC)) != SYS_CTRL_OSCS);
+                                    SYS_CTRL_CLOCK_STA_OSC)) != SYS_CTRL_OSCS) ;
```
bthebaudeau: Unwanted cosmetic change: it makes the line longer than 80 characters, and there must be no space before semicolons. See:
https://github.com/contiki-os/contiki/blob/master/doc/code-style.c

cpu/cc2538/dev/sys-ctrl.c
```diff
@@ -70,22 +70,22 @@ sys_ctrl_init()
...
 
 #if SYS_CTRL_OSC32K_USE_XTAL
   /* Wait for the 32-kHz crystal oscillator to stabilize */
-  while(REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SYNC_32K);
-  while(!(REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SYNC_32K));
+  while(REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SYNC_32K) ;
+  while(!(REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SYNC_32K)) ;
```
bthebaudeau: Unwanted cosmetic change: no space before semicolon.

cpu/cc2538/dev/sys-ctrl.c
```diff
@@ -94,7 +94,20 @@ sys_ctrl_reset()
...
   REG(SYS_CTRL_PWRDBG) = SYS_CTRL_PWRDBG_FORCE_WARM_RESET;
 }
-
+/*---------------------------------------------------------------------------*/
+uint32_t
+sys_ctrl_get_sys_clock(void)
+{
+  return SYS_CTRL_32MHZ >> (REG(SYS_CTRL_CLOCK_STA) &
+                            SYS_CTRL_CLOCK_STA_SYS_DIV);
+}
```
bthebaudeau: Good, but `get_sys_clock()` should also be removed from` i2c.c`.

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -242,6 +242,36 @@
...
 /*---------------------------------------------------------------------------*/
+/** \name System clock divisor selection
+ * @{
+ */
+#ifdef SYS_CTRL_CONF_SYS_DIV
+#if SYS_CTRL_CONF_SYS_DIV & ~SYS_CTRL_CLOCK_CTRL_SYS_DIV
+#error Invalid system clock divisor
+#else
+#define SYS_CTRL_SYS_DIV        SYS_CTRL_CONF_SYS_DIV
+#endif
```
bthebaudeau: Move this `#endif` to replace the `#else` above (this line can't be reached if the `#error` is triggered).

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -242,6 +242,36 @@
...
+#else
+#define SYS_CTRL_SYS_DIV        SYS_CTRL_CLOCK_CTRL_SYS_DIV_16MHZ
+#endif
+
+#ifdef SYS_CTRL_CONF_IO_DIV
+#if SYS_CTRL_CONF_IO_DIV & ~SYS_CTRL_CLOCK_CTRL_IO_DIV
+#error Invalid IO clock divisor
```
bthebaudeau: `IO` -> `I/O`.

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -242,6 +242,36 @@
...
+#else
+#define SYS_CTRL_IO_DIV         SYS_CTRL_CONF_IO_DIV
+#endif
```
bthebaudeau: Move this `#endif` to replace the `#else` above (this line can't be reached if the `#error` is triggered).

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -242,6 +242,36 @@
...
+#else
+#define SYS_CTRL_IO_DIV         SYS_CTRL_CLOCK_CTRL_IO_DIV_16MHZ
+#endif
+
+/* Returns actual sys clock in MHz */
```
bthebaudeau: - `sys` -> `system`.
 - This is a value in Hz, not in MHz.

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -242,6 +242,36 @@
...
+#define SYS_CTRL_SYS_CLOCK        (SYS_CTRL_32MHZ >> SYS_CTRL_SYS_DIV)
```
bthebaudeau: Align the definition with the previous and next ones.

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -242,6 +242,36 @@
...
+/* Returns actual io clock in MHz */
```
bthebaudeau: - `io` -> `I/O`.
 - This is a value in Hz, not in MHz.

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -242,6 +242,36 @@
...
+#define SYS_CTRL_IO_CLOCK       (SYS_CTRL_32MHZ >> (SYS_CTRL_IO_DIV >> 8))
+
```
bthebaudeau: Useless empty line.

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -253,6 +283,12 @@ void sys_ctrl_init();
 /** \brief Generates a warm reset through the SYS_CTRL_PWRDBG register */
 void sys_ctrl_reset();
 
+/** \brief Returns the actual system clock in MHz */
```
bthebaudeau: Hz, not MHz.

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -253,6 +283,12 @@ void sys_ctrl_init();
...
+uint32_t sys_ctrl_get_sys_clock();
+
+/** \brief Returns the actual io clock in MHz */
```
bthebaudeau: Hz, not MHz.

cpu/cc2538/dev/uart.c
```diff
@@ -136,7 +136,9 @@
  * Baud rate defines used in uart_init() to set the values of UART_IBRD and
  * UART_FBRD in order to achieve the configured baud rates.
  */
-#define UART_CLOCK_RATE       16000000 /* 16 MHz */
+
+#define UART_CLOCK_RATE			(SYS_CTRL_32MHZ/(1 << SYS_CTRL_SYS_DIV))
+
```
bthebaudeau: - Remove the useless empty lines.
 - Keep the original alignment.
 - Use the I/O clock (confirming test pending).
 - Use the existing definition:
```
#define UART_CLOCK_RATE       SYS_CTRL_IO_CLOCK
```

cpu/cc2538/lpm.c
```diff
@@ -73,13 +73,13 @@ static unsigned long irq_energest = 0;
 #define DEEP_SLEEP_PM1_THRESHOLD    10
 #define DEEP_SLEEP_PM2_THRESHOLD    100
 /*---------------------------------------------------------------------------*/
-#define assert_wfi() do { asm("wfi"::); } while(0)
+#define assert_wfi() do { asm ("wfi" ::); } while(0)
```
bthebaudeau: Unwanted cosmetic change: the Contiki coding style prefers no spaces.

cpu/cc2538/lpm.c
```diff
@@ -73,13 +73,13 @@ static unsigned long irq_energest = 0;
...
 /*---------------------------------------------------------------------------*/
 #if LPM_CONF_STATS
 rtimer_clock_t lpm_stats[3];
 
 #define LPM_STATS_INIT()         do { memset(lpm_stats, 0, sizeof(lpm_stats)); \
-  } while(0)
+} while(0)
```
bthebaudeau: This looks weird. It would be better to move the whole do/while onto its own line indented with two spaces after the `#define` line.

cpu/cc2538/lpm.c
```diff
@@ -105,7 +115,7 @@ static uint8_t max_pm;
 #endif
 
 static lpm_periph_permit_pm1_func_t
-periph_permit_pm1_funcs[LPM_PERIPH_PERMIT_PM1_FUNCS_MAX];
+  periph_permit_pm1_funcs[LPM_PERIPH_PERMIT_PM1_FUNCS_MAX];
```
bthebaudeau: Unwanted cosmetic change: no leading space here.

bthebaudeau:
@uknoblic: I have just added some comments inline. Apart from that:
 - You have not addressed my comment about `spi.c`, but it's difficult before #1042 has been merged. @g-oikonomou: Is there anything preventing from merging it right now?
 - You have not addressed my comment requesting to remove the references to 16 MHz in the comments about the system clock in `clock.c`.

cpu/cc2538/lpm.c
```diff
@@ -212,13 +222,40 @@ lpm_exit()
...
+    REG(SYS_CTRL_CLOCK_CTRL) = val;
+    reset_div = 0;
+  }
+#endif
+
+/* Restore system clock to the 32 MHz XOSC */
   select_32_mhz_xosc();
 
-  /* Restore PMCTL to PM0 for next pass */
+/* Restore PMCTL to PM0 for next pass */
```
bthebaudeau: Unwanted cosmetic change: comments should follow regular indentation.

cpu/cc2538/lpm.c
```diff
@@ -212,13 +222,40 @@ lpm_exit()
...
   REG(SYS_CTRL_PMCTL) = SYS_CTRL_PMCTL_PM0;
 
-  /* Remember IRQ energest for next pass */
+/* Remember IRQ energest for next pass */
```
bthebaudeau: Unwanted cosmetic change: comments should follow regular indentation.

cpu/cc2538/lpm.c
```diff
@@ -202,8 +212,8 @@ lpm_exit()
    * for a positive transition on the 32-kHz clock by polling the
    * SYS_CTRL_CLOCK_STA.SYNC_32K bit, before reading the sleep timer value.
    */
-  while(REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SYNC_32K);
-  while(!(REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SYNC_32K));
+  while(REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SYNC_32K) ;
+  while(!(REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SYNC_32K)) ;
```
bthebaudeau: Unwanted cosmetic change: no space before semicolon.

cpu/cc2538/lpm.c
```diff
@@ -177,13 +187,13 @@ select_16_mhz_rcosc(void)
...
   /*First, make sure there is no ongoing clock source change */
-  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SOURCE_CHANGE) != 0);
+  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SOURCE_CHANGE) != 0) ;
 
   /* Set the System Clock to use the 16MHz RC OSC */
   REG(SYS_CTRL_CLOCK_CTRL) |= SYS_CTRL_CLOCK_CTRL_OSC;
 
   /* Wait till it's happened */
-  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_OSC) == 0);
+  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_OSC) == 0) ;
```
bthebaudeau: Unwanted cosmetic change: no space before semicolon.

cpu/cc2538/lpm.c
```diff
@@ -177,13 +187,13 @@ select_16_mhz_rcosc(void)
   REG(SYS_CTRL_CLOCK_CTRL) &= ~SYS_CTRL_CLOCK_CTRL_OSC_PD;
 
   /*First, make sure there is no ongoing clock source change */
-  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SOURCE_CHANGE) != 0);
+  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SOURCE_CHANGE) != 0) ;
```
bthebaudeau: Unwanted cosmetic change: no space before semicolon.

cpu/cc2538/lpm.c
```diff
@@ -154,14 +164,14 @@ enter_pm0(void)
...
-  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SOURCE_CHANGE) != 0);
+  /* First, make sure there is no ongoing clock source change */
+  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SOURCE_CHANGE) != 0) ;
 
   /* Turn on the 32 MHz XOSC and source the system clock on it. */
   REG(SYS_CTRL_CLOCK_CTRL) &= ~SYS_CTRL_CLOCK_CTRL_OSC;
 
   /* Wait for the switch to take place */
-  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_OSC) != 0);
+  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_OSC) != 0) ;
```
bthebaudeau: Unwanted cosmetic change: no space before semicolon.

cpu/cc2538/lpm.c
```diff
@@ -154,14 +164,14 @@ enter_pm0(void)
 static void
 select_32_mhz_xosc(void)
 {
-  /*First, make sure there is no ongoing clock source change */
-  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SOURCE_CHANGE) != 0);
+  /* First, make sure there is no ongoing clock source change */
+  while((REG(SYS_CTRL_CLOCK_STA) & SYS_CTRL_CLOCK_STA_SOURCE_CHANGE) != 0) ;
```
bthebaudeau: Unwanted cosmetic change: no space before semicolon.

g-oikonomou:
>  - You have not addressed my comment about `spi.c`, but it's difficult before #1042 has been merged. @g-oikonomou: Is there anything preventing from merging it right now?

Waiting for travis to play nicely. I'm not planning to test #1042, I'll just take your word for it!

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -242,6 +242,36 @@
...
+#define SYS_CTRL_IO_DIV         SYS_CTRL_CONF_IO_DIV
+#endif
+#else
+#define SYS_CTRL_IO_DIV         SYS_CTRL_CLOCK_CTRL_IO_DIV_16MHZ
+#endif
+
+/* Returns actual sys clock in MHz */
+#define SYS_CTRL_SYS_CLOCK        (SYS_CTRL_32MHZ >> SYS_CTRL_SYS_DIV)
+/* Returns actual io clock in MHz */
+#define SYS_CTRL_IO_CLOCK       (SYS_CTRL_32MHZ >> (SYS_CTRL_IO_DIV >> 8))
```
bthebaudeau: According to §7.3.1 in the UG, the I/O clock must not exceed 5/3 of the system clock. There should be a `#error` if this is not the case.

cpu/cc2538/dev/uart.c
bthebaudeau: @g-oikonomou: No need to test. The answer is in table 7-2 and §7.3.1 in the UG. {UART_CC|SSI_CC}.CS.{DSEN|PIOSC} select the UART and SSI system and core clocks in the various modes. The UART driver uses the system clock, but the SPI driver uses the I/O clock. It would be better to also make the UART driver use the I/O clock.

cpu/cc2538/lpm.c
```diff
@@ -92,6 +92,16 @@ rtimer_clock_t lpm_stats[3];
...
 
+/* CC2538 silicon errata:
+   Possible Incorrect Value of Clock Dividers after PM2 and PM3 */
+#if SYS_CTRL_SYS_DIV == SYS_CTRL_CLOCK_CTRL_SYS_DIV_32MHZ || \
+  SYS_CTRL_IO_DIV == SYS_CTRL_CLOCK_CTRL_IO_DIV_32MHZ
+/* We use this variable in order to flag if we have adjusted either
+   one or both clock divisors upon entering PM2 or PM3. As everything
+   is done at compile time we can use defines to do the remaining stuff */
+static uint8_t reset_div;
+#endif
```
bthebaudeau: This is useless: you only need the values of `SYS_CTRL_SYS_DIV` and `SYS_CTRL_IO_DIV` for the restore stage, and restoring the same values would be harmless anyway. Moreover, the system and I/O clocks must be handled separately.

cpu/cc2538/lpm.c
```diff
@@ -212,13 +222,40 @@ lpm_exit()
    * We need to convert sleep duration from rtimer ticks to clock ticks */
   clock_adjust();
 
-  /* Restore system clock to the 32 MHz XOSC */
+  /* CC2538 silicon errata:
+     Possible Incorrect Value of Clock Dividers after PM2 and PM3 */
```
bthebaudeau: The multi-line comment style is:
```
/*
 * blah
 */
```

cpu/cc2538/lpm.c
```diff
@@ -212,13 +222,40 @@ lpm_exit()
...
+    val &= ~SYS_CTRL_CLOCK_CTRL_SYS_DIV;
+#elif SYS_CTRL_IO_DIV == SYS_CTRL_CLOCK_CTRL_IO_DIV_32MHZ
+    /* Set io clock divisor back to 0 (=32MHz) */
+    val &= ~SYS_CTRL_CLOCK_CTRL_IO_DIV;
+#endif
+
+    REG(SYS_CTRL_CLOCK_CTRL) = val;
+    reset_div = 0;
+  }
+#endif
```
bthebaudeau: `select_16_mhz_rcosc()` can no longer work properly with the dividers set for 32 MHz. The required change to fix this is also the first step of the workaround for the erratum, so the restore stage has to be in `select_32_mhz_xosc()`, which makes things very easy and avoids duplicating the restore code like you did.

Moreover, this code can be simplified a lot.

Just replace the first line of `select_16_mhz_rcosc()` with this:
```
  REG(SYS_CTRL_CLOCK_CTRL) = (REG(SYS_CTRL_CLOCK_CTRL)
#if SYS_CTRL_SYS_DIV == SYS_CTRL_CLOCK_CTRL_SYS_DIV_32MHZ
    | SYS_CTRL_CLOCK_CTRL_SYS_DIV_16MHZ
#endif
#if SYS_CTRL_IO_DIV == SYS_CTRL_CLOCK_CTRL_IO_DIV_32MHZ
    | SYS_CTRL_CLOCK_CTRL_IO_DIV_16MHZ
#endif
    ) & ~SYS_CTRL_CLOCK_CTRL_OSC_PD;
```

And replace the last line of `select_32_mhz_xosc()` with this:
```
  REG(SYS_CTRL_CLOCK_CTRL) = (REG(SYS_CTRL_CLOCK_CTRL)
#if SYS_CTRL_SYS_DIV == SYS_CTRL_CLOCK_CTRL_SYS_DIV_32MHZ
    & ~SYS_CTRL_CLOCK_CTRL_SYS_DIV
#endif
#if SYS_CTRL_IO_DIV == SYS_CTRL_CLOCK_CTRL_IO_DIV_32MHZ
    & ~SYS_CTRL_CLOCK_CTRL_IO_DIV
#endif
    ) | SYS_CTRL_CLOCK_CTRL_OSC_PD;
```

cpu/cc2538/lpm.c
```diff
@@ -212,13 +222,40 @@ lpm_exit()
...
+
+/* Restore system clock to the 32 MHz XOSC */
```
bthebaudeau: Unwanted cosmetic change: comments should follow regular indentation.

cpu/cc2538/lpm.c
```diff
@@ -212,13 +222,40 @@ lpm_exit()
...
   select_32_mhz_xosc();
```
bthebaudeau: This one should be moved before `clock_adjust()` in order to have proper, clean and stable clocks before making adjustments.

cpu/cc2538/lpm.c
```diff
@@ -284,6 +321,25 @@ lpm_enter()
...
+#elif SYS_CTRL_SYS_DIV == SYS_CTRL_CLOCK_CTRL_SYS_DIV_32MHZ
+    /* Set sys clock divisor to 16MHz */
+    REG(SYS_CTRL_CLOCK_CTRL) |= SYS_CTRL_CLOCK_CTRL_SYS_DIV_16MHZ;
+    reset_div = 1;
+#elif SYS_CTRL_IO_DIV == SYS_CTRL_CLOCK_CTRL_IO_DIV_32MHZ
+    /* Set io clock divisor to 16MHz */
+    REG(SYS_CTRL_CLOCK_CTRL) |= SYS_CTRL_CLOCK_CTRL_IO_DIV_16MHZ;
+    reset_div = 1;
+#endif
+
```
bthebaudeau: Simplify and move to `select_16_mhz_rcosc()`. See my previous comments for more details.

cpu/cc2538/lpm.c
```diff
@@ -314,6 +370,32 @@ lpm_enter()
...
+#elif SYS_CTRL_IO_DIV == SYS_CTRL_CLOCK_CTRL_IO_DIV_32MHZ
+      /* Set io clock divisor back to 0 (=32MHz) */
+      val &= ~SYS_CTRL_CLOCK_CTRL_IO_DIV;
+#endif
+
+      REG(SYS_CTRL_CLOCK_CTRL) = val;
+      reset_div = 0;
+    }
+#endif
+
```
bthebaudeau: Remove.

cpu/cc2538/lpm.c
bthebaudeau: After second thought, it's better to leave this unchanged and not to swap `clock_adjust()` and `select_32_mhz_xosc()` because this could not actually improve anything, but this could increase the wake-up latency due to the reduced time before triggering the switch to the crystal oscillator, so the switching time would be spent doing nothing but waiting instead of being used by `clock_adjust()`.

cpu/cc2538/lpm.c
uknoblic: But i need some flag to save the information wether we come from PM2 / PM3 or not when wakeing-up anyway? I don't have to restore the dividers if this is not the case?

bthebaudeau: No, you don't need such a flag. Just restore the dividers unconditionally. If they are already 0 (32 MHz), this won't change anything. And anyway, you have to make `select_32_mhz_xosc()` symmetrical to `select_16_mhz_rcosc()`because of the latter has to be changed because the dividers are not allowed to select clock frequencies higher than their source clocks (see UG), so you can't keep the system and I/O clock dividers set for 32 MHz when switching to the 16-MHz RC oscillator. See my code in my following comments on line 250 of `lpm.c`.

cpu/cc2538/dev/uart.c
uknoblic: Which would also mean (Page 414:) 
"Note: If the PIOSC is used for the UART baud clock, the system clock frequency must be at  least 9 MHz in run mode." 
What's the reason for choosing the I/O clock?

uknoblic:
Revised most of the modules. Open issues:
- Which clock to use for SPI / UART (System clock / I/O clock)?
- Streamline I2C / SPI / UART with regard to clock selection (no need for get_system_clock() in I2C)
- Compute SPI CPRS from target clock speed (merge [https://github.com/contiki-os/contiki/pull/1042] before?)
- Hardware tests pending (i will do the tests at for UART, SPI and I2C)
- George's remarks with regard to lpm-module (cc26xx)

uknoblic:
@bthebaudeau : With regard to your proposal to deduce CPSR from the I/O clock and a configured target SPI clock for SPI: As the divider is limited to even numbers the computation would lead e.g. to an divider of 4 for a system clock of 32 MHz and a SPI target speed of 15MHz (if we assume that the computation will always round up and never generate a SPI clock which is faster than what the user wants to configure). This would result in a SPI speed of 8 MHz, far away from what the user wanted to configure. For the UART, this would directly lead to a non-functional system, but not for the SPI where the system will work even with the 8MHz speed. What i want to say: if we compute such a divider there should be at least a means to inform the user about the actual SPI speed used. How to achieve this?

cpu/cc2538/dev/uart.c
bthebaudeau: Yes, it seems to come from the 5/3 rule tested in `sys-ctrl.h` for an I/O clock of 16 MHz. And 16 MHz for the SPI currently configured to use the I/O clock (page 433). Is that a problem?

The reason for choosing the I/O clock is explained in table 7-2 and in §7.3.1 (page 195). With the current `enter_pm0()`, choosing the I/O clock does not matter much, and it even prevents users from selecting slow clocks in run mode as you pointed out. However, in the future, that would allow to slow the system clock down during PM0, without disturbing UART and SSI operation, thus improving power consumption without altering the behavior.

So let's stick to the system clock for UART and SPI for now, and we'll switch to the I/O clock if `enter_pm0()` is improved in the future:
```
#define UART_CLOCK_RATE       SYS_CTRL_SYS_CLOCK
```

So the SPI driver also has to be updated to use the system clock. The 5/3 rule tested in `sys-ctrl.h` is UART-specific, so it should be dropped.

bthebaudeau:
> Revised most of the modules. Open issues:
 - Which clock to use for SPI / UART (System clock / I/O clock)?

System clock.

> - Streamline I2C / SPI / UART with regard to clock selection (no need for get_system_clock() in I2C)

Yes.

> - Compute SPI CPRS from target clock speed (merge [#1042] before?)

See the conversation in #1042: just remove the changes in `spi.h` from that PR, then it will be merged, then you can address the CPSR change here.

> - Hardware tests pending (i will do the tests at for UART, SPI and I2C)

OK.

> - George's remarks with regard to lpm-module (cc26xx)

This will be managed separately later (i.e. in a new PR).

> With regard to your proposal to deduce CPSR from the I/O clock and a configured target SPI clock for SPI: As the divider is limited to even numbers the computation would lead e.g. to an divider of 4 for a system clock of 32 MHz and a SPI target speed of 15MHz (if we assume that the computation will always round up and never generate a SPI clock which is faster than what the user wants to configure). This would result in a SPI speed of 8 MHz, far away from what the user wanted to configure. For the UART, this would directly lead to a non-functional system, but not for the SPI where the system will work even with the 8MHz speed. What i want to say: if we compute such a divider there should be at least a means to inform the user about the actual SPI speed used. How to achieve this?

You can just compute the actual SPI clock frequency, and issue a `#warning` if it is different from the requested frequency, e.g. something like:
```
#define SPI_SRC_CLOCK SYS_CTRL_SYS_CLOCK

#define SPI_CONF_CLOCK 1000000
#if SPI_SRC_CLOCK < 2 * SPI_CONF_CLOCK
#error In master mode, the SPI source clock must be at least twice faster than the SPI clock.
#endif

#define SPI_CPSDVSR ...
#define SPI_SCR ...

#define SPI_CLOCK (SPI_SRC_CLOCK / (SPI_CPSDVSR * (1 + SPI_SCR)))
#if SPI_CLOCK != SPI_CONF_CLOCK
#warning The actual SPI clock differs from the configured one.
#endif
```

This won't give the user the actual value of the SPI clock if this is what you are looking for. In that case, the only solution is a `printf()`, which you could enable or not depending on some `DEBUG` or `VERBOSE` setting like in `contiki-main.c`.

Note that setting and computing the SPI clock dynamically (i.e. passed e.g. to `spix_set_mode()` rather than defined with some `SPI_CONF_CLOCK`) might be useful, e.g. with several SPI slaves on different chip select pins having various clock requirements. The code would also be nicer. This could go into a separate PR. For the current PR, the only change required for SPI is setting SSI_CC to 0 (system clock).

cpu/cc2538/clock.c
```diff
@@ -61,8 +61,20 @@
 
 #include <stdint.h>
 /*---------------------------------------------------------------------------*/
-#define RTIMER_CLOCK_TICK_RATIO (RTIMER_SECOND / CLOCK_SECOND)
-#define RELOAD_VALUE (125000 - 1)  /** Fire 128 times / sec */
+#define RTIMER_CLOCK_TICK_RATIO			(RTIMER_SECOND / CLOCK_SECOND)
```
bthebaudeau: The assigned value is too far on the right, exceeding the 80-char limit. Moreover, only 2 of the 3 definitions are aligned.

cpu/cc2538/clock.c
```diff
@@ -61,8 +61,20 @@
...
+
+/* Prescaler for GPT0:Timer A used for clock_delay_usec(). */
+#if SYS_CTRL_SYS_CLOCK < SYS_CTRL_1MHZ
+#error System clock speeds below 1MHz are not supported
+#endif
+#define PRESCALER_VALUE					(SYS_CTRL_SYS_CLOCK / SYS_CTRL_1MHZ - 1)
```
bthebaudeau: The assigned value is too far on the right, exceeding the 80-char limit. Moreover, only 2 of the 3 definitions are aligned.

cpu/cc2538/clock.c
```diff
@@ -61,8 +61,20 @@
...
+
+/* Reload value for SysTick counter */
+#if SYS_CTRL_SYS_CLOCK % CLOCK_SECOND
+/* Clock speeds below 2 MHz will lead to reduced accurracy */
```
bthebaudeau: Remove (preferred) or update comment.

cpu/cc2538/clock.c
```diff
@@ -61,8 +61,20 @@
...
+#error System clock speed too slow for CLOCK_SECOND, leading to reduced accuracy 
+#endif
+#define RELOAD_VALUE					(SYS_CTRL_SYS_CLOCK / CLOCK_SECOND - 1)
```
bthebaudeau: The assigned value is too far on the right, exceeding the 80-char limit. Moreover, only 2 of the 3 definitions are aligned.

cpu/cc2538/clock.c
```diff
@@ -74,8 +86,7 @@ static volatile uint64_t rt_ticks_startup = 0, rt_ticks_epoch = 0;
  *
  * We also initialise GPT0:Timer A, which is used by clock_delay_usec().
  * We use 16-bit range (individual), count-down, one-shot, no interrupts.
- * The system clock is at 16MHz giving us 62.5 nano sec ticks for Timer A.
- * Prescaled by 16 gives us a very convenient 1 tick per usec
+ * The prescaler is computed according to the system clock.
```
bthebaudeau:
```
 * The prescaler is computed according to the system clock in order to get 1
 * tick per usec.
```

There is another reference to `16 MHz` in the `cc2538-clock` group description right after the file header.

cpu/cc2538/dev/i2c.c
```diff
@@ -111,7 +103,7 @@ void
 i2c_set_frequency(uint32_t freq)
 {
   /* Peripheral clock setting, using the system clock */
-  REG(I2CM_TPR) = ((get_sys_clock() + (2 * 10 * freq) - 1) /
+  REG(I2CM_TPR) = ((sys_ctrl_get_sys_clock() + (2 * 10 * freq) - 1) /
```
bthebaudeau: Or even `SYS_CTRL_SYS_CLOCK` for now (part of your driver streamlining item)?

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -242,6 +244,37 @@
...
+#define SYS_CTRL_SYS_DIV		SYS_CTRL_CLOCK_CTRL_SYS_DIV_16MHZ
+#endif
+
+#ifdef SYS_CTRL_CONF_IO_DIV
+#if SYS_CTRL_CONF_IO_DIV & ~SYS_CTRL_CLOCK_CTRL_IO_DIV
+#error Invalid I/O clock divisor
+#endif
+#define SYS_CTRL_IO_DIV			SYS_CTRL_CONF_IO_DIV
+#else
+#define SYS_CTRL_IO_DIV			SYS_CTRL_CLOCK_CTRL_IO_DIV_16MHZ
```
bthebaudeau: Align these two definitions with the other ones above and below.

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -242,6 +244,37 @@
...
+#endif
+
+/* Returns actual system clock in Hz */
+#define SYS_CTRL_SYS_CLOCK		(SYS_CTRL_32MHZ >> SYS_CTRL_SYS_DIV)
+/* Returns actual I/O clock in Hz */
+#define SYS_CTRL_IO_CLOCK		(SYS_CTRL_32MHZ >> (SYS_CTRL_IO_DIV >> 8))
```
bthebaudeau: Line exceeding the 80-char limit. You don't have to put the definition that far on the right. Just keep all definitions aligned for clarity.

cpu/cc2538/dev/sys-ctrl.h
```diff
@@ -242,6 +244,37 @@
...
+
+#if SYS_CTRL_IO_CLOCK > (5 * SYS_CTRL_SYS_CLOCK / 3)
+#error I/O clock must not exceed 5/3 of system clock
+#endif
```
bthebaudeau: This check and the empty line above it can be removed now that we keep the UART on the system clock.

cpu/cc2538/lpm.c
```diff
@@ -73,13 +73,12 @@ static unsigned long irq_energest = 0;
 #define DEEP_SLEEP_PM1_THRESHOLD    10
 #define DEEP_SLEEP_PM2_THRESHOLD    100
 /*---------------------------------------------------------------------------*/
-#define assert_wfi() do { asm("wfi"::); } while(0)
+#define assert_wfi() do { asm ("wfi"::); } while(0)
```
bthebaudeau: Remove this change: no space before opening parenthesis in Contiki.

cpu/cc2538/lpm.c
```diff
@@ -73,13 +73,12 @@ static unsigned long irq_energest = 0;
...
 /*---------------------------------------------------------------------------*/
 #if LPM_CONF_STATS
 rtimer_clock_t lpm_stats[3];
 
-#define LPM_STATS_INIT()         do { memset(lpm_stats, 0, sizeof(lpm_stats)); \
-  } while(0)
+#define LPM_STATS_INIT()  do { memset(lpm_stats, 0, sizeof(lpm_stats)); } while(0)
```
bthebaudeau: Too long. Like this:
```
#define LPM_STATS_INIT() \
  do { memset(lpm_stats, 0, sizeof(lpm_stats)); } while(0)
```

cpu/cc2538/lpm.c
```diff
@@ -163,13 +162,38 @@ select_32_mhz_xosc(void)
...
+
+  REG(SYS_CTRL_CLOCK_CTRL) = (REG(SYS_CTRL_CLOCK_CTRL)
+#if SYS_CTRL_SYS_DIV == SYS_CTRL_CLOCK_CTRL_SYS_DIV_32MHZ
+    | SYS_CTRL_CLOCK_CTRL_SYS_DIV_16MHZ
+#endif
+#if SYS_CTRL_IO_DIV == SYS_CTRL_CLOCK_CTRL_IO_DIV_32MHZ
+    | SYS_CTRL_CLOCK_CTRL_IO_DIV_16MHZ
+#endif
+    ) & ~SYS_CTRL_CLOCK_CTRL_OSC_PD;
+
```
bthebaudeau: - Useless empty lines before and after comment block.
 - This should replace the first line of the function, not be inserted before it. This also causes a comment duplicate.

cpu/cc2538/clock.c
uknoblic: All these indentation problems come either from my emacs configuration or from the uncrustify script... May i assume that blanks are used for indentation (not tabs?)

uknoblic:
Ok, new code tested using cc2538dk timer-test. Works down to 2 MHz. Two remarks:
- Watchdog has to be disabled below 16 MHz
- UART should give an error message when baud rate doesn't fit uart's clock speed

cpu/cc2538/clock.c
bthebaudeau: Yes, Contiki's whitespace rule is all spaces, no tabs (https://github.com/contiki-os/contiki/blob/master/tools/code-style/uncrustify.cfg).

bthebaudeau:
There is a merge conflict to resolve in `i2c.c`. You can just rebase your PR on the current mainline master, and you can even squash all your commits into a single one. If we assume that your local Git repository is configured with `contki-os` and `uknoblic` being the remotes for respectively `contiki-os/contiki` and `uknoblic/contiki` (see `git remote -v`), you can do this with the following commands:
```
git remote update
# You can then keep the cc2538_speed_v1 branch as a backup if anything goes wrong.
git checkout -b cc2538_speed_v1 uknoblic/cc2538_speed
git checkout -b cc2538_speed_v2 contiki-os/master
# You may have to `git add -A` and `git commit` right after merging if conflicts were resolved
# manually.
git merge cc2538_speed_v1
git reset contiki-os/master
git add -A
git commit
git push -f uknoblic cc2538_speed
```
If you do not feel comfortable with this procedure, you can also just merge the current master in your PR branch, but it would be better to keep a clean (vs. WIP) commit log before merging the PR.
You should do this only once all PR changes have been done and acked, in order not to have to do this several times.

As you said, the changes for UART and SPI are still pending. Apart from that, everything looks good.

> Ok, new code tested using cc2538dk timer-test. Works down to 2 MHz.

Good.

> Two remarks:
 - Watchdog has to be disabled below 16 MHz

Any idea why? Is it because the software is no longer resetting it quickly enough while its period is unchanged? If possible, try to configure it for a longer period, preferably depending on the system clock frequency.

>  - UART should give an error message when baud rate doesn't fit uart's clock speed

Can you give more details? Do you mean at compile time? What was your test case? Very low system clock frequency? There may be a prescaler or oversampling setting in the UART module that can be adjusted in this case to improve the baud rate accuracy (to be checked).

uknoblic:
The watchdog is already running at the longest period possible, so i guess we shouldn't solve this problem here as it is application dependent. With regard to the UART: as with the clock module which will produce an inaccurate timing (and our discussion with regard to the SPI clock rate) my proposal would be to throw an error message once the the baud rate can't be realized / doesn't match with the uart's clock speed. Maybe we should also consider to modify the HSE bit...

g-oikonomou:
Any idea where the WDT is being triggered?

bthebaudeau:
> The watchdog is already running at the longest period possible, so i guess we shouldn't solve this problem here as it is application dependent.

Since the issue is only triggered below 16 MHz, it's fine fixing it in a new PR, except if it is caused by a regression introduced by the current PR (though it's more likely in the upper layers as you pointed out). It'd be good to identify the root cause before choosing where to fix it.

Is there at least something printed on the UART before the watchdog is triggered?

> With regard to the UART: as with the clock module which will produce an inaccurate timing (and our discussion with regard to the SPI clock rate) my proposal would be to throw an error message once the the baud rate can't be realized / doesn't match with the uart's clock speed.

This does not answer my questions:
 - The UART not being available to print a message in this case, you have to issue a `#warning`, which could be more or less noticed. A `#error` would be too much since the UART is not always mandatory for an application to work, or we could compel the user to disable UART if it can't meet the baud rate configuration.
 - What were the system clock frequencies for which you got such issues?

> Maybe we should also consider to modify the HSE bit...

Yes, changing the HSE bit is what I was suggesting (prescaler / oversampling).

Whatever the change(s), this should be a new PR dedicated to it.

uknoblic:
@g-oikonomou From what i can say now: the watchdog triggers both during the call of fade_leds() (which was more or less obvious) but also depending on the baud rate - so maybe somewhere deep inside the uart driver (when waiting for the FIFO to become free?). At least for the timer-test example...

uknoblic:
@bthebaudeau 
>   Is there at least something printed on the UART before the watchdog is triggered?

Yes, the timer test starts to print some information

> What were the system clock frequencies for which you got such issues?

As soon as i went down to 8 MHz + UART clock speed of 115200. I had to comment out fade_leds(). At 460800 it still works even at 8MHz

bthebaudeau:
The `printf()` stuff uses polling and never resets the watchdog, so long text outputs will trigger it. µDMA would be much better, and this would also free CPU time for something useful. At least, it could reset the watchdog in the loop as a short-term fix.

`fade()` depends on the system clock because it uses `for` loops instead of timer functions. Also, it does not reset the watchdog.

These UART/`printf()`/`fade()` issues should be fixed in 3 new PRs, not here.

g-oikonomou:
Are we taking the CC2538 errata about clock dividers into account for this one?

bthebaudeau:
> Are we taking the CC2538 errata about clock dividers into account for this one?

Yes.

bthebaudeau:
@uknoblic Why have you closed it? I see that the rebase went terribly wrong, but it was recoverable with a forced push without closing the pull request. Maybe you can still reopen it.

Tell me if I can help. If you have kept your old branch (before rebasing) somewhere, I can help you do the rebase, e.g. by sending a pull request to your repository, then you can push this to your branch for the current PR.

There is an important feature lacking in GitHub: it would be great to have a global diff in the "Files changed" tab between each forced push in order to follow the patchset revisions when reviewing.

bthebaudeau:
And note that you can still recover your old states of your working branch from your local repository by using `git reflog <branch>`, even if you have not saved it to a new named branch. You can access older HEADs by appending `@{n}` (e.g. `@{1}`) to the branch name to access its previous n-th state.

bthebaudeau:
Looking at the commit log, actually it's not that bad. The mess comes from the fact that there is a merge commit. You can probably recover this just by cherry-picking your top commit from a new branch based on master. You could could also use `git reset`, or even `git diff` then `git apply` to reset this commit to a simple one.

uknoblic:
@bthebaudeau Thanks for offering your help, i am going to issue a new pull request from am clean and actual branch this week. Also started to read a book about Git ;-)

PR #986, opened 2015-03-09
Replace STH21 with HDC1000 for the CC2650 SensorTag

g-oikonomou:
The release version of the CC2650 SensorTag will be equipped with an HDC1000 sensor. This pull:

* Adds a driver for this sensor
* Removes the driver for the SHT21
* Adjusts platform examples to use the new sensor
* Updates the plaform README accordingly

PR #1047, opened 2015-05-03
CC26xx Ultra Low Power Operation

g-oikonomou:
This is a big pull for the CC26xx CPU/Platform code that results in major power savings. TL;DR, see the traces below

srf-before-after
Above: SmartRF+CC2650EM consumption between two ContikiMAC CCA cycles, including 1 cycle.

* Before (top trace): with all peripherals on the Srf disconnected (jumpers removed).
* After (bottom trace, with this pull in place): With all peripherals connected. In both traces we can see CCA spikes and RTC wakeup spikes.

In the top trace only, we see some even smaller spikes caused by a [known issue](http://processors.wiki.ti.com/index.php/CC26xxware) with the adaptive recharge rate calculation in current cc26xxware. The bottom trace uses the fix discussed in the linked page, hence the recharge spikes have been eliminated.

srf-ticks-before-after
Above: SmartRF+CC2650EM zoom in between two RTC ticks, including a wakeup and back to sleep cycle. The smaller spikes in the top trace are due to the CC26xx recharge, as above.

tag-cca-pull
Above: Sensortag consumption between two consecutive CCA cycles.

tag-ticks-pull
Above: Sensortag zoom in to show consumption between RTC ticks, including 1 wakeup cycle.

In more detail:
* New logic on how to handle power domains and module clocks:
  * The PERIPH PD is always on when the CM3 is running
  * The clock to GPIO is always enabled when the CM3 is running.
  * The SERIAL PD is _not_ on when the CM3 is running. We explicitly power it up only when we need it (I2C and UART)
  * I2C and UART are _not_ clocked automatically when the CM3 is running. We only enable module clocks when we need them (and we disable them afterwards).
  * The GPT0 module clock is _not_ enabled all the time, we only enable it when we need it (sensortag buzzer, `clock_delay_usec()`) and we make sure to disable it afterwards
* New logic for power domain locks. In other words, a new way of keeping a PD powered when the CM3 is dropping to deep sleep and of releasing it when we no longer need it.
* Oscillator control changes:
  * The CC26xxware oscillator control API requires two modules within AUX to be clocked: The oscillator module (OSC) and the semaphore module (SMPH). Failing to clock both results in a bus fault when trying to manipulate oscillators. CC26xxware oscillator control functions automatically enable the OSC clock, but they do not check SMPH. We add our own oscillator control wrappers, which make sure the SMPH module is clocked before calling CC26xxware osc functions.
  * Furthermore, these wrappers allow us to optimise usage of the HF XOSC. We request it as the source for the HF clock, which triggers its startup sequence. We do other stuff while it's starting and we only perform the actual source switch when we need it (just before using the RF's Freq. Synth.). This reduces the time time we spend blocked waiting for it to be ready, which in turn reduces power consumption.
* We handle all pins explicitly and make sure they are in a low leakage state when we deep sleep / shutdown. This includes GPIO pins (e.g. ACC_PWR, ALS_PWR for Srf+EM, the tag's MPU_PWR and TMP_RDY), but also pins under peripheral control (UART, MISO, MOSI, SDA, SCL etc)
* Improve all peripheral drivers (tag sensors, external flash) to make sure they are in their deepest sleep mode / lowest consumption state when not in use.
* Re-written UART driver: The driver is now clever enough to do everything that's required in order to keep UART RX working irrespective of the chip's state:
  * When there is a registered input handler, the driver will automatically clock the module under sleep and deep sleep, it will request the SERIAL PD to stay powered under all power states and it will enable the RX interrupt. Thus, all the caller has to do is to register an input handler (e.g. serial line or slip).
  * When there is no input handler, the driver will disable the RX interrupt, it will release the lock to the SERIAL PD and it will disable module clocks under sleep and deep sleep.
  * This should solve all UART-related and BR-related problems discussed in http://thread.gmane.org/gmane.os.contiki.devel/24767
* New `lpm_shutdown()` implementation: We use our own rather than the one provided by CC26xxware. This allows us to optimise the sequence for Contiki.
* Some minor improvements to `lpm_drop()` and `wake_up()`
* The CC26xx simple demo now demonstrates ultra low power operation off the shelf.

This pull requires #986, so I've included the respective history here.

This pull obsoletes #988 and #993, so I'll be closing them.

### DIFF
--- a/cpu/cc26xx/Makefile.cc26xx
+++ b/cpu/cc26xx/Makefile.cc26xx
@@ -81,7 +81,7 @@ CONTIKI_CPU_SOURCEFILES += clock.c rtimer-arch.c cc26xx-rtc.c uart.c
 CONTIKI_CPU_SOURCEFILES += cc26xx-rf.c contiki-watchdog.c
 CONTIKI_CPU_SOURCEFILES += putchar.c ieee-addr.c batmon-sensor.c
 CONTIKI_CPU_SOURCEFILES += slip-arch.c slip.c cc26xx-uart.c lpm.c
-CONTIKI_CPU_SOURCEFILES += gpio-interrupt.c
+CONTIKI_CPU_SOURCEFILES += gpio-interrupt.c oscillators.c
 
 DEBUG_IO_SOURCEFILES += dbg-printf.c dbg-snprintf.c dbg-sprintf.c strformat.c
 
--- a/cpu/cc26xx/clock.c
+++ b/cpu/cc26xx/clock.c
@@ -155,10 +155,18 @@ clock_wait(clock_time_t i)
 void
 clock_delay_usec(uint16_t len)
 {
+  uint32_t clock_status;
+
   if(ti_lib_prcm_power_domain_status(PRCM_DOMAIN_PERIPH) !=
      PRCM_DOMAIN_POWER_ON) {
     power_domain_on();
   }
+
+  clock_status = HWREG(PRCM_BASE + PRCM_O_GPTCLKGR) & PRCM_GPIOCLKGR_CLK_EN;
+
+  ti_lib_prcm_peripheral_run_enable(PRCM_PERIPH_TIMER0);
+  ti_lib_prcm_load_set();
+  while(!ti_lib_prcm_load_get());
 
   ti_lib_timer_load_set(GPT0_BASE, TIMER_B, len);
   ti_lib_timer_enable(GPT0_BASE, TIMER_B);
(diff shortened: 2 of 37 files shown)


jonnteolsson:
This has been a big task for us internally at TI, a lot of work has gone into making sure this is the best power performance we can get out of cc26xx running Contiki. 

Thoroughly tested!

alignan:
There are some uncrustify whitespaces leftovers from previous commits.

g-oikonomou:
Thanks @alignan I pushed some tidying-up. Let me know if you can spot anything else

alignan:
:+1:

PR #1048, opened 2015-05-05
Fix serialdump.c

cmorty:
_GNU_SOURCE is needed for O_DIRECT
time.h is needed for strftime


cmorty:
Please don't merge yet. There is a patch to come.

msloth:
May I propose to make 115200 default instead of 57600 while you're at it? 115200 seems to be the de facto standard for most platforms.

cmorty:
@msloth There you go. :)

nvt:
Good to have a clean up of serialdump, but on a first look I noticed that the PR adds a space after some of the <code>for</code> and <code>switch</code> keywords.

simonduq:
Very welcome fix and cleanup, and it seems @nvt's comment is addressed, so :+1: from me

alignan:
:+1: as well, one less argument to my aliases :-)

PR #1050, opened 2015-05-07
Added missing sanity check in uip-ds6-route.c

simonduq:
We have no guarantee that `current_nexthop` is not null even when `r` is a valid entry. On platforms with memory protection, `memcpy` on a null pointer can hurt. Found this bug while playing with a regression test with native compilation.


joakimeriksson:
Great fix! These are hard to find when they occur.

adamdunkels:
:+1:

PR #1053, opened 2015-05-07
RPL probing

simonduq:
This PR implements an optional probing mechanism for RPL. Nodes will periodically send a unicast message (DIO or DIS) to one of their best parents (mostly the current preferred and second best parent, sometimes also the third best), to keep link metric estimates up to date. The main benefit is that it avoids switching to a parent which is seemingly good but has an outdated link metric (or never updated, stuck at its initial value). 

When probing and refreshing our link estimate to the second best parent **S**, there are three possible consequences:
(1) the link estimate remains roughly unchanged, **S** can be used safely as a backup in case the current preferred parent fails
(2) the link proves better than we thought, RPL may chose to switch parent a now use **S**
(3) the link proves worse than we thought, with its new link estimate **S** will no longer be the second best parent, we will be able to probe another parent until we find a safe backup to our current preferred parent

Through many deployments in different testbeds, with different platforms and MAC layers (Nullrdc, ContikiMAC, TSCH), we found that probing was a must when aiming for reliable RPL operation.

### DIFF
--- a/core/net/rpl/rpl-conf.h
+++ b/core/net/rpl/rpl-conf.h
@@ -227,6 +227,15 @@
 #endif
 
 /*
+ * RPL probing expiration time.
+ * */
+#ifdef RPL_CONF_PROBING_EXPIRATION_TIME
+#define RPL_PROBING_EXPIRATION_TIME RPL_CONF_PROBING_EXPIRATION_TIME
+#else
+#define RPL_PROBING_EXPIRATION_TIME (10 * 60 * CLOCK_SECOND)
+#endif
+
+/*
  * Function used to select the next parent to be probed.
  * */
 #ifdef RPL_CONF_PROBING_SELECT_FUNC
--- a/core/net/rpl/rpl-conf.h
+++ b/core/net/rpl/rpl-conf.h
@@ -249,4 +249,14 @@
 #define RPL_PROBING_SEND_FUNC(instance, addr) dio_output((instance), (addr))
 #endif
 
+/*
+ * Function used to calculate next RPL probing interval
+ * */
+#ifdef RPL_CONF_PROBING_DELAY_FUNC
+#define RPL_PROBING_DELAY_FUNC RPL_CONF_PROBING_DELAY_FUNC
+#else
+#define RPL_PROBING_DELAY_FUNC() ((RPL_PROBING_INTERVAL / 2) \
+    + random_rand() % (RPL_PROBING_INTERVAL))
+#endif
+
 #endif /* RPL_CONF_H */
--- a/core/net/rpl/rpl-conf.h
+++ b/core/net/rpl/rpl-conf.h
@@ -207,4 +207,46 @@
 #define RPL_INSERT_HBH_OPTION       1
 #endif
 
+/*
+ * RPL probing. When enabled, probes will be sent periodically to keep
+ * parent link estimates up to date.
+ * */
+#ifdef RPL_CONF_WITH_PROBING
+#define RPL_WITH_PROBING RPL_CONF_WITH_PROBING
+#else
+#define RPL_WITH_PROBING 1
+#endif
+
+/*
+ * RPL probing interval.
+ * */
+#ifdef RPL_CONF_PROBING_INTERVAL
+#define RPL_PROBING_INTERVAL RPL_CONF_PROBING_INTERVAL
+#else
+#define RPL_PROBING_INTERVAL (120 * CLOCK_SECOND)
+#endif
+
+/*
+ * Function used to select the next parent to be probed.
+ * */
+#ifdef RPL_CONF_PROBING_SELECT_FUNC
+#define RPL_PROBING_SELECT_FUNC RPL_CONF_PROBING_SELECT_FUNC
+#else
+#define RPL_PROBING_SELECT_FUNC(dag) get_probing_target((dag))
+#endif
+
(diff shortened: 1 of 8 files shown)


joakimeriksson:
I do fully support this PR! We have been doing a similar implementation for Yanzi's products to get a significantly improved parent selection. :+1:

laurentderu:
Great job ! This will greatly improve network resilience (and save me the job of finishing and testing my own implementation ;) )

Just two remark, in your PR comment you talk about DIS, but I don't see this in the commits. Also, you have included the commit from #1050

simonduq:
* for using DIS instead of unicast DIO, set the `RPL_CONF_PROBING_SEND_FUNC` macro (there is an example on how to do this in comment, see `rpl-conf.h`)
* the commit from https://github.com/contiki-os/contiki/pull/1050 is intentionally included, as changes from this PR incidentally revealed the `uip-ds6-route` bug in a particular regression test, causing Travis to fail (cooja segfault when doing a memcmp over address 0). I'll close the other PR if this one gets merged first.

adamdunkels:
Great work @simonduq! Probing definitely is a must, RPL easily falls into situations where nodes are stuck with suboptimal parents just because they heard of them first and has a high initial ETX on the better parents.

A couple of questions (I didn't check the code thoroughly yet, so please forgive me if they are already answered there): 
* Does the probing interval in any way depend on network density or traffic intensity? When we've been trying out different probing techniques we've found it to be really tricky to find a good balance between probing traffic and network density. It is easy to get into situations where large, dense networks have too much probing traffic. But we've been using ICMP pings for probes, maybe using RPL messages would make it easier to find a good balance because they are handled by the RPL code already?
* The current parent typically gets a good ETX estimate just from all the data traffic. Does this patch take this into consideration when choosing to whom to send the next probe?
* Do you have a good simulation setup / regression test for this? Would be extremely useful to ensure that the mechanism doesn't degrade over time.

uknoblic:
Will this probing also help in the following situation ( which i discovered during multiple real-live installation runs): Consider a "moving" node N which comes out of range of its initial parent P1 and switches to P2. If the node again approaches P1 it does (up to now) not switch back to P1 even if this would be the better alternative?

simonduq:
@adamdunkels 
* The (mean) probing interval is fixed, here 2 minutes by default. Every nodes transmits a link-local unicast every 2 minutes. Scales better than multi-hop probing, or than probing of every neighbor at a given interval. But it's true that very dense networks might lead to to much probing load. How replacing this periodic interval by a user-defined function (`RPL_CONF_PROBING_something` in `rpl-conf.h`) that returns the next delay?
* Yes. The patch keeps track of the last unicast sent callback for every parent, and always probes the best parent not updated in the last (by default) 4 minutes.
* Good idea, I'll try to come up with something.

simonduq:
@uknoblic yes it can help in this situation. However the default probing target selection function (set via `RPL_PROBING_SELECT_FUNC`) focuses on the parents that currently appear to be the best, and might miss out `P1` if it's link estimate has become too bad. I'll think about possible a enhancement to better handle this case.

simonduq:
BTW, there is one regression test which I believe should be disabled: `06-rpl-temporary-root-loss.csc`.
It removes the root after some time and verifies that multi-hop routing in a sub-tree still works without the root. IMO, the expected behavior here is that nodes should notice that the root is gone, they should leave the network and become unable to route. Note that the current test fails if we make it last longer, because nodes eventually send new DAOs and realize there is no connectivity to the root. With probing, this happens even sooner (not soon enough to see the test fails, fortunately ;)).

simonduq:
One more thing: I suspect a bug in the way losses are counted in regression tests. `numMissed = num - lastMsg` seems to be one to many, shouldn't it be replaced by `numMissed = num - lastMsg - 1`?

simonduq:
I addressed the comments above:
* the probing interval function is now user-configurable
* the default probing selection process now not only probes the best neighbors, but also other neighbors (lest recently updated) from time to time
* added a new regression test where a sender **S** transmits to the root **R** and there is also an intermediate node **I** in range. After some time, **R** is moved away from **S** but still in range of **I**. Later, **R** is moved back to its original location. The test verifies two things: (1) no more than 2 losses occur in the process of switching to **I** (it takes 4 losses in the non-probing case, because **S** does not now it has a 100% link to **I**). (2) **S** eventually switches back to **R** as preferred parent (without probing, it never attempts to transmit again to **R** because the link had worsened earlier).

nvt:
:+1:

adamdunkels:
Nice to see the travis test in there!

g-oikonomou:
@simonduq 

For the CM3 platforms:

```
  CC        ../../core/net/rpl/rpl-dag.c
../../core/net/rpl/rpl-dag.c: In function 'rpl_print_neighbor_list':
../../core/net/rpl/rpl-dag.c:103:11: warning: format '%u' expects argument of type 'unsigned int', but argument 7 has type 'long unsigned int' [-Wformat=]
           (now - p->last_tx_time) / (60 * CLOCK_SECOND));
           ^
../../core/net/rpl/rpl-dag.c:103:11: warning: format '%u' expects argument of type 'unsigned int', but argument 7 has type 'long unsigned int' [-Wformat=]
```

simonduq:
should get fixed with https://github.com/contiki-os/contiki/pull/1067

PR #1032, opened 2015-04-20
Wismote external flash driver

sumanpanchal:
Contiki 2.7 does not have driver support for WISMOTE external flash. I have updated the driver of external flash for WISMOTE platform.

PR #845, opened 2014-11-03
Changed code region to any

alignan:
Original discussion [Here](http://sourceforge.net/p/contiki/mailman/message/32628670/)

Tested with json-ws test application at examples/ipv6/json-ws.

Original:
> websense-z1.z1 section `.far.text' will not fit in region `far_rom'
> region `far_rom' overflowed by 376 bytes

Amended:
msp430-size websense-z1.z1 
>   text    data	    bss	    dec	    hex	filename
>  48260	    594	   5936	  54790	   d606	websense-z1.z1

PR #1055, opened 2015-05-08
Added MSP430x ELFLOADER support to load image with large memory model.

sumanpanchal:


g-oikonomou:
No idea whether this works, but I do know that it needs lots of code-style love.

sumanpanchal:
@g-oikonomou  
I have tested this code on a WISMOTE to load module from external memory to main memory dynamically.

simonduq:
You'll find more info on code style here: https://github.com/contiki-os/contiki/wiki/Code-Contributions#code-formatting

Out of curiosity (and might be worth some clarification in the code), what is the difference between "20-bit addressing" and "20-bit addressing with large memory model"?

sumanpanchal:
@simonduq 
Thanks. I will look into the code style.
Few references to understand MSP430X with large memory model.
http://www.ti.com/lit/ml/slap109/slap109.pdf
https://github.com/contiki-os/contiki/wiki/MSP430X

sumanpanchal:
@simonduq 
I have updated the code as per Contiki code style.

simonduq:
Thanks! Some more questions:
* What is the reason to change the Wismote compile flags from memory model `medium` to `large` and both ram and rom regions from `any` to `far`? Is this needed for the loadable elf only or also for the main firmware? Were the previous flags incorrect?
* Are there other msp430x platforms that could benefit from the new `CPU_HAS_MSP430X_LARGE`? (z1 or exp5438)

alignan:
See https://github.com/contiki-os/contiki/pull/845, there were problems with the `-mcode-region=far` but I was using the `medium` memory model.  Have you tested this PR on examples like the `examples/ipv6/json-ws` application? I can test this on Z1 motes next week.

sumanpanchal:
@simonduq 

Medium memory model use 16-bit Data pointer and 20-bit Function pointer so it will allow only single Data segment and multiple Code segments.

Large memory model use 20-bit Data pointer and 20-bit Function pointer so it will allow multiple Data segments and Code segments.

With regions "any" compiler place code in either NEAR or FAR region based on elf size where as with regions "FAR" compiler place code/data above 64KB boundary. 

Large memory model needed for not only loadable elf but also for the main firmware. Previous flags were not incorrect but as I said earlier, with large memory model it will allow multiple code/data segments and utilise ROM storage efficiently if image size is larger than 64KB. 

Large memory model with define regions "far" works with exp5438 and Wismote but not with Z1. For Z1, we have to define region "any" with Large memory model as it has less memory.

simonduq:
OK but then, given the wismote has only 16kB RAM, isn't the medium memory model sufficient? We probably want to force the code to FAR, but no need for large I would say. I did some tests on json-ws, and noticed the switch from medium to far comes at a high price on footprint: an extra ~8kB ROM and 1kB RAM on wismote.

sumanpanchal:
@simonduq 
Thanks for the input. I agree with you. As large memory model uses 20-bit data pointers and add other new segments, size of footprint increase compare to medium memory model. 
I will try to modify elf-loader code for the medium memory model. so that it will work for all the msp430 platforms.

simonduq:
Sounds great, thanks! :)

sumanpanchal:
As per the information given in https://github.com/contiki-os/contiki/wiki/The-dynamic-loader, when I generated the firmware with symbol table for the ./app/codeprop/codeprop-tmp.c example using medium memory model, I got the error : 

"contiki-wismote.a(symbols.o):(.rodata.symbols+0x166): additional relocation overflows omitted from the output".

I noticed that symtab.c has 300+ const variables which are stored in .rodata or .far.rodata segment based on the selected memory model. 

const int symbols_nelts = 348;
const struct symbols symbols[348] = {
{ "autostart_processes", (void *)&autostart_processes },
  ...
}

As the medium memory model allows 16-bit Data pointer, the compiler failed to generate firmware due to lack of space for accommodating all the const variables within a single .rodata segment.  Whereas, in the Large memory model, these const variables get stored in .far.rodata segments, and the firmware gets generated successfully.

I think, you have to provide both the memory models in Contiki.

alignan:
How about providing both models and make the following overridable at compile?

CPU_HAS_MSP430X
CPU_HAS_MSP430X_LARGE

So each platform defines a default, but depending on the application and capabilities, one alternatively could change the memory model.

simonduq:
@alignan sounds good!

sumanpanchal:
@alignan 
I agree with you. 
:+1:

sumanpanchal:
Hi @simonduq 
Can you look into my following PR also as I did not get any response from @nvt yet. 
https://github.com/contiki-os/contiki/pull/1032

sumanpanchal:
Maintainers,
Let me know is there anything to be done in this PR to make it suitable for merge in Contiki master branch.

simonduq:
It seems to me that the problem remains: wismote will now always use the large memory model, which is apparently unneeded in most cases and comes with significant overhead

sumanpanchal:
For dynamic elf loading, LARGE MEMORY MODEL is needed. Otherwise ./example/codeprop itself doesnt work. As per @alignan suggestion, why dont we make both MEDIUM and LARGE MEMORY MODEL available in Contiki so depending on application requirement one can change the model and can use.

simonduq:
Yes that sounds good. But the default for Wismote should remain medium, i.e. non-elfloader-users are unaffected with the change. The elfloader example can be update to set the model to large.

sumanpanchal:
Thanks. I will make all necessary changes for that and update you soon.

sumanpanchal:
@simonduq 
I have made all the necessary changes. Now large memory model is application specific. Just look into it and provide your feedback.

apps/codeprop/Makefile.codeprop-tmp
```diff
@@ -0,0 +1,9 @@
+codeprop-tmp_src = codeprop-tmp.c
+
+# Enable large memory model supports for WISMOTE and EXP5438 platform
```
alignan: I can also test for the Z1 mote, but I don't like platform specific defines... maybe we could check if CPU_HAS_MSP430X (the default case) is defined, then override with CPU_HAS_MSP430X_LARGE? or check if CPU_HAS_MSP430X_LARGE is not defined and throw an error? what do you think it would be the best approach @simonduq ?

simonduq: As app makefiles are included before platform ones, it looks to me like we cannot check for CPU_HAS_MSP430X` here.

My only concern is the naming. `_HAS_` defines are usually defined by platforms, telling whether the platform has a given feature. A proposal would be to always keep `CPU_HAS_MSP430X` defined, and then set `TARGET_MEMORY_MODEL` to identify the memory model (and re-adjust the msp430 makefile)

sumanpanchal: @Simonduq
I have done few other changes as per your proposal. Please look into it.

simonduq:
Looks good to me! :+1: 
Waiting for @alignan 's approval

sumanpanchal:
@alignan ..Please provide your feedback on this PR.

alignan:
Sorry for the delay, seems OK to me :+1:

sumanpanchal:
Thanks @alignan  and @simonduq.

PR #1056, opened 2015-05-08
[Cooja/cc2420] Check the frame preamble and MPDU length before parsing an outgoing packet

tim-ist:
Due to errors in mspsim and/or radio drivers, packets of incorrect length are sometimes transmitted. The length might be larger than the 127-byte maximum (considered negative in the current code) or not matching the actual number of transmitted bytes. This leads to wrong packet delimiting when converting from the mspsim-level stream of bytes to Cooja-level packets, causing unhandled exceptions that terminate the simulation.

This patch checks the frame preamble (0000007A) and the length field. If they are wrong, no decoding attempt is done.

The transmitted bytes are still delivered to the receivers untouched. The connection is terminated when the radio state is changed (which always happens when TX is done).


fros4943:
Great! :+1:

Issue #1058, opened 2015-05-09
Cortex M3 .hex files


g-oikonomou:
The CC2538 and CC26xx ports are identical in terms of how they produce the final .bin and .hex

```
arm-none-eabi-objcopy -O ihex cc2538-demo.elf cc2538-demo.hex
arm-none-eabi-objcopy -O binary --gap-fill 0xff cc2538-demo.elf cc2538-demo.bin
```

In a nutshell, we create a .elf and from it we create a .bin and a .hex

The `.bin` works perfectly with CC2538 as well as with CC26xx. The .hex works fine with CC2538, but it fails miserably with CC26xx. TI's flash prog complains about an unknown record type, but I suspect this is not the primary cause.

I ran objdump on both hex files

```
$ arm-none-eabi-objdump -h cc26xx-demo.hex 

cc26xx-demo.hex:     file format ihex

Sections:
Idx Name          Size      VMA               LMA               File off  Algn
  0 .sec1         0000d6bd  00000000  00000000  00000000  2**0
                  CONTENTS, ALLOC, LOAD
  1 .sec2         00000008  0000d6c0  0000d6c0  00025bf6  2**0
                  CONTENTS, ALLOC, LOAD
  2 .sec3         00000054  0001ffac  0001ffac  00025c24  2**0
                  CONTENTS, ALLOC, LOAD
```

and this looks normal to me.

Any ideas?

bthebaudeau:
You can first test if TI's flash prog is lying or not. Just use a text editor to remove all the records of each type, type after type, until TI's flash prog no longer complains:
http://en.wikipedia.org/wiki/Intel_HEX#Record_types

The record type 0x00 should be fine, but maybe some of the other record types are not correctly supported by TI's flash prog. Does it have a command line option to force the input file format? Is there a big difference between the flash address ranges of the CC2538/CC26xx?

g-oikonomou:
> Is there a big difference between the flash address ranges of the CC2538/CC26xx?

CC26xx starts at 0x0 (which is what Arm likes AFAIK), whereas CC2538 starts at 0x20000

> You can first test if TI's flash prog is lying or not.

This is the offending record type with CC26xx:

```
:04000003000000E910
```

whereas with the CC2538 we get

```
:020000040027D3
```

So objcopy ultimately gives us one record type for one platform, but it gives a different record type for the other. /stumped

g-oikonomou:
[...]

PR #1059, opened 2015-05-10
Update CC26xxware download instructions

g-oikonomou:
TI released a new version of CC26xxware this week and our CC26xx port will fail to build with this version - see the discussion in #1058.

This pull merely updates the platform's README:
* It states the correct version number explicitly
* It provides a link to a page where the user can find a download link for this version

I am becoming increasingly convinced that we should pull CC26xxware as a submodule, which is something I will probably do when I update the port to use the latest version. So this is just a temporary counter-measure.

### DIFF
--- a/platform/srf06-cc26xx/README.md
+++ b/platform/srf06-cc26xx/README.md
@@ -75,10 +75,12 @@ the section ["Drivers" in the CC2538DK readme](https://github.com/contiki-os/con
 
 Environment
 ===========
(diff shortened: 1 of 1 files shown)


bthebaudeau:
:+1:

jonnteolsson:
Please merge this

g-oikonomou:
Documentation stuff. Merging

PR #1060, opened 2015-05-10
Produce ihex files usable with TI's Flash Programmer

### LINKED ISSUE #1058: Cortex M3 .hex files

g-oikonomou:
As per the discussion in #1058, TI's Flash Programmer seems to dislike 03 records in ihex files. As a result, the `.hex` file generated by the CC26xx build system cannot be used with Flash Programmer.

This pull changes the CC26xx build system as follows:
* We `objcopy` from the `.elf` to an intermediate `.hex` ~~which has all addresses offset by `0x200000`. This forces `objcopy` to only use 04/05 records instead of 02/03.~~
* We use `srecord` to post-process this intermediate hex. This produces a final hex file that uses 32-bit linear addresses ~~by offsetting everything back by `-0x200000`. This produces a final `.hex` which has correct addresses, but which only uses 04/05 records.~~

This final `.hex` was used successfully to program Srf+CC2650 EM as well as the SensorTag with cc26xx-demo as well as with cc26xx-web-demo.

Closes #1058

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,10 +41,11 @@ before_script:
 
   ## Install mainline ARM toolchain.  gcc-arm-none-eabi is available
   ## in Ubuntu >= 14.04, but this external PPA is needed for 12.04.
+  ## Install srecord
   - if [ ${BUILD_ARCH:-0} = arm-aapcs ] ; then
       sudo add-apt-repository -y ppa:terry.guo/gcc-arm-embedded &&
       sudo apt-get -qq update &&
-      sudo apt-get -qq install gcc-arm-none-eabi &&
+      sudo apt-get -qq install gcc-arm-none-eabi srecord &&
       arm-none-eabi-gcc --version ;
     fi
 
--- a/cpu/cc26xx/Makefile.cc26xx
+++ b/cpu/cc26xx/Makefile.cc26xx
@@ -6,6 +6,7 @@ OBJCOPY = arm-none-eabi-objcopy
 OBJDUMP = arm-none-eabi-objdump
 NM      = arm-none-eabi-nm
 SIZE    = arm-none-eabi-size
+SREC_CAT = srec_cat
 
 ### TI CC26xxware out-of-tree
 ### TI_CC26XXWARE is the home directory of the cc26xxware
@@ -106,8 +107,11 @@ CUSTOM_RULE_LINK=1
 	$(TRACE_LD)
 	$(Q)$(LD) $(LDFLAGS) ${filter-out $(LDSCRIPT) %.a,$^} ${filter %.a,$^} $(TARGET_LIBFILES) -lm -o $@
 
-%.hex: %.elf
+%.i16hex: %.elf
 	$(OBJCOPY) -O ihex $< $@
+
+%.hex: %.i16hex
+	$(SREC_CAT) $< -intel -o $@ -intel
 
 %.bin: %.elf
 	$(OBJCOPY) $(OBJCOPY_FLAGS) $< $@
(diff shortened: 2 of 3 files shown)


bthebaudeau:
Why do you use `0x200000` instead of `0x100000`, which is the minimal offset doing the trick? We don't care about the start of flash on CC2538 here. But anyway, you don't need to use any offset here because `srec_cat` 32-bit linear addresses by default (`--address-length=4`: `i32hex`).

You could use the stem automatic variable, i.e. `$*-offset.hex` instead of `$(basename $@)-offset.hex`, twice.

So to sum up, your `.hex` make rule could be simplified like this:
```
%.hex: %.elf
	$(OBJCOPY) -O ihex $< $*-i16hex.hex
	$(SREC_CAT) $*-i16hex.hex -intel -o $@ -intel
```

g-oikonomou:
Yep, you are right, that works too. Updated.

bthebaudeau:
:+1:

jonnteolsson:
Very nice, I will take an action at TI internally to make sure we improve the Flash programmer.

bthebaudeau:
Thanks @jonnteolsson. If I may ask, unrelated to this one, but related to Contiki, maybe you can have some influence internally for this one too:
http://e2e.ti.com/support/wireless_connectivity/f/155/p/308156/1105746#1105746
I never got any answer (and I no longer need one), but for me this is a silicon erratum, which of course should be documented for the TI community.

vsaw:
:+1: Another handy feature of this is that the BSL also uses the iHex format in case anyone tries to flash Contiki using the BSL.

jonnteolsson:
@bthebaudeau I have pinged the right resources regarding your issue.

bthebaudeau:
@jonnteolsson Thanks!

g-oikonomou:
There is also this alternative:

```
%.i16hex: %.elf
       $(OBJCOPY) -O ihex $< $@

%.hex: %.i16hex
       $(SREC_CAT) $< -intel -o $@ -intel
```

Two good things about it:
* Make identifies .i16hex as intermediate and deletes it after the build, not leaving junk laying around.
* Users needing to use this file for whatever reason can make it explicitly by running `make foo-example.i16hex`

bthebaudeau:
Yes, that'd be good, and that would also simplify the `$*-i16hex.hex` and avoid repeating it twice.

g-oikonomou:
Pushed. Also a commit to install srecord when running aapcs tests on travis

bthebaudeau:
:+1:

bthebaudeau:
:+1:

Nothing more to change? Ready to merge?

g-oikonomou:
I made a change to travis.yml which triggered a build. If/when it passes then yes!

Issue #980, opened 2015-02-26
Problem with travis job 5 (BUILD_TYPE='rpl').


marcas756:
Hello,

following builds failed (I did not touch relevant code i think):

https://travis-ci.org/contiki-os/contiki/builds/51772784
https://travis-ci.org/contiki-os/contiki/builds/52170020
https://travis-ci.org/contiki-os/contiki/builds/52175890

@ BUILD_TYPE='rpl' 	linux

I went through 52175890 and assume the reason to be in line 621:

```
Rearranging network
==== Files used for simulation (sha1sum) ====

xargs: unmatched single quote; by default quotes are special to xargs unless you use the -0 option

FAIL ಠ_ಠ

Running test 08-rpl-dao-route-loss-0 with random Seed 1: ................... OK

Running test 08-rpl-dao-route-loss-1 with random Seed 1: ................./../regression-tests/simexec.sh: line 70: 328 Aborted java -Xshare:on -jar $CONTIKI/tools/cooja/dist/cooja.jar -nogui=$CSC -contiki=$CONTIKI -random-seed=$RANDOMSEED > $BASENAME.log
```
Can someone have a look on it (I am not familiar with regression testing in contiki).

Regards Marcas

g-oikonomou:
Something wrong with that job, yes. Lost track of how many times I've had to restart it

marcas756:
FAIL ಠ_ಠ

No emoticons in code please

marcas756:
No further fails in the meantime ... seems to be solved somehow ... close

g-oikonomou:
So the build failed on master after the recent merge of #990 and I had to restart job 5. I'm re-opening this for further investigation.

Always good to have authors close their issues when they no longer need to be open by the way! (even though in this case it seems we may have jumped the gun :()

marcas756:
Currently there seems also to be a connectivity issue with travis (fails to download):
https://travis-ci.org/contiki-os/contiki/jobs/56321513

sumanpanchal:
Problem is still persist. Any suggestion please.

g-oikonomou:
I have not seen this occur in ages. Closing

PR #1062, opened 2015-05-10
Fix log messages of Travis tests in 12-rpl

simonduq:
This PR fixes the output of RPL-related regression tests. Should have no other effect on the tests.


adamdunkels:
I was about to say :+1: to this since this looks good and travis was very green and happy, but then I wanted to see the new output and looked at the travis log.

To my horror the log says that the RPL test fails, but travis gives it a green light anyway!

https://travis-ci.org/contiki-os/contiki/jobs/61989649

No idea what the problem is, but there is something in there that looks like an error:
`xargs: unmatched single quote; by default quotes are special to xargs unless you use the -0 option`
Maybe that has something to do with this?

In any case, this is pretty serious as it makes us merge code that doesn't pass the travis tests.

simonduq:
Ooops...
Good (and important!) catch :)
No idea what this can be, I'll look into this.

simonduq:
Found out! It's expected behavior since this commit https://github.com/contiki-os/contiki/commit/74b741bad20397c2bcbe59abe9ba96f956cc9b2d. The first test with seed 1 fails, then we have a new run with seed 5 which succeeds, and the conclusion if one seed works then the test passes. Seems like a bad idea to me, I'd suggest to regard any failure as global failure.

Now, looking at the recent history, it seems the failure on first seed for this particular root loss test came with my RPL probing https://github.com/contiki-os/contiki/pull/1053. I did struggle with getting probing to pass the root loss test, but of course did not notice there was a Travis failure left; sorry about that.

What was the root loss test originally intended for? It is called "temporary root loss" but the test ends before the root is back. I don't think continued traffic without a border router is really wanted -- and probing does help figuring our the root is gone quicker, resulting in nodes leaving the network and test failing. I suggest to either remove or fix the test so that is lasts long enough to see the root back (but then, not sure what condition to use for success).

simonduq:
Added a commit that modifies the root loss test -- thoughts are welcome!

g-oikonomou:
Related to #980 ?

simonduq:
@g-oikonomou I don't think so. However, #980 might be fixed by https://github.com/contiki-os/contiki/pull/1050. The problem addressed in the latter pull was occasional memcmp at NULL, causing Cooja nodes to segault. The symptoms were exactly the same as in #980: simulation stopping before 100% and Cooja attempting to issue a core dump (failed attempt in Travis, but can work on local setups). Might be worth rebasing the code in #980 to rerun Travis and see if the problem persists.

nvt:
:+1:

PR #1051, opened 2015-05-07
Fixed a number of compiler warnings

simonduq:
This PR simply fixes some compiler warnings.

PR #1064, opened 2015-05-11
Fixed compiler warnings in core/lib/petsciiconv.c

kasunch:
Fixed compiler warnings related with assigning signed pointer to unsigned pointer.


cmorty:
See https://github.com/contiki-os/contiki/pull/1051

kasunch:
So, my request can be ignored.

PR #412, opened 2013-11-08
cc2538: Add crypto drivers and examples for AES-CCM and SHA-256

bthebaudeau:
Signed-off-by: Benoît Thébaudeau <benoit.thebaudeau@advansee.com>

PR #1071, opened 2015-05-15
cc2538: Add crypto drivers and examples for AES-CCM and SHA-256

bthebaudeau:
This is a rebased and updated re-pull request of #412.


bthebaudeau:
Of course, LLSEC could be configured to use that by default on CC2538 (see `AES_128_CONF` and `CCM_STAR_CONF`).

alignan:
Thanks for the update! I will give it a go!

alignan:
Tested with the Re-Mote, worked nicely!

I would suggest to document the expected test result for each test vector for the sha256-test (success, fail), I think would help to go through the example.

:+1:

bthebaudeau:
> Tested with the Re-Mote, worked nicely!

Great!

> I would suggest to document the expected test result for each test vector for the sha256-test (success, fail), I think would help to go through the example.

What do you mean exactly? This example already contains the expected test result for each test vector in the `sha256` array, and it prints either `Computed SHA-256 hash OK` or `Computed SHA-256 hash does not match expected hash` for respectively success or failure (similarly to `ccm-test`). Are you just suggesting to reword these messages with `success`/`failure`? Why does your comment only apply to `sha256-test`, and not to `ccm-test`?

alignan:
I remember (maybe mistaken) when I ran the `ccm-test` the results were all positive, but upon running the `sha256-test` there were test vectors showing both results, that's why I made the comment about this test only.  

As this test is meant to show the implementation of sha-256 as well as validate the implementation, my suggestion is to add a single line comment in each vector to specify the expected result, either `Computed SHA-256 hash OK` or `Computed SHA-256 hash does not match expected hash`, so the reader can easily pin-point this even before running the example (or having to use a SHA-256 calculator as I did to assert the results).

Anyway this is just a suggestion to improve the readability of the example, not a biggie, I'm happy to merge as it is.

joakimeriksson:
Nice to see more security related stuff coming in, but should we not also consider making a more generic security layer for Contiki? Maybe like core/lib/aes-128.h could serve as example? If we start naming the modules and functions like crypto.c and crypto_init, crypto_enable, etc I am sure we will soon have name conflicts. Maybe functions and files should be prefixed cc2538?

bthebaudeau:
> I remember (maybe mistaken) when I ran the `ccm-test` the results were all positive, but upon running the `sha256-test` there were test vectors showing both results, that's why I made the comment about this test only.

I see. It must be a mistake, because this test is supposed to print `Computed SHA-256 hash OK` for all vectors, hence our misunderstanding. Please run it again.

> Nice to see more security related stuff coming in, but should we not also consider making a more generic security layer for Contiki? Maybe like core/lib/aes-128.h could serve as example? If we start naming the modules and functions like crypto.c and crypto_init, crypto_enable, etc I am sure we will soon have name conflicts. Maybe functions and files should be prefixed cc2538?

The code in this PR is CC2538-specific. It is supposed to be directly used only from CC2538-related code. It is a low-level layer providing these features. Platform-agnostic Contiki modules are not supposed to use this directly. Instead, as I explained above (https://github.com/contiki-os/contiki/pull/1071#issuecomment-102522384), a glue layer should be added to make modules like LLSEC benefit from hardware acceleration.

joakimeriksson:
>The code in this PR is CC2538-specific. It is supposed to be directly used only from CC2538-related >code. It is a low-level layer providing these features. Platform-agnostic Contiki modules are not >supposed to use this directly. Instead, as I explained above (#1071 (comment)), a glue layer should be >added to make modules like LLSEC benefit from hardware acceleration.

Absolutely and that is why it needs to be prefixed with cc2538 (or similar). No generic modules should use it directly, but it should also try to avoid having names that these other module are likely to define. crypto_init, crypto_enabled are such names I would say - or?

If we do a crypto module that is generic and where it is possible to plug-in modules implementing the crypto functions such as these ones it would be nice if that module could use the generic names - and therefore it would be good if all implementations have more "prefixed" and more specific names.

bthebaudeau:
> Absolutely and that is why it needs to be prefixed with cc2538 (or similar). No generic modules should use it directly, but it should also try to avoid having names that these other module are likely to define. crypto_init, crypto_enabled are such names I would say - or?

> If we do a crypto module that is generic and where it is possible to plug-in modules implementing the crypto functions such as these ones it would be nice if that module could use the generic names - and therefore it would be good if all implementations have more "prefixed" and more specific names.

Yeah, theoretically, but I'm not sure fixing a hypothetical future name conflict is worth it, all the more this prefix would break Contiki's naming rules, it would make names longer, and it could also be added to all drivers in order to differentiate the generic and platform-specific APIs (e.g. spi vs. spi-arch, or LPM management features that could be factored out into the core modules, so why add this prefix only for crypto?).

Also, a generic crypto module would not make much sense. You would have AES, RSA, SHA256, etc., but not simply `crypto`. So we could think about this prefix for AES/CCM/SHA256 here. But still, there is nothing defined about how this should be done, and the usual way of doing such things in Contiki is with <module>-arch, `#define` or weak functions, not with a CPU prefix.

So maybe it's better to just wait here. @g-oikonomou, @alignan: What do you think?

adamdunkels:
I think `crypto.c` is fine in this context. There is a minor naming issue with the `aes_load_keys()` function though (maybe `crypto_load_aes_keys()` would work?). Also, there is an awful lot of `#define`s in the `crypto.h` file. Are they used outside of the `crypto.c` file (e.g. like parameters to its various functions)?

In any case, this is platform code which doesn't need to be as strict on naming as in the core code.

joakimeriksson:
Sure it is platform - I have recently been working a bit with TinyDTLS that happens to use exactly the same type of naming which for sure would cause conflicts if both this platform code is used and TinyDTLS at the same time. But I guess the TinyDTLS code should have dtls_init_crypto, etc. which then should resolve the issue. I do think that we need to be careful when naming any code that can happen to be used the same time as the core code. I do think there are a lot of cpu-prefixing - take a look at cc26xx for example. Anyway - better to get it in sooner rather than later - we can always solve the problems when they happen to occur.

alignan:
@bthebaudeau I ran the test again and observed the same result as before, that's why I was confused about the intention of the test vectors, sorry for not making this clear before, here is a log (sorry for the long post).  I checked the results of test vector 0 and indeed the provided sha256 matches the expected, so no clue what might be wrong (haven't gone too deep).

Haven't tested with the CC2538dk thou, I'm using the Re-Mote (CC2538SF53), any hint?

-----------------------------------------
Initializing cryptoprocessor...
-----------------------------------------
Test vector 0:
sha256_init(): success, 0 us
Buffer #0 (length: 3):
sha256_process(): success, 0 us
sha256_done(): success, 30 us
Computed SHA-256 hash does not match expected hash
Total duration: 30 us
-----------------------------------------
Test vector 1:
sha256_init(): success, 0 us
Buffer #0 (length: 56):
sha256_process(): success, 0 us
sha256_done(): success, 30 us
Computed SHA-256 hash OK
Total duration: 30 us
-----------------------------------------
Test vector 2:
sha256_init(): success, 0 us
Buffer #0 (length: 130):
sha256_process(): success, 61 us
sha256_done(): success, 30 us
Computed SHA-256 hash OK
Total duration: 91 us
-----------------------------------------
Test vector 3:
sha256_init(): success, 0 us
Buffer #0 (length: 128):
sha256_process(): success, 61 us
sha256_done(): success, 30 us
Computed SHA-256 hash OK
Total duration: 91 us
-----------------------------------------
Test vector 4:
sha256_init(): success, 0 us
Buffer #0 (length: 64):
sha256_process(): success, 0 us
sha256_done(): success, 30 us
Computed SHA-256 hash does not match expected hash
Total duration: 30 us
-----------------------------------------
Test vector 5:
sha256_init(): success, 0 us
Buffer #0 (length: 66):
sha256_process(): success, 30 us
sha256_done(): success, 30 us
Computed SHA-256 hash does not match expected hash
Total duration: 61 us
-----------------------------------------
Test vector 6:
sha256_init(): success, 0 us
Buffer #0 (length: 14):
sha256_process(): success, 30 us
Buffer #1 (length: 42):
sha256_process(): success, 0 us
sha256_done(): success, 30 us
Computed SHA-256 hash OK
Total duration: 61 us
-----------------------------------------
Test vector 7:
sha256_init(): success, 0 us
Buffer #0 (length: 64):
sha256_process(): success, 30 us
Buffer #1 (length: 64):
sha256_process(): success, 30 us
sha256_done(): success, 30 us
Computed SHA-256 hash OK
Total duration: 91 us
-----------------------------------------
Test vector 8:
sha256_init(): success, 0 us
Buffer #0 (length: 60):
sha256_process(): success, 0 us
Buffer #1 (length: 4):
sha256_process(): success, 0 us
Buffer #2 (length: 64):
sha256_process(): success, 30 us
sha256_done(): success, 30 us
Computed SHA-256 hash does not match expected hash
Total duration: 61 us
-----------------------------------------
Disabling cryptoprocessor...
Done!
-----------------------------------------

bthebaudeau:
@adamdunkels

> I think crypto.c is fine in this context. There is a minor naming issue with the aes_load_keys() function though (maybe crypto_load_aes_keys() would work?). Also, there is an awful lot of #defines in the crypto.h file. Are they used outside of the crypto.c file (e.g. like parameters to its various functions)?

I think that I should move all the AES stuff to its own `aes.c`/`h`. The crypto module provides the AES and SHA256 features. Then, AES can be used for several algorithms, one of them being CCM (the other possibilities are not implemented in this PR). So, yes, the AES definitions are supposed to be shared by CCM and other future algorithms, but some of them are also needed to be passed to the functions. Currently, the register definitions used by the CC2538 drivers (e.g. UART) are in their header files.

bthebaudeau:
@alignan

> I ran the test again and observed the same result as before, that's why I was confused about the intention of the test vectors, sorry for not making this clear before,

No worries. You said that you had used a SHA256 calculator to check the expected results placed in each vector, and you deduced from that that everything was correct, but all results should have been `OK` in that case.

> here is a log (sorry for the long post). I checked the results of test vector 0 and indeed the provided sha256 matches the expected, so no clue what might be wrong (haven't gone too deep).

> Haven't tested with the CC2538dk thou, I'm using the Re-Mote (CC2538SF53), any hint?

That's pretty weird. The version that I had tested on CC2538DK (CC2538SF53) is this one:
https://github.com/bthebaudeau/contiki/tree/cc2538-crypto-v1
If it works in your environment, then the issue probably comes from the rebase (because of something that changed in Contiki in the meantime, `git bisect` wanted). If not, then it might be the board or the toolchain. I think that my toolchain was either Sourcery CodeBench Lite 2013.05-23 arm-none-eabi (http://sourcery.mentor.com/public/gnu_toolchain/arm-none-eabi/) or the latest (at that time) gcc-arm-embedded from Linaro.

I don't see any common pattern in the failing vectors. Are the failing vectors always the same from one run to another, or is it random?

bthebaudeau:
@drandreas #412 worked well for you. Did you test with `sha256-test` too, or only with `ccm-test`? Were the results OK for all test vectors (see the discussion above)? Can you test #1071 to see if there is any difference for you?

@alignan If you swap a working test vector with a non-working one, what do you get? Does the failure follow the vector #, or the vector contents? Or do you get twice a failure or success?

drandreas:
@bthebaudeau 
Sorry for the late replay. I re-run the tests on my OpenMote.
https://github.com/bthebaudeau/contiki/tree/cc2538-crypto-v1 works
https://github.com/bthebaudeau/contiki/tree/cc2538-crypto shows the errors described above

bthebaudeau:
@drandreas Thanks for testing. Don't be sorry: you've been pretty quick! :wink:

examples/cc2538dk/crypto/sha256-test.c
```diff
@@ -0,0 +1,264 @@
...
+      }
+    }
+  };
+  static sha256_state_t state;
+  static int i, j;
+  static uint8_t ret;
+  static rtimer_clock_t total_time;
+  rtimer_clock_t time;
+  size_t len;
+  uint8_t sha256[32];
```
drandreas: I suspect this declaration to be the issue 
sha256 is not static hence its state has to be assumed invalid after line 239 PROCESS_PAUSE()

bthebaudeau: Correct. That might be it, and it should anyway be fixed. Does it work if you make it static?

drandreas: currently, the only target at hand refuses to enter the DFU boot loader :-/

drandreas: making it static fixes the issue on my target

bthebaudeau: Great! Thanks! I'll push an update.

alignan:
Nice catch @drandreas! I was re-testing and changing compilers and optimizations, but you suggestion fixed the problem, fine to me now.

bthebaudeau:
I've just pushed an update. The previous version is here for reference:
https://github.com/bthebaudeau/contiki/tree/cc2538-crypto-v2

Changes since v2:
 - Split off AES in order to follow Contiki's naming rules. The cryptoprocessor is nicknamed AES in the user's guide, but it would be confusing for users to have to use AES stuff for SHA, so the common cryptoprocessor features are named CRYPTO in the drivers. Then, CRYPTO provides AES and SHA, and AES provides CCM. The MCU definitions (registers, etc.) are still named `AES_*` like in the user's guide, and they are placed in `aes.h`.
 - Use the new definitions from `sys/cc.h` for `MIN()`/`MAX()` instead of redefining them.
 - Use the functions from `rom-util.h` for `memset`/`cpy`/`cmp` instead of those from `string.h`.
 - Reorganize the crypto header files so that users need to include only one file to get the whole feature set, e.g. only `#include "dev/ccm.h"`.
 - Reorganize the crypto return codes per driver in order to make the error causes clear.
 - Make `sha256` `static` in `sha256-test.c` in order to fix a context corrupted when yielding.
 - Some cosmetic cleanup.

alignan:
:+1:

platform/cc2538dk/README.md
```diff
@@ -29,6 +29,7 @@ In terms of hardware support, the following drivers have been implemented:
     * Low Power Modes
     * General-Purpose Timers. NB: GPT0 is in use by the platform code, the remaining GPTs are available for application development.
     * ADC
+    * Cryptoprocessor (AES-CCM-256, SHA-256)
```
alignan: Do you mind also including this note for the Re-Mote platform README?

bthebaudeau:
> Do you mind also including this note for the Re-Mote platform README?

Done. The previous version is here for reference:
https://github.com/bthebaudeau/contiki/tree/cc2538-crypto-v3

Changes since v3:
 - Rebase.
 - Add cryptoprocessor support note to the Re-Mote platform README.

alignan:
Merging already, this is platform dependant and has been tested by +2 developers, thanks for the work @bthebaudeau, @drandreas could you please update #1078 afterwards?

drandreas:
@alignan yes, but it will take a couple of days until I can get to it.

drandreas:
@alignan I rebased #1078 and fixed all conflicts with #1071

alignan:
Great! I will take a look when I get back to the office

PR #1074, opened 2015-05-17
Update to latest CC26xxware and add it as a submodule

g-oikonomou:
Our CC26xx port relies on TI's [CC26xxware](http://www.ti.com/tool/cc26xxware). Currently, our users need to manually download and extract the correct version and to set an environment variable. Furthermore, due to changes in the CC26xxware API with the recent release, our sources no longer compile and our users need to download an older version.

This pull:

* Automates the process by adding CC26xxware as a module. From now on a simple clone will automatically get the correct version from [g-oikonomou/cc26xxware](https://github.com/g-oikonomou/cc26xxware) and put it under `cpu/cc26xx/lib/cc26xxware`. No user intervention required; things will 'just work'.
* Updates our sources so that Contiki/cc26xx will compile with the latest release of CC26xxware.
* CC26xxare sources have doxygen comments, but our doc system can't build this. Therefore we configure doxygen to exclude docs under `cpu/cc26xx/lib/cc26xxware`

CC26xxware is distributed under the 3-clause BSD, so no problems there.


alignan:
:+1:

PR #1075, opened 2015-05-17
cc2538: Fix and improve .data/.bss initialization

bthebaudeau:
Make the initialization of the .data and .bss output sections safer, faster, smaller, simpler and more maintainable. See the commit messages for details.

This is a follow-up of #581.

For `memcpy()` vs. `rom_util_memcpy()` speed/size comparison, see https://github.com/contiki-os/contiki/pull/426#issuecomment-28699372.

### DIFF
--- a/cpu/cc2538/cc2538.lds
+++ b/cpu/cc2538/cc2538.lds
@@ -91,6 +91,7 @@ SECTIONS
         *(.data*)
         _edata = .;
     } > SRAM AT > FLASH
+    _ldata = LOADADDR(.data);
 
     .ARM.exidx :
     {
--- a/cpu/cc2538/cc2538.lds
+++ b/cpu/cc2538/cc2538.lds
@@ -85,7 +85,7 @@ SECTIONS
         *(.udma_channel_control_table)
     } > SRAM
 
-    .data :
+    .data : ALIGN(4)
     {
         _data = .;
         *(.data*)
--- a/cpu/cc2538/cc2538.lds
+++ b/cpu/cc2538/cc2538.lds
@@ -106,6 +106,11 @@ SECTIONS
         _ebss = .;
     } > SRAM
 
+    .stack (NOLOAD) :
+    {
+        *(.stack)
+    } > SRAM
+
 #if (LPM_CONF_MAX_PM==2) && (LPM_CONF_ENABLE != 0)
     .nrdata (NOLOAD) :
     {
--- a/cpu/cc2538/startup-gcc.c
+++ b/cpu/cc2538/startup-gcc.c
@@ -89,7 +89,7 @@ void uart1_isr(void) WEAK_ALIAS(default_handler);
 #endif
 /*---------------------------------------------------------------------------*/
 /* Allocate stack space */
-static unsigned long stack[512];
+static unsigned long stack[512] __attribute__ ((section(".stack")));
 /*---------------------------------------------------------------------------*/
 /* Linker construct indicating .text section location */
 extern uint8_t _text[0];
--- a/cpu/cc2538/startup-gcc.c
+++ b/cpu/cc2538/startup-gcc.c
@@ -275,7 +275,7 @@ void(*const vectors[])(void) =
 };
 /*---------------------------------------------------------------------------*/
 /* Linker constructs indicating .data and .bss segment locations */
-extern unsigned long _etext;
+extern unsigned long _ldata;
 extern unsigned long _data;
 extern unsigned long _edata;
 extern unsigned long _bss;
@@ -303,7 +303,7 @@ reset_handler(void)
   REG(SYS_CTRL_EMUOVR) = 0xFF;
 
   /* Copy the data segment initializers from flash to SRAM. */
-  pul_src = &_etext;
+  pul_src = &_ldata;
 
   for(pul_dst = &_data; pul_dst < &_edata;) {
     *pul_dst++ = *pul_src++;
--- a/cpu/cc2538/startup-gcc.c
+++ b/cpu/cc2538/startup-gcc.c
@@ -40,6 +40,7 @@
 #include "reg.h"
 #include "flash-cca.h"
 #include "sys-ctrl.h"
+#include "rom-util.h"
 
 #include <stdint.h>
 /*---------------------------------------------------------------------------*/
@@ -275,11 +276,11 @@ void(*const vectors[])(void) =
 };
 /*---------------------------------------------------------------------------*/
 /* Linker constructs indicating .data and .bss segment locations */
-extern unsigned long _ldata;
-extern unsigned long _data;
-extern unsigned long _edata;
-extern unsigned long _bss;
-extern unsigned long _ebss;
+extern uint8_t _ldata;
+extern uint8_t _data;
+extern uint8_t _edata;
+extern uint8_t _bss;
+extern uint8_t _ebss;
 /*---------------------------------------------------------------------------*/
 /* Weak interrupt handlers. */
 void
@@ -298,26 +299,13 @@ default_handler(void)
 void
 reset_handler(void)
 {
-  unsigned long *pul_src, *pul_dst;
-
   REG(SYS_CTRL_EMUOVR) = 0xFF;
 
   /* Copy the data segment initializers from flash to SRAM. */
-  pul_src = &_ldata;
-
-  for(pul_dst = &_data; pul_dst < &_edata;) {
-    *pul_dst++ = *pul_src++;
-  }
+  rom_util_memcpy(&_data, &_ldata, &_edata - &_data);
 
   /* Zero-fill the bss segment. */
-  __asm("    ldr     r0, =_bss\n"
-        "    ldr     r1, =_ebss\n"
-        "    mov     r2, #0\n"
-        "    .thumb_func\n"
-        "zero_loop:\n"
-        "        cmp     r0, r1\n"
-        "        it      lt\n"
-        "        strlt   r2, [r0], #4\n" "        blt     zero_loop");
+  rom_util_memset(&_bss, 0, &_ebss - &_bss);
 
   /* call the application's entry point. */
   main();


g-oikonomou:
This looks very useful, but unfortunately I can't get my Srf+CC2538EM to start properly :disappointed: 

```
$ arm-none-eabi-gcc -v
Using built-in specs.
COLLECT_GCC=arm-none-eabi-gcc
COLLECT_LTO_WRAPPER=/usr/local/gcc-arm-none-eabi-4_9-2015q1/bin/../lib/gcc/arm-none-eabi/4.9.3/lto-wrapper
Target: arm-none-eabi
Configured with: 
[...]
Thread model: single
gcc version 4.9.3 20150303 (release) [ARM/embedded-4_9-branch revision 221220] (GNU Tools for ARM Embedded Processors) 
```

g-oikonomou:
Update: The first two commits work fine, it's the third one (a9a2dec) that breaks things for me.

bthebaudeau:
Thanks for testing.

Unfortunately, I don't have hardware to test. The generated binary seems healthy.

Does it work better if `memcpy()` and `memset()` from `<string.h>` are used instead of `rom_util_memcpy()` and `rom_util_memset()` from `"rom-util.h"`?

Can you send me a memory dump of the ROM (MCU addresses from 0x00000000 to 0x0001FFFF: 131072 bytes)? You can get it using `printf()` (and clearing the watchdog), or with TI CCS IDE and a dummy application for CC2538 with just a `while (1);` in `main()` (with the memory debug window).

If you're curious, you can go further by programming the `.hex` from Contiki and using TI CCS IDE again, disabling the Flash programming in the options, to asm-step through `reset_handler()` (without the source code) and the ROM functions, and see what's going on.

bthebaudeau:
It may also be possible to dump the ROM using TI's flash prog, but I've never tried that.

g-oikonomou:
```
$ git diff
diff --git a/cpu/cc2538/startup-gcc.c b/cpu/cc2538/startup-gcc.c
index 82253e3..0239ece 100644
--- a/cpu/cc2538/startup-gcc.c
+++ b/cpu/cc2538/startup-gcc.c
@@ -302,10 +302,10 @@ reset_handler(void)
   REG(SYS_CTRL_EMUOVR) = 0xFF;
 
   /* Copy the data segment initializers from flash to SRAM. */
-  rom_util_memcpy(&_data, &_ldata, &_edata - &_data);
+  memcpy(&_data, &_ldata, &_edata - &_data);
 
   /* Zero-fill the bss segment. */
-  rom_util_memset(&_bss, 0, &_ebss - &_bss);
+  memset(&_bss, 0, &_ebss - &_bss);
 
   /* call the application's entry point. */
   main();
```

And this works just fine. Hmmm

bthebaudeau:
So the issue probably comes from the ROM functions, which are doing something non standard. A dump of the ROM would probably tell what.

If you mix up both libraries (i.e. `rom_util_memcpy()` + `memset()`, or `memcpy()` + `rom_util_memset()`), is there any combination working?

g-oikonomou:
Yep, I was just trying this.

The culprit seems to be the memset. So now I'm looking at this version working:

```
$ git diff
diff --git a/cpu/cc2538/startup-gcc.c b/cpu/cc2538/startup-gcc.c
index 82253e3..db942ea 100644
--- a/cpu/cc2538/startup-gcc.c
+++ b/cpu/cc2538/startup-gcc.c
@@ -305,7 +305,7 @@ reset_handler(void)
   rom_util_memcpy(&_data, &_ldata, &_edata - &_data);
 
   /* Zero-fill the bss segment. */
-  rom_util_memset(&_bss, 0, &_ebss - &_bss);
+  memset(&_bss, 0, &_ebss - &_bss);
 
   /* call the application's entry point. */
   main();
```

g-oikonomou:
But the other way round does not.

bthebaudeau:
Interesting. I'm certain to have tested `rom_util_memcpy()` in the past, but not `rom_util_memset()`. I have double-checked the offset of `rom_util_memset()` in the ROM API table, and it seems correct. If you align everything with words for `rom_util_memset()`, does it work?

```
rom_util_memset((void *)((uint32_t)&_bss & ~3), 0, ((uint32_t)(&_ebss - &_bss) + 3) & ~3);
```

g-oikonomou:
Nope, that didn't work either.

I checked my .map file:

```
.bss            0x0000000020000110     0x14d5
                0x0000000020000110                _bss = .
...
                0x00000000200015e5                _ebss = .
```
Cannot see anything wrong with the `Length` argument.

Simple poor man's test of rom_util_memset() on a test array didn't seem to overshoot the `[start, start + length)` boundaries. I didn't play around with alignments.

Funnily though:

* `rom_util_memset(&_bss, 0, 1);`: Contiki starts and everything seems to work normally for the first time after programming the EM. If we reset or power-cycle even once then Contiki starts but gets reset by the WDT pretty much immediately after booting.
* `rom_util_memset(&_bss, 0, 0x7FF);`: Same behaviour as above
* `rom_util_memset(&_bss, 0, 0x800;`: Contiki won't start at all

Not sure whether that's a coincidence, but the numbers do smell a bit

bthebaudeau:
> Cannot see anything wrong with the `Length` argument.

Sure. My test was just aimed at detecting possibly broken optimizations depending on alignments in this function.

> Not sure whether that's a coincidence, but the numbers do smell a bit

For the first two cases, this behavior could make sense because of the changes in the initial values in the uninitialized part of .bss. Restarting the board after a long period powered off (I'd say at least 15 minutes) could help make it start normally again without reprogramming it, but this may not be enough because of some random bits left in the SRAM. The programming operation probably clears the SRAM at some point, clearing .bss by the way.

The difference with the third case could indeed suggest a size somewhat limited to 11 bits, which is a strange value.

The only solution is now a ROM dump to clearly open the black box.

g-oikonomou:
> The only solution is now a ROM dump to clearly open the black box.

Any preference formatwise?

bthebaudeau:
No, it doesn't matter. `.bin` or `.hex` would be fine, but I would also be OK with an ASCII UART dump.

bthebaudeau:
Here is the disassembly of the implementation of `rom_util_memset()` in the ROM:
```
    142e:	f001 03ff 	and.w	r3, r1, #255	; 0xff
    1432:	4611      	mov	r1, r2
    1434:	b501      	push	{r0, lr}
    1436:	ea43 2203 	orr.w	r2, r3, r3, lsl #8
    143a:	ea42 4202 	orr.w	r2, r2, r2, lsl #16
    143e:	f000 f886 	bl	0x154e
    1442:	bd01      	pop	{r0, pc}
	[...]
    154e:	2904      	cmp	r1, #4
    1550:	f0c0 8012 	bcc.w	0x1578
    1554:	f010 0c03 	ands.w	ip, r0, #3
    1558:	f000 808a 	beq.w	0x1670
    155c:	f1cc 0c04 	rsb	ip, ip, #4
    1560:	f1bc 0f02 	cmp.w	ip, #2
    1564:	bf18      	it	ne
    1566:	f800 2b01 	strbne.w	r2, [r0], #1
    156a:	bfa8      	it	ge
    156c:	f820 2b02 	strhge.w	r2, [r0], #2
    1570:	eba1 010c 	sub.w	r1, r1, ip
    1574:	f000 b87c 	b.w	0x1670
    1578:	ea5f 7cc1 	movs.w	ip, r1, lsl #31
    157c:	bf24      	itt	cs
    157e:	f800 2b01 	strbcs.w	r2, [r0], #1
    1582:	f800 2b01 	strbcs.w	r2, [r0], #1
    1586:	bf48      	it	mi
    1588:	f800 2b01 	strbmi.w	r2, [r0], #1
    158c:	4770      	bx	lr
	[...]
    1670:	b500      	push	{lr}
    1672:	4613      	mov	r3, r2
    1674:	4694      	mov	ip, r2
    1676:	4696      	mov	lr, r2
    1678:	3920      	subs	r1, #32
    167a:	bf22      	ittt	cs
    167c:	e8a0 500c 	stmiacs.w	r0!, {r2, r3, ip, lr}
    1680:	e8a0 500c 	stmiacs.w	r0!, {r2, r3, ip, lr}
    1684:	f1b1 0120 	subscs.w	r1, r1, #32
    1688:	f4bf aff7 	bcs.w	0x167a
    168c:	0709      	lsls	r1, r1, #28
    168e:	bf28      	it	cs
    1690:	e8a0 500c 	stmiacs.w	r0!, {r2, r3, ip, lr}
    1694:	bf48      	it	mi
    1696:	c00c      	stmiami	r0!, {r2, r3}
    1698:	f85d eb04 	ldr.w	lr, [sp], #4
    169c:	0089      	lsls	r1, r1, #2
    169e:	bf28      	it	cs
    16a0:	f840 2b04 	strcs.w	r2, [r0], #4
    16a4:	bf08      	it	eq
    16a6:	4770      	bxeq	lr
    16a8:	bf48      	it	mi
    16aa:	f820 2b02 	strhmi.w	r2, [r0], #2
    16ae:	f011 4f80 	tst.w	r1, #1073741824	; 0x40000000
    16b2:	bf18      	it	ne
    16b4:	f800 2b01 	strbne.w	r2, [r0], #1
    16b8:	4770      	bx	lr
```

This seems to be AAPCS-compliant.

This could be translated like this in C:
```
#include <stddef.h>
#include <stdint.h>

void rom_util_memset(void *s, int c, size_t n)
{
	union {
		uint8_t *pu8;
		uint16_t *pu16;
		uint32_t *pu32;
		uint32_t u32;
	} _s = {.pu8 = s};

	c &= 0xff;
	c |= c << 8;
	c |= c << 16;

	if (n < 4) {
		if (n & 2) {
			*_s.pu8++ = c;
			*_s.pu8++ = c;
		}
		if (n & 1)
			*_s.pu8++ = c;
	} else {
		int s_w_al = _s.u32 & 3;
		if (s_w_al) {
			s_w_al = 4 - s_w_al;
			if (s_w_al != 2)
				*_s.pu8++ = c;
			if (s_w_al >= 2)
				*_s.pu16++ = c;
			n -= s_w_al;
		}

		while (n >= 32) {
			*_s.pu32++ = c;
			*_s.pu32++ = c;
			*_s.pu32++ = c;
			*_s.pu32++ = c;
			*_s.pu32++ = c;
			*_s.pu32++ = c;
			*_s.pu32++ = c;
			*_s.pu32++ = c;
			n -= 32;
		}
		n -= 32;

		if (n & 0x10) {
			*_s.pu32++ = c;
			*_s.pu32++ = c;
			*_s.pu32++ = c;
			*_s.pu32++ = c;
		}
		if (n & 8) {
			*_s.pu32++ = c;
			*_s.pu32++ = c;
		}
		if (n & 4)
			*_s.pu32++ = c;
		if (n & 3) {
			if (n & 2)
				*_s.pu16++ = c;
			if (n & 1)
				*_s.pu8++ = c;
		}
	}
}
```

Maybe I have to look at it again, but at first sight it seems correct to me.

I still have to check the Cortex-M3 errata, just in case.

Maybe we can also try to run this PIC binary blob from flash to see if it makes any difference. `rom_util_memset()` could also be tested from an application in current `master`, rather than from `reset_handler()`, dumping the target SRAM area (initialized with specific, recognizable data) to check the results.

bthebaudeau:
I've found! The bug is in Contiki, not in the ROM. I've used QEMU to run the ROM and cc2538-demo, and the bug is that `startup-gcc.c` puts the stack in .bss, which is total nonsense. Because of that, the initialization of .bss destroys the stack contents, so any code initializing .bss and using the stack (just like `rom_util_memset()` does) is doomed to crash. This also explains the size threshold that you found while testing, because it corresponds to the first useful stack entries being destroyed.

I'll send a fix for that.

bthebaudeau:
Update pushed.

g-oikonomou:
Splendid. Confirmed to work and travis happy :+1:

PR #1077, opened 2015-05-20
Remove redundant assignment

rajithr:
Remove redundant assignment (reported by CppCheck)


cmorty:
Nope, this isn't redundant. There might be a better way of doing it though.

rajithr:
I think the assignment will get overwritten by the code below it, or am I missing something?

g-oikonomou:
I'm reading it as redundant too, @cmorty what makes you say it is not?

Once upon a time, the call to `read_log_page()` was wrapped inside an `#if` guard, hence the need to set `r` to a default value. No longer required as far as I can see

:+1: @nvt

cmorty:
Mea culpa! You're right.

nvt:
:+1:

PR #1081, opened 2015-05-22
Remove old unused light drivers

pablocorbalan:
There is an old deprecated light sensor (light.h and light.c) that was replaced by light-sensor.c. This driver is not even included in the Makefile of any of the platforms that have the driver (Sky, Wismote, Z1) and does not follow the lib/sensors library, so I think we should remove it.


alignan:
I couldn't find any example relying on the light driver anymore, at least for the Z1 mote can be removed safely.

:+1:

nvt:
:+1:

PR #1083, opened 2015-05-22
cc1200: New device driver for TI sub-ghz radio

uknoblic:
In our company's effort to develop a new platform (CC2538 + CC1200 + ENC28j60) i hereby commit the driver for the CC1200 radio. Even if it borrows some ideas e.g. from the cc1120 or cc2538 radio driver, it was completely re-worked, is highly configurable and also supports 802.15.4g (packet length > 127 bytes). It also uses the channel spacing defined in 802.15.4g for ETSI and FCC. It is currently running stable at 50kbps (ETSI) using nullrdc mac-driver. To my knowledge it will NOT WORK with the cc1120 or cc1101 drivers at it uses at least another channel spacing. Hopefully we will be able to achieve compatibility with old (and upcoming) sub-ghz radios one day...

I would again like to raise the question if we shouldn't consider a re-work of way we handle rf transmission within contiki. As we are currently blocking within "transmit()", we will hardly be able to use features like bigger packets or wake-up preambles!

I would be happy to get feedback!

### DIFF
--- a/dev/cc1200/cc1200-802154g-863-870-fsk-50kbps.c
+++ b/dev/cc1200/cc1200-802154g-863-870-fsk-50kbps.c
@@ -25,7 +25,7 @@
 #define RF_CFG_CCA_THRESHOLD            (-91)
 
 /*---------------------------------------------------------------------------*/
-static const char rf_cfg_descriptor[] = "802.15.4g 863-870MHz MR-FSK mode #1\n";
+static const char rf_cfg_descriptor[] = "802.15.4g 863-870MHz MR-FSK mode #1";
 /*---------------------------------------------------------------------------*/
 /* 
  * Register settings exported from SmartRF Studio using the standard template
@@ -119,7 +119,7 @@ static const registerSetting_t preferredSettings[]=
 /*---------------------------------------------------------------------------*/
 /* Global linkage: symbol name must be different in each exported file! */
 const cc1200_rf_cfg_t cc1200_802154g_863_870_fsk_50kbps = {
-  .rf_cfg_descriptor = rf_cfg_descriptor,
+  .cfg_descriptor = rf_cfg_descriptor,
   .register_settings = preferredSettings,
   .size_of_register_settings = sizeof(preferredSettings),
   .tx_pkt_lifetime = (RTIMER_SECOND / 20),
--- a/dev/cc1200/cc1200-802154g-863-870-fsk-50kbps.c
+++ b/dev/cc1200/cc1200-802154g-863-870-fsk-50kbps.c
@@ -57,36 +57,6 @@
 #define RF_CFG_MAX_TXPOWER              CC1200_CONST_TX_POWER_MAX
 /* The carrier sense level used for CCA in dBm */
 #define RF_CFG_CCA_THRESHOLD            (-91)
-/*
- * ACK_WAIT_TIME is the time between sending a packet and detecting
- * an arbitrary incoming packet (hopefully an ACK). 
- * An incoming packet is detected using either
- * - receiving_packet()
- * - pending_packet()
- * - channel_clear()
- * Let's assume we use receiving_packet() as the latest (reliable) 
- * indication for an incoming packet. It returns 1 as soon as the sync word
- * is detected. We can then estimate this time (assuming an 802.15.4-2011 frame)
- * by:
- * - 1ms RX/TX turnarround (also FIFO read-out -> depends on packet length!)
- * + length of preamble (4 bytes)
- * + length of sync word (2 bytes)
- * -> let's use 2.5 ms for 50kbps
- */
-#define NULLRDC_CONF_ACK_WAIT_TIME      RTIMER_SECOND/400
-/*
- * NULLRDC_AFTER_ACK_DETECTED_WAIT_TIME is the time after an incoming
- * paket is detected (using the means described above) and the ACK is
- * available for read-out by the MAC layer (pending_packet(), read()).
- * If we assume that channel_clear() indicated the incoming packet, the maximum
- * time it takes for the ACK to be available calulates as following:
- * - length of preamble (4 bytes)
- * + length of ACK (3 bytes)
- * + length of CRC (2 bytes)
- * - 1ms safety margin
- * -> let's use 2.5 ms for 50kbps
- */
-#define NULLRDC_CONF_AFTER_ACK_DETECTED_WAIT_TIME   RTIMER_SECOND/400
 /*---------------------------------------------------------------------------*/
 static const char rf_cfg_descriptor[] = "802.15.4g 863-870MHz MR-FSK mode #1";
 /*---------------------------------------------------------------------------*/
--- a/dev/cc1200/cc1200-802154g-863-870-fsk-50kbps.c
+++ b/dev/cc1200/cc1200-802154g-863-870-fsk-50kbps.c
@@ -1,3 +1,37 @@
+/*
+ * Copyright (c) 2015, Weptech elektronik GmbH Germany
+ * http://www.weptech.de
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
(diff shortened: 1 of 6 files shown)


jonnteolsson:
Great addition! I know you have made tests with other data rates, it would be useful for people to understand how other RF settings can be exported from SmartRF Studio.

alignan:
Great contribution!

There are some files without proper code style formatting (the cc1200.c has many), please run uncrustify and push -f.

g-oikonomou:
Same as  #1002, this feature is mega-welcome. I shall try it as soon as I get a chance!

> I would again like to raise the question if we shouldn't consider a re-work of way we handle rf transmission within contiki. As we are currently blocking within "transmit()", we will hardly be able to use features like bigger packets or wake-up preambles!

This is something that has been spinning around in my head for a long while now. IMHO we should be allowed to pass a frame to the DMA controller if one exists and then let rip! But this is probably a discussion more suitable for another pull/issue.

g-oikonomou:
This looks overall very very tidy. As Antonio said, it needs a little uncrustify love.

Can you briefly explain the situation about the .inc file? How do things work? Is it a direct dump from SRF Studio? I see that we are including one among a list of possible .inc files, but this pull only presents one of them. Is this because of where you have focused your testing efforts?

It should be easy to `typedef registerSetting_t register_setting_t` within the driver's main file in order to get rid of the camelCase naming. But before spending any time on that, I first want to understand how those inc files work.

Thanks for your efforts!

g-oikonomou:
> Can you briefly explain the situation about the .inc file? How do things work? Is it a direct dump from SRF Studio?

OK now I see how it works.

As a bare minumum, I would change `.inc` to `.h`. I also want to clarify copyright etc for stuff exported by SRF Studio. Ideally we can have confirmation that it's OK to redistribute under the 3-clause BSD. Have you spoken to TI about this?

Beyond that, it depends how clever we want to get. For example, we can specify a contiki template for SRF Studio that will export a global symbol including register settings as well as other stuff such as packet lifetimes and whatever else changes across modes. With this in place, we can then get rid of the mega `#if` block in the main driver file and replace it with something as simple as a configuration directive.

Totally untested, but here's the rough idea:
* In `cc1200.h` we specify

```
typedef struct cc1200_register_setting {
  uint16_t addr;
  uint8_t val;
} cc1200_register_setting_t;

typedef struct cc1200_mode {
  cc1200_register_setting_t *register_settings;
  rtimer_clock_t packet_lifetime;
  uint32_t whatever_else_we_need;
} cc1200_mode_t;
```

* In our settings file exported by SRF Studio, we do something like:
```
#include "cc1200.h"
static const cc1200_register_setting_t register_setting[] = {
  {CC1200_IOCFG2,            0x06},
  {CC1200_DEVIATION_M,       0xD1},
  {CC1200_XOSC1,             0x03},
};
/* Global linkage: symbol name must be different in each exported file */
const cc1200_mode_t cc1200_mode_etsi_gfsk_50kpbs = {
  register_setting,
  (RTIMER_SECOND / 20),
  0x1234ABCD
};
```
* In our configuration file, we do `#define CC1200_CONF_MODE cc1200_mode_etsi_gfsk_50kpbs`.

alignan:
That's actually how we did in TinyOS and my idea of having for the cc1120, as we had a project in which we had to support support for different configurations (FCC, CEPT, ARIB) and then variant-specifics (Thailand, Australia, etc).

uknoblic:
I agree with the idea to use a single file to put the mode specific stuff into, but maybe it's not a good idea to use a SmartRF template for this purpose. As a compromise i would propose to put all the relevant stuff into on file for each mode, and let the user copy / paste the SmartRF settings using a default template.

uknoblic:
By the way: in the meantime i fixed the indentations problems and a bug in the sniffer functionality.

laurentderu:
With a good template, SmartRF can create files almost compliant with Contiki coding style. And, as I learned the hard way, some of the reserved register of the CC1200 must be set to certains values that depends on the configuration (and that smartRF knows) otherwise the radio does not work properly.

What could be done is include the configuration file using a macro so you don't have to use a huge #if switch case, for example : 
```
  #ifdef RADIO_PROFILE_H
  #include RADIO_PROFILE_H
  #else
  #error RADIO_PROFILE not defined
  #endif
```

Also it would be nice to print this profile name in the init() function, it's much easier to detect that you are trying to communicate to an EU profile with an US profile :)

Small remark about the comments, there are quite a bunch of "/* func(), uk 21.05.2015 */" at the end of functions, I don't think they are really useful.

uknoblic:
> With a good template, SmartRF can create files almost compliant with Contiki coding style. And, as I learned the hard way, some of the reserved register of the CC1200 must be set to certains values that depends on the configuration (and that smartRF knows) otherwise the radio does not work properly.

Ok, but wheres the problem when i use the registers exported from smartRF? I shouldn't miss one if smartRF works reliable? I am aware of the fact that one should not try to specify the registers from the user's guide alone...
> Small remark about the comments, there are quite a bunch of "/* func(), uk 21.05.2015 */" at the end of functions, I don't think they are really useful.

I will remove them in one of my next commits. It's just a reminder for me that i have reviewed the corresponding function.
> I agree with the idea to use a single file to put the mode specific stuff into, but maybe it's not a good idea to use a SmartRF template for this purpose. As a compromise i would propose to put all the relevant stuff into on file for each mode, and let the user copy / paste the SmartRF settings using a default template.

Comming back to this proposal: From my experience there is sometimes the need to re-export the settings from smartRF. If we put all the stuff (like rf timeouts, min-channel, max-channel, max txpower...) in a smartrf-template, then you have to re-edit all this (basic) stuff each time you change a small parameter which to my felling is error-prone. Comments are welcome!

uknoblic:
I have added two files taking into account your proposals regarding the mode configuration. I didn't test / compile it (i am on my windows pc), is just for review...

uknoblic:
> I have added two files taking into account your proposals regarding the mode configuration. I didn't test / compile it (i am on my windows pc), is just for review...

Revised my last commit. Comments are welcome.

g-oikonomou:
> Comming back to this proposal: From my experience there is sometimes the need to re-export the settings from smartRF. If we put all the stuff (like rf timeouts, min-channel, max-channel, max txpower...) in a smartrf-template, then you have to re-edit all this (basic) stuff each time you change a small parameter which to my felling is error-prone. Comments are welcome!

I am not sure I understand what you are trying to say here, Ulf. Do you mean the stuff not exported by SRF Studio like channel min, max etc?

My original thinking about a SRF Studio template was under the provision that the user would only have to change 1, max 2 things after export. If we cannot achieve that, then we should forget about it and proceed as you suggest with register settings copy-paste.

Some additional comments:

* Sniffer stuff is problematic I'm afraid, and this will start to surface when people start tryng to build for different platforms. For example, platforms that don't have a `usb-serial.h` or platforms that use `uart1_write(b)` instead of `uart_write(1, b)`. I don't have a solution for that I'm afraid, otherwise I'd have put it forward for border routers and similar examples too. This file being in core though, warrants a platform-independent approach.

Some minor stuff inline

dev/cc1200/cc1200.c
```diff
@@ -0,0 +1,2471 @@
...
+#include "cc1200-conf.h"
+#include "cc1200-arch.h"
+
+#include "net/netstack.h"
+#include "net/packetbuf.h"
+#include "net/rime/rimestats.h"
+
+#if defined(CC1200_TX_LEDS) || defined(CC1200_RX_LEDS)
+#include "dev/leds.h"
+#endif
```
g-oikonomou: No need for this conditional here. You can simply just include it

laurentderu:
I'm going to test it on exp5438 in the following days, I got it compiling for it already. There are a few compiler warnings that you should remove :

cc1200_driver.prepare
cc1200_driver.transmit
cc1200_driver.send
cc1200_driver.read

have incompatible signatures wrt struct radio_driver.

line 2453: the second argument for burst_read has incompatible pointer type.

uknoblic:
Sorry, i was to fast pushing my new versions. I will squash all commits considering your remarks.
>>   Comming back to this proposal: From my experience there is sometimes the need to re-export the settings from smartRF. If we put all the stuff (like rf timeouts, min-channel, max-channel, max txpower...) in a smartrf-template, then you have to re-edit all this (basic) stuff each time you change a small parameter which to my felling is error-prone. Comments are welcome!

> I am not sure I understand what you are trying to say here, Ulf. Do you mean the stuff not exported by SRF Studio like channel min, max etc?

Yes, this is what i wanted to say. Have a look at the two new files (cc1200-rf-cfg.h + cc1200-802154g-863-870-fsk-50kbps.c)

uknoblic:
> Sniffer stuff is problematic I'm afraid, and this will start to surface when people start tryng to build for different platforms. For example, platforms that don't have a usb-serial.h or platforms that use uart1_write(b) instead of uart_write(1, b). I don't have a solution for that I'm afraid, otherwise I'd have put it forward for border routers and similar examples too. This file being in core though, warrants a platform-independent approach.

So just move the sniffer output from cc1200.c and  place it in stub-rdc.c? This would be easy...

g-oikonomou:
> So just move the sniffer output from cc1200.c and place it in stub-rdc.c? This would be easy...

There have been some discussions about knocking together a generic sniffer and the first step (out of many) would be exactly what you suggest.

In the context of this pull request, I'd move the sniffer stuff out of the driver to start with. I suspect there will be an example of how to use this driver at some point in the future with the Weptech platform. An accompanying device-specific sniffer example would be a nice addition to that bundle.

I'll take a closer look at your updated proposal for the settings, but at first glance it looks much better this way.

Thanks!

dev/cc1200/cc1200-arch.h
```diff
@@ -0,0 +1,157 @@
...
+#ifndef CC1200_ARCH_H
+#define CC1200_ARCH_H
+
+#include <stdint.h>
+/*---------------------------------------------------------------------------*/
+/*
+ * Initialize SPI module & Pins.
+ *
+ * The function has to accomplish the following tasks:
+ * - Enable SPI and configure SPI (CPOL = 0, CPHA = 0)
```
alignan: Tested with CPOL=1 and CPHA=0 and worked, checked with a logic analyser, otherwise failed.

alignan:
I have tested the PR over some time now, I have some enhancements as well.  I'm ready to merge as it is (I have only made a small comment).

dev/cc1200/cc1200-802154g-863-870-fsk-50kbps.c
```diff
@@ -0,0 +1,196 @@
...
+ * Let's assume we use receiving_packet() as the latest (reliable) 
+ * indication for an incoming packet. It returns 1 as soon as the sync word
+ * is detected. We can then estimate this time (assuming an 802.15.4-2011 frame)
+ * by:
+ * - 1ms RX/TX turnarround (also FIFO read-out -> depends on packet length!)
+ * + length of preamble (4 bytes)
+ * + length of sync word (2 bytes)
+ * -> let's use 2.5 ms for 50kbps
+ */
+#define NULLRDC_CONF_ACK_WAIT_TIME      RTIMER_SECOND/400
```
uknoblic: This define doesn't work at this place. Remove it from here.

dev/cc1200/cc1200-802154g-863-870-fsk-50kbps.c
```diff
@@ -0,0 +1,196 @@
...
+ * available for read-out by the MAC layer (pending_packet(), read()).
+ * If we assume that channel_clear() indicated the incoming packet, the maximum
+ * time it takes for the ACK to be available calulates as following:
+ * - length of preamble (4 bytes)
+ * + length of ACK (3 bytes)
+ * + length of CRC (2 bytes)
+ * - 1ms safety margin
+ * -> let's use 2.5 ms for 50kbps
+ */
+#define NULLRDC_CONF_AFTER_ACK_DETECTED_WAIT_TIME   RTIMER_SECOND/400
```
uknoblic: The same for this

alignan:
:+1:

PR #1084, opened 2015-05-22
Z1: tmp102: simple: fixed wrong cast

matttbe:
Hello,

When using this code:

    (int8_t)tmp102_read_temp_x100() / 100

Only the first value is casted into a `int8_t` type.

`tmp102_read_temp_x100()` returns the temperature in Celcius * 100. Most of
the time this value will be lower than -2^7 or higher than 2^7 (+/- 1.27°C).

Best Regards,

### DIFF
--- a/platform/z1/dev/tmp102.c
+++ b/platform/z1/dev/tmp102.c
@@ -185,5 +185,6 @@ tmp102_read_temp_x100(void)
 int8_t
 tmp102_read_temp_simple(void)
 {
-  return (int8_t)tmp102_read_temp_x100() / 100;
+  /* Casted to int8_t: We don't expect temperatures outside -128 to 127 C */
+  return tmp102_read_temp_x100() / 100;
 }


cmorty:
Is that cast needed at all? Though might be good to make this more explicit! Maybe add a comment. :+1:

matttbe:
> Is that cast needed at all? Though might be good to make this more explicit!

No, that cast is not needed at all but I guess it has been added to explicitly inform about the reduction from `int16_t` to `int8_t`. (Except that it caused a bug)

> Maybe add a comment.

I can do that but what kind of comment? :-)
Or maybe you prefer adding a comment than using that cast?

cmorty:
Yes, something like "We don't expect temperatures outside -128 to 127 C".

matttbe:
Thanks.

I just removed the explicit cast and added this comment:

> Casted to int8_t: We don't expect temperatures outside -128 to 127 C

alignan:
Nice catch, simple enough to merge already, thanks for adding the comment!

Issue #972, opened 2015-02-22
Cannot modify make variables for travis compile tests


g-oikonomou:
For travis compile tests, it would be nice to be able to run:
```
make VARIABLE=something
```
rather than simply `make`.

Motivation: To use one example to test two (or more) different compile configurations for the same platform (e.g. 6LoWPAN fragmentation on/off, RPL on/off, RS/RA on/off, platform-specific options etc)

I would probably configure this within the job's Makefile.

I had a look at the various Makefiles of our testing framework and I couldn't figure out a straightforward way to do this, nor could I think of a simple patch that could extend the framework to do something like that.

I wouldn't call this a bug, but certainly a limitation that IMHO needs addressed. Any ideas on how to patch things? I'll label it 'enhancement'.

markushx:
Is
```make DEFINES=VARIABLE=something```
as mentioned in README-BUILDING.md working for you?

**Comment by bthebaudeau** [comment]
Yes, `make DEFINES=VARIABLE=something` should work. There are even commands using it under `regression-tests/`.

The only exception is for the make files that already use it, because the command line variable would prevent the make file variable from working. In that case, the command line should use the value from the make file and add what's needed to it. Another solution would be to rewrite the make files using `DEFINE=`with `override DEFINE+=`, in which case the command line and the make file variables would be concatenated.

**Comment by g-oikonomou** [comment]
So I've either articulated the question incorrectly, or I'm being very blind here. I am talking about getting travis to run a test with a specific value for a variable, not running the test myself in my PCs command line. As far as I can tell travis simply runs `make TARGET=x` for compile tests?

**Comment by bthebaudeau** [comment]
If we take `regression-tests/Makefile.compile-test` and `regression-tests/18-compile-arm-ports/Makefile`, instead of putting a set of paths into the `EXAMPLES` variable, you should define a set of variables for each example, e.g.:
```
EXAMPLE_CC2538DK_DEFAULT_PATH=cc2538dk/cc2538dk
EXAMPLE_CC2538DK_DEFAULT_DEFINE=
EXAMPLE_CC2538DK_6LOWPAN_FRAG_ON_PATH=cc2538dk/cc2538dk
EXAMPLE_CC2538DK_6LOWPAN_FRAG_ON_DEFINE=SICSLOWPAN_CONF_FRAG=1
```

Then:
```
[...]

PR #1088, opened 2015-05-26
Implement generic Newlib syscalls

drandreas:
I implemented the platform specific system call for heap support.
I didn't implement the remaining platform independent system calls like _stat _fstat _open...

They are currently implemented in 
cpu/stm32w108/e_stdio/src/syscalls.c

I suggest moving the file to a location shared between all platforms.


cpu/cc2538/Makefile.cc2538
```diff
@@ -84,7 +85,7 @@ CUSTOM_RULE_LINK=1
 
 %.elf: $(CPU_STARTFILES) %.co $(PROJECT_OBJECTFILES) $(PROJECT_LIBRARIES) contiki-$(TARGET).a $(LDSCRIPT)
 	$(TRACE_LD)
-	$(Q)$(LD) $(LDFLAGS) ${filter-out $(LDSCRIPT) %.a,$^} ${filter %.a,$^} $(TARGET_LIBFILES) -o $@
+	$(Q)$(LD) $(LDFLAGS) -Wl,--start-group ${filter-out $(LDSCRIPT) %.a,$^} ${filter %.a,$^} $(TARGET_LIBFILES) -lc -Wl,--end-group -o $@
```
bthebaudeau: This would make more sense:
```
	$(Q)$(LD) $(LDFLAGS) ${filter-out $(LDSCRIPT) %.a,$^} -Wl,--start-group ${filter %.a,$^} $(TARGET_LIBFILES) -lc -Wl,--end-group -o $@
```

cpu/cc2538/cc2538.lds
```diff
@@ -69,6 +69,9 @@ MEMORY
     SRAM (RWX) : ORIGIN = SRAM_START, LENGTH = SRAM_LEN
 }
 
+//End of HEAP, BEGIN is relative, see below.
```
bthebaudeau: Comments in Contiki should be C-style, not C++. See [Contiki's code formatting rules](https://github.com/contiki-os/contiki/wiki/Code-Contributions#code-formatting).

cpu/cc2538/cc2538.lds
```diff
@@ -69,6 +69,9 @@ MEMORY
...
+_eheapdata = SRAM_START + SRAM_LEN;
```
bthebaudeau: Naming this `_eheap` would be simpler.

cpu/cc2538/cc2538.lds
```diff
@@ -103,6 +106,8 @@ SECTIONS
         *(.bss*)
         *(COMMON)
         _ebss = .;
+        //Begin of HEAP, END is absolute, see above.
+        _heapdata  = ABSOLUTE(.);
```
bthebaudeau: This badly needs to be rebased because of the latest stack changes. Put the heap after the stack, filling the end of the SRAM. Or, the required heap size could be a configuration option given as an application-specific estimated value, in order to fail at build time rather than at run time if the heap is too small for the application requirements, but this is probably too much (it's up to the application developers to test the pointers returned by the allocation functions).

`_heap` rather than `_heapdata`.

Why `ABSOLUTE()`? Where are you using relocation?

In the end, the comments are no longer needed, and I would simplify the changes in this file like this:
```
     .stack (NOLOAD) :
     {
         *(.stack)
     } > SRAM
 
+    _heap = .;
+    _eheap = SRAM_START + SRAM_LEN;
+
 #if (LPM_CONF_MAX_PM==2) && (LPM_CONF_ENABLE != 0)
     .nrdata (NOLOAD) :
     {
         _nrdata = .;
```

cpu/cc2538/syscalls.c
```diff
@@ -0,0 +1,60 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
```
bthebaudeau: Doxygen documentation missing.

cpu/cc2538/syscalls.c
```diff
@@ -0,0 +1,60 @@
...
+#include <sys/stat.h>
```
bthebaudeau: Why not `#include <sys/types.h>` rather than that?

cpu/cc2538/syscalls.c
```diff
@@ -0,0 +1,60 @@
...
+
+#define DEBUG 0
+#if DEBUG
+#include <stdio.h>
+#define PRINTF(...) printf(__VA_ARGS__)
+#else
+#define PRINTF(...)
+#endif
+
+static char *allocated = 0;
```
bthebaudeau: This has no reason to be at file scope, so move it into `_sbrk()`.

cpu/cc2538/syscalls.c
```diff
@@ -0,0 +1,60 @@
...
+
+caddr_t _sbrk(int incr) {
```
bthebaudeau: A function definition should not start like this. See the code formatting rules.

cpu/cc2538/syscalls.c
```diff
@@ -0,0 +1,60 @@
...
+  extern char _heapdata;  // symbols are defined in cc2538.lds
+  extern char _eheapdata; // they mark end of bss and end of memory.
```
bthebaudeau: `uint8_t` would be cleaner than `char`. This requires `#include <stdint.h>`.

C-style comments please.

cpu/cc2538/syscalls.c
```diff
@@ -0,0 +1,60 @@
...
+
+  if (allocated == 0) {
+    memset(&_heapdata, 0, (unsigned int)&_eheapdata - (unsigned int)&_heapdata);
```
bthebaudeau: I don't see any specification stating that the heap should be initially cleared. You can remove this and `#include <string.h>`.

cpu/cc2538/syscalls.c
```diff
@@ -0,0 +1,60 @@
...
+    allocated = &_heapdata;
+  }
+
+  if (allocated + incr > &_eheapdata) {
```
bthebaudeau: No space before opening parenthesis.

cpu/cc2538/syscalls.c
```diff
@@ -0,0 +1,60 @@
...
+    PRINTF("Out of heap space!\n");
+    return (caddr_t)0;
```
bthebaudeau: This should probably be `-1`, and `errno` should be set to `ENOMEM`.

cpu/cc2538/syscalls.c
```diff
@@ -0,0 +1,60 @@
...
+  } else {
```
bthebaudeau: Useless `else`: unreachable if you already have returned in the `if`.

cpu/cc2538/syscalls.c
```diff
@@ -0,0 +1,60 @@
...
+    allocated += incr;
+    return (caddr_t) allocated - incr;
+  }
+}
```
bthebaudeau: Why don't you just follow more or less the example provided by Newlib's documentation? This would give:
```
#include <sys/types.h>
#include <errno.h>
#include <stdint.h>

#define DEBUG 0
#if DEBUG
#include <stdio.h>
#define PRINTF(...) printf(__VA_ARGS__)
#else
#define PRINTF(...)
#endif

caddr_t
_sbrk(int incr)
{
  /* Heap boundaries defined in cc2538.lds. */
  extern uint8_t _heap;
  extern uint8_t _eheap;

  static uint8_t *heap_end = &_heap;
  uint8_t *prev_heap_end = heap_end;

  if(heap_end + incr > &_eheap) {
    PRINTF("Out of heap space!\n");
    errno = ENOMEM;
    return (caddr_t)-1;
  }

  heap_end += incr;
  return (caddr_t)prev_heap_end;
}
```

+ doxygen.

drandreas: I had copy and pasted the implementation of one of the other platforms.

cpu/cc2538/syscalls.c
drandreas: Sorry for that. It’s a residue from additional code not part of the pull request (heap usage statistics used in our project). For the same reason *allocated* is outside the function.

cpu/cc2538/Makefile.cc2538
drandreas: I have to revisit that one.

You might remember our conversation on the mailing list. It took me quite a while to get the linking with an external library, included with PROJECT_LIBRARIES += working.

cpu/cc2538/syscalls.c
drandreas: This is also a residue from our internal code. _fstat required sys/types.h.

This brings us back to my question/suggestion in the initial pull request. Shouldn't all "dummy" system calls like _fstat, _open and _unlink ... be implemented in a common file. Currently they are added bit by bit to every platform.

cpu/cc2538/Makefile.cc2538
bthebaudeau: Yes, I do remember that. `-Wl,--start-group` and `-Wl,--end-group` are meant to be used to enclose a list of archive files to be searched in a loop until all symbols are resolved. So object files have nothing to do in that list.

Normally, ld goes through the objects and archives in the order they are passed. The objects within archives that do not contain definitions unresolved by previously kept objects are discarded (even without using ld garbage collection, which is completely unrelated).

The issue here is that Contiki puts nearly everything into an archive. Because of that, without the archive group trick, `libc.a` is implicitly linked at the end, so `malloc()` and `_sbrk_r()` are kept because the objects and archives previously linked need them (if `malloc()` is used somewhere in an example). But in this process, `_sbrk()` was discarded with your `syscall.o` within `contiki-cc2538dk.a` because nothing directly needed it at the time it was linked (`libc.a` containing `malloc()` needing `_sbrk_r()` needing `_sbrk()` is linked later). Also, `contiki-cc2538dk.a` needs `libc.a` to resolve many other definitions. In the end, we have `contiki-cc2538dk.a` and `libc.a` depending on each other, which is impossible to link unless they are grouped in a symbol resolution loop.

It would be possible to avoid the archive groups, and even passing `-lc`, if `syscall.o` were not included in `contiki-cc2538dk.a`, or even better, if `contiki-cc2538dk.a` were removed (e.g. changed into a big `contiki-cc2538dk.o`). `contiki-cc2538dk.a` is not a real feature-centric library built in order to stay and be linked with applications at any time later, but just a convenient object bag to ease the build process of several examples at once, and in this case it complicates things. Also, being limited to a single instance of a given file name is not clean, but this is another topic.

There are also some magical dangerous tricks used in Contiki that depend on this behavior of ld towards archives. E.g., the commit f7baf5aba2cb2916eb6e44c8ed40bcc70e145cd3 (@g-oikonomou) has introduced weak aliases for the USB and UART ISRs that can only work because of that. With the current code, if UART is not used, then nothing needs `uart.o` within `contiki-cc2538dk.a`, so `uart.o` is fully discarded, even though it contains the strong `uart0_isr()` definition, because ld is satisfied at this point with the weak alias. Instead, with a big `contiki-cc2538dk.o`, the strong definition would win, so the wrong ISR would be linked in, which could be fixed by keeping the strong definition of `uart0_isr()` in `uart.c` only `#if UART_CONF_ENABLE`. It is probably safer not to rely on the current implicit behavior, because changing the build system would introduce the issue described here, and it would remain unnoticed for a very long time. With this example, we also see that the build system builds a lot of things that are using for a given example, which is cool for testing with Travis, but this slows down the build.

cpu/cc2538/syscalls.c
bthebaudeau: If I take `_open()`, I see it defined in `cpu/arm/at91sam7s/newlib-syscalls.c`, `cpu/stm32w108/e_stdio/src/syscalls.c` and `cpu/stm32w108/hal/micro/cortexm3/stm32w108/crt-stm32w108.c`. They're indeed dummy, which means that they probably had to be introduced only because the build system is broken (see the archive group build trick just discussed, and try to put `-lc` at various places without the archive group, and ld will eventually complain about `_open()` and the like not being defined). Newlib does not provide these.

If dummy implementations are needed, something has to be wrong somewhere, because it means that an unsupported feature of Newlib is used, or that the build system is broken. If something unsupported is used, then we want to detect it at build time, so using dummy implementations to fix the build process is wrong. By not doing this on CC2538, the link will fail if something unsupported is used, which is what we want.

Apart from that, there are many things duplicated on many platforms, including only on ARM platforms, not only to implement common features, but also because of diverging local APIs for the same features. All common stuff should be factored out and reused as much as possible, but doing this from the current state would be a huge work.

cpu/cc2538/syscalls.c
bthebaudeau: I have checked: I confirm that Newlib needs `-1` here.

As to `errno`, it needs something, but it's more complicated, because the `errno` from `errno.h`is not the same as the `errno` used by `_sbrk_r()`. See https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=newlib/libc/include/sys/errno.h;h=a72c37320a7681ae28da520ad719139236e378fc;hb=HEAD vs. https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=newlib/libc/reent/sbrkr.c;h=44e42b57f6ad0d6f395fe29bf6a1294fc64f4462;hb=HEAD. The former uses the function from https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=newlib/libc/errno/errno.c;h=fd1743d73625bf098fc23af9c06924381f840139;hb=HEAD to access the `errno` field inside `_REENT`, while the latter uses a global `errno` variable. To make this work, we have to do like `_sbrk_r()` since we are constrained by it, even if this is not very portable, but we are already implementing a Newlib-specific function, so it's not that bad. This would also probably require to add a comment. I update my suggested code at the bottom accordingly.

cpu/cc2538/Makefile.cc2538
g-oikonomou: Thanks for taking the time to share your thoughts and for the in-depth analysis.

I'll shit this from a line note to the PR's discussion, cause a `pull -f` will hide it.

g-oikonomou:
So, allow me to think aloud about our requirements here:

* Dumping everything into a huge `.a` is not strictly needed. CC2538 does it because it was common practice in other platforms, but if you look at CC2430 and CC2530 (cc253x), they do not. If we have good reasons to get rid of it, I cannot think of any reasons preventing us from doing so.
* If an external library provides a symbol that is also provided by Contiki, the latter should win
* Strong symbols should win over weak ones if - and only if - they are required. @bthebaudeau 's comments about interrupt handlers is a perfect example.
* We should be explicit wherever possible, rather than relying on implicit behaviour. At the same time though, I see nothing immediately wrong with relying on linker behaviour, as long as i) it's a documented linker behaviour and one that is therefore predictable, reliable and unlikely to change in the future, ii) we rely on it intentionally because we want to exploit it towards a specific goal (as opposed to fluking things). If both are true, then we should be able to pick up problems during PR review.

So: Assuming we got rid of the archive and we passed all objects under `obj_cc2538dk` to the linker one-by-one, what would happen? I've not tried, but my guess is that we would always end up linking the strong symbols, leading to the wrong interrupt handlers making their way into the image, in a similar fashion to the discussion earlier. We can address this by wrapping handlers within `#if` guards. How would this PR be influenced if we were to do this? How would linking external libraries be influenced by this? Would satisfying our requirements become easier? Harder? The same?

Regarding dummy implementations, my guess is that those are there because of { poor ; quick 'n' dirty ; not very well thought-of } porting efforts. I am not familiar with how `cpu/arm/common` got to its current state, but I do know that i) the code therein looks a little messy and that ii) we don't have a dedicated maintainer for it. I am 100% against dummy implementations and 100% in agreement with Benoît: If `something` requires `something_else` and `something_else` does not exist, the build should fail. The only exception to this are implementations required by Contiki's core APIs (and then again we should arguably fix the API so they can be avoided, but that's another story). A classic example is `watchdog_stop()`, which is something that is plain impossible for some CPUs, yet we have to provide an implementation for it. Same applies for some of the `clock_delay_` stuff.

Regarding re-using code:

* We should obviously do this as much as we can and I can immediately think of two examples where CC2538 does exactly that:
  * It compiles the USB library under `cpu/cc253x/usb/common`, rather than copying it over. These files are platform-independent enough and could potentially be moved to `core` one day.
  * It compiles `dbg-printf.c` and friends from `arm/common/dbg-io` rather than copying them over. These files are IMHO not generic-enough to be moved outside `cpu`.
* Unfortunately, there is no immediate and well-established way to share code across _some_ but not _all_ architectures. The discussion about nested platforms is also part of this same situation. One idea could be `cpu/lib/arm`. My current thinking is that `syscalls.c` proposed here would also belong somewhere in that directory.

> All common stuff should be factored out and reused as much as possible, but doing this from the current state would be a huge work.

Agreed, but then again we can always create an issue and start knocking together a list of those occurrences, in view of fixing them one-by-one in the longer term. Creating an issue would have many benefits:
* We won't forget anything
* The issue will be there, annoying us to close it :)
* It will be more visible to the community than this discussion, so it could attract contributions.

I see the fix to this as a three-pronged approach:
* Prevent new such occurrences during PR review (once we have agreed on an acceptable way of re-using stuff).
* Fix some existing occurrences for some platforms.
* Allow some existing occurrences to disappear as older platforms are being phased out. A prime example is the Sensinode platform, which I am planning to blow to kingdom come at some point in the near future, and I would therefore discourage anyone from putting any effort to improve it.

Moving stuff to core must always respect the need that Contiki must keep building successfully for all N different toolchains used by the various targets, many of which are not `-gcc`.

g-oikonomou:
> I am not familiar with how `cpu/arm/common` got to its current state, but I do know that i) the code therein looks a little messy and that ii) we don't have a dedicated maintainer for it.

In reality I should have written "`cpu/stm32w108` and most of `cpu/arm`"

bthebaudeau:
>  - We should be explicit wherever possible, rather than relying on implicit behaviour. At the same time though, I see nothing immediately wrong with relying on linker behaviour, as long as i) it's a documented linker behaviour and one that is therefore predictable, reliable and unlikely to change in the future,

This last condition is often difficult to predict. If we take the handling of libraries, it has already changed in GCC in the recent past.

> ii) we rely on it intentionally because we want to exploit it towards a specific goal (as opposed to fluking things). If both are true, then we should be able to pick up problems during PR review.

I would add that we should also take care about portability when doing such things, especially if we want to support several toolchains. Also, the issue with implicit toolchain behaviors used intentionally is that anything going wrong might be unnoticed, and people will forget about how it works after some time, or even forget that such a thing was done.

> So: Assuming we got rid of the archive and we passed all objects under `obj_cc2538dk` to the linker one-by-one, what would happen? I've not tried, but my guess is that we would always end up linking the strong symbols, leading to the wrong interrupt handlers making their way into the image, in a similar fashion to the discussion earlier. We can address this by wrapping handlers within `#if` guards.

Correct, or with even more infrastructure rework, we could just not build at all the unused drivers/modules. Something like kbuild + kconfig would be great for that. The ld garbage collector would do the rest. This would also provide a menuconfig to ease and partly automate the configuration process, rather than having tons of configuration files.

> How would this PR be influenced if we were to do this?

The new make rule after this change for `%.elf` would no longer have to be changed by this PR in order to link the local implementations of the system calls used by Newlib.

> How would linking external libraries be influenced by this?

It could only help to avoid the same kind of link nightmare as in this PR. All the unused stuff would be garbage-collected.

> Would satisfying our requirements become easier? Harder? The same?

I would say rather a bit easier:
 - This would naturally prevent the use of this implicit linker behavior.
 - Regarding strong vs. weak symbols:
    * With only objects, strong symbols always win, so they should not build compiled in if unwanted.
    * With archives, considering the philosophy of libraries, I think that it would only make sense to have weak symbols in libraries, not strong ones, and everything works fine if we follow this rule. The other way around, the strong symbols are kind of indirectly weakened, which can be confusing, like for the ISR example.

> If `something` requires `something_else` and `something_else` does not exist, the build should fail. The only exception to this are implementations required by Contiki's core APIs (and then again we should arguably fix the API so they can be avoided, but that's another story). A classic example is `watchdog_stop()`, which is something that is plain impossible for some CPUs, yet we have to provide an implementation for it. Same applies for some of the `clock_delay_` stuff.

I have not looked into this, but how can these exceptions work if the core needs some behaviors that are unsupported by some architectures? E.g. clearing a watchdog would make sense for architectures with a watchdog and a short watchdog period compared to some core operations, in which case it is more for the platform than for core requirements, so weak empty functions could be added to the core for such cases. But why would the core need to stop a watchdog? For the clock delay functions, e.g. if no timer is available on some platforms, weak default functions using loops could be added to the core.

> Regarding re-using code:
> - We should obviously do this as much as we can and I can immediately think of two examples where CC2538 does exactly that:
>    * It compiles the USB library under `cpu/cc253x/usb/common`, rather than copying it over. These files are platform-independent enough and could potentially be moved to `core` one day.

Indeed.

>    * It compiles `dbg-printf.c` and friends from `arm/common/dbg-io` rather than copying them over. These files are IMHO not generic-enough to be moved outside `cpu`.

I don't see why these files even exist in Contiki, all the more they do not even support `%f` though they have an aborted implementation in `cpu/arm/common/dbg-io/strformat.c`. For me, `printf()` should always use the implementation provided by the toolchain (which can often be tuned to be more or less complete in order to reduce size). Then, the system calls required by the toolchain should be implemented in order to actually print something to UART, USB CDC, etc. This would also automatically give access to `perror()`. Even `fprintf()` could be handled in this way to support printing to several devices, or even to CFS.

On top of that, instead of `PRINTF()` everywhere for debug, we should have a common `debug.h` or `utils.h` header file defining `DBG()` if `DEBUG` was defined prior to including this header file. This macro would automatically print the file name, the function name and the line number. There could also be more advanced variations, like `warn()`/`warnx()`/`err()`/`errx()` functions similar to the BSD extensions (http://linux.die.net/man/3/errx).

> - Unfortunately, there is no immediate and well-established way to share code across some but not all architectures. The discussion about nested platforms is also part of this same situation. One idea could be `cpu/lib/arm`.

If the unused modules were just not built, there would be no issues here.

> My current thinking is that `syscalls.c` proposed here would also belong somewhere in that directory.

This is not ARM-specific. This is not even CPU-specific. It's just Newlib-specific, so it should ideally be placed in a new common place for toolchain-specific things. The only thing to take care of is the standardization of the `_heap` and `_eheap` symbol names among all the linker scripts using this.

> Agreed, but then again we can always create an issue and start knocking together a list of those occurrences, in view of fixing them one-by-one in the longer term. Creating an issue would have many benefits:
[...]
I see the fix to this as a three-pronged approach:
[...]

Sure.

> Moving stuff to core must always respect the need that Contiki must keep building successfully for all N different toolchains used by the various targets, many of which are not `-gcc`.

Yes. This is also why no toolchain-specific features should be directly used anywhere in the code, in order to make all the code as portable as possible, e.g. to easily support a new toolchain later on a platform: Such toolchain-specific features should be abstracted, and the definition of the abstracting macros should be placed in a common place. Platform code should be allowed to directly use toolchain-specific features only if there is no other way, and in such cases the toolchain should be detected and a `#error` should be issued for unsupported toolchains (a bit like for the discussion about the `WEAK_ALIAS()` macro).

Here is the Contiki tree hierarchy that I think would be great, copying Linux's or U-Boot's:
 - `apps`: the current `examples`, not the current `apps`, which are actually more library features.
 - `arch/arm/cpu/armv7m/cc2538`: very specific things like initialization stuff, custom APIs for specfic things like clock handling, etc., the more common things being in the upper folders. Folders could also be created under `armv7m` for things common to `cc2538` and `cc26xx` if any.
 - `board/cc2538dk`: a bit like the current platforms, but with a lot of things factored out, keeping here only the board-specific initialization.
 - `doc`: documentation.
 - `drivers`: all the drivers: MCU peripherals (UART, USB, SPI, I²C, GPIO, etc., making things easy if they are shared by several MCUs), but also external devices (sensors, extenders, interfaces like ENC28J60, etc.) or the common USB stack.
 - `fs`: all file systems, including CFS.
 - `include`: common header files.
 - `kernel`: more or less the current `core`, but focusing more on the OS kernel features (processes, threads, etc.).
 - `lib`: the current `apps`, but also the current `core/lib` and other common things like toolchain-specific abstractions.
 - `net`: all the network protocols, including the 6lowpan stack, so the current `core/net`.
 - `test`: the current `regression-tests`.
 - `tools`: like now.

g-oikonomou:
So let's create a pull to get rid of the archive and tidy-up the linking of handlers, then we can rebase this one here.

Regarding building vs not building drivers/modules etc, there have been various discussions in the past and the outcome has always been that we want to do things with `#define`s to take advantage of automatic dependency generation and to avoid having to `make clean` after changes.

> But why would the core need to stop a watchdog? 

Beats me. This is only used in 1 place in the core in `cxmac.c`, and from a quick look at the code it is probably being used because the author selected to "stop, do lengthy stuff, start it again" instead of refreshing it inside the code.

So if we are not using cxmac, we don't _strictly_ have to provide a dummy. However, the function is declared unconditionally in the core `watchdog.h` and therefore one would normally expect an implementation to be provided.

> For the clock delay functions, e.g. if no timer is available on some platforms, weak default functions using loops could be added to the core.

`void clock_delay(unsigned int delay)` was deprecated as "platform-specific" and also because nobody really knew what the unit of `delay` was. It was obsoleted in favour of the more sane `clock_delay_usec()`.

That was the theory. Unfortunately, `leds.c`, `dev/cc2420.c` and `dev/cc2520.c` still call the former. We (CC2538) don't really care about those two RF drivers, but we have to provide the symbol because of `leds.c`. Thus, in practice, what we have is a "deprecated" function that must be implemented by _all_ platforms, therefore perhaps not that deprecated after all ;)

> I don't see why these files even exist in Contiki, all the more they do not even support %f though they have an aborted implementation in cpu/arm/common/dbg-io/strformat.c.

Not sure why they are there either, but philosophical discussion aside, they do exist and they do work. So the immediate decision here should be "where do we put those things in order to be shared more easily, but without having to do a huge re-write?"

> This is not ARM-specific. This is not even CPU-specific. It's just Newlib-specific, so it should ideally be placed in a new common place for toolchain-specific things. The only thing to take care of is the standardization of the _heap and _eheap symbol names among all the linker scripts using this.

Agreed. So where do we want said common place to be? `cpu/lib`? `/lib`?

bthebaudeau:
> So let's create a pull to get rid of the archive and tidy-up the linking of handlers, then we can rebase this one here.

I have something ready for that. But I have a question. If we put the ISRs within `#if`s, then the USB ISR will never be built by Travis (because no example is configured to use it AFAIK). Is this an issue (not all C preprocessor paths are tested anyway because there are many possible configurations that are not present in the tree)? Maybe it would be better to just revert the `WEAK_ALIAS()` change. What do you think? We will always have this issue with weak vs. strong symbols if we get rid of the archive (and even worse with the archive if a strong symbol with a weak counterpart somewhere else is alone in its `.c` file, because there would be no way to link it), but it's healthier to deal with it explicitly as discussed before.

> Regarding building vs not building drivers/modules etc, there have been various discussions in the past and the outcome has always been that we want to do things with `#define`s to take advantage of automatic dependency generation and to avoid having to `make clean` after changes.

Linux handles dependencies pretty well, while only building what's needed. A set of configurations dedicated to testing can then be used to maximize the code coverage at build time (or there can simply be a rule enforcing each piece of code to be enabled in at least one standard configuration). But switching to such a build system would require massive changes in Contiki.

It would also be possible to simply automatically generate make variables from the configuration files in order to build only what's required. The configuration files and all the pieces of make files would be global make dependencies. This is what U-Boot did in the past, before switching to kbuild + kconfig.

In all these cases, no `make clean` is required after changes.

The concept of module recently introduced in Contiki could also help here, but it is at folder level rather than at file level, and it has the disadvantage of splitting this part of the configurations.

If everything is built, then the choices apart from Contiki modules are GCC archives, or LD garbage collection, or putting whole source files within `#if`s, only the latter saving a bit of build time.

> > But why would the core need to stop a watchdog?

> Beats me. This is only used in 1 place in the core in `cxmac.c`, and from a quick look at the code it is probably being used because the author selected to "stop, do lengthy stuff, start it again" instead of refreshing it inside the code.

Then `cxmac.c` could be changed to rather call `watchdog_periodic()` somewhere in the loop.

> So if we are not using cxmac, we don't *strictly* have to provide a dummy. However, the function is declared unconditionally in the core `watchdog.h` and therefore one would normally expect an implementation to be provided.

The declarations in `watchdog.h` are just here to impose an API. If one calls `watchdog_stop()` but this feature is unavailable on some platform while the build succeeds, the user should expect that this function actually stops the watchdog. If it does not, then nasty bugs will eventually appear somewhere. Unsupported feature that are used should be detected ASAP, so the best is to have a build failure in such a case, so `watchdog_stop()` should never be defined if unsupported.

Only features that are transparent if unavailable should have weak dummy implementations in the core.

> > For the clock delay functions, e.g. if no timer is available on some platforms, weak default functions using loops could be added to the core.

> `void clock_delay(unsigned int delay)` was deprecated as "platform-specific" and also because nobody really knew what the unit of `delay` was. It was obsoleted in favour of the more sane `clock_delay_usec()`.

> That was the theory. Unfortunately, `leds.c`, `dev/cc2420.c` and `dev/cc2520.c` still call the former. We (CC2538) don't really care about those two RF drivers, but we have to provide the symbol because of `leds.c`. Thus, in practice, what we have is a "deprecated" function that must be implemented by all platforms, therefore perhaps not that deprecated after all ;)

OK. It would be worth cleaning this too by converting everything to the new API.

> > I don't see why these files even exist in Contiki, all the more they do not even support %f though they have an aborted implementation in cpu/arm/common/dbg-io/strformat.c.

> Not sure why they are there either, but philosophical discussion aside, they do exist and they do work. So the immediate decision here should be "where do we put those things in order to be shared more easily, but without having to do a huge re-write?"

Once the current PR is completed and merged, the syscall that it implements will make it possible to use the set of `iprintf()` functions from Newlib, which do exactly the same, without complicating things with files. There could even be a configuration to choose between `printf()` and `iprintf()` to save size or not depending on the application requirements. This would have avoided the introduction of `print_float()` in #677.

But in the meantime, the current code has nothing ARM-specific, so it should be moved to a new `/lib/`, or simply to `/core/lib/`. This should be kept there as a folder in order to be able to add it as a module for the build, or `#define`s would have to be added everywhere in those file. Besides the `printf()` helpers, there are just some obscure APIs that may be improved for a better sharing, but it does not matter much until there are new users.

> > This is not ARM-specific. This is not even CPU-specific. It's just Newlib-specific, so it should ideally be placed in a new common place for toolchain-specific things. The only thing to take care of is the standardization of the _heap and _eheap symbol names among all the linker scripts using this.

> Agreed. So where do we want said common place to be? cpu/lib? /lib?

New `/lib/`, or simply to `/core/lib/`, maybe inside a `newlib` sub-folder, which would become a module.

g-oikonomou:
> I have something ready for that. But I have a question. If we put the ISRs within `#ifs`, then the USB ISR will never be built by Travis (because no example is configured to use it AFAIK). Is this an issue (not all C preprocessor paths are tested anyway because there are many possible configurations that are not present in the tree)? Maybe it would be better to just revert the `WEAK_ALIAS()` change. What do you think?

That's a very good point about travis. I wouldn't call it a show-stopping issue, but it would be nice to keep compile-testing as much as we can. I kinda like weak aliases, what if we were to rename ISRs when not in use to prevent the linker from selecting them. Something like `void usb_isr_unused(void)`. Too ugly? The only other alternative that I can think of would be to create specific "examples" (e.g. `examples/cc2538/travis-tests` to compile-test specific parts of the code.

Also see #972 in a similar context. 

> Then `cxmac.c` could be changed to rather call `watchdog_periodic()` somewhere in the loop.

Looks like cxmac stops the dog when it starts sending strobes and restarts it afterwards. It should be easy to simply refresh as part of the loop. I'll have a look.

> so the best is to have a build failure in such a case, so watchdog_stop() should never be defined if unsupported

Point taken, I'll open a pull for this

> OK. It would be worth cleaning this too by converting everything to the new API.

Yes, however there is a slight problem. Some CPUs don't provide `clock_delay_usec`... Crucially, all msp430 flavours are among them.

So before we can convert to the new API, we would first need all CPUs to implement `_usec()`. Either that or we'll have to come up with something very ugly to conditionally choose between the two.

Once we have that then `leds.c` would be a start. I am unfortunately neither interested nor familiar in the CC2x20 drivers.

bthebaudeau:
> That's a very good point about travis. I wouldn't call it a show-stopping issue, but it would be nice to keep compile-testing as much as we can.

Sure.

> I kinda like weak aliases,

Me too, but they don't fit here. Again, they would be perfect if unused code were not built. Besides the lack of testing by Travis, I do not like putting whole UART/USB drivers within `#if`s just for this purpose, because it's `#if` bloat and we should do it for all files in that case, not making a specific case for files containing ISRs. This should really be done at Makefile level.

Putting each driver in a module would not be cool either, because that would mean writing the configuration file twice, once for the C preprocessor, and once for make. So the Makefile should really be changed automatically depending on the configuration file.

> what if we were to rename ISRs when not in use to prevent the linker from selecting them. Something like `void usb_isr_unused(void)`. Too ugly?

Yes, it would really be a makeshift for the sole purpose of keeping the weak aliases. And it would not require more lines of code to remove the weak aliases.

> The only other alternative that I can think of would be to create specific "examples" (e.g. `examples/cc2538/travis-tests` to compile-test specific parts of the code.

That would make sense, but rather if the whole build system were changed to only build the needed parts.

> Yes, however there is a slight problem. Some CPUs don't provide `clock_delay_usec`... Crucially, all msp430 flavours are among them.

> So before we can convert to the new API, we would first need all CPUs to implement `_usec()`. Either that or we'll have to come up with something very ugly to conditionally choose between the two.

> Once we have that then `leds.c` would be a start. I am unfortunately neither interested nor familiar in the CC2x20 drivers.

Are there official maintainers for these parts of the code?

g-oikonomou:
>  Besides the lack of testing by Travis, 
> [...]
> This should really be done at Makefile level.

So assuming we could come up with a way to build stuff conditionally without having to re-write significant portions of Contiki's build system: How would that help us teach travis to compile-test the handlers in question when they are not in use?

> Are there official maintainers for these parts of the code?

Which parts? The CPUs that do not provide `_delay_usec()` or the two RF drivers? For the latter most commits point to Adam, with contributions by Joakim and Niclas Finne.

bthebaudeau:
> So assuming we could come up with a way to build stuff conditionally without having to re-write significant portions of Contiki's build system: How would that help us teach travis to compile-test the handlers in question when they are not in use?

That would not help for that. That would help to keep the weak aliases without adding `#if`s everywhere. For Travis, an all-yes-config would have to be create somewhere.

> Which parts? The CPUs that do not provide `_delay_usec()` or the two RF drivers?

Both.

> For the latter most commits point to Adam, with contributions by Joakim and Niclas Finne.

Maybe we can contact them to see if they can help, at least with some information to ease those changes.

bthebaudeau:
After #1102, the current PR should become:
```
    Implemented CPU specific newlib syscalls

---
 cpu/cc2538/Makefile.cc2538 |  1 +
 cpu/cc2538/cc2538.lds      |  3 ++
 cpu/cc2538/syscalls.c      | 69 ++++++++++++++++++++++++++++++++++++++++++++++
 3 files changed, 73 insertions(+)
 create mode 100644 cpu/cc2538/syscalls.c

diff --git a/cpu/cc2538/Makefile.cc2538 b/cpu/cc2538/Makefile.cc2538
index 56ce4a2..32e63e3 100644
--- a/cpu/cc2538/Makefile.cc2538
+++ b/cpu/cc2538/Makefile.cc2538
@@ -54,6 +54,7 @@ CONTIKI_CPU_SOURCEFILES += cc2538-rf.c udma.c lpm.c
 CONTIKI_CPU_SOURCEFILES += dbg.c ieee-addr.c
 CONTIKI_CPU_SOURCEFILES += slip-arch.c slip.c
 CONTIKI_CPU_SOURCEFILES += i2c.c cc2538-temp-sensor.c vdd3-sensor.c
+CONTIKI_CPU_SOURCEFILES += syscalls.c
 
 DEBUG_IO_SOURCEFILES += dbg-printf.c dbg-snprintf.c dbg-sprintf.c strformat.c
 
diff --git a/cpu/cc2538/cc2538.lds b/cpu/cc2538/cc2538.lds
index cfd4cb4..579bc33 100644
--- a/cpu/cc2538/cc2538.lds
+++ b/cpu/cc2538/cc2538.lds
@@ -111,6 +111,9 @@ SECTIONS
         *(.stack)
     } > SRAM
 
+    _heap = .;
+    _eheap = SRAM_START + SRAM_LEN;
+
 #if (LPM_CONF_MAX_PM==2) && (LPM_CONF_ENABLE != 0)
     .nrdata (NOLOAD) :
     {
diff --git a/cpu/cc2538/syscalls.c b/cpu/cc2538/syscalls.c
new file mode 100644
index 0000000..ef33b62
--- /dev/null
+++ b/cpu/cc2538/syscalls.c
@@ -0,0 +1,69 @@
+/*
+ * Copyright (c) 2014, Institute for Pervasive Computing, ETH Zurich.
+ * All rights reserved.
+ *
+ * Author: Andreas Dröscher <contiki@anticat.ch>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+/* + doxygen */
+#include <sys/types.h>
+#include <errno.h>
+#include <stdint.h>
+
+#define DEBUG 0
+#if DEBUG
+#include <stdio.h>
+#else
+#define PRINTF(...)
+#endif
+
+caddr_t
+_sbrk(int incr)
+{
+  /*
+   * Newlib's _sbrk_r() assumes that this global errno variable is used here,
+   * which is different from the errno definition provided by <errno.h>.
+   */
+#undef errno
+  extern int errno;
+
+  /* Heap boundaries defined in cc2538.lds. */
+  extern uint8_t _heap;
+  extern uint8_t _eheap;
+
+  static uint8_t *heap_end = &_heap;
+  uint8_t *prev_heap_end = heap_end;
+
+  if(heap_end + incr > &_eheap) {
+    PRINTF("Out of heap space!\n");
+    errno = ENOMEM;
+    return (caddr_t)-1;
+  }
+
+  heap_end += incr;
+  return (caddr_t)prev_heap_end;
+}
```

`/* + doxygen */` should be replaced with a doxygen documentation throughout this file.

This definition of `PRINTF()` is really dirty because it is a new copy of the gazillion similar definitions of `PRINTA()`/`ANNOTATE()`/`PRINTF()`/`PRINTD()`/`PRINTDEBUG()`/etc. that have been allowed to accumulate in Contiki. All these definitions should be factored out to a common `/core/lib/debug-print.h` that would be similar to `/core/net/ip/uip-debug.h`, but improved with a `static inline int dummy_printf(const char *format, ...) {return 0;}` without `DEBUG_*` in order not to generate warnings for the passed variables that become unused in that case, and with GCC format checking (see the `format` attribute) if GCC is detected, in order to test the format compliance even without `DEBUG_*`. Unfortunately, this change would be a huge move, so maybe it would be better to just include `/core/net/ip/uip-debug.h` in the meantime, but the issue is that this feature has nothing to do with `uip`. @g-oikonomou: What do you think?

It would be great if `syscalls.c` were moved to a common place like `/core/lib/newlib/`, which would be added as a module in the CPU make files. It would be even better if this were a core module always built in, detecting the presence of Newlib in order to avoid adding it for all platforms using Newlib, and making all its functions weak (Newlib requires GCC, so this attribute is always available here) so that some platforms can override some functions but not all. This would probably require to fix some conflicts with existing syscalls implementations on other platforms.

g-oikonomou:
Regarding the debug situation, I thought I had commented on it already. This has been discussed in the past repeatedly, on occasion leading to [epic discussion threads](http://thread.gmane.org/gmane.os.contiki.devel/13625/focus=13626) (also [this page](https://github.com/contiki-os/contiki/wiki/Debugging-system%3A-ideas-on-how-to-improve-the-debugging-system) in the wiki). This didn't really get resolved, and I think we should avoid discussing it here.

For the purposes of this pull: Some people seem to prefer the `uip-debug.h` approach. I personally don't hugely mind those local `#if` blocks and I would not stop a pull because of using either one approach instead of the other one. If and when we decide to change the situation, both are likely to need replaced anyway. I normally `#include <stdio.h>` outside the `#if` block though, so I can easily enable a single `PRINTF` without having to touch the conditional block.

> but improved with a `static inline int dummy_printf(` ...

`static inline` used to be very suboptimal with SDCC (probably still is), so adding instances of it under `core` will make yours truly cry :)

> It would be great if `syscalls.c` were moved to a common place

Yes, but I am still a little uneasy about putting it under core - not sure whether it would go down well with other maintainers.

> ... in order to avoid adding it for all platforms using Newlib

So just to be clear, how many platforms are we really talking about here? Even though we've established that this stuff is not arm-specific, I am only seeing them defined by some of the arm CPUs. If that is the case, adding this as a `MODULE` would simply require changing a very small number of makefiles (5 or so I would think).

bthebaudeau:
> Regarding the debug situation, I thought I had commented on it already.

Not about the same thing. Previously, I was suggesting to completely replace the existing mechanism, but my last comment was only about factoring out and merging the existing definitions without changing them.

> This has been discussed in the past repeatedly, on occasion leading to epic discussion threads (also this page in the wiki). This didn't really get resolved, and I think we should avoid discussing it here.

OK.

> For the purposes of this pull: Some people seem to prefer the `uip-debug.h` approach. I personally don't hugely mind those local #if blocks and I would not stop a pull because of using either one approach instead of the other one. If and when we decide to change the situation, both are likely to need replaced anyway.

Agreed.

> I normally `#include <stdio.h>` outside the `#if` block though, so I can easily enable a single `PRINTF` without having to touch the conditional block.

If you prefer. One way or the other does not matter much here since there is no possible issue with `<stdio.h>` without `DEBUG`, so let's move it out of the `#if`.

> > It would be great if syscalls.c were moved to a common place

> Yes, but I am still a little uneasy about putting it under core - not sure whether it would go down well with other maintainers.

Then, under a new `/lib/newlib/`?

> > ... in order to avoid adding it for all platforms using Newlib

> So just to be clear, how many platforms are we really talking about here? Even though we've established that this stuff is not arm-specific, I am only seeing them defined by some of the arm CPUs. If that is the case, adding this as a `MODULE` would simply require changing a very small number of makefiles (5 or so I would think).

It's about all platforms using Newlib that might be interested in using `malloc()` or standard functions requiring it, not only about the platforms that currently define `_sbrk()`. And we are even talking about a global mechanism for Newlib system calls, not only about `malloc()`. And the number should not be a criterion of choice. But for this PR and for now, I agree that moving this file to a common place and enabling it as a module on CC2538 should be enough. It's not even the purpose of this PR to take care about other platforms. This PR introduces a common syscall, and other platforms can switch to it when they want. An issue could be created to track the switch of all platforms to this new syscall.

g-oikonomou:
> Then, under a new `/lib/newlib/`?

That might make more sense.

@nvt @adamdunkels Short story: This is about providing implementations of some Newlib system calls (e.g. `_sbrk()` here, but possibly more in the future). We want those to be shared across multiple CPUs, but they won't be relevant to all CPUs and they won't necessarily be portable. One idea is `core/lib/newlib` (which makes me feel a little uneasy in terms of non-portability), another idea is `lib/newlib`. These will be pulled in as `MODULE`s at this stage. Thoughts about where to place them?

alignan:
as the new PR policy, I'm merging this in a month if no other comment or change is made.

bthebaudeau:
@alignan: Please read these comments:
https://github.com/contiki-os/contiki/pull/1088#issuecomment-109452920
https://github.com/contiki-os/contiki/pull/1088#issuecomment-109484360

Basically, we're waiting for some feedback about #1102, and for answers from the core maintainers about where they would like to put things.

alignan:
Nice! then let's wait for updates.

bthebaudeau:
@drandreas Now that #1102 has been merged, please rebase and take all the comments above into account, which should change your commitdiff to:
```
commit d86c2329a48338f1ab469347a9d01a0ba372fb7e
Author: Andreas Dröscher <contiki@anticat.ch>
Date:   Tue May 26 20:03:37 2015 +0200

    Implement generic Newlib syscalls

diff --git a/cpu/cc2538/Makefile.cc2538 b/cpu/cc2538/Makefile.cc2538
index f37e34d..01acb4e 100644
--- a/cpu/cc2538/Makefile.cc2538
+++ b/cpu/cc2538/Makefile.cc2538
@@ -77,6 +77,8 @@ CPU_STARTFILES = ${addprefix $(OBJECTDIR)/,${call oname, $(CPU_START_SOURCEFILES
 CONTIKI_SOURCEFILES += $(CONTIKI_CPU_SOURCEFILES) $(DEBUG_IO_SOURCEFILES)
 CONTIKI_SOURCEFILES += $(USB_CORE_SOURCEFILES) $(USB_ARCH_SOURCEFILES)
 
+MODULES += lib/newlib
+
 .SECONDEXPANSION:
 
 ### Don't treat the .elf as intermediate
diff --git a/cpu/cc2538/cc2538.lds b/cpu/cc2538/cc2538.lds
index ce90ebf..1899687 100644
--- a/cpu/cc2538/cc2538.lds
+++ b/cpu/cc2538/cc2538.lds
@@ -102,6 +102,9 @@ SECTIONS
         *(.stack)
     } > FRSRAM
 
+    _heap = .;
+    _eheap = ORIGIN(FRSRAM) + LENGTH(FRSRAM);
+
     .nrdata (NOLOAD) :
     {
         _nrdata = .;
diff --git a/lib/newlib/syscalls.c b/lib/newlib/syscalls.c
new file mode 100644
index 0000000..72b7161
--- /dev/null
+++ b/lib/newlib/syscalls.c
@@ -0,0 +1,92 @@
+/*
+ * Copyright (c) 2014, Institute for Pervasive Computing, ETH Zurich.
+ * All rights reserved.
+ *
+ * Author: Andreas Dröscher <contiki@anticat.ch>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+/**
+ * \addtogroup lib
+ * @{
+ *
+ * \defgroup newlib Generic Newlib customizations
+ *
+ * Library providing generic implementations of Newlib features for Contiki
+ * @{
+ *
+ * \file
+ * System calls
+ */
+#include <sys/types.h>
+#include <errno.h>
+#include <stdint.h>
+#include <stdio.h>
+#define DEBUG 0
+#if DEBUG
+#else
+#define PRINTF(...)
+#endif
+/**
+ * \brief Enlarges the allocated heap space
+ * \param incr Number of bytes by which to increase the heap space
+ * \return The previous end of heap on success (which is also a pointer to the
+ *         start of the newly allocated memory if \p incr is positive), or
+ *         <tt>(caddr_t)-1</tt> with \c errno set to \c ENOMEM on error
+ */
+caddr_t
+_sbrk(int incr)
+{
+  /*
+   * Newlib's _sbrk_r() assumes that this global errno variable is used here,
+   * which is different from the errno definition provided by <errno.h>.
+   */
+#undef errno
+  extern int errno;
+
+  /* Heap boundaries from linker script. */
+  extern uint8_t _heap;
+  extern uint8_t _eheap;
+
+  static uint8_t *heap_end = &_heap;
+  uint8_t *prev_heap_end = heap_end;
+
+  if(heap_end + incr > &_eheap) {
+    PRINTF("Out of heap space!\n");
+    errno = ENOMEM;
+    return (caddr_t)-1;
+  }
+
+  heap_end += incr;
+  return (caddr_t)prev_heap_end;
+}
+
+/**
+ * @}
+ * @}
+ */
```

You can just take this patch as is to update your PR (or I can create my own PR to supersede yours if you prefer; just tell me).

@adamdunkels Are you fine with `lib/newlib` in this patch, or do you prefer `core/lib/newlib` or some other location (see @g-oikonomou's summary of the issue in https://github.com/contiki-os/contiki/pull/1088#issuecomment-109484360)?

bthebaudeau:
I've pushed my suggestion here to make things easier:
https://github.com/bthebaudeau/contiki/tree/newlib-syscalls
https://github.com/bthebaudeau/contiki/commit/d86c2329a48338f1ab469347a9d01a0ba372fb7e

AntiCatCH:
@bthebaudeau thank you for your effort.

drandreas:
@bthebaudeau thank you for your effort.

bthebaudeau:
:+1:

No answers to the remaining questions since June 6, and @alignan has triggered the time-out procedure on August 28, so merging. It will still be easy to move `syscalls.c` anywhere else once in `master` if someone is not happy with its current location.

Ashokrlz:
Hi I'm trying to use malloc() with cc2650 but I'm getting this error 

"/usr/bin/../lib/gcc/arm-none-eabi/4.9.3/../../../../arm-none-eabi/lib/armv7-m/libc.a(lib_a-sbrkr.o): In function `_sbrk_r':
sbrkr.c:(.text._sbrk_r+0xc): undefined reference to `_sbrk'"

can anyone help me with this?  

Can i implement the same way as "https://github.com/bthebaudeau/contiki/tree/newlib-syscalls" for cc2650?

Thanks

PR #1090, opened 2015-05-27
Protection against possible buffer overflow

rajithr:
Protection against possible buffer overflow

### DIFF
--- a/core/net/packetbuf.c
+++ b/core/net/packetbuf.c
@@ -141,6 +141,7 @@ packetbuf_copyto(void *to)
     int i;
     char buffer[1000];
     char *bufferptr = buffer;
+    int  bufferlen = 0;
     
     bufferptr[0] = 0;
     for(i = hdrptr; i < PACKETBUF_HDR_SIZE; ++i) {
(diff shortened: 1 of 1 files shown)


aguirrem:
Where does the  < 980  come from?

rajithr:
Related to [1000]

sieben:
Could it be a macro instead of an hard coded value?

joakimeriksson:
What about using sizeof() on the buffer and subtract 10 to get the value? (it seems to add only 6 chars so even 6 might work?).

rajithr:
@joakimeriksson, OK, can do sizeof() on buffer and minus 10, would that be better?

joakimeriksson:
sure it is better since that would work even if someone changed the buffer size.

simonduq:
I suggest using `snprintf` instead of `sprintf` and keep the for loop condition as it was. Break the loop whenever `snprintf` wrote fewer bytes than you asked.

simonduq:
@rajithr can you please address our concern, either as @joakimeriksson suggested or as I did? Ready to merge then.

rajithr:
@simonduq I have replaced the magic number with sizeof, is it OK now?

simonduq:
:+1: if Travis is green (and merging in a month if so, unless somebody has something more to say against this fix)

PR #1091, opened 2015-06-01
Enhanced RPL loop repair

simonduq:
This PR improves RPL loop repair by:
* Sending unicast DIO to the packet source when detecting a loop from ext header;
* Fixing rank along the path (update neighbor rank from ext header when rank error is signaled).

The unicast DIO transmission is delayed to after relaying the packet being forwarded. This is implemented on top of the existing probing mechanism. Testbed in different testbeds, TelosB and NXP JN5168.

### DIFF
--- a/core/net/rpl/rpl-ext-header.c
+++ b/core/net/rpl/rpl-ext-header.c
@@ -73,6 +73,7 @@ rpl_verify_header(int uip_ext_opt_offset)
   uint16_t sender_rank;
   uint8_t sender_closer;
   uip_ds6_route_t *route;
+  rpl_parent_t *sender = NULL;
 
   if(UIP_HBHO_BUF->len != RPL_HOP_BY_HOP_LEN - 8) {
     PRINTF("RPL: Hop-by-hop extension header has wrong size\n");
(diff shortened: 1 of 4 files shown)


laurentderu:
It looks really interesting :)
Still looking at it, but first quick question, why are you relaying on rpl probing and rpl_schedule_probing_immediatly() instead of immediately invoke dio_output() ? If you have deactivated RPL Probing or, using DIS instead of DIO as probe, no DIO will ever be send. Could we instead trigger this unicast DIO using a dedicated ctimer() ?

simonduq:
The problem with `dio_output` is that this would write a DIO to `uipbuf` and drop the packet being forwarded.

Note that the target is specified through a field `unicast_dio_target`, which, when set, forces the probing to be a unicast DIO rather than generated with `RPL_PROBING_SEND_FUNC`.

The other way is what you mention: a new dedicated ctimer. There are pros and cons: I agree this would be cleaner and probing-independent, but on the other hand, that's maybe much additions (new ctimer per instance and new schedule function) for use only in this loop repair case. Cannot really make up my mind, I'm open to both options (leaving as it is, or adding a ctimer).

laurentderu:
We could move probing_timer and unicast_dio_target out of the RPL_PROBING guard and have a dedicated scheduling function and callback for the dio sending. So no new timer and the code would be simpler as handle_probing_timer() don't have to check for the actual action to perform (and probably offset a bit the code increase of a new small function).

simonduq:
Sounds almost good to me, except for using `probing_timer` in two different probing function. If we want to make this feature independent from probing, I'd rather have an extra timer for the sake of clarity.

simonduq:
@laurentderu reworked this PR to make it independent from probing as discussed above (now with an additional ctimer and a `rpl_schedule_unicast_dio_immediately` function)

simonduq:
scheduling for merge in a month

core/net/rpl/rpl-ext-header.c
```diff
@@ -132,10 +133,25 @@ rpl_verify_header(int uip_ext_opt_offset)
 	 instance->current_dag->rank
 	 );
 
+  sender = nbr_table_get_from_lladdr(rpl_parents, packetbuf_addr(PACKETBUF_ADDR_SENDER));
+
+  if(UIP_EXT_HDR_OPT_RPL_BUF->flags & RPL_HDR_OPT_RANK_ERR) {
```
nfi: There should also be a check that sender is not `NULL`.

simonduq: Thanks for the catch, pushed a fix!

simonduq:
merging after timeout

PR #1092, opened 2015-06-01
Remove CC2530 and CC2538 dummy watchdog_stop()

g-oikonomou:
The CC2538 WDT cannot be stopped once it has been started.

The CC2530/1 WDT can be stopped if it is running in timer mode, but it cannot be stopped once it has been started in watchdog mode.

Both chips currently provide "dummy" implementations of `watchdog_stop()`, one does nothing and the other one basically re-maps `_stop()` to `_periodic()`.

This was originally done in order to provide implementations for all prototypes declared in `core/dev/watchdog.h`. In hindsight and as per the discussion in #1088, this is bad practice since, if the build succeeds, the caller will expect that the WDT has in fact been stopped, when in reality it has not.

Since the feature (stopping the WDT) is unsupported by the hardware, this pull removes those dummy implementations. Thus, we will now be able to reliably detect - at build time - attempts at using this unsupported feature.


bthebaudeau:
:+1:

alignan:
:+1:

PR #1093, opened 2015-06-01
Simple implementation of the Bloom filter

ardarico:
In core/lib/, I have placed a simple implementation (bloom.h/.c) of the Bloom filter, based on Contiki's implementation of CRC16. 

This implementation of the Bloom filter was the starting point of the research project called DINAS, described is available in the following paper: 
M. Amoretti, O. Alphand, G. Ferrari, F. Rousseau, A. Duda
DINAS: a DIstributed NAming Service for All-IP Wireless Sensor Networks
IEEE Wireless Communications and Networking Conference (WCNC), Istanbul, Turkey, April 2014.

I am going to push the implementation of DINAS to Contiki Projects. However, as the Bloom filter is a useful tool in general, I think it could be included into Contiki's libs.

### DIFF
--- a/core/lib/bloom.c
+++ b/core/lib/bloom.c
@@ -1,0 +1,129 @@
+/* 
+ * Copyright (c) 2015, Michele Amoretti.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+*/
+
+/** \addtogroup bloom
+ * @{ */
+
+/**
+ * \file
+ *         Implementation of the Bloom filter
+ * \author
+ *         Michele Amoretti <michele.amoretti@unipr.it> 
+ */
+
+#include <limits.h>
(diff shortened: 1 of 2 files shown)


nvt:
I think that we should avoid adding modules in core/lib that are not used anywhere in the main source tree. Furthermore, there are multiple issues with this implementation.

* The code style does not conform with that of Contiki.
* The module is not flexible as to what hash functions can be used -- the hash functions are merely calls to crc16_data() with different parameter values.
* bloom_create returns an object that is allocated on the stack.
* The size parameters (BLOOM_NFUNCS and BLOOM_SIZE) apply to all Bloom filters (had it been possible to actually have more than one Bloom filter.)

For these reasons (mainly the one at the top), I think that this PR should be closed, and that the module would be more suited to be distributed along with the DINAS source code instead of with Contiki.

ardarico:
Hi Nicolas,

thank you for the feedback.

I have a couple of questions:

On Tue, Jun 23, 2015 at 4:55 PM, Nicolas Tsiftes <notifications@github.com>
wrote:

> I think that we should avoid adding modules in core/lib that are not used
> anywhere in the main source tree. Furthermore, there are multiple issues
> with this implementation.
>
>    - The code style does not conform with that of Contiki.
>
> I have carefully checked this point, before submitting the code and the
PR. I would like to know exactly what I missed.



>
>    - The module is not flexible as to what hash functions can be used --
>    the hash functions are merely calls to crc16_data() with different
>    parameter values.
>
> ok


>
>    - bloom_create returns an object that is allocated on the stack.
>
>
any suggestions on how to solve this issue?

>
>    - The size parameters (BLOOM_NFUNCS and BLOOM_SIZE) apply to all Bloom
>    filters (had it been possible to actually have more than one Bloom filter.)
>
>
good point, I will improve this aspect

>
>
> For these reasons (mainly the one at the top), I think that this PR should
> be closed, and that the module would be more suited to be distributed along
> with the DINAS source code instead of with Contiki.
>
ok


> —
> <https://github.com/contiki-os/contiki/pull/1093#issuecomment-114533601>.
>



-- 
Michele Amoretti, Ph.D.

nvt:
Regarding the code style, I'll point out a few things below.

For bloom_create(), you can either make it possible to have the user pass an object as an argument, that bloom_create() then initializes. Another possibility is to have a MEMB block of bloom objects that the function allocates from.

core/lib/bloom.c
```diff
@@ -0,0 +1,129 @@
...
+#include <limits.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <string.h>
+#include "bloom.h"
+#include "lib/crc16.h"
+#define DEBUG DEBUG_PRINT
+#include "net/ip/uip-debug.h"
+
+#define SETBIT(a, n) (a[n/CHAR_BIT] |= (1<<(n%CHAR_BIT)))
```
nvt: Missing whitespace beteen operators and operands. E.g., this line should be written as
<code>#define SETBIT(a, n) (a[n / CHAR_BIT] |= (1 << (n % CHAR_BIT)))</code>

core/lib/bloom.c
```diff
@@ -0,0 +1,129 @@
...
+#include <string.h>
+#include "bloom.h"
+#include "lib/crc16.h"
+#define DEBUG DEBUG_PRINT
+#include "net/ip/uip-debug.h"
+
+#define SETBIT(a, n) (a[n/CHAR_BIT] |= (1<<(n%CHAR_BIT)))  
+#define GETBIT(a, n) (a[n/CHAR_BIT] & (1<<(n%CHAR_BIT)))
+
+const unsigned char oneBits[] = {0,1,1,2,1,2,2,3,1,2,2,3,2,3,3,4};
```
nvt: <code>oneBits</code> looks like Java names. In Contiki, it should typically be written as <code>one_bits</code>. Furthermore, this should probably be declared as <code>static</code> also.

core/lib/bloom.c
```diff
@@ -0,0 +1,129 @@
...
+
+
+/*---------------------------------------------------------------------------*/
+BLOOM
```
nvt: All-caps are typically not used for type names in Contiki. This could be named as <code>bloom_t</code>.

core/lib/bloom.h
```diff
@@ -0,0 +1,152 @@
...
+#define BLOOM_SIZE 40
+#endif /* BLOOM_SIZE_PARAM */
+
+#ifdef BLOOM_NFUNCS_PARAM
+#define BLOOM_NFUNCS BLOOM_NFUNCS_PARAM
+#else /* BLOOM_NFUNCS_PARAM */
+#define BLOOM_NFUNCS 7
+#endif /* BLOOM_NFUNCS_PARAM */
+
+#define A_SIZE (BLOOM_SIZE+CHAR_BIT-1)/CHAR_BIT
```
nvt: Defined macros should preferably have a name that ties them to the module in which they are used; e.g., BLOOM_A_SIZE. It is unclear what "A" refers to in the name without looking further.

PR #1094, opened 2015-06-01
Fix z1-bsl-nopic

OlegHahm:
The Zolertia Z1 bootloader script will work only with Python2 and was broken due to mixing spaces and tabs.


alignan:
:+1:

PR #1096, opened 2015-06-01
Makefile.include: Remove unused MODULESSUBST

bthebaudeau:
The commit bddd96d5c83323fb3f99835149af5ab1058c10bb "Removed all module
makefiles. Instead, all .c files in a module directory are compiled."
made the MODULESSUBST variable useless, but it did not remove it, so do
it now.


adamdunkels:
:+1:

PR #1104, opened 2015-06-08
generalised CCM module into lib

phoenix-frozen:
CCM_STAR was specific to the noncoresec module, and not usable for other things. This pull request generalises it, along the same lines as AES_128.


laurentderu:
The file ccm-star.c should have been moved instead of deleted/added, we are losing the commit history. Also, Travis fails with compilation errors in ccm-star.c

phoenix-frozen:
Major bugs. Closing pull request and reopening after cleanup.

PR #1108, opened 2015-06-10
generalised CCM module into lib

phoenix-frozen:
also introduced a convenience module for LLSEC modules that automatically construct an appropriate IV to pass in

### DIFF
--- a/core/lib/ccm-star.c
+++ b/core/lib/ccm-star.c
@@ -45,30 +45,26 @@
 /* see RFC 3610 */
 #define CCM_STAR_AUTH_FLAGS(Adata, M) ((Adata ? (1u << 6) : 0) | (((M - 2u) >> 1) << 3) | 1u)
 #define CCM_STAR_ENCRYPTION_FLAGS     1
-#define CCM_STAR_NONCE_MAX_IV_LENGTH  13
 
 /*---------------------------------------------------------------------------*/
 static void
-set_nonce(uint8_t *nonce,
+set_nonce(uint8_t *iv,
     uint8_t flags,
-    const uint8_t *iv, uint8_t iv_len,
+    const uint8_t *nonce,
     uint8_t counter)
 {
   /* 1 byte||          8 bytes        ||    4 bytes    || 1 byte  || 2 bytes */
   /* flags || extended_source_address || frame_counter || sec_lvl || counter */
 
-  if(iv_len > CCM_STAR_NONCE_MAX_IV_LENGTH)
-    iv_len = CCM_STAR_NONCE_MAX_IV_LENGTH;
-
-  nonce[0] = flags;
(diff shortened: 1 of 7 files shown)


phoenix-frozen:
Sorry about the repeated open/closes, still getting the hang of GitHub.
However, this now passes all tests, and gives us a generic CCM* library.

laurentderu:
Looks good to me, just one or two minor coding style issues, but someone with more crypto background should have a look at it. @kkrentz ?

kkrentz:
That's a good idea, but I would like to realize it in a different way. The main reason why I opted to leave the CCM* driver exchangeable is that some 802.15.4 transceivers feature a hardware-accelerated CCM* implementation (e.g. CC2520, CC2531, CC2538). I think that my CCM* interface could be implemented by such a hardware-accelerated CCM* implementation, whereas the new CCM interface may be too general. For this reason, I suggest to do it the other way around: We can place a general `ccm-aes-128.c` in lib and change the current software CCM* implementation to use `ccm-aes-128.c` instead.

phoenix-frozen:
I'm not sure I understand the criticism; what you're describing is exactly what I've done (as well as a minor adapter module for llsec/noncoresec).

kkrentz:
Seems like. So am I right that `ccm-star.c` should become more general, but not a fully-fledged CCM implementation as of RFC 3610? (My concern was that a fully-fledged CCM implementation may have to accept messages that are longer than 127 bytes, which is not supported by hardware-accelerated CCM* implementations.) In this case, however, I would suggest to fix the IV length to 13 bytes in order to get rid of the additional iv_length parameter. Additionally, I would suggest the parameter `iv` to be renamed `nonce` so as to stay in line with the terminology of CCM* and CCM.

phoenix-frozen:
That's the intention.
I've made the changes you've suggested, tested, and pushed.

kkrentz:
:+1:

simonduq:
Great to get CCM* independent from packetbuf! I've tested this on two different platforms and am very happy with it.
One suggestion, but this could be in a future PR, would be to add new functions in the CCM* driver to do CTR+MIC, to better exploit H/W co-processors that are able to perform both as a single integrated operation.
:+1:

phoenix-frozen:
I'm actually surprised it wasn't already architected that way, but I'm gonna leave that to a future PR, I think.

kkrentz:
Well, I avoided a general version of CCM* so as to save program memory and I also needed (and still need) the flexibility to only generate a CCM*-MIC.

simonduq:
Note that even the integrated CCM* operation can be used for CTR only or MIC only, if playing a bit with `a` and `m` :)

kkrentz:
Unfortunately, that all costs program memory :( Can we at least avoid the duplicate code in `ccm-star-packetbuf.c`.

PR #1109, opened 2015-06-10
CC26XX - Test IEEE_ADDR_NODE_ID at compile time

bkozak-scanimetrics:
Moves test of IEEE_ADDR_NODE_ID out of preprocessor and into regular conditional.

This allows defining IEEE_ADDR_NODE_ID to a variable (I use this to set the node id statically in applications that build multiple binaries like in examples/ipv6/rpl-udp) ,fits with the style of ieee-addr.c (IEEE_ADDR_CONF_HARDCODED is tested at compile time) and also allows the compiler to check the code inside the conditional.

### DIFF
--- a/cpu/cc26xx/ieee-addr.c
+++ b/cpu/cc26xx/ieee-addr.c
@@ -81,10 +81,10 @@ ieee_addr_cpy_to(uint8_t *dst, uint8_t len)
     }
   }
 
-#if IEEE_ADDR_NODE_ID
-  dst[len - 1] = IEEE_ADDR_NODE_ID & 0xFF;
-  dst[len - 2] = IEEE_ADDR_NODE_ID >> 8;
-#endif
+  if(IEEE_ADDR_NODE_ID){
+    dst[len - 1] = IEEE_ADDR_NODE_ID & 0xFF;
+    dst[len - 2] = IEEE_ADDR_NODE_ID >> 8;
+  }
 }
 /*---------------------------------------------------------------------------*/
 /** @} */
--- a/cpu/cc26xx/ieee-addr.h
+++ b/cpu/cc26xx/ieee-addr.h
@@ -85,6 +85,10 @@
 #endif
 /** @} */
 /*---------------------------------------------------------------------------*/
+#ifndef IEEE_ADDR_NODE_ID
+#define IEEE_ADDR_NODE_ID (0)
+#endif
+/*---------------------------------------------------------------------------*/
 /**
  * \brief Copy the node's IEEE address to a destination memory area
  * \param dst A pointer to the destination area where the IEEE address is to be


g-oikonomou:
I am not 100% certain I understand the motivation behind this pull (other than compile-testing the code inside the the CPP conditional).

You can use the `NODEID` make variable in order to achieve something that sounds exactly what you are trying to do. Does it not satisfy your requirements?

https://github.com/contiki-os/contiki/tree/master/platform/cc2538dk#node-ieeerimeipv6-addresses

bkozak-scanimetrics:
In my application I have something like

In project-conf.h
```C
#ifdef APPLICATION_NODEID
extern const unsigned short app_nodeid;
#define IEEE_ADDR_NODE_ID app_nodeid
#define NODE_ID app_nodeid
#endif
```

In C file for node type 1:
```C
unsigned const short app_nodeid = 0x0001;
```

In C file for node type 2:
```C
unsigned const short app_nodeid = 0x0000;
```

If IEEE_ADDR_NODE_ID is checked by preprocessor then `#if IEEE_ADDR_NODE_ID` fails when IEEE_ADDR_NODE_ID evaluates to a global variable.

g-oikonomou:
What if we forget about the code and we think about what it is you are trying to achieve?

I understand you have nodes of type 1 (eg. `node-1.c`) and type 2 (`node-2.c`). I am guessing you also have a common file `node-common.c` where you need this `extern`.

Are you then trying to set the node's IPv6 address based on what's encountered in `app_nodeid`, which you want to control at source level rather than during build?

Am I on the right track here?

bkozak-scanimetrics:
You've exactly got it.

g-oikonomou:
`node-common.c` being `ieee-addr.c` as far as I can tell.

Why do you need an additional layer of node IDs? Why not do things the other way round, i.e. set `app_nodeid` from `IEEE_ADDR_NODE_ID` or from the final 2 bytes of either the link-layer address or of the IPv6 address (assuming you are using v6 in the first place)?

So, something like:

```
uint16_t app_nodeid = IEEE_ADDR_NODE_ID;
```

```
make NODEID=1 node-1.bin && make NODEID=2 node-2.bin
```

The main idea behind passing `NODEID` make variable is that you can build/deploy images without having to change the sources. Would this not work?

bkozak-scanimetrics:
Well actually it seems that would work just fine.

I have to admit that I was very surprised to see that cpu/cc26xx/ieee-addr.c is re-built every time that make is run (otherwise this, of course, wouldn't work). I assume that this happens precisely because there is a need to support setting NODEID like this.

Also, although I new that NODEID was used on other platforms, I didn't think to look in the makefile to find that IEEE_ADDR_NODE_ID is set from NODEID (until you mentioned it earlier).

So it seems that I've somewhat wasted both your time and mine. I suppose that there may still be some (small) value in this change so I'll let you have the final word on whether I should just close this.

g-oikonomou:
None of my time wasted! Thanks for sharing your effort, as ever.

I'm happy to consider this if we come up with a user requirement which is not already covered by the current code base. Not quite there yet IMHO!

PR #1116, opened 2015-06-15
Add uip_clear_buf() macro and replace all instances of uip_len = 0

laurentderu:
This PR adds a macro uip_clear_buf() that set uip_len to 0 and also uip_ext_len to 0 when in IPv6 mode. This is the result of the discussion in #927.

All instances of uip_len = 0 have been replaced by uip_clear_buf();


simonduq:
:+1:

adamdunkels:
:+1:

PR #1118, opened 2015-06-15
made AES-128.set_key's argument const

phoenix-frozen:
Made AES-128.set_key's argument const, thus helping to avoid compiler warnings.


lebrush:
:+1: makes sense

laurentderu:
:+1:

PR #1119, opened 2015-06-16
Estimated fragments count should take into account 802.15.4 header size

laurentderu:
Since #557, MAC_MAX_PAYLOAD does not include 802.15.4 header size anymore, that means the fragments count estimation is wrong again in sicslowpan.c (it requires too few queuebuf for a given packet).

This PR add the 802.15.4 header length again as committed by the framer.

### DIFF
--- a/core/net/ipv6/sicslowpan.c
+++ b/core/net/ipv6/sicslowpan.c
@@ -1455,7 +1455,7 @@ output(const uip_lladdr_t *localdest)
      * IPv6/HC1/HC06/HC_UDP dispatchs/headers.
      * The following fragments contain only the fragn dispatch.
      */
-    int estimated_fragments = ((int)uip_len) / ((int)MAC_MAX_PAYLOAD - SICSLOWPAN_FRAGN_HDR_LEN) + 1;
+    int estimated_fragments = ((int)uip_len) / (max_payload - SICSLOWPAN_FRAGN_HDR_LEN) + 1;
     int freebuf = queuebuf_numfree() - 1;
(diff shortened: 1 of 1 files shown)


simonduq:
good catch :+1:

simonduq:
scheduling this for merging (unless something happens) in a month

simonduq:
Question, shouldn't we use `max_payload` instead of `MAC_MAX_PAYLOAD - framer_hdrlen`?

laurentderu:
max_payload includes the overhead due to security layer, but it is present only once in a packet. Here we are using MAC_MAX_PAYLOAD - framer_hdrlen - SICSLOWPAN_FRAGN_HDR_LEN instead to have a rough estimate of the number of fragments needed. To take into account the extra bytes needed by the security layer, they should be added to the left side of the division, with uip_len :

<pre>
    int estimated_fragments = ((int)uip_len + NETSTACK_LLSEC.get_overhead()) / ((int)MAC_MAX_PAYLOAD - framer_hdrlen - SICSLOWPAN_FRAGN_HDR_LEN) + 1;
</pre>

simonduq:
Wait, no, this is link-layer security, the header is per-802-15.4 frame, not per IP datagram

laurentderu:
oops, indeed, I was mixing layers :) So your original remark is totally correct, I'll update it.

simonduq:
the fix, with `max_payload`, is now also part of https://github.com/contiki-os/contiki/pull/1258

simonduq:
ping @laurentderu

laurentderu:
Fixed as suggested

simonduq:
I see no reason to keep this bug laying around longer, merging

PR #1120, opened 2015-06-16
Cooja: Clone packet data before modifying it in packet analyser

laurentderu:
THe packet analyser plugin is modifying the original packet data instead of working on it's own copy. This leads to broken cap files when more than one packet analyser is running.


laurentderu:
@fros4943 ping :)

fros4943:
:+1:

PR #1122, opened 2015-06-16
Add CRC to packets send by CoojaMote

laurentderu:
Packets sent from CoojaMote do not have a CRC. CoojaRadio simply send the packet to the RfListener as is. Also the adaptation layer between CC2420 emulator and RfListener drops and recreates the CRC for outgoing and incoming packets.

This PR ensure that all packets have (or keep) their CRC and so are exportable into valid PCAP.

Notes: I had to copy CCITT_CRC from MSPSIM as I did not see a way to cleanly reference it from Cooja, also, other radio driver might need adaptation, so feedback is most welcome.


laurentderu:
@fros4943 ping :)

fros4943:
Great! :+1:

PR #1123, opened 2015-06-16
Check pointer in uip_debug_lladdr_print

laurentderu:
If DEBUG is enabled in uip-ds6-nbr, the code crashes when trying to print a NULL lladdr. A NULL lladdr occurs with INCOMPLETE neighbor state.

### DIFF
--- a/core/net/ip/uip-debug.c
+++ b/core/net/ip/uip-debug.c
@@ -75,6 +75,10 @@ void
 uip_debug_lladdr_print(const uip_lladdr_t *addr)
 {
   unsigned int i;
+  if(addr == NULL) {
+    printf("(NULL LL addr)");
+    return;
+  }
   for(i = 0; i < sizeof(uip_lladdr_t); i++) {
     if(i > 0) {
       PRINTA(":");


core/net/ip/uip-debug.c
```diff
@@ -74,6 +74,10 @@ uip_debug_ipaddr_print(const uip_ipaddr_t *addr)
 void
 uip_debug_lladdr_print(const uip_lladdr_t *addr)
 {
+  if(addr == NULL) {
+    printf("(NULL LL addr)");
+    return;
+  }
   unsigned int i;
```
nvt: The variable definition should come before the added code, otherwise some older compilers (for platforms using the IPv4 stack) will fail here.

laurentderu:
Indeed I overlooked that, it's fixed now.

nvt:
:+1:

core/net/ip/uip-debug.c
```diff
@@ -75,6 +75,10 @@ void
 uip_debug_lladdr_print(const uip_lladdr_t *addr)
 {
   unsigned int i;
+  if(addr == NULL) {
+    printf("(NULL LL addr)");
```
oliverschmidt: Is there a specific reason not to use `PRINTA` like in the rest of the file?

laurentderu: That happens when you blindly copy&paste the code from the function uip_debug_ipaddr_print() above :) I'll fix both and also move the check after the local declaration in uip_debug_ipaddr_print()

PR #1125, opened 2015-06-18
Fixed TX/RX pin define in board.h of the cc2538dk

davidrojas:
SPI0_RX_PIN and SPI0_TX_PIN in board.h of the cc2548dk platform are swapped.

### DIFF
--- a/platform/cc2538dk/dev/board.h
+++ b/platform/cc2538dk/dev/board.h
@@ -199,11 +199,11 @@
 /** TX port SPI0 (master mode: MOSI) */
 #define SPI0_TX_PORT            GPIO_A_NUM
 /** TX pin SPI0 */
-#define SPI0_TX_PIN             4
+#define SPI0_TX_PIN             5
 /** RX port SPI0 (master mode: MISO */
 #define SPI0_RX_PORT            GPIO_A_NUM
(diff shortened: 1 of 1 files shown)


g-oikonomou:
In the SmartRF06EB User Guide I am seeing MOSI on PA4 and MISO on PA5, which at a first glance matches what we have in `board.h`. You seem to be inverting that. Is there a documentation error?

davidrojas:
We have tested in a custom board using the CC2538. If you go to the page 8 of the datasheet: http://www.ti.com/lit/ds/swrs096d/swrs096d.pdf
There it is PA5 = SSI_TX and PA4 = SSI_RX

g-oikonomou:
Those datasheet values are referring to the mappings used by the bootloader, in order to allow programming over SSI.

As far as the SmartRF06EB goes, MISO and MOSI (e.g. for the LCD and the SDCard reader) are connected to the pins defined in our current `board.h`.

davidrojas:
Yes, you are right, the CC2538 has a mux matrix for the serial peripherals, they can be routed to any gpio pin, it is indeed routed that way in the CC2538dk, although it feels strange using the same pins as in the SSI bootloader but swapped.

PR #1126, opened 2015-06-18
Change all printf into PRINTA in uip-debug.c and remove compilation warnings

laurentderu:
Change all printf into PRINTA in uip-debug.c; remove compilation warning (an array can never be a NULL pointer) and move addr check after local variables definition in uip_debug_ipaddr_print() for older compiler.


nvt:
:+1:

PR #1131, opened 2015-06-19
Minor improvements on native platform

zehortigoza:


platform/native/contiki-main.c
```diff
@@ -242,6 +242,8 @@ main(int argc, char **argv)
 
   serial_line_init();
 
+  process_start(&sensors_process, NULL);
```
vsaw: I would not start this for any native target. Instead the projects can include the sensors based on their preferences.

zehortigoza: Without this, a call to sensors_changed() would not cause any event been generated and send to another process, I will add this to the commit message.

vsaw: I understand what this does. I just don't think everybody wants to use this sensors. So instead of assuming the sensors are always on, applications that require sensors should start them manually.

zehortigoza: I see your point but several platforms do it (cc2530dk, cc2538dk, cooja, econotag, remote, seedeye, sensinode, sky, srf06-cc25xx, wismote and z1), so the default is let applications start the sensors_process? If yes, those platforms should be fixed to avoid mistakes like this.

PR #1132, opened 2015-06-19
LRU deletion strategy for rime routes

tpetry:
Actually contiki removes the first added (FIFO) rime route if any module tries to add a new route to a full routing table. For large networks the route discovery module (used by the mesh module) saves many newly learned routes. But this amount of newly learned routes may not be as relevant as already existing routes the node is aware of.
With large sensor networks my most important routes have been constantly removed because of many learned "useless" route discovery routes, triggering route discovery indefinitely.

The suggested change will change the route deletion strategy from FIFO to LRU so any actively used routes which are refreshed by ``route_refresh()`` are not deleted "by accident". If neither route is refreshed by the application the LRU implementation is completely backwards compatible to the contiki 2.7 release because of:
* every routing table entry has the same ``time=0`` value (all are same "old")
* routing table entries added first (entries at the end of the list) have a higher precedence to be removed if their ``time`` value is same.

### DIFF
--- a/core/net/rime/route.c
+++ b/core/net/rime/route.c
@@ -121,7 +121,7 @@ int
 route_add(const linkaddr_t *dest, const linkaddr_t *nexthop,
 	  uint8_t cost, uint8_t seqno)
 {
-  struct route_entry *e;
+  struct route_entry *e, *oldest = NULL;
 
   /* Avoid inserting duplicate entries. */
   e = route_lookup(dest);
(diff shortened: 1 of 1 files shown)


simonduq:
Seems reasonable to me, indeed the current `route_refresh()` doesn't seem to prevent deletion.
:+1: and scheduling for merging in a month

nvt:
:+1:

PR #1133, opened 2015-06-19
put gcc specific stuff into seperate .h

bkozak-scanimetrics:
I think that it does not make sense to define functionality specific to compilers in platform headers or as needed in C files. I'd like to propose that we have specific header files for compiler specific functionality and have created such a header for gcc.

In addition to the above, this will also have the tangible effect of defining CC_CONF_INLINE on platforms that build with gcc and which do not already define it (such as CC2538 and CC2650) without having to add more to the platform specific header files.

### DIFF
--- a/core/sys/cc-gcc.h
+++ b/core/sys/cc-gcc.h
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2015, Scanimetrics - http://www.scanimetrics.com
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+/*---------------------------------------------------------------------------*/
+#ifndef _CC_GCC_H_
+#define _CC_GCC_H_
+#ifdef __GNUC__
+
+#ifndef CC_CONF_INLINE
+/* use __inline__ in case "inline" is not available for any reason */
+#define CC_CONF_INLINE __inline__
+#endif
+
+#define CC_CONF_ALIGN(n) __attribute__((__aligned__(n)))
+
+#endif /* __GNUC__ */
+#endif /* _CC_GCC_H_ */
--- a/core/sys/cc.h
+++ b/core/sys/cc.h
@@ -46,6 +46,7 @@
 #define CC_H_
 
 #include "contiki-conf.h"
+#include "sys/cc-gcc.h"
 
 /**
  * Configure if the C compiler supports the "register" keyword for
@@ -108,6 +109,10 @@
 #define CC_INLINE CC_CONF_INLINE
 #else /* CC_CONF_INLINE */
 #define CC_INLINE
+#endif /* CC_CONF_INLINE */
+
+#ifdef CC_CONF_ALIGN
+#define CC_ALIGN(n) CC_CONF_ALIGN(n)
 #endif /* CC_CONF_INLINE */
 
 /**
--- a/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
+++ b/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
@@ -75,12 +75,6 @@
 #include <stdio.h>
 #include <stdbool.h>
 /*---------------------------------------------------------------------------*/
-#ifdef __GNUC__
-#define CC_ALIGN_ATTR(n) __attribute__ ((aligned(n)))
-#else
-#define CC_ALIGN_ATTR(n)
-#endif
-/*---------------------------------------------------------------------------*/
 #define DEBUG 0
 #if DEBUG
 #define PRINTF(...) printf(__VA_ARGS__)
@@ -202,7 +196,7 @@ const output_config_t *tx_power_current = &output_power[0];
  * A buffer to send a CMD_IEEE_RX and to subsequently monitor its status
  * Do not use this buffer for any commands other than CMD_IEEE_RX
  */
-static uint8_t cmd_ieee_rx_buf[RF_CMD_BUFFER_SIZE] CC_ALIGN_ATTR(4);
+static uint8_t cmd_ieee_rx_buf[RF_CMD_BUFFER_SIZE] CC_ALIGN(4);
 /*---------------------------------------------------------------------------*/
 #define DATA_ENTRY_LENSZ_NONE 0
 #define DATA_ENTRY_LENSZ_BYTE 1
@@ -210,10 +204,10 @@ static uint8_t cmd_ieee_rx_buf[RF_CMD_BUFFER_SIZE] CC_ALIGN_ATTR(4);
 
 #define RX_BUF_SIZE 140
 /* Four receive buffers entries with room for 1 IEEE802.15.4 frame in each */
-static uint8_t rx_buf_0[RX_BUF_SIZE] CC_ALIGN_ATTR(4);
-static uint8_t rx_buf_1[RX_BUF_SIZE] CC_ALIGN_ATTR(4);
-static uint8_t rx_buf_2[RX_BUF_SIZE] CC_ALIGN_ATTR(4);
-static uint8_t rx_buf_3[RX_BUF_SIZE] CC_ALIGN_ATTR(4);
+static uint8_t rx_buf_0[RX_BUF_SIZE] CC_ALIGN(4);
+static uint8_t rx_buf_1[RX_BUF_SIZE] CC_ALIGN(4);
+static uint8_t rx_buf_2[RX_BUF_SIZE] CC_ALIGN(4);
+static uint8_t rx_buf_3[RX_BUF_SIZE] CC_ALIGN(4);
 
 /* The RX Data Queue */
 static dataQueue_t rx_data_queue = { 0 };
@@ -225,7 +219,7 @@ volatile static uint8_t *rx_read_entry;
 #define TX_BUF_PAYLOAD_LEN 180
 #define TX_BUF_HDR_LEN       2
 
-static uint8_t tx_buf[TX_BUF_HDR_LEN + TX_BUF_PAYLOAD_LEN] CC_ALIGN_ATTR(4);
+static uint8_t tx_buf[TX_BUF_HDR_LEN + TX_BUF_PAYLOAD_LEN] CC_ALIGN(4);
 /*---------------------------------------------------------------------------*/
 /* Overrides for IEEE 802.15.4, differential mode */
 static uint32_t ieee_overrides[] = {
--- a/cpu/cc26xx-cc13xx/rf-core/prop-mode.c
+++ b/cpu/cc26xx-cc13xx/rf-core/prop-mode.c
@@ -80,12 +80,6 @@
 #include <stdio.h>
 #include <stdbool.h>
 /*---------------------------------------------------------------------------*/
-#ifdef __GNUC__
-#define CC_ALIGN_ATTR(n) __attribute__ ((aligned(n)))
-#else
-#define CC_ALIGN_ATTR(n)
-#endif
-/*---------------------------------------------------------------------------*/
 #define DEBUG 0
 #if DEBUG
 #define PRINTF(...) printf(__VA_ARGS__)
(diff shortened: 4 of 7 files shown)


bkozak-scanimetrics:
This Travis failure seems to be because doxygen does not specify any directories in its preprocessor include path. When CC_ALIGN_ATTR is moved to a header file doxygen no longer recognizes certain variable declarations (which use CC_ALIGN_ATTR) and produces warnings about undefined variables.

bthebaudeau:
> I think that it does not make sense to define functionality specific to compilers in platform headers or as needed in C files.

I totally agree. See my comment here:
https://github.com/contiki-os/contiki/pull/966#discussion_r30322788

> I'd like to propose that we have specific header files for compiler specific functionality and have created such a header for gcc.

I don't know if it's preferable to group these definitions by compiler or by feature as per my suggestion, but we definitely need something like that.

We need such macros at least for inline, alignments, packed structs, weak stuff, aliases, and sections. Some of these macros (typically those affecting only performance, but not behavior, e.g. for inline) can default to an empty definition if this is not dangerous for the resulting behavior, but the other ones (e.g. for mandatory alignments) should have no default definition in order to trigger an error if they are used. And some macros need several variants, e.g. an alignment can be requested by the user of a macro to be either compulsory or a performance hint, in which case there should be several sets of macros. Alignment macros should also make it clear in their name or in their documentation whether the passed value is a number of bytes or a bit-shift to generate a power of 2.

Then, the toolchains should be auto-detected as you did. Keeping configurations for these macros does not make much sense if all the toolchains used by Contiki can be detected in such a way. Otherwise, and until all Contiki code has been converted, we need to keep them.

These are comments for long-term changes. For this PR, see my comments in the diff view.

> This Travis failure seems to be because doxygen does not specify any directories in its preprocessor include path. When CC_ALIGN_ATTR is moved to a header file doxygen no longer recognizes certain variable declarations (which use CC_ALIGN_ATTR) and produces warnings about undefined variables.

Please retry with `sys/` (see my comments in the diff view).

core/sys/cc-gcc.h
```diff
@@ -0,0 +1,43 @@
...
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+/*---------------------------------------------------------------------------*/
+#ifndef _CC_GCC_H_
+#define _CC_GCC_H_
+#ifdef __GNUC__
+
+#ifndef CC_CONF_INLINE
+/* use _inline in case "inline" is not available for any reason */
+#define CC_CONF_INLINE __inline
```
bthebaudeau: Even `__inline__`.

We should also distinguish soft and hard inline requests. This one would be soft. But that's OK for this PR.

core/sys/cc-gcc.h
```diff
@@ -0,0 +1,43 @@
...
+#endif
+
+#define CC_CONF_ALIGN_ATTR(n) __attribute__ ((aligned(n)))
```
bthebaudeau: - Why `_ATTR` (except because it is already used by `cc26xx-rf.c`)?
 - `__aligned__`.

core/sys/cc-gcc.h
```diff
@@ -0,0 +1,43 @@
...
+
+#endif //__GNUC__
+#endif //_CC_GCC_H_
```
bthebaudeau: Missing newline at EoF.

core/sys/cc.h
```diff
@@ -46,6 +46,7 @@
 #define CC_H_
 
 #include "contiki-conf.h"
+#include "cc-gcc.h"
```
bthebaudeau: `#include "sys/cc-gcc.h"`.

core/sys/cc.h
```diff
@@ -110,6 +111,12 @@
 #define CC_INLINE
 #endif /* CC_CONF_INLINE */
 
+#ifdef CC_CONF_ALIGN_ATTR
+#define CC_ALIGN_ATTR CC_CONF_ALIGN_ATTR
+#else /* CC_CONF_INLINE */
+#define CC_ALIGN_ATTR
```
bthebaudeau: If the requested alignment is a requirement and not a simple performance hint, this will silently generate a runtime bug.

core/sys/cc-gcc.h
bkozak-scanimetrics: @bthebaudeau 

I don't know of any reason to prefer __inline__ over __inline in gcc but I've taken your word for it and made the change.

I agree that it would be useful to distinguish between soft and hard inline in the future.

core/sys/cc-gcc.h
bkozak-scanimetrics: @bthebaudeau 

I've kept ATTR only because of cc26xx-rf.c and haven't given the name much thought.

What would your prefer?
CC_ALIGN() perhaps?

bthebaudeau:
That was not enough. The doxygen test still fails.

bthebaudeau:
Try with `#define CC_ALIGN_ATTR(n)` (the parameter is missing in the fallback definitions).

bkozak-scanimetrics:
@bthebaudeau 

> I don't know if it's preferable to group these definitions by compiler or by feature as per my suggestion, but we definitely need something like that.

I personally prefer to group by compiler (easier to search for a particular definition IMO) but I can see it going either way.

> We need such macros at least for inline, alignments, packed structs, weak stuff, aliases ... 

> These are comments for long-term changes. For this PR, see my comments in the diff view.

I agree with all of this.

> Please retry with sys/ (see my comments in the diff view).

This still fails when I run the test on my local machine. I've kept the change because it seems to be more in keeping with the rest of Contiki (and maybe Travis disagrees with my machine).

bthebaudeau:
> > Please retry with sys/ (see my comments in the diff view).

> This still fails when I run the test on my local machine. I've kept the change because it seems to be more in keeping with the rest of Contiki (and maybe Travis disagrees with my machine).

Yes, see my previous comment. I don't think that doxygen has been configured to use `__GNUC__` in Contiki, so the broken fallback definitions apply for it.

bkozak-scanimetrics:
@bthebaudeau

> That was not enough. The doxygen test still fails.

> Try with #define CC_ALIGN_ATTR(n) (the parameter is missing in the fallback definitions).

While debugging the Travis failure earlier, I already tried this and found that it didn't work. So far the only way I have found to fix this is to edit docs/Doxyfile as:
```diff
--- a/doc/Doxyfile
+++ b/doc/Doxyfile
@@ -1932,7 +1932,7 @@ SEARCH_INCLUDES        = YES
 # preprocessor.
 # This tag requires that the tag SEARCH_INCLUDES is set to YES.
 
-INCLUDE_PATH           =
+INCLUDE_PATH           = ../core
```

bkozak-scanimetrics:
>Yes, see my previous comment. I don't think that doxygen has been configured to use __GNUC__ in Contiki, so the broken fallback definitions apply for it.

You're right about that much, doxygen doesn't seem to have __GNUC__ defined. The fallback definition is necessary for this purpose (although, strangely, doxygen doesn't seem to care if the fallback takes a parameter or not).

bthebaudeau:
> While debugging the Travis failure earlier, I already tried this and found that it didn't work. So far the only way I have found to fix this is to edit docs/Doxyfile as:
[...]

This makes sense, but Travis should also fail because of all the headers files in there (maybe it's the case for some of the other warnings). Travis is just complaining about 5 new doxygen warnings, but there were already tons of other warnings, so failing for 5 new warnings is a strange rule. @g-oikonomou What are the doxygen rules supposed to be in Contiki for such issues? Should the doxygen configuration be fixed as suggested by @bkozak-scanimetrics, or should these new warnings just be ignored (which would mean accepting a PR despite a failure of Travis)?

bkozak-scanimetrics:
@bthebaudeau 

> This makes sense, but Travis should also fail because of all the headers files in there (maybe it's the case for some of the other warnings

Yes, you're quite right, if we edit the Doxyfile we actually lose some warnings.

> Travis is just complaining about 5 new doxygen warnings, but there were already tons of other warnings, so failing for 5 new warnings is a strange rule. 

It even fails in the case of reduced warnings. It seems to expect that the exact number of warnings stays the same as the value in regression-tests/00-doxygen/doxyerrors.cnt

Alternatively, the number in doxyerrors.cnt could just be increased.

core/sys/cc.h
bkozak-scanimetrics: @bthebaudeau 

I agree with you here. I've only kept it this way because that's how it was in cc26xx-rf.c and I wanted to avoid changing runtime behaviour for this PR. Note also that in the case of cc26xx-rf.c it really would be a bug to not align any of these variables.

Now that you've mentioned this specifically, however, I think that I will simply git rid of the fallback definition so that this generates an error. It really just doesn't make sense to do otherwise.

bthebaudeau:
> It even fails in the case of reduced warnings.

Yes, but suggesting to adjust `regression-tests/00-doxygen/doxyerrors.cnt` in the PR.

> It seems to expect that the exact number of warnings stays the same as the value in regression-tests/00-doxygen/doxyerrors.cnt

> Alternatively, the number in doxyerrors.cnt could just be increased.

That would work, but that would not be satisfactory. Fixing the doxygen configuration (which should ideally be done in another PR if we decide to do so) is by far better.

`regression-tests/00-doxygen/Makefile` should probably be reworked. For example, it could automatically generate a temporary `doxyerrors.cnt` from the master branch before running doxygen again on the PR branch.

bkozak-scanimetrics:
> Fixing the doxygen configuration (which should ideally be done in another PR if we decide to do so) is by far better.

Completely agree

g-oikonomou:
The idea with the `doxyerrors.cnt` was so that we can test documentation builds but without first having to fix all existing warnings.

If a pull actually reduces the warning count (either intentionally or as a side-effect), then the `.cnt` should be updated to reflect the new lower value. But if a PR adds new warnings, then I personally _strongly_ oppose increasing the count simply so that we can accept the pull. If we start doing so we set a bad precedent and the entire doxygen testing will become moot. I still dream of the day when we will have 0 warnings and we'll be able to remove the `.cnt` test altogether.

To get back on track, these exact warnings are the reason why I started working on #971 in the first place. `#define CC_ALIGN_ATTR(n)` should work, I am not sure why it does not in this instance and I'm not quite getting why changing `INCLUDE_PATH` actually fixes it.

bkozak-scanimetrics:
> #define CC_ALIGN_ATTR(n) should work, I am not sure why it does not in this instance and I'm not quite getting why changing INCLUDE_PATH actually fixes it.

I'm no expert with doxygen but from what I've read at [the doxygen manual](http://www.stack.nl/~dimitri/doxygen/manual/config.html#cfg_include_path) I wouldn't understand how this could work without adding to INCLUDE_PATH.

My understanding is that the doxygen preprocessor will not be able to find "core/sys/cc.h" to expand the CC_ALIGN_ATTR macro unless core is added to the include path.

However, after reading through #971 I think that maybe the best solution is just to add CC_ALIGN to PREDEFINED in the Doxyfile. Adding to INCLUDE_PATH is not even a possible solution (on its own) now that I've removed the fallback definition of CC_ALIGN (see [above](https://github.com/contiki-os/contiki/pull/1133#issuecomment-113620479)).

I've tested this and it gets rid of the warnings.

g-oikonomou:
Yeah the situation with includes is a bit of a train wreck at the moment. I experimented with adding `../core` to `INCLUDE_PATH` as you suggest and by putting the fallback define back in place, but this seems to introduce new warnings in `uip.h`. Those two:

```
/home/travis/build/g-oikonomou/contiki/core/net/ip/uip.h:1039: warning: argument 'dest' of command @param is not found in the argument list of uip_ip4addr_cmp(addr1, addr2)
/home/travis/build/g-oikonomou/contiki/core/net/ip/uip.h:1039: warning: argument 'src' of command @param is not found in the argument list of uip_ip4addr_cmp(addr1, addr2)
```
These happen because this entire block gets removed by the doxygen preprocessor:
```
#ifndef uip_ipaddr_copy
#define uip_ipaddr_copy(dest, src) (*(dest) = *(src))
#endif
#ifndef uip_ip4addr_copy
#define uip_ip4addr_copy(dest, src) (*((uip_ip4addr_t *)dest) = *((uip_ip4addr_t *)src))
#endif
#ifndef uip_ip6addr_copy
#define uip_ip6addr_copy(dest, src) (*((uip_ip6addr_t *)dest) = *((uip_ip6addr_t *)src))
#endif
```
... and doxygen tries to apply the respective documentation block to `uip_ip4addr_cmp`, which is what follows immediately after the removed block. I've no clue why the block is getting removed though.

There is also this one:
```
/home/travis/build/g-oikonomou/contiki/doc/sicslowpan-doc.txt:592: warning: explicit link request to 'sicslowpan_buf' could not be resolved
```
which is not appearing when I build locally due to having a different version of doxygen here.

I'm not quite sure why you removed the fallback for `CC_ALIGN`. Did I miss something in the discussion?

Here's the build for my branch:
https://travis-ci.org/g-oikonomou/contiki/jobs/67607488

In terms of adding `CC_ALIGN` to `PREDEFINED`, I wouldn't be opposed

bthebaudeau:
#1135

I've experimented adding all the paths of header files to `INCLUDE_PATH`: this should definitely not be done because there are some `#include` within doxygen blocks, so it makes doxygen complain about group breakage.

Giving all attribute-like macros empty definitions for doxygen is the way to go. Using `:=` prevents the code from overriding these definitions.

@g-oikonomou:
> I'm not quite sure why you removed the fallback for `CC_ALIGN`. Did I miss something in the discussion?

Yes: click the last "Show outdated diff".

bthebaudeau:
@bkozak-scanimetrics
Now that #1135 has been merged, you can:
 - rebase,
 - remove the space following `__attribute__` (this space is unneeded, and this practice prevents the default definition in `doc/Doxyfile` from working, even if this does not apply here because `__attribute__` would not be expanded in that case),
 - use `__aligned__` for `CC_CONF_ALIGN()` in order to avoid issues if `aligned` has been defined to something,
 - fix the missing macro parameter (i.e. write `#define CC_ALIGN(n) CC_CONF_ALIGN(n)`) in `core/sys/cc.h`,
 - replace `CC_ALIGN_ATTR(n)` with `CC_ALIGN(n)` in `doc/Doxyfile`,
 - squash your commits.

bthebaudeau:
:+1:

core/sys/cc-gcc.h
```diff
@@ -0,0 +1,43 @@
...
+
+#ifndef CC_CONF_INLINE
+/* use __inline__ in case "inline" is not available for any reason */
+#define CC_CONF_INLINE __inline__
+#endif
+
+#define CC_CONF_ALIGN(n) __attribute__((__aligned__(n)))
+
+#endif //__GNUC__
+#endif //_CC_GCC_H_
```
simonduq: for the two lines above, please use `/* */` comments

bthebaudeau: Right, and GitHub says that a rebase is needed too now.

It would be great if the merging rules were relaxed for such things. IMO, these are typical trivial fixes that should be part of a manual merge commit or of some fixup PR edit or of a new PR made by a maintainer because they are not worth bothering the original PR author, especially after several months of inactivity. What do you think?

Also, we are still waiting for an upvote from the core maintainers for this PR. There have been nearly 3 months of inactivity on this PR, so we should merge this PR in one month according to the new PR policy (after fixup of the above). We should probably not reset the PR counter to 2 months just for the little change requested here.

bkozak-scanimetrics:
Re-base done and comments fixed.

core/sys/cc-gcc.h
simonduq: Yes sure, the counter needs no reset. If you still want merge just add a comment that you will merge in a month if nobody speaks against this PR (needs to be made explicit, then 1 month wait time). I also agree on relaxing things a bit, would have been ok to see a merge as is, but when the PR author is willing to help that's also welcome :)

simonduq: Also, please add a "timeout" tag when doing so (makes it easy to spot pull request with scheduled action within a month)

simonduq:
@bthebaudeau, not sure if you get notifications for comments on outdated diffs, pinging you just in case.

bthebaudeau:
@bkozak-scanimetrics :+1:

I approve this PR. If nobody disapproves within one month, I will merge it.

@simonduq Yes, notifications work for comments on outdated diffs too. Thanks.

simonduq:
@bthebaudeau still willing to merge this?

bthebaudeau:
@simonduq Yes. :+1:

PR #1135, opened 2015-06-20
doxygen: Fix some warnings

bthebaudeau:


bthebaudeau:
Maybe the expected warning count will need to be adjusted because I do not get the same number as Travis for the master branch locally.

g-oikonomou:
Defining `DOXYGEN_SHOULD_SKIP_THIS` (which is a good thing), makes 20 new warnings appear. Thus, this pull fixes 76, but adds 20. I've opened a pull on top of this PR to explicitly fix those 20.

g-oikonomou:
In addition to fixing explicit doxygen warnings, this pull also fixes the situation below (documentation of `cc2538/clock.c`). See those `< GPT[3:0] clocks - active mode` artifacts that are not meant to be there.

This is a good improvement - thanks :+1: 

```
Arch-specific implementation of clock_init for the cc2538.

Initialize the clock library.

We initialise the SysTick to fire 128 interrupts per second, giving us a value of 128 for CLOCK_SECOND

We also initialise GPT0:Timer A, which is used by clock_delay_usec(). We use 16-bit range (individual), count-down, one-shot, no interrupts. The prescaler is computed according to the system clock in order to get 1 tick per usec.

< GPT[3:0] clocks - active mode

< GPT0 clock enable, CPU running

< GPTIMER0

< GPTM control

< GPTIMER0

< GPTM configuration

< GPTIMER0

< GPTM Timer A mode

< GPTIMER0

< GPTM Timer A prescale

Definition at line 93 of file clock.c.

References GPT_0_BASE, GPTIMER_CFG, GPTIMER_CTL, GPTIMER_TAMR, GPTIMER_TAPR, SYS_CTRL_RCGCGPT, and SYS_CTRL_RCGCGPT_GPT0.
```

bthebaudeau:
Yes, these artifacts were caused by the doxygen comments expanded through macro expansion, because doxygen does not handle this like the C preprocessor. This PR limits macro expansion to the macros predefined in Doxyfile. Keeping the other macro expansions did not make sense because all conditional definitions and inclusions are broken in the Contiki doxygen context.

PR #1140, opened 2015-06-22
CC26xx fix corrupt frame propagation

bkozak-scanimetrics:
I've found that, when using RIME with CC26XX, I occasionally see corrupted packets make it up to the application layer. I can make this worse by putting the nodes in a noisy environment.

With some debugging I was able to find that the radio module was detecting CRC errors in packets but that the driver was allowing them to propagate up to the application. This PR fixes this, making the driver's behaviour more similar to that of the cc2538.

The use of checksums at higher layers may explain why this problem is not seen with IPv6.

In addition, I believe that there is a race condition between the radio co-processor and the radio driver which can occur when DATA_ENTRY_STATUS_PENDING is set before the driver is completely finished with the packet (I wasn't able to find a clear statement in the CC26xx data sheet that this would be the case however). I've also fixed this.

After making these changes I haven't yet seen the same sort of problem again.

### DIFF
--- a/cpu/cc26xx/dev/cc26xx-rf.c
+++ b/cpu/cc26xx/dev/cc26xx-rf.c
@@ -1309,25 +1309,32 @@ static int
 read_frame(void *buf, unsigned short buf_len)
 {
   int len = 0;
+  uint8_t rx_status_bf;
 
   if(GET_FIELD_V(rx_read_entry, dataEntry, status) == DATA_ENTRY_STATUS_FINISHED) {
-    /* Set status to 0 "Pending" in element */
-    GET_FIELD_V(rx_read_entry, dataEntry, status) = DATA_ENTRY_STATUS_PENDING;
 
     if(rx_read_entry[8] > 0) {
-      memcpy(buf, (char *)&rx_read_entry[9], buf_len);
-
-      /* Remove the footer */
       len = MIN(buf_len, rx_read_entry[8] - 4);
+      rx_status_bf = rx_read_entry[9 + len + 3];
 
-      int rssi = (int8_t)rx_read_entry[9 + len + 2];
+      if(rx_status_bf & 0x80) {
+        len = 0;
+        RIMESTATS_ADD(badcrc);
+        PRINTF("RF: Bad CRC\n");
+      } else {
+        int rssi = (int8_t)rx_read_entry[9 + len + 2];
 
-      packetbuf_set_attr(PACKETBUF_ATTR_RSSI, rssi);
-      RIMESTATS_ADD(llrx);
+        memcpy(buf, (char *)&rx_read_entry[9], buf_len);
 
+        packetbuf_set_attr(PACKETBUF_ATTR_RSSI, rssi);
+        RIMESTATS_ADD(llrx);
+      }
       /* Clear the length byte */
       rx_read_entry[8] = 0;
     }
+
+    /* Set status to 0 "Pending" in element */
+    GET_FIELD_V(rx_read_entry, dataEntry, status) = DATA_ENTRY_STATUS_PENDING;
 
     /* Move read entry pointer to next entry */
     rx_read_entry = GET_FIELD_V(rx_read_entry, dataEntry, pNextEntry);


g-oikonomou:
Billy, thanks.

While I'm reviewing this, please have a look at https://github.com/contiki-os/contiki/wiki/Code-Contributions#code-formatting

This code needs an uncrustify pass. Also, please bear in mind that we_prefer_this_naming overThisNaming.

Cheers

bkozak-scanimetrics:
@g-oikonomou 

Sorry to bother you but does this look aright to you now?

Thanks

bkozak-scanimetrics:
I have now uploaded a [repository to my github](https://github.com/bkozak-scanimetrics/RIME-Throughput-Test) which clearly demonstrates the issue I am describing. This is a simple Contiki application which streams packets over RIME. I have found it to be invaluable in testing the CC26XX platform.

If you run the test as per the instructions in the README of this repository you will find that packet errors are detected on ContikiOS/master but that they go away after making the changes in this PR.

I encourage anyone who is interested in the cc26xx platform to give this a try so that we can test that this PR is doing what it is supposed to do.

Thanks

bkozak-scanimetrics:
I've found some more problems here.
1. read_frame should not be returning non-zero if buf_len < len
2. in my code here I am indexing to get the status bitfield (**rx_status_bf**) using **len** which may be truncated to **buf_len** (and will therefore be wrong in this case).
3. We are probably memcpying unnecessary bytes
4. In my opinion this code should be cleaned up to be made more readable

@g-oikonomou how do you feel about rolling all of these changes into this PR versus creating a new PR for the additional changes (point 2 above I think obviously needs to be fixed before this code can be merged however as the current implementation of the CRC check is just wrong).

g-oikonomou:
Billy hi, swamped with work at the moment, but here's a thought: How about simply setting `bAutoFlushCrc` to 1, so that CRC errors are auto-dropped by the CPE?

bkozak-scanimetrics:
Using `bAutoFlushCrc` seems like a good idea. With `bAutoFlushCrc`, however, I think we should also enable RX_NOK (maybe optionally based on some preprocessor macro) so that we can still detect CRC errors for debugging purposes.

If you don't have any more ideas I'll start work on this when I have some more time.

Thanks

g-oikonomou:
I'm not hugely bothered about putting RX_NOK in the main repo: I am guessing it would only be a very small number of users who might actually use it, and they would most likely all know what to do to enable it by themselves.

In terms of changing `bAutoFlushCrc`, feel free to isolate it in its own pull and I'll be happy to merge it immediately.

Would anything else proposed in this pull be useful if we changed `bAutoFlushCrc`? If not, feel free to just re-use this pull.

bkozak-scanimetrics:
> I'm not hugely bothered about putting RX_NOK in the main repo: I am guessing it would only be a very small number of users who might actually use it, and they would most likely all know what to do to enable it by themselves.

I don't really see any downsides to enabling RX_NOK behind a macro like CC26XX_RF_DEBUG or even just DEBUG (except maybe that this source file is already quite long).  But it's certainly your call.

> Would anything else proposed in this pull be useful if we changed `bAutoFlushCrc`? If not, feel free to just re-use this pull.

I think that there may be a possible race condition with setting the `dataEntry` status back to pending before we are done using the `dataEntry`. I haven't found any clear information in the TRM that says exactly how we signal to radio CPU that we are done with the `dataEntry` but I would think that setting status to pending would be how we do this.

g-oikonomou:
> I don't really see any downsides to enabling `RX_NOK`...

OK, if you feel strongly about it I certainly won't object. In fact I used it myself internally for a little while for debugging purposes too.

> I think that there may be a possible race condition...

Yes, I remember you mention that before and you are right. The data entry should only be released after we are done with it. We want your proposed change. Given how our queue is 4 entries long I don't think this would manifest itself under normal day-to-day scenarios, but under your speed tests it might certainly come up. The TRM is quite clear in that the CPE will only touch the entry if its `status` is `DATA_ENTRY_STATUS_PENDING`.

I'll leave the pull breakdown to you, feel free to use as many as you think is more suitable. Let me just re-iterate that `bAutoFlushCrc` I'll merge immediately, whereas the rest I'll need to look at.

Thanks a lot for your efforts!

g-oikonomou:
Is this obsolete now after opening #1173 and #1174?

bkozak-scanimetrics:
I believe it is.

PR #1141, opened 2015-06-23
nat64: Handle "Well-Known Prefix" 64:ff9b::/96

uknoblic:
I added the well-known-prefix to the list of allowed prefixes in ip64-addr.c. Private IPv4 addresses are dropped as specified in RFC 6052.


alignan:
Seems legit to me, are there any comments about this?

I would suggest squashing into a single commit and merge.

core/net/ip/ip64-addr.c
```diff
@@ -36,6 +36,61 @@
...
+
+  /* 192.0.2.0/24 */
+  uip_ipaddr(&ipaddr, 192, 0, 2, 0);
+  uip_ipaddr(&mask, 255, 255, 255, 0);
+  if(uip_ipaddr_maskcmp(ipv4addr, &ipaddr, &mask)) {
+    return 1;
+  }
+
+  /* 198.51.100.0/24 */
+  uip_ipaddr(&ipaddr, 198, 51, 100, 0);
```
vsaw: I would add the `uip_ipaddr(&mask, 255, 255, 255, 0);` line here and in the following check as well.
Otherwise these will quickly fail if someone misses to see the link and adds a check in between somewhere.

greg-king5: I think that some code doesn't need to be added here.  I think that some comments will be enough.  They should say that those checks depend on the `uip_ipaddr(&mask, 255, 255, 255, 0);` from the previous check.

laurentderu:
This code will break existing NAT64 example as the prefix used is ::ffff:0, the examples should also be updated, or make the NAT64 prefix configurable (This is what we did in 6LBR)

Also, forbiding private network is not a good idea, in several deployment NAT64 is used to allow communication between the WSN and the LAN which is using private addresses.

alignan:
@laurentderu thanks! I forgot about the NAT64 example.

alignan:
@uknoblic any update about this PR?

PR #1146, opened 2015-06-27
LLSEC as FRAMER

kkrentz:
Here are two twists to link layer security:
* I removed the functions `on_frame_created` and `get_overhead` of `llsec_driver`s by letting `llsec_driver`s implement a special `framer` instead. A downside of this approach is that users have to configure both a `framer` and an `llsec_driver`. On the other hand, this simplifies the integration of link layer security in Contiki. Also, this enables us to unsecure frames within `framer.parse`. For example, using the `noncoresec_driver` works as follows:
```
#undef NETSTACK_CONF_LLSEC
#define NETSTACK_CONF_LLSEC noncoresec_driver
#undef NETSTACK_CONF_FRAMER
#define NETSTACK_CONF_FRAMER noncoresec_framer
```
* I replaced the `bootstrap` function of `llsec_driver` with an `init` function. The `bootstrap` function had a parameter `on_bootstrapped`, which was intended for deferring the startup of upper layers until the `llsec_driver` has established keys. However, integrating such a function into all main files is a lot of work. Moreover, a better approach might be to move the whole startup code to a protothread and place PROCESS_WAIT_UNTILs therein.


simonduq:
That's a huge improvement IMO, this makes link-layer security in Contiki so much neater and clearer.

Now I would like to take this one step further. I would be in favor of moving all security operations at the framer level, rather than having the current split LLSEC (above MAC) -- FRAMER. We might even not need a layer at all -- only a driver to manage security policies in a flexible manner. Security could simply be set by selecting a secured framer, or, even simpler, be handled transparently by the `frame802154.c` and `framer-802154.c`.

The downsides of having a LLSEC layer above MAC:
* The send function does not do anything but set security level and frame counter as attributes. Both could be done at frame creation time (next to or in the FRAMER). There is no point maintaining a 4-byte frame counter in every queuebuf when we only need it at the frame level.
* The input function checks and decrypts incoming frames only above MAC, and passes the packet to NETWORK. This means that MAC layer packets such as BEACON, ACK or CMD can not be used directly at MAC.
* Makes everything packetbuf-dependent. This means frames can not be secured or parsed from interrupt. That is quite limiting for the MAC layers in 802.15.4e (soon to be 802.15.4-2015), such as TSCH, where frames often need to be secured and parsed from interrupt (e.g. for preparing an ACK and sending it withing required timing).

What I would be in favor of (and I'm ready to prepare a PR):
* Simply extend the 802.15.4 framer with security. Parse incoming frames with any security level, depending on their aux security header. Create outgoing frames with a security level defined as packetbuf attribute OR directly in `frame802154_fcf_t`.
* To see what this would look like, I'd invite you to check out my preliminary version of it, currently TSCH-specific but I can work on generalizing it: https://github.com/EIT-ICT-RICH/contiki/blob/rich-3.x/core/net/mac/tsch/tsch-security.c (note that the functions `secure_frame` resp. `parse_frame` are meant to be called right after frame creation or before Tx resp. right after parsing)

kkrentz:
* An aspect I have not mentioned so far is that an `llsec_framer` could also write its own security header by calling `hdralloc/hdrreduce`. That is, there is no need to move everything to `framer_802154` from my perspective. Furthermore, in #880 I surrounded the security-related packetbuf attributes with `#if ... #endif`, such that there is no overhead when disabling this option in `framer_802154`.
* The changes in #1108 already allow you to do CCM* calls within interrupts. 9dbd1adb9271f5f080ce2b814b6cc5e1fbf3496d also adds support for doing authentication and encryption at once.

simonduq:
For this PR, :+1: (I'll open an issue or PR later addressing the discussion above)

simonduq:
Oh, I'm noticing only now that the PR was updated since last time I checked. It has gotten even better, now with encryption / decryption going on in the framer rather than llsec layer. :+1:

core/lib/ccm-star.c
```diff
@@ -144,13 +140,38 @@ ctr(uint8_t *m, uint8_t m_len, const uint8_t* nonce)
...
 /*---------------------------------------------------------------------------*/
-static void set_key(const uint8_t *key) {
-    AES_128.set_key((uint8_t*)key);
+static void
+set_key(const uint8_t *key)
+{
+  AES_128.set_key(key);
+}
+/*---------------------------------------------------------------------------*/
+void
```
nvt: Shouldn't aead() be declared static as set_key() is?

core/lib/ccm-star.c
```diff
@@ -144,13 +140,38 @@ ctr(uint8_t *m, uint8_t m_len, const uint8_t* nonce)
...
+    const uint8_t* a, uint8_t a_len,
+    uint8_t *result, uint8_t mic_len,
+    int forward)
+{
+  if(!forward) {
+    /* decrypt */
+    ctr(nonce, m, m_len);
+  }
+  
+  mic(nonce,
```
nvt: Maybe the multiple line breaks aren't necessary in this function call since the identifiers are short.

platform/sky/contiki-sky-main.c
```diff
@@ -189,65 +189,6 @@ set_gateway(void)
 }
 #endif /* NETSTACK_CONF_WITH_IPV4 */
 /*---------------------------------------------------------------------------*/
-static void
```
nvt: There are some modifications of contiki-sky-main.c that might be good to mention in the PR description as well.

core/lib/ccm-star.c
kkrentz: yes

platform/sky/contiki-sky-main.c
kkrentz: These changes are restoring the version of the pre-llsec era.

simonduq:
@kkrentz simply rebasing now should help get travis green again :)

kkrentz:
I am on vacation - will do that asap

simonduq:
I like everything in this PR:
* the fact all crypto processing is now done at/below MAC rather than above
* bootstrap callback replaced with simpler init
* no more create_and_secure
* the new ccm-star interface with combined crypto and mic (aead)

I also tested and it works nicely for me.
:+1: and merging in a month if nobody disagrees

simonduq:
This has conflicts now and needs rebasing.
Please also cherry-pick https://github.com/simonduq/contiki/commit/9bb3427e98a29aff7954c1887828c6058e887571 to port the JN516x ccm-star implementation to the new interface (Travis would fail otherwise).

simonduq:
great, thanks for the extra jn516x change, merging

PR #1152, opened 2015-07-06
turned on correlator in radio packet filter

bkozak-scanimetrics:
After making this change I am able to get ~3x higher max data throughput when sending packets with RIME (as quickly as I can) on ContikiMAC (with duty cycling enabled on the transmitter, duty cycling disabled on the receiver and a CHANNEL_CHECK_RATE of 8) than without these changes.

This improvement seems to be caused by significantly reducing the number of false positive CCAs. These false positives then limit the TX rate by causing ContikiMAC to detect collisions before sending.

In addition, I have found that this also reduces the number of false positives in the receiver routines of ContikiMAC which will result in less power consumption on duty cycling receivers.

I am somewhat concerned that these changes will lead to less reliable operation in very noisy environments but I think that these improvements are too large to ignore so I want the community to have a look.

### DIFF
--- a/cpu/cc26xx/dev/cc26xx-rf.c
+++ b/cpu/cc26xx/dev/cc26xx-rf.c
@@ -892,10 +892,11 @@ init_rf_params(void)
   /* Configure CCA settings */
   GET_FIELD(cmd_ieee_rx_buf, CMD_IEEE_RX, ccaOpt) =
     BITVALUE(CMD_IEEE_RX, ccaOpt, ccaEnEnergy, 1) |
-    BITVALUE(CMD_IEEE_RX, ccaOpt, ccaEnCorr, 0) |
+    BITVALUE(CMD_IEEE_RX, ccaOpt, ccaEnCorr, 1) |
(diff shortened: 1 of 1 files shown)


g-oikonomou:
@jonnteolsson any thoughts?

jonnteolsson:
This should make the system more stable and better performing (as proved) in an IEEE802.15.4 dominated radio environment.  
With new settings radio will only report medium busy (CCA failure) if both RSSI is above threshold AND the correlator has indicated 'success' (at least 3 of the last 8 symbols found in the preamble):

If ccaEnEnergy = 1 and ccaEnCorr = 1 and ccaCorrOp = 1, then:
– If either ccaEnergy or ccaCorr is IDLE, then ccaTmp = IDLE;
– Otherwise, if either ccaEnergy or ccaCorr is Invalid, then ccaTmp = INVALID;
– Otherwise, ccaTmp = BUSY

However if the interfering signal is not IEEE802.15.4 based (e.g. Wi.Fi or other 'energy') Contiki will not 'see' this (CCA true) and try to send anyway, which might result in real transmission failure.

jonnteolsson:
(Y)

PR #1156, opened 2015-07-09
Stunicast timer fix

amitgeron:
In core/net/rime/stunicast.c:

When calling stunicast_send_stubborn, a clock_time_t argument is passed to determine re-transmission time. The connection's ctimer is set with this time to call the send function.
When the timer expires for the first time and the send function is called, it sets the re-transmission timer to CLOCK_SECOND, regardless of the time that was passed to  stunicast_send_stubborn.

In this fix, I simply restart the connection's ctimer, i.e. setting it to the same given clock_time_t argument as passed to stunicast_send_stubborn.


simonduq:
looks ok to me, scheduling for merging in a month :+1:

PR #1158, opened 2015-07-10
Fixed wrong first cycle_start

bkozak-scanimetrics:
The first time that powercycle() runs, cycle_start is incremented
by CHECK_TIME twice which causes the second cycle to be late.

This commit fixes this.

### DIFF
--- a/core/net/mac/contikimac/contikimac.c
+++ b/core/net/mac/contikimac/contikimac.c
@@ -289,9 +289,15 @@ off(void)
   }
 }
 /*---------------------------------------------------------------------------*/
-static volatile rtimer_clock_t cycle_start;
 static void powercycle_wrapper(struct rtimer *t, void *ptr);
 static char powercycle(struct rtimer *t, void *ptr);
+/*---------------------------------------------------------------------------*/
+static volatile rtimer_clock_t cycle_start;
+#if SYNC_CYCLE_STARTS
+static volatile rtimer_clock_t sync_cycle_start;
+static volatile uint8_t sync_cycle_phase;
+#endif
+/*---------------------------------------------------------------------------*/
 static void
 schedule_powercycle(struct rtimer *t, rtimer_clock_t time)
 {
@@ -340,7 +346,7 @@ powercycle_turn_radio_off(void)
 #if CONTIKIMAC_CONF_COMPOWER
   uint8_t was_on = radio_is_on;
 #endif /* CONTIKIMAC_CONF_COMPOWER */
-  
+
   if(we_are_sending == 0 && we_are_receiving_burst == 0) {
     off();
 #if CONTIKIMAC_CONF_COMPOWER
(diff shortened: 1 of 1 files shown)


bkozak-scanimetrics:
actually, the more I look at this the more I think that I've stumbled on something a little bigger than just fixing the first inter-CCA period.

I don't see how the timing can fix itself unless the device falls way behind on its timing (because interrupts are being masked for a long time for example).

My testing with real hardware (CC26xx) actually shows some performance gains after making this change.

simonduq:
@g-oikonomou, do you have any opinion?

g-oikonomou:
No opinion. I leave ContikiMAC to its experts :)

simonduq:
Haven't used ContikiMAC in a couple years and am not aware of this particular bug.
@bkozak-scanimetrics sorry, we need you to spoon-feed us ;) Can you please explain the problem and fix in greater detail?

core/net/mac/contikimac/contikimac.c
```diff
@@ -319,13 +323,32 @@ powercycle_turn_radio_on(void)
   }
 }
 /*---------------------------------------------------------------------------*/
+static void advance_cycle_start(void){
```
greg-king5: A little coding style issue:
```c
static void
advance_cycle_start(void)
{
```

bkozak-scanimetrics:
The issue is that after the very first run through of the while loop in ```power_cycle()``` routine, the next call to ```power_cycle()``` (i.e. the next CCA) is scheduled on an rtimer at the wrong point in time.

Here's how it happens:

First, cycle_start gets initialized to the current time as we start the current CCA.
```C
cycle_start = RTIMER_NOW();
```
Next (assuming ```SYNC_CYCLE_STARTS``` is false), we move cycle_start forward
```C
cycle_start += CYCLE_TIME;
```
When the CCA has completed (bottom of the loop), we schedule the next run of ```power_cycle()```
```C
schedule_powercycle_fixed(t, CYCLE_TIME + cycle_start);
```

Therefore, we have scheduled the second CCA to occur at time:
 T<sub>C0</sub> + (2 *  ```CYCLE_TIME)```
Where  T<sub>C0</sub> is the time at which the first cycle_start started.

Each CCA is meant to occur at T<sub>Cn</sub> + ```CYCLE_TIME``` (as all of the others will) so this behaviour is wrong. To summarize, the problem here is that, while a CCA is running, ```cycle_start``` is meant to represent  T<sub>Cn</sub> ```power_cycle``` runs it actually represents  what would be T<sub>Cn + 1</sub>.

bkozak-scanimetrics:
The rebase is complete and I've fixed the style error that @greg-king5 pointed out.

core/net/mac/contikimac/contikimac.c
```diff
@@ -483,22 +492,25 @@ powercycle(struct rtimer *t, void *ptr)
...
-	 be blocked until a packet is detected */
+      until then.  Sleeping will not exit from this interrupt, so
+      ensure an occasional wake cycle or foreground processing will
+      be blocked until a packet is detected */
 #if RDC_CONF_MCU_SLEEP
+
       static uint8_t sleepcycle;
       if((sleepcycle++ < 16) && !we_are_sending && !radio_is_on) {
-        rtimer_arch_sleep(CYCLE_TIME - (RTIMER_NOW() - cycle_start));
+        rtimer_arch_sleep( RTIMER_NOW() - cycle_start );
```
simonduq: there should be no space after `(` and before `)`

simonduq:
ok, sounds reasonable. Comments from any other ContikiMAC users are very welcome/needed here

bkozak-scanimetrics:
Fixed the style issue pointed out by @simonduq...

> there should be no space after ( and before )

simonduq:
Tried in Cooja and I saw both the problem and the fix. The code looks good and seems correct to me.
:+1:

simonduq:
merging after timeout

nvt:
@ghavag It would be good if you could open an issue (or even a PR) about this matter.

ghavag:
@nvt Thank you for your advise! I just opened issue #2190.

PR #1160, opened 2015-07-11
Make ContikiMAC more configurable

g-oikonomou:
This pull changes ContikiMAC so that it allows the project / platform to provide values for `CCA_SLEEP_TIME` and `LISTEN_TIME_AFTER_PACKET_DETECTED`.

This is useful for sub-ghz operation.

This has been shamelessly stolen from the [Mountain Sensing project](https://github.com/feshie/contiki)

@heliosfa @kmartinez

### DIFF
--- a/core/net/mac/contikimac/contikimac.c
+++ b/core/net/mac/contikimac/contikimac.c
@@ -136,11 +136,15 @@ static int we_are_receiving_burst = 0;
 
 /* CCA_SLEEP_TIME is the time between two successive CCA checks. */
 /* Add 1 when rtimer ticks are coarse */
+#ifdef CONTIKIMAC_CONF_CCA_SLEEP_TIME
+#define CCA_SLEEP_TIME CONTIKIMAC_CONF_CCA_SLEEP_TIME
+#else
 #if RTIMER_ARCH_SECOND > 8000
 #define CCA_SLEEP_TIME                     RTIMER_ARCH_SECOND / 2000
 #else
 #define CCA_SLEEP_TIME                     (RTIMER_ARCH_SECOND / 2000) + 1
-#endif
+#endif /* RTIMER_ARCH_SECOND > 8000 */
+#endif /* CONTIKIMAC_CONF_CCA_SLEEP_TIME */
 
 /* CHECK_TIME is the total time it takes to perform CCA_COUNT_MAX
    CCAs. */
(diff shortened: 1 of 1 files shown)


alignan:
+1

kmartinez:
yes +1 !
timing is different and critical for things like the CC1120 - @heliosfa tweaked them to get our Z1+CC1120+6LowPAN+CoAP working on the Scotish mountains.

nvt:
:+1:

LanshunNie:
+1 !
May I ask how to tweak the timing parameters such as CCA_SLEEP_TIME, CCA_CHECK_TIME for contikimac effectively?
Thanks a lot. 
We are working on port CC1120 (434MHz, 100kbps) to contikimac, and fail to make it work.

kmartinez:
Hi LanshunNie - we already have an 868MHz  CC1120 port for contikimac - working and test deployed: see http://mountainsensing.org/
The main issue will be adjusting the timings for the lower bit rate of 433MHz (I do not think you will get 100 kbps)

LanshunNie:
Hi, kmartinez:
Nice work on mountain sense.

Thanks for your reply.
I don't understand well "I do not think you will get 100 kbps". Would you please explain it more?

With 50ksps+4-GFSK+434MHz, our motes seem work well with nullrdc and CC1120 driver from Thingsquare mist 1.1.0. 
However, they achieve poor packet receiption rate (less than 20%) with contikimac even we tried a bunch of timing parameters. (When calculate these parameters,  we learn a lot from your project. Thanks again.)

Am I missing something? Is 100kbps totally wrong?
Hope for your reply.

g-oikonomou:
I've got pretty stable settings for ContikiMAC at 50kbps in the 863 15.4g band (much better than the 20% you are reporting), they should be a reasonable starting point for you.

```
#define CONTIKIMAC_CONF_CCA_CHECK_TIME            (RTIMER_ARCH_SECOND / 1600)
#define CONTIKIMAC_CONF_CCA_SLEEP_TIME            (RTIMER_ARCH_SECOND / 210)
#define CONTIKIMAC_CONF_LISTEN_TIME_AFTER_PACKET_DETECTED  (RTIMER_ARCH_SECOND / 20)
#define CONTIKIMAC_CONF_SEND_SW_ACK               1
#define CONTIKIMAC_CONF_AFTER_ACK_DETECTECT_WAIT_TIME (RTIMER_SECOND / 1000)
#define CONTIKIMAC_CONF_INTER_PACKET_INTERVAL     (RTIMER_SECOND / 280)
```
Keep in mind that I am using a different radio, not the CC1120.

You need a sniffer: Check whether the receiver is sending out ACKs that the sender misses due to not staying awake long enough to pick them up. If you do sniff the ACKs, then you need to fiddle with sender post-TX timings (stay on for longer). If you do not see the ACKs then you need to fiddle with receiver pre-RX timings.

.4g specifies 100kbps as one of the operating modes for the 470 band (China) (Band 2, Op Mode 2), so I guess that if you get the timings right, and assuming the driver is not letting you down, you should be OK. You can always experiment at 50kbps, which has been used successfully. Once you have those right, then pump up the data rate. You will obviously need to adjust timings again, but it will give you good experience and you will also be able to determine driver stability.

LanshunNie:
Thank you so much, George. 
We would definitely study your instruction and experiment at 50kbps.
Thanks again.

LanshunNie:
Hi, George, thank you for your strong support. 
When we start from 50kbps@434MHz with your timing parameters, and borrow register settings from Adam's new driver, the packet recepition rate becomes pretty good. : )

As pointed in CC1120 new driver, CC1120 spends at least 1.2ms to get a good CCA readout if it was turned off recently. This is fixed by the statement "BUSYWAIT_UNTIL((state()==CC1120_STATE_RX), RTIMER_SECOND/100);" inside channel_clear funtion. So shall we adjust CCA_CHECK_TIME parameter? Or we just keep it as (RTIMER_ARCH_SECOND/ 1600)?  
Thanks.

smpj:
How do I change frequency to 865 M Hz in CC1200 driver

PR #1161, opened 2015-07-11
Implement TX of S/W ACKs with ContikiMAC

g-oikonomou:
This PR adds support for S/W ACK TX in ContikiMAC.

This is handy when the radio cannot send H/W ACKs. This is quite common for sub-ghz operation.

### DIFF
--- a/core/net/mac/contikimac/contikimac.c
+++ b/core/net/mac/contikimac/contikimac.c
@@ -195,6 +195,12 @@ static int we_are_receiving_burst = 0;
    to a neighbor for which we have a phase lock. */
 #define MAX_PHASE_STROBE_TIME              RTIMER_ARCH_SECOND / 60
 
+#ifdef CONTIKIMAC_CONF_SEND_SW_ACK
+#define CONTIKIMAC_SEND_SW_ACK CONTIKIMAC_CONF_SEND_SW_ACK
+#else
+#define CONTIKIMAC_SEND_SW_ACK 0
+#endif
+
 #define ACK_LEN 3
 
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


simonduq:
Nice feature but isn't the ACK timing going to suffer a lot? I feel this should be done from `powercycle`, right after reception, while you're still in interrupt.

g-oikonomou:
#1160 takes care of timings - correct configuration required but this has been tested extensively at 50kbps.

I am not sure I understand where in `powercycle` this could be done. The frame needs to be read out from the radio and parsed before we can determine whether an ACK should be sent.

simonduq:
I meant, using  `NETSTACK_RADIO.read` just like Contikimac already does for reading ACKs. But maybe this is not interrupt-proof on some platforms? Or am I missing something else?

As for timing, I'm not worried about the extra delay as much as the added jitter. `packet_input` might be called with arbitrary delay.

g-oikonomou:
OK so perhaps it's me missing something :)

We receive a frame as normal: interrupt polls RF driver process calls `NETSTACK_RDC.read()`. Within `read()` and once all checks have succeeded, the s/w sends an ACK if the radio cannot do this by itself (configurable). Unless I'm very mistaken, `powercycle` never reads out frames from the RF FIFO.

The logic suggested here is identical to nullrdc s/w ACKs. As you say, timing will have some jitter, but surely we cannot start reading out frames from within the RF interrupt just to support this?

simonduq:
We're on the same page, and I agree doing all from `powercycle` is probably too much of a change just for this feature.
:+1:

alignan:
:+1:

Issue #1165, opened 2015-07-13
ContikiMAC - qsend_list doesn't account for changed packetbuf from mac_call_sent_callback


bkozak-scanimetrics:
I have a test application (which can be found [here](https://github.com/bkozak-scanimetrics/RIME-Throughput-Test)) that attempts to burst through a number of packets at once by sending new packets in RIME's unicast "sent" callback.

I had expected that this would cause ContikiMAC to send the packets through faster by skipping the CCA check (because is_receiver_awake will be true) but I have found that this does not happen.

I believe that this problem is caused because the packetbuf is modified in the call to `mac_call_sent_callback` in the function `qsend_list()` in contikimac.c. After modifying the packetbuf PACKETBUF_ATTR_PENDING is now false and so qsend_list exits early . I'm not entirely sure, but I believe that this behaviour is not intended.

Thanks.

Issue #1166, opened 2015-07-13
CSMA driver slows down queued packets


bkozak-scanimetrics:
This is related to the test I am performing in #1165.

I have found that when using ContikiMAC with RIME (I think this may also apply to other network stack configurations), if I try to queue up packets one after another using the unicast "sent" callback, I actually get a severe throughput decrease.

After some debugging, I found that the majority of this slowdown occurs because of the behaviour of `free_packet()` in csma.c. What happens is that, if there are a number of packets queued together, `free_packet()` will setup a ctimer to transmit after `default_timebase()` time ticks have elapsed. This will only happen if there are packets in CSMA's neighbour queue but ContikiMAC does not send them all at once (this is the case if the queued packets do not have the PENDING attribute set).

So, we can actually have (much) faster sending by sending packets one at a time in the application such that we never have a queue in the csma driver. If the receiver is performing duty cycling (which is not the case in my test) I don't think that there will be any slowdown because `default_timebase()` will wait for only one `channel_check_interval()`.

I don't see any good reason for this behaviour (although perhaps someone can tell me why it exists) and suspect that it may be a bug.

Thanks.

cmorty:
It's very likely that it's a bug. After all this is all growing software, reusing old parts. Try fixing it, run the test-suite and if it works fine, send a pull-request. You're more likely to get decent feedback then.

simonduq:
I don't think that's a bug, but I agree things can be improved in the particular case described here (packets en-queued from packet_sent callback)
I added a comment under https://github.com/contiki-os/contiki/pull/1167

PR #1167, opened 2015-07-13
Fix Mishandling of PENDING/Queued Packets at ContikiMAC and CSMA

bkozak-scanimetrics:
Fix for #1165

This fixes this problem by storing pending status in a local variable before calling mac_call_sent_callback.

Since this is (in my opinion) the obvious way to fix this problem I have gone ahead and simply implemented (and tested) this solution.

### DIFF
--- a/core/net/mac/contikimac/contikimac.c
+++ b/core/net/mac/contikimac/contikimac.c
@@ -805,6 +805,7 @@ qsend_list(mac_callback_t sent, void *ptr, struct rdc_buf_list *buf_list)
   struct rdc_buf_list *next;
   int ret;
   int is_receiver_awake;
+  int pending;
   
   if(buf_list == NULL) {
     return;
@@ -849,7 +850,9 @@ qsend_list(mac_callback_t sent, void *ptr, struct rdc_buf_list *buf_list)
 
     /* Prepare the packetbuf */
     queuebuf_to_packetbuf(curr->buf);
-    
+
+    pending = packetbuf_attr(PACKETBUF_ATTR_PENDING);
+
     /* Send the current packet */
     ret = send_packet(sent, ptr, curr, is_receiver_awake);
     if(ret != MAC_TX_DEFERRED) {
@@ -866,7 +869,7 @@ qsend_list(mac_callback_t sent, void *ptr, struct rdc_buf_list *buf_list)
       /* The transmission failed, we stop the burst */
       next = NULL;
     }
-  } while((next != NULL) && packetbuf_attr(PACKETBUF_ATTR_PENDING));
+  } while((next != NULL) && pending);
 }
 /*---------------------------------------------------------------------------*/
 /* Timer callback triggered when receiving a burst, after having
--- a/core/net/mac/csma.c
+++ b/core/net/mac/csma.c
@@ -171,8 +171,10 @@ transmit_packet_list(void *ptr)
 }
 /*---------------------------------------------------------------------------*/
 static void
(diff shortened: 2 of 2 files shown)


bkozak-scanimetrics:
I've found that my solution for #1166 actually relies on my initial commit in this PR. I've therefore decided to change the title of this PR and add my fix for #1166 here.

After making these changes I've found that I can achieve dramatically higher data throughput (over 2x higher) in my [RIME throughput test](https://github.com/bkozak-scanimetrics/RIME-Throughput-Test) when multiple (I've tested up to 4) packets are queued at once.

simonduq:
Can you please point out exactly where in `mac_call_sent_callback` `PACKETBUF_ATTR_PENDING` is unset? Might be non-intended behavior.

simonduq:
There are a number of code style issues in this PR which need addressing

simonduq:
The `ctimer` fix in `csma.c` seems dangerous to me. The original idea was to not reschedule immediately, because we might be in the case where ContikiMAC failed to transmit, Csma decided to drop the packet, and in this case we want a backoff even though there are more packets pending. I would be ok with a fix that sets `ctimer` to 0 iff there are packets pending AND the current transmit callback is `MAC_TX_OK`

bkozak-scanimetrics:
> There are a number of code style issues in this PR which need addressing

Style issues should be good now.

> Can you please point out exactly where in mac_call_sent_callback PACKETBUF_ATTR_PENDING is unset? Might be non-intended behavior.

As far as I know this will only occur if the application sets a callback which, as far as In know, is only possible when using RIME. I do this in the application I linked to above.

>  I would be ok with a fix that sets ctimer to 0 iff there are packets pending AND the current transmit callback is MAC_TX_OK

This makes sense to me. I'll make this change when I have some more time.

bkozak-scanimetrics:
Made the changes and did some quick tests to make sure that everything is still working.

Seems ok now as far as I'm concerned.

simonduq:
How will this play with nullrdc, which transmits all packets in `buf_list` in a burst and without setting the packet pending bit?

bkozak-scanimetrics:
Only had a quick look but I'm pretty sure the behaviour is unchanged for nullrdc.

Since the ctimer won't be able to run while nullrdc is still sending packets it won't matter that packets with MAC_OK status are scheduled to occur immediately.

Instead what will happen is that nullrdc will just continue to send packets until either:
A. It encounters an error, and the ctimer will be bumped ahead into the future
B. All packets finish sending in which case the ctimer is cancelled before it has a chance to run

simonduq:
I agree. Similarly, I think you should keep setting ctimer independent of the packet pending attribute. I would remove the conditional `if(!packetbuf_attr(PACKETBUF_ATTR_PENDING)) {`. This way, csma makes sure it is always scheduled as long as there is something in the list, regardless of the rdc layer.

Looking closer at the current Contikimac, it seems the packet pending bit is first set on all packets in the list, and then only transmission starts. In case of failure, it will stop the burst, call the csma callback with `MAC_TX_NOACK`, but with packet pending bit set... would end up not re-scheduling ctimer with your current patch, right?

bkozak-scanimetrics:
Yes it looks like you're right.

I'll update when I have time.

bkozak-scanimetrics:
I've made the change.

I believe that this still accomplishes this PR's original intent.

simonduq:
Grea, thanks! :+1:

simonduq:
merging in a month if nobody finds a flaw in this fix

simonduq:
merging after timeout

PR #1169, opened 2015-07-16
Fix multiple warnings (TCPIP, coap-observe, msp430, z1/tmp102, ContikiMAC)

arthurfabre:
These 5 commits fix compiler warnings for TCPIP, coap-observe, msp430 mtarch, z1/tmp102 driver, and ContikiMAC.
Details are included in the commit messages.

I can submit these all as separate PRs if that's preferred.

### DIFF
--- a/apps/er-coap/er-coap-observe.c
+++ b/apps/er-coap/er-coap-observe.c
@@ -234,8 +234,6 @@ coap_observe_handler(resource_t *resource, void *request, void *response)
   coap_packet_t *const coap_res = (coap_packet_t *)response;
   coap_observer_t * obs;
 
-  static char content[16];
-
   if(coap_req->code == COAP_GET && coap_res->code < 128) { /* GET request and response without error code */
     if(IS_OPTION(coap_req, COAP_OPTION_OBSERVE)) {
       if(coap_req->observe == 0) {
@@ -249,13 +247,14 @@ coap_observe_handler(resource_t *resource, void *request, void *response)
            * A subscription should return the same representation as a normal GET.
            * Uncomment if you want an information about the avaiable observers.
            */
-          /*
-           * coap_set_payload(coap_res,
-           *                  content,
-           *                  snprintf(content, sizeof(content), "Added %u/%u",
-           *                           list_length(observers_list),
-           *                           COAP_MAX_OBSERVERS));
-           */
+#if 0
+          static char content[16];
+          coap_set_payload(coap_res,
+                           content,
+                           snprintf(content, sizeof(content), "Added %u/%u",
+                                    list_length(observers_list),
+                                    COAP_MAX_OBSERVERS));
+#endif
         } else {
           coap_res->code = SERVICE_UNAVAILABLE_5_03;
           coap_set_payload(coap_res, "TooManyObservers", 16);
--- a/core/net/ip/tcpip.c
+++ b/core/net/ip/tcpip.c
@@ -155,6 +155,7 @@ unsigned char tcpip_is_forwarding; /* Forwarding right now? */
 PROCESS(tcpip_process, "TCP/IP stack");
 
 /*---------------------------------------------------------------------------*/
+#if UIP_TCP || UIP_CONF_IP_FORWARD
 static void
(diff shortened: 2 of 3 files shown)


arthurfabre:
Is there anything I can do to help get this merged in?

core/net/ip/tcpip.c
```diff
@@ -155,6 +155,7 @@ unsigned char tcpip_is_forwarding; /* Forwarding right now? */
 PROCESS(tcpip_process, "TCP/IP stack");
 
 /*---------------------------------------------------------------------------*/
+#if UIP_TCP
```
simonduq: shouldn't this be `#if UIP_TCP || UIP_CONF_IP_FORWARD`?
(see https://github.com/feshie/contiki/blob/pr-warnings-fix/core/net/ip/tcpip.c#L180)

platform/z1/dev/tmp102.c
```diff
@@ -43,6 +43,12 @@
 #include "i2cmaster.h"
 #include "tmp102.h"
 
+#ifdef TMP_DEBUG
+#define PRINTFDEBUG(...) printf(__VA_ARGS__)
+#else
+#define PRINTFDEBUG(...)
+#endif
+
```
simonduq: isn't `PRINTFDEBUG` defined in: https://github.com/contiki-os/contiki/blob/master/platform/z1/dev/i2cmaster.h ?

arthurfabre:
Good catch on the TCP_IP ifdef guard, I've updated that.

PRINTFDEBUG is indeed defined in `platform/z1/dev/i2cmaster.h`, it justwasn't defined in the version for our current platform (based on the z1 platform) - I've removed that commit.

simonduq:
:+1:

alignan:
:+1:

cpu/msp430/mtarch.c
```diff
@@ -66,10 +66,10 @@ mtarch_start(struct mtarch_thread *t,
 
   t->sp = &t->stack[MTARCH_STACKSIZE - 1];
 
-  *t->sp = (unsigned short)mt_exit;
+  *t->sp = (unsigned short)(void *)mt_exit;
```
laurentderu: Question : is it allowed on 20bits version of msp ? I believe far pointers are to be stored in 32 bits integers.

simonduq: That's a good question :) If this is not allowed, then having `sp` in `struct mtarch_thread` as `unsigned short *` is wrong, right? Should be `void **` instead or?

arthurfabre: I don't know if it's allowed, but we're using it without issues on a 20bit msp430

laurentderu: Actually the original warning is pretty clear : <pre>warning: cast from pointer to integer of different size</pre>
I just tested :
<pre>
  printf("%d\n", sizeof(int *));
  printf("%d\n", sizeof(testfunc));
</pre>
gives : 
<pre>
2
4
</pre>
A pointer to a function will be 16bits only if the compiler knows that the function is in the near region. But there is no issue because mtarch is not used in vanilla Contiki.

@simonduq void * can only be used reliably for data pointer, not for function pointer

simonduq: My suggestion then is to revert this particular fix to leave the warning in there, as long as we don't have a better fix (actually fixing or removing mtarch in the 20-bit case)

arthurfabre: That sounds sane, I've reverted this "fix"

apps/er-coap/er-coap-observe.c
```diff
@@ -250,6 +248,7 @@ coap_observe_handler(resource_t *resource, void *request, void *response)
            * Uncomment if you want an information about the avaiable observers.
```
cmorty: If you are touching this anyway. Why not put it in an #if 0 .... #endif. - I know its a bit off topic and only a suggestion.

arthurfabre: Done!

PR #1170, opened 2015-07-16
CC26xx autoflush bad crc packets

bkozak-scanimetrics:
Setting bAutoFlushCrc=1 so that the radio co-processor will automatically drop packets with bad crc.

Created as per discussion in #1140. 

I've confirmed that this fixes the problem of corrupt frames propagating up to the application layer.


g-oikonomou:
:+1: 

I'll just wait for travis

PR #1173, opened 2015-07-20
CC26xx - Enable RX_NOK on DEBUG builds

bkozak-scanimetrics:
Enabled the RX_NOK interrupt, if DEBUG or CC26XX_DEBUG_CRC are set true, for the purpose of monitoring packets with bad CRC.

With bAutoFlushCrc now set true, we can't monitor bad CRC packets without enabling RX_NOK (we could also monitor the radio's  nRxNok counter but then we can't see the bad CRCs as they occur).

### DIFF
--- a/cpu/cc26xx/dev/cc26xx-rf.c
+++ b/cpu/cc26xx/dev/cc26xx-rf.c
@@ -92,6 +92,12 @@
 #define PRINTF(...)
 #endif
 /*---------------------------------------------------------------------------*/
+#ifdef CC26XX_CONF_DEBUG_CRC
+#define CC26XX_DEBUG_CRC CC26XX_CONF_DEBUG_CRC
+#else /* CC26XX_CONF_DEBUG_CRC */
+#define CC26XX_DEBUG_CRC DEBUG
+#endif /* CC26XX_CONF_DEBUG_CRC */
+/*---------------------------------------------------------------------------*/
 /* Data entry status field constants */
 #define DATA_ENTRY_STATUS_PENDING    0x00 /* Not in use by the Radio CPU */
 #define DATA_ENTRY_STATUS_ACTIVE     0x01 /* Open for r/w by the radio CPU */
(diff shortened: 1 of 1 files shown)


cpu/cc26xx/dev/cc26xx-rf.c
```diff
@@ -266,11 +272,16 @@ const output_config_t *tx_power_current = &output_power[0];
 /*---------------------------------------------------------------------------*/
 /* RF interrupts */
 #define RX_IRQ       IRQ_IEEE_RX_ENTRY_DONE
+#define RX_NOK_IRQ   IRQ_IEEE_RX_NOK
 #define TX_ACK_IRQ   IRQ_IEEE_TX_ACK
 #define ERROR_IRQ    IRQ_INTERNAL_ERROR
 
 /* Those IRQs are enabled all the time */
-#define ENABLED_IRQS (RX_IRQ + ERROR_IRQ)
+#if !(CC26XX_DEBUG_CRC)
```
g-oikonomou: I would prefer inverting the blocks here, staring with `#if CC26XX_DEBUG_CRC`

cpu/cc26xx/dev/cc26xx-rf.c
```diff
@@ -92,6 +92,12 @@
 #define PRINTF(...)
 #endif
 /*---------------------------------------------------------------------------*/
+#if !defined(CC26XX_DEBUG_CRC) && defined(DEBUG)
```
g-oikonomou: How about something like this instead of !checks:

```
#ifdef CC26XX_CONF_DEBUG_CRC
#define CC26XX_DEBUG_CRC CC26XX_CONF_DEBUG_CRC
#else /* CC26XX_CONF_DEBUG_CRC */
#define CC26XX_DEBUG_CRC DEBUG
#endif /* CC26XX_CONF_DEBUG_CRC */
```

cpu/cc26xx/dev/cc26xx-rf.c
bkozak-scanimetrics: Fair enough

PR #1174, opened 2015-07-20
Cc26xx fix read frame implementation

bkozak-scanimetrics:
Fixes various bugs and stylistic issues with the `read_frame` function.

In particular:
- Fixed a race condition caused by using the dataEntry after setting status to PENDING
- Fixed general misuse of the packet length
- made some stylistic changes

### DIFF
--- a/cpu/cc26xx/dev/cc26xx-rf.c
+++ b/cpu/cc26xx/dev/cc26xx-rf.c
@@ -1350,15 +1350,26 @@ read_frame(void *buf, unsigned short buf_len)
     return 0;
   }
 
-  if(!rx_read_entry[8]) {
+
+  if(rx_read_entry[8] < 4) {
+    PRINTF("RF: too short\n");
+    RIMESTATS_ADD(tooshort);
+
     release_data_entry();
     return 0;
   }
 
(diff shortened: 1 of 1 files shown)


g-oikonomou:
This is useful stuff as discussed, but not hugely keen on the `goto`. Can we put that in a separate function that bumps the read entry? Then we can call this function and return instead of `goto`

bkozak-scanimetrics:
> Can we put that in a separate function that bumps the read entry? 

Sure thing.

I quite like goto for just this sort of thing but I'll make the change as soon as I have a chance.

bkozak-scanimetrics:
Should be good now

g-oikonomou:
Just tested this :+1: 

Please, can you fix the naming convention for `release_dataEntry` -> `release_data_entry` then we are ready for this to be merged

bkozak-scanimetrics:
I've got the rename change in my local repo.

Before I push, just want to make it clear that I broke the naming convention in order to match the naming of the ```dataEntry``` structure.

If that's not justification enough for you just say the word and I'll push the changes.

g-oikonomou:
Please push the change, thanks.

PR #1177, opened 2015-07-24
Stm32nucleo spirit1

STclab:
The port supports the following boards from ST:
-	NUCLEO-L152RE board, based on the STM32L152RET6 ultra-low power microcontroller
-	X-NUCLEO-IDS01A4 based on sub-1GHz SPSGRF-868 SPIRIT1 module (operating at 868 MHz)
-	X-NUCLEO-IDS01A5 based on sub-1GHz SPSGRF-915 SPIRIT1 module (operating at 915 MHz)
-   X-NUCLEO-IKS01A1 featuring motion MEMS and environmental sensors (optional)

ST CLAB team


g-oikonomou:
In addition to our discussion over email:

* I see you are trying to add CMSIS sources directly to the Contiki repo. This has been discussed before and the outcome was against due to licensing. One solution could be to add them as a submodule, in which case I'd be more keen to consider it.
* I see that, similar to other ports for ST devices, you are adding your existing sources/headers etc to Contiki, unchanged. This is something that I _personally_ dislike for a couple of reasons, code style and naming being among them. They just look totally out of place. As per my previous mail, one option is to pull those in a sub-module.
* In terms of regression warnings: Assuming you decide to pull in the libraries as a sub-module, then you can just exclude their documentation from the doxygen build (see CC26xxware). Otherwise I can (again, _personally_) see no other route than to fix the warnings

STclab:
Thank you, we got your points and we are going to modify the port including our drivers and HAL components as sub-module.

STclab:
We have refactored the code, please let us know your opinion.

alignan:
I can't test, don't have the hardware.

STclab:
Hello,
we can ship a kit for you to test, just let us know the address.

STclab:
We have just published a “quick start guide” presentation to help using this port (http://www.st.com/st-web-ui/static/active/en/resource/sales_and_marketing/presentation/product_presentation/x-nucleo-ids01a4_contiki_6lowpan_quick_guide.pdf). The presentation outlines all the steps to have Contiki running on our STM32 Nucleo (and expansion) boards. It is concluded by the description of a demo showing the access to the sensors data of a wireless node using CoAP (er-rest-example). As this is an official ST port, this presentation (and other collaterals) is located on st.com in the product page of our sub-1 GHz SPIRIT1 RF communication expansion boards for STM32 Nucleo (http://www.st.com/web/catalog/tools/FM146/CL2167/SC2006/PF261982). 

We think that the presentation can be helpful for the evaluation of this port, please let us know if you need more information. We shipped a test kit to Antonio a few weeks ago and, needless to say, we are eager to receive feedbacks about it :-)

alignan:
Finally got some time to review :-)

Instead of building with `USE_SUBGHZ_BOARD` why not change it to `BOARD`? right now we have the cc26xx/cc13xx platforms following this approach, and soon the Zolertia ones.  I prefer to have a consistent approach for current and new platforms.

More feedback to come!

alignan:
BTW drag-n-drop FTW :-)

alignan:
The heading of the license files are a bit off in some files, like ` cpu/arm/stm32l152/console.c` vs ` cpu/arm/stm32l152/rtimer-arch.c `, please keep consistency (the licensing in the later file is preferred).

alignan:
The files also lacks Doxygen source documentation, this is not mandatory as currently stated in the [Code Contributions guideline](https://github.com/contiki-os/contiki/wiki/Code-Contributions) but we encourage new platforms to provide this.

cpu/arm/stm32l152/sysmem.c
```diff
@@ -0,0 +1,67 @@
...
+**                of any kind.
+**
+**  (c)Copyright System Workbench for MCU.
+**  You may use this file as-is or modify it according to the needs of your
+**  project. Distribution of this file (unmodified or modified) is not
+**  permitted. System Workbench for MCU permit registered System Workbench(R) users the
+**  rights to distribute the assembled, compiled & linked contents of this
+**  file as part of an application binary file, provided that it is built
+**  using the System Workbench for MCU toolchain.
+**
```
alignan: Is this 3-BSD compatible?

platform/stm32nucleo-spirit1/contiki-conf.h
```diff
@@ -0,0 +1,152 @@
...
+#define NETSTACK_CONF_RDC     nullrdc_driver
+#define NETSTACK_RDC_HEADER_LEN 0
+
+#undef NETSTACK_CONF_MAC
+#define NETSTACK_CONF_MAC csma_driver
+#define NETSTACK_MAC_HEADER_LEN 0
+
+#define SICSLOWPAN_CONF_MAC_MAX_PAYLOAD \
+  (NETSTACK_RADIO_MAX_PAYLOAD_LEN - NETSTACK_MAC_HEADER_LEN - \
+   NETSTACK_RDC_HEADER_LEN )
```
alignan: extra white space in `NETSTACK_RDC_HEADER_LEN )`

platform/stm32nucleo-spirit1/contiki-conf.h
```diff
@@ -0,0 +1,152 @@
...
+
+#define RIMESTATS_CONF_ENABLED                 0
+#define RIMESTATS_CONF_ON                      0
+
+
```
alignan: Remove extra spaces

platform/stm32nucleo-spirit1/dev/acceleration-sensor.c
```diff
@@ -0,0 +1,140 @@
...
+/*---------------------------------------------------------------------------*/
+#if COMPILE_SENSORS
+/*---------------------------------------------------------------------------*/
+#include "lib/sensors.h"
+#include "acceleration-sensor.h"
+#include "st-lib.h"
+/*---------------------------------------------------------------------------*/
+static int _active = 1;
+/*---------------------------------------------------------------------------*/
+static void init(void)
```
alignan: Change code style, functions should be:
```
static void
deactivate(void)
```

alignan:
Please run `uncrustify`, the port needs some code format loving (i.e `button-sensor.c`, etc).

STclab:
Thank you,
we have updated the license headers and formatted the code with uncrustify. Regarding the BOARD / USE_SUBGHZ_BOARD issue, we are of course willing to be standard. Just note that that flag is not used to specify a subplatform but to select the proper expansion board for radio, since we have different ones for the different frequencies (868 MHz, 915 MHz, ...). We are also using the (not mandatory) USE_SENSOR_BOARD flag (=1 or 0) to enable the support for the (optional) X-NUCLEO-IKS01A1 expansion board. What is the best approach, in your opinion, taking also in account that we may add the support for other expansion boards, of various kind, in the future?

alignan:
I'm `BOARD` as a subplatform/expansion board, so you specify in `TARGET` the platform you wish to compile, then pull in any specific files from `BOARD`, like:

`make TARGET=zoul BOARD=remote`

Then I'm pulling files and config from `Makefile.zoul` and `Makefile.remote`.

I think this is consistent to what you currently have, being in your case `BOARD` a reference to any of your expansion boards.  Do you have a case for which this wouldn't work?

STclab:
Thanks for the clarifications.
If we understand correctly, for example in our case, considering a setup made of a STM32 Nucleo board with expansion boards, the resulting command line would look like:

when using the X-NUCLEO-IDS04A1 (or 05) together with the X-NUCLEO-IKS01A1 sensors expansion board:

make TARGET=stm32nucleo-spirit1 BOARD=ids04a1-iks01a1
make TARGET=stm32nucleo-spirit1 BOARD=ids05a1-iks01a1

when NOT using the optional X-NUCLEO-IKS01A1 sensors expansion board:

make TARGET=stm32nucleo-spirit1 BOARD=ids04a1
make TARGET=stm32nucleo-spirit1 BOARD=ids05a1

Would this be in line with your expectations?

alignan:
This is OK, as you can pull-in any dependencies and specifics in a `Makefile`, you can then regex as much as required.  

Note that I would also be happy also if using `BOARD` for the `ids04a1` and `ids05a1` and `SENSORBOARD` for the expansion, I was mostly concerned about the `USE_SUBGHZ_BOARD` flag.

@g-oikonomou?

STclab:
Ok, so probably an even more consistent approach could be (as an example):

make TARGET=stm32nucleo-spirit1 BOARD=ids04a1 SENSORBOARD=iks01a1

where BOARD is mandatory and can currently be ids04a1 or ids05a1, depending on the version of the SPIRIT1 board, and SENSORBOARD is optional and can currently be only iks01a1.

What do you think?

alignan:
Cool, it is OK to me.

STclab:
Thank you, we updated the repository reflecting these changes. BTW there was a typo in the previous messages (with regard to the BOARD possible values), so the actual flags are:

make TARGET=stm32nucleo-spirit1 BOARD=ids01a4 SENSORBOARD=iks01a1
make TARGET=stm32nucleo-spirit1 BOARD=ids01a5 SENSORBOARD=iks01a1

SENSORBOARD is optional.

alignan:
Great! let's wait for travis to finish and I will review then

STclab:
Hi, the regression test failed because we are now using the BOARD flag and we set it in the Makefile.compile-test file, too: this causes problems to the tests when compiling for srf06-cc26xx platform. Since we would prefer not to set a default value for BOARD (in order to avoid possible troubles to users), it would be ok for you if, in the Makefile.compile-test, we set (instead of a value for BOARD) something like:

 export REGRESSIONTEST=1; \

and then we set things accordingly in our Makefiles?

nfi:
In the long term, I would suggest changing the regression test so it is possible to specify make variables for each compile test. That would make it possible to compile for different boards with same target:

hello-world/stm32nucleo-spirit1:BOARD=ids01a4:SENSORBOARD=iks01a1
hello-world/stm32nucleo-spirit1:BOARD=ids01a5:SENSORBOARD=iks01a1

I created a pull request #1327 for this.

STclab:
The approach proposed by @nfi seems definitely ok to me, so how do you suggest to proceed?

alignan:
We can either wait for #1327 to be merged, or you can default to a `BOARD` if none is selected, your call.

STclab:
Ok, we went for a default value for BOARD anyway, this also fixes other conflicts. Let us know!

alignan:
:+1:

STclab:
Thanks for the thumb up!

This has been a long but fruitful process... At this point, is there anything left to prevent the merging of this pull request?

alignan:
I would like to have at least another contributor or maintainer to take a look, as merges in Contiki normally require 2 approvals as minimum.  I can vouch the PR is functional.

simonduq:
One thing we ask now for new platforms is to have someone commit as a long-term maintainer. If you have one please add his/her contact information in the platform readme file.

Other than that, this satisfies new platform requirements and the code looks very good, thanks for the contribution!

STclab:
Thanks for the review and the positive comments :-) 

We have added the information about the long term mantainers and contributors in the readme file. 

Hopefully we should be done at this point!

simonduq:
Thanks! :+1:

PR #1179, opened 2015-07-27
Changed RPL IPv6 NBR Callback Print Statement

pablocorbalan:
The function ```rpl_ipv6_neighbor_callback``` is called when a neighbour has been removed or a new one has been added to the ```ds6_neighbors``` table. However, the ```rpl_ipv6_neighbor_callback``` always prints ```RPL: Removing a neighbor ...``` when ```DEBUG``` is enabled in ```rpl.c```. This PR changes the print statement to don't misunderstand the NBR event.

Perhaps, it would be nicer to change the definition of ```NEIGHBOR_STATE_CHANGED(n)``` to ```NEIGHBOR_STATE_CHANGED(neighbor, state)``` and then within the RPL callback function we could understand if a NBR has been added or removed. I could change the functions to work that way, but in that case when removing a NBR which state should we send in the ```NEIGHBOR_STATE_CHANGED``` function? I guess we could send ```NBR_INCOMPLETE``` or create a new state for non-neighbours but AFAIK that is not specified in IPv6 ND.

### DIFF
--- a/core/net/rpl/rpl.c
+++ b/core/net/rpl/rpl.c
@@ -277,9 +277,9 @@ rpl_ipv6_neighbor_callback(uip_ds6_nbr_t *nbr)
   rpl_instance_t *instance;
   rpl_instance_t *end;
 
-  PRINTF("RPL: Removing neighbor ");
+  PRINTF("RPL: Neighbor state changed for ");
   PRINT6ADDR(&nbr->ipaddr);
-  PRINTF("\n");
+  PRINTF(", nscount=%u, state=%u\n", nbr->nscount, nbr->state);
   for(instance = &instance_table[0], end = instance + RPL_MAX_INSTANCES; instance < end; ++instance) {
     if(instance->used == 1 ) {
       p = rpl_find_parent_any_dag(instance, &nbr->ipaddr);


nvt:
The state is available in the nbr structure that is passed as an argument. A second argument that indicates whether it is being removed or not could be an option, but this should not be added unless there is an immediate use for it. If there is a candidate parent that corresponds to the changed nbr entry, the RPL callback function will set the rank of that parent to infinite rank regardless of the state.

For this PR, I suggest changing the string to "Neighbor state changed for ".

pablocorbalan:
Thanks for the clarification @nvt . Sorry, I didn't realise that the state was actually sent within the ```uip_ds6_nbr_t``` structure. I have already updated the print statement. I guess that in fact checking out the neighbour state you can realise if the neighbour has beed added or removed. In any case, I have no immediate use for it, so I think we just need to change the print statement.

simonduq:
To make things even clearer, I would add the current state to the printout (`state` field of the sturct as a `%u`)

pablocorbalan:
@simonduq I have updated the print statement so that the ```nbr->nscount``` and ```nbr->state``` parameters are printed as well. As you can add and remove a neighbour with state ```NBR_INCOMPLETE```, it is necessary to check out the ```nbr->nscount``` to figure it out if the neighbour has been added or removed. In case it has been added, the ```nbr->nscount``` will always be ```0```.

simonduq:
:+1:

nvt:
:+1: The PR can be merged as it is, but I'll first leave a suggestion to make the output slightly more clear by adding at least a comma after the IP address, and possibly print the variable names (nscount, state) before the values. E.g., "RPL: Neighbor state changed for <IP address>: nscount=%d, state=%d".

pablocorbalan:
@nvt I have updated the print statement according to your suggestion. Let me know if you want me to change something else.

nvt:
:+1:

Issue #1147, opened 2015-06-30
AES-128 timing vulnerability


akdjka:
AES128 implementation does not run in constant time, the culprit being a branch in galois_mul2().
I suggest to change galois_mul2 to the following.

static uint8_t
galois_mul2(uint8_t value)
{
unsigned xor_val = (value>>7) * 0x1b;
	return ((value << 1) ^ xor_val);
}

As a side effect, on Cortex M0+ the code is both smaller and faster by a tiny amount.

cmorty:
Please make a pull-request with your chages. There are instructions on the wiki how to do so.

akdjka:
Forking your project is much further off my way than copying and pasting a piece of code is for a contiki developer and I don't intend to do it.

simonduq:
I created a PR https://github.com/contiki-os/contiki/pull/1183

simonduq:
now merged!

PR #1183, opened 2015-08-03
galois_mul2 conditional-free implementation

simonduq:
@akdjka's fix for https://github.com/contiki-os/contiki/issues/1147 as a PR.

Looks legit to me, even though I don't think this gives 100% constant time guarantee (that being compiler- and platform-dependent)


vsaw:
:+1:

nvt:
I tested this with examples/llsec/ccm-star-tests/encryption/tests.c, and it still gave a successful results. :+1:

PR #1185, opened 2015-08-04
Fixed the issue with CC2520_READ_REG macro to read correct value of register.

sumanpanchal:
Fixed the issue with CC2520_READ_REG macro to read correct value of register.

### DIFF
--- a/dev/cc2520/cc2520.h
+++ b/dev/cc2520/cc2520.h
@@ -118,6 +118,7 @@ void cc2520_set_cca_threshold(int value);
     CC2520_SPI_ENABLE();                                                \
     SPI_WRITE((CC2520_INS_MEMRD | ((adr>>8)&0xFF)));                    \
     SPI_WRITE((adr & 0xFF));                                            \
+    (void)SPI_RXBUF;                                                    \
     SPI_READ(data);                                                     \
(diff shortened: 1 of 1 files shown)


sumanpanchal:
The following function belongs to CC2520 driver returns wrong value of register. It always returns 128. 
static uint8_t
getreg(uint16_t regname)
{
  uint8_t reg;
  CC2520_READ_REG(regname, reg);
  return reg;
}
/*----------------------------------------------------------------------------*/

The basic problem is with CC2520_READ_REG macro definition. After calling SPI_RXBUF before SPI_READ, it returns correct value of register.

sumanpanchal:
Dear Maintainers,
Please provide your feedback on this PR.

nvt:
The fix appears to make getreg return the correct values, but I don't think that a one-line change is substantial enough for inclusion in the author list.

cmorty:
I wonder why such a severe bug wasn't caught earlier. Shouldn't it break basic functionality?

sumanpanchal:
@nvt Thanks for the feedback. I have removed my name from author list.

nvt:
:+1:

sumanpanchal:
@nvt Thanks.

PR #1189, opened 2015-08-12
Make TX power configurable for CC26XX

fengshenyue:
By default the TX power cannot be modified by the developers.


g-oikonomou:
Sorry, I do not like this. This  assume that the developer knows the position in the array that corresponds to the desired TX power and this PR is not adding any documentation about this.

I'd prefer setting it somewhere in `init()` by calling the respective function and passing a dBm value as argument. Then the standard contiki configuration convention should be used to make said dBm value configurable.

nvt:
Closing, since this is apparently replaced by #1190.

PR #1190, opened 2015-08-12
Make TX power configurable for CC26XX

fengshenyue:
By default the TX power cannot be modified.

### DIFF
--- a/cpu/cc26xx/dev/cc26xx-rf.c
+++ b/cpu/cc26xx/dev/cc26xx-rf.c
@@ -263,9 +263,8 @@ static const output_config_t output_power[] = {
 #define OUTPUT_POWER_MAX     (output_power[0].dbm)
 #define OUTPUT_POWER_UNKNOWN 0xFFFF
 
-/* Default TX Power - position in output_power[] */
-#define CC26XX_RF_TX_POWER 0
-const output_config_t *tx_power_current = &output_power[0];
+/* TX Power Setting - position in output_power[] */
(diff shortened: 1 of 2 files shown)


g-oikonomou:
My comment in #1189 has not been addressed by this revised pull. Please adjust accordingly if you still wish this to be considered. Rebase needed now.

Please push your changes in the same branch - do not start a new pull.

Cheers

tonnenpinguin:
Sorry for being late to this discussion.
@g-oikonomou what would be you suggestions on what has to be done to get this merged?
Would it be sufficient to add some documentation?
Would you add some defines or enums stating the actual dB value?

g-oikonomou:
Please see my comment in #1259, the configuration value has to be in dBm, not a magic number (position in array with TX power configurations).

The IEEE mode driver already has a mapping between dBm and configuration values to be used in `CMD_RADIO_SETUP`, have a look at at the `output_power[]` variable. The values in this array are taken directly from Smart RF Studio and for that reason we don't support all possible values in dBm. This is why `set_tx_power` sets TX power to 'at least' the dBm value requested by the user. For example, if the user requests -2, we will select 0. This has to be somehow documented.

In principle having this configurable is a desirable feature if done properly. Happy to re-open if you are gong to work on this.

g-oikonomou:
I meant #1189

tonnenpinguin:
I fully agree with your point of not using magic numbers.
I'll put it on my personal to do list and hopefully I'll get to it in the near future.
Thanks for your input!

PR #1192, opened 2015-08-12
changed tcpip static temporaries to stack vars

bkozak-scanimetrics:
A number of local variables in tcpip.c were declared with static storage for no good reason (they do not ever preserve state between function calls).

This patch removes static declaration from these variables.

### DIFF
--- a/core/net/ip/tcpip.c
+++ b/core/net/ip/tcpip.c
@@ -249,7 +249,7 @@ tcp_connect(const uip_ipaddr_t *ripaddr, uint16_t port, void *appstate)
 void
 tcp_unlisten(uint16_t port)
 {
-  static unsigned char i;
+  unsigned char i;
   struct listenport *l;
 
   l = s.listenports;
@@ -267,7 +267,7 @@ tcp_unlisten(uint16_t port)
 void
 tcp_listen(uint16_t port)
 {
-  static unsigned char i;
+  unsigned char i;
   struct listenport *l;
 
   l = s.listenports;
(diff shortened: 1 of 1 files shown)


oliverschmidt:
> ... for no good reason ...

The good reason was that the 6502 compiler creates more effective code for static variables. So once one used them for all non-recursive functions.

bkozak-scanimetrics:
> The good reason was that the 6502 compiler creates more effective code for static variables. So once one used them for all non-recursive functions. 

That seems very odd but you're right, the 6502 actually uses more memory with my changes.

Of course, gcc is using less memory with my changes.

What are your thoughts on how to resolve this conflict?

oliverschmidt:
That's a misunderstanding. I did NOT want to question your change. Therefore I wrote "...was..." and "...once...". I just wanted to oppose to the term "...for no good reason...". There has been a discussion on a different pull request and it's clear that "ancient" static variables are to be removed.

> the 6502 actually uses more memory with my changes

Thanks for taking the time to actually check :-) The Travis barriers btw. only build the 6502 targets but don't run them. Therefore they won't fail unless a change makes things a lot bigger - which already happened...

bkozak-scanimetrics:
> I just wanted to oppose to the term "...for no good reason...". There has been a discussion on a different pull request and it's clear that "ancient" static variables are to be removed.

I see. Thanks for clearing that up for me.

simonduq:
:+1:

bkozak-scanimetrics:
Any reasons why this shouldn't be merged now?

There is someone on the mailing list asking if it would be appropriate to drop the static designation from some other local variables in a different part of the code. If a decision is made to not merge this PR then this would also be relevant to them.

Thanks.

simonduq:
Waiting for another approval. Let me set a timeout

alignan:
I thought this was already merged as it was commented by @simonduq and @oliverschmidt, seems like a straightforward fix to me

alignan:
:+1:

PR #1193, opened 2015-08-12
Handle possible reetrancy in contikimac protothread

bkozak-scanimetrics:
On platforms which allow nested interrupts (I believe all of the cortex m3 based platforms fall into this category) it may be possible for the ```powercycle()``` function to be re-entered in between the ```rtimer_set```s and ``PT_YIELD``s. This will wreck the protothread's state. 

This PR fixes this problem and does so by adding the PT_YIELD_ON macro which handles statements which cause possible re-entry to the calling protothread. PT_YIELD_ON will also have utility in user app code and may find use in other parts of core Contiki.

### DIFF
--- a/core/net/mac/contikimac/contikimac.c
+++ b/core/net/mac/contikimac/contikimac.c
@@ -388,8 +388,8 @@ powercycle(struct rtimer *t, void *ptr)
         }
         powercycle_turn_radio_off();
       }
-      schedule_powercycle_fixed(t, RTIMER_NOW() + CCA_SLEEP_TIME);
-      PT_YIELD(&pt);
+      PT_YIELD_ON(&pt,
+                  schedule_powercycle_fixed(t, RTIMER_NOW() + CCA_SLEEP_TIME));
     }
 
     if(packet_seen) {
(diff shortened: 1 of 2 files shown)


g-oikonomou:
I'm not sure I understand the motivation here. An rtimer interrupt cannot preempt an rtimer interrupt. What have I missed?

bkozak-scanimetrics:
> I'm not sure I understand the motivation here. An rtimer interrupt cannot preempt an rtimer interrupt. What have I missed?

Can't it? Maybe I'm mistaken but I think interrupts can self preempt in Cortex m3 if you clear the pending bit first.

g-oikonomou:
CC26xx TRM (sec. 4.1.5):

> When the processor is executing an exception handler, the exception handler is preempted if a higher- priority exception occurs. If an exception occurs with the same priority as the exception being handled, the handler is not preempted, irrespective of the exception number. However, the status of the new interrupt changes to pending.

bkozak-scanimetrics:
Well I'll be damned

bkozak-scanimetrics:
Closing because I don't believe there are any platforms where this will occur either.

PR #1194, opened 2015-08-14
Fix Rime attribute serialization in chameleon-bitopts for big endian systems

atiselsts:
When testing Contiki on a big endian platform we discovered that Rime protocols do not work correctly even both communicating nodes are based on a big endian platform. This patch fixes this problem allowing the Rime attributes to be serialized correctly.

This patch does just what it promises, in particular it does not address the interoperability between little-endian and big-endian nodes using Rime.


simonduq:
:+1:

PR #1198, opened 2015-08-18
new tunslip6 features

simonduq:
This PR adds the following (optional) features:
* set the interface MTU via -M
* enable software flow (XON/XOFF) control via -X
* support for baudrates up to 1000000 mbps


laurentderu:
Sadly this won't compile on MacOS, BSD and some versions of Linux as the macros added do not always exists. I have a similar patch for the slip-radio with all the existing baud rate macros redefined if they do not exists, I think it could be reused here.

simonduq:
I'm interested, please share!
Do you think `IXON`, `IXOFF`, `IXANY` might also be a problem, or are you referring to the baudrates only?

laurentderu:
IXON, IXOFF, IXANY are fine, they are pretty much standard (and I just checked, it compiles on mac)

laurentderu:
I created https://github.com/contiki-os/contiki/pull/1205 for the baud rate definition for native-border-router.

atiselsts:
`serialdump` (and possibly other tools) is going to have the same problem. Maybe better to create a single header file and `#include` it in all C apps that require these definitions?

simonduq:
Absolutely, but then there is also the big switch which is going to be common in all tools. We could have this as a function. How about a static function declared in the .h? (the tools are simple .c files, let's keep their compilation simple)

laurentderu:
Personally I don't like the idea of code in a .h file. However, what we could do is have a tools-utils.c which is compiled trivially in the Makefile like this :

<pre>
tunslip6: tools-utils.c tunslip6.c
</pre>

simonduq:
sounds good, I'm on it :)

simonduq:
I added the module, used by tunslip6 only in this PR

laurentderu:
Compile nicely on Mac :+1: What are you using to test XON/XOFF ?

simonduq:
I've tested XON/XOFF on the JN516x platform (intend to PR it in the coming weeks). We have boards with HW flow control, other without -- for the latter we decided to implement SW flow control. Tested and stable :)

simonduq:
Sorry the boards all support HW flow control, it was a particular gateway that didn't. Will think about contributing a platform-independent XON-XOFF on the Contiki side.

simonduq:
I'm not expecting much more activity here, will merge in a month so we can also get https://github.com/contiki-os/contiki/pull/1205 in next

simonduq:
merging after timeout

PR #1199, opened 2015-08-18
ContikiMAC: add CONTIKIMAC_RTIMER_GUARD_TIME #define

atiselsts:
CONTIKIMAC_RTIMER_GUARD_TIME is the minimum amount of rtimer ticks between the current time and the future time when the ContikiMAC rtimer is scheduled.

It is necessary to avoid accidentally scheduling a rtimer in the past  on platforms with fast rtimer ticks (e.g. 16 MHz).

### DIFF
--- a/core/net/mac/contikimac/contikimac.c
+++ b/core/net/mac/contikimac/contikimac.c
@@ -296,14 +296,18 @@ static void
 schedule_powercycle(struct rtimer *t, rtimer_clock_t time)
 {
   int r;
+  rtimer_clock_t now;
 
   if(contikimac_is_on) {
 
-    if(RTIMER_CLOCK_LT(RTIMER_TIME(t) + time, RTIMER_NOW() + 2)) {
-      time = RTIMER_NOW() - RTIMER_TIME(t) + 2;
+    time += RTIMER_TIME(t);
+    now = RTIMER_NOW();
+    if(RTIMER_CLOCK_LT(time, now + RTIMER_GUARD_TIME)) {
+      time = now + RTIMER_GUARD_TIME;
     }
 
-    r = rtimer_set(t, RTIMER_TIME(t) + time, 1, powercycle_wrapper, NULL);
+    r = rtimer_set(t, time, 1, powercycle_wrapper, NULL);
+
     if(r != RTIMER_OK) {
       PRINTF("schedule_powercycle: could not set rtimer\n");
     }
@@ -314,11 +318,13 @@ static void
 schedule_powercycle_fixed(struct rtimer *t, rtimer_clock_t fixed_time)
 {
   int r;
+  rtimer_clock_t now;
 
   if(contikimac_is_on) {
 
-    if(RTIMER_CLOCK_LT(fixed_time, RTIMER_NOW() + 1)) {
-      fixed_time = RTIMER_NOW() + 1;
+    now = RTIMER_NOW();
+    if(RTIMER_CLOCK_LT(fixed_time, now + RTIMER_GUARD_TIME)) {
+      fixed_time = now + RTIMER_GUARD_TIME;
     }
 
     r = rtimer_set(t, fixed_time, 1, powercycle_wrapper, NULL);
--- a/core/sys/rtimer.h
+++ b/core/sys/rtimer.h
@@ -151,6 +151,16 @@ void rtimer_arch_schedule(rtimer_clock_t t);
 
 #define RTIMER_SECOND RTIMER_ARCH_SECOND
 
+/* RTIMER_GUARD_TIME is the minimum amount of rtimer ticks between
+   the current time and the future time when a rtimer is scheduled.
+   Necessary to avoid accidentally scheduling a rtimer in the past
+   on platforms with fast rtimer ticks. Should be >= 2. */
+#ifdef RTIMER_CONF_GUARD_TIME
+#define RTIMER_GUARD_TIME RTIMER_CONF_GUARD_TIME
+#else /* RTIMER_CONF_GUARD_TIME */
+#define RTIMER_GUARD_TIME (RTIMER_ARCH_SECOND >> 14)
+#endif /* RTIMER_CONF_GUARD_TIME */
+
 #endif /* RTIMER_H_ */
 
 /** @} */


bkozak-scanimetrics:
In my opinion this is an important problem to address but I would add some things to your fix to make it more robust (this is just my opinion please do point out where you disagree).

First, I would move the RTIMER_LT() check into a public function in rtimer and then have Contikimac call this function. This will allow all code which calls the rtimer to use RTIMER_GUARD_TIME and not just Contikimac.

Second, I would mask interrupts before the RTIMER_NOW() and unmask only after the scheduling is complete. This would allow you to select a very conservative RTIMER_GUARD_TIME and be fully confident that it's sufficient (otherwise the schedule time can become stale while the CPU is busy servicing interrupts).

Third, I think that you should add a way of allowing RTIMER_GUARD_TIME to be set on a per platform basis similar to how other configuration constants in Contiki are set with a macro like <MODULE>_CONF_<NAME>.  In some platforms it might make sense to override your default settings.

Finally, you should never allow CONTIKIMAC_RTIMER_GUARD_TIME to be 1 because you never have any way of knowing just how close you are to the next timer tick.


Thoughts?

g-oikonomou:
Just shooting from the hip here: I've not given this much thought, but should this not be implemented by each platform's rtimer_arch module? (i.e. make sure to not schedule a trigger in the past etc). I find it a little moot for core components to try and second-guess the requirements of various devices.

atiselsts:
I moved the guard define to `core/sys/rtimer.h` and added the comment that it should always be >= 2.

Masking interrupts is a complex issue, there is no standard API for that and if the setting is done from a rtimer callback its not required anyway unless nested interrupts are enabled.

Checking whether the time is in past is also a complex issue. On Tmote Sky the used could easily add the timer ~2 seconds in the future. An automated check that would not allow to do that (counting it as "past time") is not a good idea.

bkozak-scanimetrics:
> Masking interrupts is a complex issue, there is no standard API for that. And if rtimer_set is called from a rtimer callback it's not required anyway unless nested interrupts are enabled.

Seems a good enough justification to me. Too complex a change for this little patch (although I would like to eventually see this case handled in the future).

> Checking whether the time is in past is also a complex issue. On Tmote Sky the user could easily add the timer ~2 seconds in the future. An automated check that would not allow to do that (counting it as "past time") is not a good idea.

Yes I agree that this should not be done for every single call to rtimer_set(). I still think it would be appropriate to add a new function to rtimer.c which implemented the check against ```RTIMER_GUARD_TIME``` however.

> I moved the guard define to core/sys/rtimer.h and added the comment that it should always be >= 2

I think this is important enough to enforce with the pre-processor. Even if you want to allow ```RTIMER_CONF_GUARD_TIME``` to be set < 2 (I don't see why this would be useful though) you should make sure that your default value for ```RTIMER_GUARD_TIME``` can never be < 2.

atiselsts:
I'd be ok to move this `#define` to `rtimer.c`, and create a function called let's say `rtimer_set_guarded` or `rtimer_set_safe`, with the same arguments as `rtimer_set`. Then simply call this function from ContikiMAC. That would additionally simplify the code of ContikiMAC and free it from some low-level details.

About enforcing the minimal value, do you think this code is OK? Issuing a `#warning` or `#message` would be nice, but hard to do in a compiler-independent way.

    #if RTIMER_GUARD_TIME < 2
    #undef RTIMER_GUARD_TIME
    #define RTIMER_GUARD_TIME 2
    #endif /* RTIMER_GUARD_TIME < 2 */

@g-oikonomou - enforcing this check in platform-specific files would likely result in different behavior of core components or applications on different platforms. E.g. scheduling in 1 tick in past on Sky or Z1 (with no check enabled) would make the rtimer fire after 2 seconds, while scheduling it in past on a 32-bit rtimer platform (with the check enabled) would make it fire immediately. Such an inconsistency is not desirable from user's point of view IMHO.

g-oikonomou:
> Such an inconsistency is not desirable from user's point of view IMHO.

You are very right, it is not. But I'm not proposing an inconsistent approach here. Again IMHO: Setting an rtimer in the past should never happen. If it does get requested for whatever reason, then it should fire immediately. If the CPU code schedules it at any point in the distant future, then said CPU code is broken and needs improved.

> E.g. scheduling in 1 tick in past on Sky or Z1 (with no check enabled) would make the rtimer fire after 2 seconds

So, IMHO, it's the respective MSP430 code that needs to be improved here

simonduq:
@g-oikonomou I agree `rtimer` should prevent setting a timer in the past, but by "fire immediately" do you mean run the callback function directly from `rtimer_set`, or schedule a few ticks in the future (e.g. `RTIMER_GUARD_TIME`)? The former is scary because it results in the callback not being executed consistently from interrupt.

Another option is to have `rtimer_set` fail when asked to schedule in the past, and return `RTIMER_ERR_TIME`. This allows the caller to react accordingly (e.g. skip `PT_YIELD`).

g-oikonomou:
> but by "fire immediately" do you mean run the callback function directly from `rtimer_set`, or schedule a few ticks in the future

Agree, this shouldn't happen before `_set` returns. Ideally there would be a CPU-specific guard inside the `arch_set` implementation that schedules in something like "no less than X ticks in the future". For example, the CC2538 user guide is says that when scheduling a new compare value for the Sleep Timer, the new value must be at least 5 ticks more than the current counter. It's just that I'm seeing too many arch-specific considerations here to try to resolve this with a platform-independent approach.

simonduq:
How about, in `rtimer_set`, scheduling the next wakeup at `max(time, now + RTIMER_GUARD_TIME)`?

g-oikonomou:
I fear we may end up with many platforms simply using a default (and most likely erroneous) `RTIMER_GUARD_TIME`.

At the end of the day though, I was thinking aloud there: I don't feel very strongly about any of that.

simonduq:
I'd say that's not a perfect solution, but definitely an improvement over ContikiMAC's current way of adding `+1` and `+2` to rtimer clocks :)

bkozak-scanimetrics:
> do you think this code is OK 
> ...
> Issuing a #warning or #message would be nice but hard to do in a compiler-independent way.
```C
#if RTIMER_GUARD_TIME < 2
#undef RTIMER_GUARD_TIME
#define RTIMER_GUARD_TIME 2
#endif /* RTIMER_GUARD_TIME < 2 */
```

Agree about the warning; too bad we don't have a portable way to do that. Otherwise this is fine to me but, if you don't mind me nitpicking (and maybe you'd even have done this yourself before you made the actual change), probably *slightly* better to `#define` to 2**U**.


Of course, given all the discussion going on here, it seems we may end up with something quite different from your original `RTIMER_GUARD_TIME` based solution.

atiselsts:
Since there is no single opinion about the best way how to do it... I add my own.

What I would like to do is to avoid breaking existing code, in particular currently valid code on Sky that schedules the rtimer exactly 2 seconds in future. Why would not an user be allowed to do that? If the max bound is 1 seconds it means the application may needs twice as many wakeups in the worst case -> not as energy efficient.

On the other hand, platforms such as CC2538 with 32-bit rtimer could easily enforce the 5 tick minimal value without breaking any applications (assuming nobody needs to schedule a rtimer more than 65535 seconds in the future).

So my current proposal is to:
* make a new macro `RTIMER_CLOCK_IS_PAST(t)` always `#defined` as `0` on msp430 platforms and `#defined` as `RTIMER_CLOCK_LT(t, now + guard_time)` on 32-bit rtimer platforms.
* fix the wakeup time in `rtimer_set` if and only if `RTIMER_CLOCK_IS_PAST(t)` evaluates to true
* additionally, introduce `rtimer_set_safe` to enforce this check even on if explictily requested by the user. The `rtimer_set_safe` would schedule the callback immediately if expired and therefore allow to simplify ContikiMAC code without changing its semantics.
* `#define` the guard timer to 2 by default, to 6 on CC2538, and to some larger value on the 16 Mhz rtimer platform.

simonduq:
Sounds complicated, and the semantics of `RTIMER_CLOCK_IS_PAST` is confusing (it's not past, it's only too close to be scheduled safely).

I'm actually happy with the PR in its current form, it is simple, fixes a real problem, and clarifies the ContikiMAC code. The rtimer design can still be improved in another PR.
:+1:

nvt:
@atiselsts I think that unacceptable configurations should result in an <code>#error</code> rather than a warning or a silent reconfiguration. This is also supported by C99.

simonduq:
@nvt are you opposed to this PR in its current form or? (the configuration thing was comment-only, not in the code). The PR is ok for me.

simonduq:
This now has a conflict, can you please rebase?

atiselsts:
@simonduq : rebased

simonduq:
Thanks!

PR #1202, opened 2015-08-19
Fix Travis icon in README.md

simonduq:
This PR updates `README.md` to show the build status of the current master rather than latest build status (can be any pull request, i.e. can show "fail" while the master does pass the test). The link also points to Contiki branches rather than latest build.


alignan:
+1

PR #1205, opened 2015-08-20
Add more baudrates to native-border-router and add definition for McOS-X for native-border-router

laurentderu:
Some baudrate (especially the higher ones) are not defined on some Linux platforms (raspbian, Angstrom, ...). Also on MacOS-X some definition are lacking.

This PR add the missing definitions for MacOS-X and add a guard for each baud rate.


simonduq:
https://github.com/contiki-os/contiki/pull/1198 is now merged, you probably want to use the new `tools-utils` module instead now

alignan:
Where are we on this one?

PR #1207, opened 2015-08-20
ND6 alignment bugfix

simonduq:
This PR ensures the second parameter of `uip_ds6_nbr_add` is word-aligned, which is required as the parameter is directly accessed as `uip_lladdr_t*` (not as a byte array).

Bug report from Deepak Sudhakar

### DIFF
--- a/core/net/ipv6/uip-nd6.c
+++ b/core/net/ipv6/uip-nd6.c
@@ -130,6 +130,16 @@ static uip_ipaddr_t ipaddr;
 static uip_ds6_prefix_t *prefix; /**  Pointer to a prefix list entry */
 #endif
 
+#if UIP_ND6_SEND_NA || UIP_ND6_SEND_RA || !UIP_CONF_ROUTER
+/*------------------------------------------------------------------*/
+/* Copy link-layer address from LLAO option to a word-aligned uip_lladdr_t */
+static void
+extract_lladdr_aligned(uip_lladdr_t *dest) {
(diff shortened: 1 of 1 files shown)


laurentderu:
:+1: 

I found a similar issue in ns_input() when UIP_LLH_LEN is != 0 :

<pre>
uip_ipaddr_copy(&UIP_IP_BUF->srcipaddr, &UIP_ND6_NS_BUF->tgtipaddr);
</pre>

makes a hardware fault on CC2538. I'll make a similar patch.

nvt:
Maybe we could have a macro or a static function that does the copy here, and move the repeated comment to the definition of that macro or function.

simonduq:
Looking closer at the code, I found this https://github.com/contiki-os/contiki/issues/1277

An added function or macro might create confusion: it would copy the address to a struct lladdr, which is needed only in particular cases (here, because it is used as parameter to function expecting a `uip_lladdr_t *`). There are many places in the file where the lladdr is accessed directly e.g. for a `memcmp`, and the reader might wonder why we sometimes do direct access, sometimes do a copy. So it will be hard to eliminate the duplicate comments.

nvt:
I thought that the naming of the macro (e.g., to extract_lladdr_aligned()) would help to improve the understanding, but the current solution is also OK for me.

simonduq:
Point taken, I'll try that!

simonduq:
done :)

nvt:
:+1:

PR #1212, opened 2015-08-23
Add support for the CC13xx

### LINKED ISSUE #1191: CC2650 - Several Race Conditions in Radio Driver
The CC2650 radio driver has multiple potential race conditions associated with the ```cmd_immediate_buf```. This variable is used both from interrupt context (see ```powercycle()``` in contikimac.c) and from the main thread without first protecting by masking interrupts. Most public  radio driver functions will result in an unprotected use of cmd_immediate_buf.

jonnteolsson:
Recently Texas Instruments announced our new SimpleLink ultra-low power wireless MCU platform CC13xx. This pull adds support for the CC13xx platform to Contiki, by adding it to the existing CC26xx port.

The CC13xx family consist of ultra-low power ARM Cortex-M3 based sub-GHz RF devices, first device out is CC1310 with support for IEEE802.15.4g FSK.

In short, this pull;

* Adds support for the CC1310. This includes the addition of a IEEE802.15.4g-compliant sub-ghz RF driver that has been tested in the 863MHz band. For the 863MHz band at 50kbps, the driver is ContikiMAC-ready off-the-shelf. Settings have been added for other bands (e.g. 470MHz, 915MHz), but those are not thoroughly tested.
* Restructures RF-related code to improve clarity: IEEE, sub-Ghz and BLE functionalities are now implemented within their own modules and there is an additional module that takes care of common RF functionality and the interface between the CM3 and the Radio Core.
* Adds support for tickless mode, applicable on all supported chips: The chip will now only come out of deep sleep just before an etimer is about to expire, instead of waking up on every tick simply to update a software counter.
* Adds a new example (very-sleepy-demo) to demonstrate how to achieve even longer battery lifetime with very sleepy nodes.
* Updates CC26xxware to the latest version and adds the newly-released CC13xxware as a submodule.
* Re-works the cpu/platform code in order to prevent code duplication. 
* Fixes #1191
* Tidies up documentation
* Improves the SPI and external flash driver for the Sensortag

This pull changes how we build for the various board/chip combos, so make sure to have a look at the updated READMEs

For more information on the platform, please see;
* http://processors.wiki.ti.com/index.php/Contiki-6LOWPAN
* http://www.ti.com/product/CC1310

Enjoy!

Jonas Olsson 
Senior Applications Engineer 
Texas Instruments, Low Power RF


adamdunkels:
Very cool! Looking forward to testing this - very nice to see tickless clock mode for the cc26xx/cc13xx. We've tested the CC26xx port with the Thingsquare platform and while we're able to go to > 0.1% RF duty cycle when in dead leaf mode, the CPU never falls below 4.4%. This should make the CPU go down too.

I just looked very briefly, but it seems like the cc13xx defaults to using the `prop` mode of the radio and not the `ieee` mode. Does this mean that the default configuration is not IEEE 802.15.4g compliant?

jonnteolsson:
yes it will help reducing CPU activity :)

The CC13xx prop API/mode is IEEE802.15.4g compliant, the IEEE API/mode is 2.4GHz only. A bit confusing but that's how the device APIs are... :)

The default prop settings are fully IEEE802.15.4g compliant, with 50kbps data rate, 25kHz deviation, 32 bit CRC, all using IEEE802.15.g frame (PHDR).

jonnteolsson:
The CC13xx prop driver (IEEE802.15.4g nullRDC) has also been verified with the Weptech Gateway (https://www.weptech.de/products/id-6lowpan-ipv6-iot/easy-wep-ipv6-gateway.html) which uses the newly merged CC1200 driver.

g-oikonomou:
ContikiMAC-ready sub-ghz .15.4g mode is super sweet and so is the interoperability with CC1200. Some of the collateral features are also great.

Thoroughly tested!

:+1:

arurke:
I am curious about the rationale behind the CONTIKIMAC_CONF_CCA_SLEEP_TIME. It seems to be disproportionately large compared to e.g. the check time and the values used on 250 kbps platforms. What am I missing?

g-oikonomou:
@arurke  Hi Andreas that was briefly discussed under #1878 too (see [here](https://github.com/contiki-os/contiki/issues/1878#issuecomment-267609228)). It was mainly trial and error, one that probably causes problems. We should be adjust all those moving forward.

arurke:
Ah, yes I remember it now when you point it out. I was doing some calc. on capacity and noticed the CCA procedure is quite dominant, especially since it by default is done 6 times before TX.

PR #1213, opened 2015-08-25
End-to-End DAO and DAO ACK handling for Contiki RPL.

joakimeriksson:
This PR adds full handling of end-to-end DAO and DAO ACK for
Contiki RPL and makes Contiki RPL significantly more stable and
scalable. The PR includes the changes needed to how neighbors and
routes are handled to enable DAO ACK and DAO NACK. The baseline
of this work have been tested in small and large networks and is
work done at Yanzi Networks.

- Full handling of DAO and DAO_ACK with end-to-end ACK all the
  way up to the RPL ROOT to ensure that the route is installed
  all the way (including retransmission of DAO when no ACK is
  received).
- Neighbor policy that decides when it is ok to add a new
  neighbor (to avoid throwing out neighbors that are useful).
- Configuration option on the routing module so that it is
  possible to select either remove least recently used route or
  only remove when route entry times out (needed for proper
  handling of DAO_ACK/NACK).
- Added automatical renewal of DAO routes from the registering node
  rather than forcing renewal at each broadcast DIO.
- Change the way DTSN is increased so that it is only increased
  when local repair occurs.
- Updated RPL configuration and other related configuration to
  make use of the new features (30 minutes routing lifetime for
  GC of routes over time).

The new default behavior of the Contiki RPL is to make use of
DAO_ACK/NACK for more stable and scalable networks.

There is also a few other needed changes for things to work like
making some erbium tests a bit smaller and increasing number of
routes in the RPL ROOT of the large network test for RPL.

The PR represents work done by the authors (Joakim, Niclas, Oriol)
but the design is a team effort - contributors include Lars and
Flag at Yanzi Networks.

### DIFF
--- a/core/net/ipv6/uip-ds6-nbr.c
+++ b/core/net/ipv6/uip-ds6-nbr.c
@@ -83,7 +83,7 @@ uip_ds6_nbr_add(const uip_ipaddr_t *ipaddr, const uip_lladdr_t *lladdr,
                 void *data)
 {
   uip_ds6_nbr_t *nbr = nbr_table_add_lladdr(ds6_neighbors, (linkaddr_t*)lladdr
-                                            ,reason, data);
+                                            , reason, data);
   if(nbr) {
     uip_ipaddr_copy(&nbr->ipaddr, ipaddr);
 #if UIP_ND6_SEND_NA || UIP_ND6_SEND_RA || !UIP_CONF_ROUTER
--- a/core/net/ipv6/uip-ds6-nbr.c
+++ b/core/net/ipv6/uip-ds6-nbr.c
@@ -79,9 +79,11 @@ uip_ds6_neighbors_init(void)
 /*---------------------------------------------------------------------------*/
 uip_ds6_nbr_t *
 uip_ds6_nbr_add(const uip_ipaddr_t *ipaddr, const uip_lladdr_t *lladdr,
-                uint8_t isrouter, uint8_t state)
+                uint8_t isrouter, uint8_t state, nbr_table_reason_t reason,
+                void *data)
 {
-  uip_ds6_nbr_t *nbr = nbr_table_add_lladdr(ds6_neighbors, (linkaddr_t*)lladdr);
+  uip_ds6_nbr_t *nbr = nbr_table_add_lladdr(ds6_neighbors, (linkaddr_t*)lladdr
+                                            ,reason, data);
   if(nbr) {
     uip_ipaddr_copy(&nbr->ipaddr, ipaddr);
 #if UIP_ND6_SEND_NA || UIP_ND6_SEND_RA || !UIP_CONF_ROUTER
--- a/core/net/ipv6/uip-ds6-nbr.c
+++ b/core/net/ipv6/uip-ds6-nbr.c
@@ -115,7 +115,7 @@ uip_ds6_nbr_add(const uip_ipaddr_t *ipaddr, const uip_lladdr_t *lladdr,
 }
 
 /*---------------------------------------------------------------------------*/
-void
+int
 uip_ds6_nbr_rm(uip_ds6_nbr_t *nbr)
 {
   if(nbr != NULL) {
@@ -123,9 +123,9 @@ uip_ds6_nbr_rm(uip_ds6_nbr_t *nbr)
     uip_packetqueue_free(&nbr->packethandle);
 #endif /* UIP_CONF_IPV6_QUEUE_PKT */
     NEIGHBOR_STATE_CHANGED(nbr);
-    nbr_table_remove(ds6_neighbors, nbr);
+    return nbr_table_remove(ds6_neighbors, nbr);
   }
-  return;
+  return 0;
 }
 
 /*---------------------------------------------------------------------------*/
--- a/core/net/ipv6/uip-ds6-route.c
+++ b/core/net/ipv6/uip-ds6-route.c
@@ -348,7 +348,8 @@ uip_ds6_route_add(uip_ipaddr_t *ipaddr, uint8_t length,
          initialize this pointer with the list of routing entries that
          are attached to this neighbor. */
       routes = nbr_table_add_lladdr(nbr_routes,
-                                    (linkaddr_t *)nexthop_lladdr);
+                                    (linkaddr_t *)nexthop_lladdr,
+                                    NBR_TABLE_REASON_ROUTE, NULL);
       if(routes == NULL) {
         /* This should not happen, as we explicitly deallocated one
            route table entry above. */
--- a/core/net/ipv6/uip-ds6-route.c
+++ b/core/net/ipv6/uip-ds6-route.c
@@ -207,6 +207,12 @@ uip_ds6_route_next(uip_ds6_route_t *r)
 }
 /*---------------------------------------------------------------------------*/
 int
+uip_ds6_route_is_nexthop(const uip_lladdr_t *lladdr)
+{
+  return nbr_table_get_from_lladdr(nbr_routes, (linkaddr_t *)lladdr) != NULL;
+}
+/*---------------------------------------------------------------------------*/
+int
 uip_ds6_route_num_routes(void)
 {
   return num_routes;
@@ -307,11 +313,16 @@ uip_ds6_route_add(uip_ipaddr_t *ipaddr, uint8_t length,
        least recently used one we have. */
 
     if(uip_ds6_route_num_routes() == UIP_DS6_ROUTE_NB) {
+      uip_ds6_route_t *oldest;
+      oldest = NULL;
+#if UIP_DS6_ROUTE_REMOVE_LEAST_RECENTLY_USED
       /* Removing the oldest route entry from the route table. The
          least recently used route is the first route on the list. */
-      uip_ds6_route_t *oldest;
-
-      oldest = list_tail(routelist); /* uip_ds6_route_head(); */
+      oldest = list_tail(routelist);
+#endif
+      if(oldest == NULL) {
+        return NULL;
+      }
       PRINTF("uip_ds6_route_add: dropping route to ");
       PRINT6ADDR(&oldest->ipaddr);
       PRINTF("\n");
--- a/core/net/ipv6/uip-ds6-route.c
+++ b/core/net/ipv6/uip-ds6-route.c
@@ -207,8 +207,15 @@ uip_ds6_route_next(uip_ds6_route_t *r)
 }
 /*---------------------------------------------------------------------------*/
 int
-uip_ds6_route_is_nexthop(const uip_lladdr_t *lladdr)
+uip_ds6_route_is_nexthop(const uip_ipaddr_t *ipaddr)
 {
+  const uip_lladdr_t *lladdr;
+  lladdr = uip_ds6_nbr_lladdr_from_ipaddr(ipaddr);
+
+  if(lladdr == NULL) {
+    return 0;
+  }
+
   return nbr_table_get_from_lladdr(nbr_routes, (linkaddr_t *)lladdr) != NULL;
 }
 /*---------------------------------------------------------------------------*/
--- a/core/net/ipv6/uip-ds6-route.h
+++ b/core/net/ipv6/uip-ds6-route.h
@@ -106,6 +106,9 @@ void uip_ds6_notification_rm(struct uip_ds6_notification *n);
 #define RPL_ROUTE_SET_NOPATH_RECEIVED(route) do {                       \
     (route)->state.state_flags |= RPL_ROUTE_ENTRY_NOPATH_RECEIVED;      \
   } while(0)
+#define RPL_ROUTE_CLEAR_NOPATH_RECEIVED(route) do {                       \
+    (route)->state.state_flags &= ~RPL_ROUTE_ENTRY_NOPATH_RECEIVED;      \
+  } while(0)
 
 #define RPL_ROUTE_IS_DAO_PENDING(route)                                 \
   ((route->state.state_flags & RPL_ROUTE_ENTRY_DAO_PENDING) != 0)
--- a/core/net/ipv6/uip-ds6-route.h
+++ b/core/net/ipv6/uip-ds6-route.h
@@ -197,7 +197,7 @@ uip_ipaddr_t *uip_ds6_route_nexthop(uip_ds6_route_t *);
 int uip_ds6_route_num_routes(void);
 uip_ds6_route_t *uip_ds6_route_head(void);
 uip_ds6_route_t *uip_ds6_route_next(uip_ds6_route_t *);
-
+int uip_ds6_route_is_nexthop(const uip_lladdr_t *lladdr);
 /** @} */
 
 #endif /* UIP_DS6_ROUTE_H */
--- a/core/net/ipv6/uip-ds6-route.h
+++ b/core/net/ipv6/uip-ds6-route.h
@@ -124,6 +124,9 @@ void uip_ds6_notification_rm(struct uip_ds6_notification *n);
 #define RPL_ROUTE_SET_DAO_NACKED(route) do {                            \
     (route)->state.state_flags |= RPL_ROUTE_ENTRY_DAO_NACK;             \
   } while(0)
+#define RPL_ROUTE_CLEAR_DAO_NACKED(route) do {                          \
+    (route)->state.state_flags &= ~RPL_ROUTE_ENTRY_DAO_NACK;            \
+  } while(0)
 
 #define RPL_ROUTE_CLEAR_DAO(route) do {                                 \
     (route)->state.state_flags &= ~(RPL_ROUTE_ENTRY_DAO_NACK|RPL_ROUTE_ENTRY_DAO_PENDING); \
@@ -200,7 +203,7 @@ uip_ipaddr_t *uip_ds6_route_nexthop(uip_ds6_route_t *);
 int uip_ds6_route_num_routes(void);
 uip_ds6_route_t *uip_ds6_route_head(void);
 uip_ds6_route_t *uip_ds6_route_next(uip_ds6_route_t *);
-int uip_ds6_route_is_nexthop(const uip_lladdr_t *lladdr);
+int uip_ds6_route_is_nexthop(const uip_ipaddr_t *ipaddr);
 /** @} */
 
 #endif /* UIP_DS6_ROUTE_H */
--- a/core/net/ipv6/uip-ds6-route.h
+++ b/core/net/ipv6/uip-ds6-route.h
@@ -97,11 +97,42 @@ void uip_ds6_notification_rm(struct uip_ds6_notification *n);
 #ifndef UIP_DS6_ROUTE_STATE_TYPE
 #define UIP_DS6_ROUTE_STATE_TYPE rpl_route_entry_t
 /* Needed for the extended route entry state when using ContikiRPL */
+#define RPL_ROUTE_ENTRY_NOPATH_RECEIVED   0x01
+#define RPL_ROUTE_ENTRY_DAO_PENDING       0x02
+#define RPL_ROUTE_ENTRY_DAO_NACK          0x04
+
+#define RPL_ROUTE_IS_NOPATH_RECEIVED(route)                             \
+  (((route)->state.state_flags & RPL_ROUTE_ENTRY_NOPATH_RECEIVED) != 0)
+#define RPL_ROUTE_SET_NOPATH_RECEIVED(route) do {                       \
+    (route)->state.state_flags |= RPL_ROUTE_ENTRY_NOPATH_RECEIVED;      \
+  } while(0)
+
+#define RPL_ROUTE_IS_DAO_PENDING(route)                                 \
+  ((route->state.state_flags & RPL_ROUTE_ENTRY_DAO_PENDING) != 0)
+#define RPL_ROUTE_SET_DAO_PENDING(route) do {                           \
+    (route)->state.state_flags |= RPL_ROUTE_ENTRY_DAO_PENDING;          \
+  } while(0)
+#define RPL_ROUTE_CLEAR_DAO_PENDING(route) do {                         \
+    (route)->state.state_flags &= ~RPL_ROUTE_ENTRY_DAO_PENDING;         \
+  } while(0)
+
+#define RPL_ROUTE_IS_DAO_NACKED(route)                                  \
+  ((route->state.state_flags & RPL_ROUTE_ENTRY_DAO_NACK) != 0)
+#define RPL_ROUTE_SET_DAO_NACKED(route) do {                            \
+    (route)->state.state_flags |= RPL_ROUTE_ENTRY_DAO_NACK;             \
+  } while(0)
+
+#define RPL_ROUTE_CLEAR_DAO(route) do {                                 \
+    (route)->state.state_flags &= ~(RPL_ROUTE_ENTRY_DAO_NACK|RPL_ROUTE_ENTRY_DAO_PENDING); \
+  } while(0)
+
+struct rpl_dag;
 typedef struct rpl_route_entry {
   uint32_t lifetime;
-  void *dag;
-  uint8_t learned_from;
-  uint8_t nopath_received;
+  struct rpl_dag *dag;
+  uint8_t dao_seqno_out;
+  uint8_t dao_seqno_in;
+  uint8_t state_flags;
 } rpl_route_entry_t;
 #endif /* UIP_DS6_ROUTE_STATE_TYPE */
 
--- a/core/net/nbr-table.c
+++ b/core/net/nbr-table.c
@@ -47,10 +47,19 @@
 static void handle_periodic_timer(void *ptr);
 static struct ctimer periodic_timer;
 static uint8_t initialized = 0;
+static void print_table();
 #define PRINTF(...) printf(__VA_ARGS__)
 #else
 #define PRINTF(...)
 #endif
+
+/* This is the callback function that will be called when there is a
+ *  nbr-policy active
+ **/
+#ifdef NBR_TABLE_FIND_REMOVABLE
+const linkaddr_t *NBR_TABLE_FIND_REMOVABLE(nbr_table_reason_t reason, void *data);
+#endif /* NBR_TABLE_FIND_REMOVABLE */
+
 
 /* List of link-layer addresses of the neighbors, used as key in the tables */
 typedef struct nbr_table_key {
@@ -170,7 +179,7 @@ nbr_set_bit(uint8_t *bitmap, nbr_table_t *table, nbr_table_item_t *item, int val
 }
 /*---------------------------------------------------------------------------*/
 static nbr_table_key_t *
-nbr_table_allocate(void)
+nbr_table_allocate(nbr_table_reason_t reason, void *data)
 {
   nbr_table_key_t *key;
   int least_used_count = 0;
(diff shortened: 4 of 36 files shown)


adamdunkels:
This looks very useful, and something that certainly has been missing in terms of functionality.

I tested this with one of our more challenging RPL tests that connects a large number of nodes and then disconnects them again by turning on the RPL feather mode. The network is larger than what the local neighbor and routing tables can handle so the nodes use DAO packets with a zero lifetime to deregister routes from the network. The test runs fine with the current contiki-os master RPL, but fails with these patches. I tried with `RPL_DEFAULT_ROUTE_INFINITE_LIFETIME` set to 0 too, with the same results.

Is there anything in this patch that prevents DAO packets from deregistering routes by sending a DAO with lifetime 0?

joakimeriksson:
I do not think so - at least that is not intended. Can you share the test so that I can debug it and figure out what is happening?

The whole idea with this fix is that you should be able to have more active routes that any of the local neighbor and routing tables can hold. We have been running around 100 nodes in a network configured with 20 routes and 10 neighbors. Only the RPL root have had space for storing all the routes and it have worked fine in these cases.

There is a useful debug-printout in the neighbor policy that will how number of free neighbors and number of free routes each minute - that should show if the de-registration works.

adamdunkels:
Our test cases are pretty extreme, but it works with the existing code: 20-100 nodes with only 3 neighbor table entries and 3 routing entries each.The root node has 8 neighbor table entries and 8 routes. So the root is not able to maintain routes for the full network. As the network boots up, all nodes attempt to make a TLS connection to an Internet server. When a connection is set up, they exchange a bit of information, close the connection, and go into feather mode. This causes a DAO with lifetime 0 to be sent, releasing routing table entries along the routes, making it possible for another node to set up a route and connect.

These tests are pretty tightly bound to the Thingsquare platform though, but I'll see if I can set up a similar test using the Contiki `http-socket` module.

joakimeriksson:
Ok, so it might even be problems with the neighbor table also then - 3 neighbors is a very low config - I will try to look into that also. Maybe even any of the rpl-udp examples also would fail with this configuration. What is the timing for these connections? Is there some random wait so that there is no risk that more than 8 nodes tries to connect at the same time?

simonduq:
That's an awesome addition, will help get RPL reliability to the next level! A few early comments (haven't had time to look at details nor test yet):
* locking children: the PR removes the lock on all children [here](https://github.com/contiki-os/contiki/pull/1213/files#diff-bcecf0c639175aab93b833b5b5ddc37eL774). I believe the lock to be still needed to guarantee connectivity to the current sub-DODAG.
* having a neighbor table insertion policy is great, but in the PR the policy is used only by RPL, i.e. other modules might mess with it. If another module inserts neighbors aggressively, RPL neighbors will be deleted regardless of the policy. I'd suggest connecting the policy to the `nbr-table` module instead. We'd have `nbr-table` consistently check which neighbor to add/remove according to the policy in use. The policy itself can be used almost unmodified IMO, whenever RPL is running.
* to make DAO handling fully bullet-proof, it would be nice, when switching parent, to make sure the new route is installed (end-to-end ACKed) before even sending a nopath DAO to the old parent. But this can definitely go in a future PR.

joakimeriksson:
I agree on that we should look at generalising the policy to be plugged into the nbr-table module instead - this is a nice generalisation and make the policy a bit safer. The locking still confuse me a bit. We are locking the rpl-parent table when making this neighbor a child. But I can no where see that it is explicitly unlocked when we remove it as a child (purge-routes will remote routes from ds6-route module) - but will it also unlock the rpl-parent entry? The only places I see that are really removing (therefore unlocking) are global_repair and free_dag - but I might have missed something.

simonduq:
Indeed, there seems to be something fishy with the children locking. Looks like we're not removing nor unlocking nodes when they are not a child any longer. (note: remove would unlock automatically)

joakimeriksson:
I am doing a separate pull request where I move the locking to the ds6-routing module which already handle removal so it might as well do the locking for more consistent behavior.

simonduq:
sounds good!

adamdunkels:
All nodes will try to connect at the same time, but there is a backoff mechanism for the retries. The configuration is intentionally low to stress the system.

I created a pull request with a similar test here: https://github.com/contiki-os/contiki/pull/1218

adamdunkels:
#1218 is in now, so it should be possible to rebase and retest this one now

joakimeriksson:
Great! Will rebase and resubmit tonight!

mcr:
I will attempt to test this code in interop next week.

joakimeriksson:
That would be very interesting to know the results of! I have tried to follow the RPL RFC when it comes to DAO and DAO ACK but that is the most weak part of the whole RFC in my view.

joakimeriksson:
I have gotten this to work fine in simulation in Instant Contiki 3.0 but in travis it usually ends with something like:
 Data compelete 40, heard 43 in total

Is there a speed-difference in travis vs Cooja in instant contiki?

simonduq:
There are differences in tool versions, e.g. in this case probably JVM and gcc

joakimeriksson:
But is that really changing the speed of simulation that much? I usually ends up "complete" (TEST OK) after 23-24 minutes in COOJA but on travis it fails (should be after 40 minutes) with only 44 nodes "done".

adamdunkels:
The test is set up to run at 200% speed, so it shouldn't differ between travis and local Cooja. The random seed may be different when running in Cooja and through `make`on the command line.

FWIW, the test fails for me (only 36 heard), when I check out your branch locally and run in Cooja.

joakimeriksson:
It turned out not to be the random seed but in fact the travis environment and its speed to contiki-os.org or something. It ended up "green" in travis suddenly - now in the morning when I restarted. Likely less load, higher bit-rate, etc. Not really an optimal test since it is depending on external sites - but I guess I am very close now at least.

joakimeriksson:
I did update this PR now - I have refactored the neighbor policy to be called by the nbr-table module. I have also changed how the test-code for the 50-http gets a bit - to make it play a bit more nicely. It will now jump into feather as soon as it either fails or succeeds. Then when starting again it will first set mesh mode and then wait for it to have a downlink. This is to minimise traffic when we do know that the DAO handling has not yet been completed.

joakimeriksson:
I did the fix according to the above and now it is "green" but travis failed on something else instead (or errored).

joakimeriksson:
I have found one reason for the PR to make the test a bit slower than it was without the DAO and DAO ACK handling. It is due to Contiki RPLs handling of parent switching and local repair. Each time either of the two happens to a node it will quickly send a NO PATH DAO to get its route removed at the old parent and then send a DAO to the new parent immediately. It will also inform its children to re-register their routes (DTSN++) - and at local-repair it will tell its children to get lost (INFINITE RANK). In both cases the children's routes will never be remove from the old parent (and that sub-trees path to ROOT) since there is no forwarding of the NO PATH DAO's in that direction for the children. This will leave a large number of routes "hanging" until they timeout (30 minutes in the default configuration using this PR). I am looking into how to fix this so that parent switching and local repair is handled in a nicer way in Contiki RPL.

adamdunkels:
Sounds good! Great that this part of the code gets a pair of eyes looking at it. I have a feeling that even though the test runs nicely on the current RPL code, this may in some cases be just because of a lucky combination of parameters. It will be very good to have a proper implementation of this.

Also, I noticed that you (probably through a left-over debugging edit) changed the test so that it no longer tests what it is supposed to test: that the network can work with way too few routes at the root node.

regression-tests/12-rpl/04-rpl-large-network.csc
```diff
@@ -51,7 +51,7 @@ make sender-node.cooja TARGET=cooja</commands>
       <description>RPL root</description>
       <source>[CONTIKI_DIR]/regression-tests/12-rpl/code/root-node.c</source>
       <commands>make TARGET=cooja clean
-make root-node.cooja TARGET=cooja</commands>
+make root-node.cooja DEFINES=UIP_CONF_MAX_ROUTES=400 TARGET=cooja</commands>
```
adamdunkels: This is the problematic change: this should not be added here (the test is explicitly set up to have only 8 routes at the root node).

joakimeriksson: Ok! I'll revert that back to what it was! Will get back with more fixes for this PR soon (including reverting back to 8 here).

joakimeriksson:
I found a few bugs including one state flag (no-path) related one and I have also rebased. I also have removed the configuration of 400 routes at root in the large travis test (was 300) and set back to default configuration - e.g. no longer very short route lifetimes - of the 50 http downloads travis test.

adamdunkels:
@joakimeriksson: I'm running this one in our test suite too and things look good so far!

joakimeriksson:
@adamdunkels: I do hope it ends upp much better since one of the fixes really was a bug fix. Now it runs much smoother than before. The bug caused the second no-path DAO from a node to not be forwarded which of course caused "locking" of routes at RPL root and therefore caused the problem in the 50 http-client test.

adamdunkels:
The previous weekend's test was _almost_ green now! We have one test that randomly pings nodes in a multi-hop mesh and reboots them that failed, but that may be because the test is old and faulty since all the other tests worked. Will look into it more deeply as quickly as I can, but overall this looks very good now.

joakimeriksson:
Thanks for the effort in testing this! If you have any tests (cooja files, etc) that you like to share I can take a look at it also.

core/net/nbr-table.c
```diff
@@ -179,39 +188,66 @@ nbr_table_allocate(void)
...
+      return NULL;
+    } else {
+      /* used least_used_key to indicate what is the least useful entry */
+      int index;
+      int locked;
+      if((index = index_from_lladdr(lladdr)) != -1) {
+        least_used_key = key_from_index(index);
+        locked = locked_map[index];
+      }
+      /* Allow delete of locked item? */
```
simonduq: seems ok to me to allow deletion of a locked item (but I'd remove the question mark for the comment!)

core/net/nbr-table.c
```diff
@@ -179,39 +188,66 @@ nbr_table_allocate(void)
...
+      if(least_used_key != NULL && locked) {
+        PRINTF("Deleting locked item!\n");
+        locked_map[index] = 0;
+      }
+    }
+#endif /* NBR_TABLE_FIND_REMOVABLE */
+
+    if(least_used_key == NULL) {
```
simonduq: How about removing this conditional, and instead, have the whole block in `#else /* NBR_TABLE_FIND_REMOVABLE */`?

simonduq: I'd also add a line of comment both at `#ifdef NBR_TABLE_FIND_REMOVABLE` and `#else /* NBR_TABLE_FIND_REMOVABLE */` to clarify that in the former we use a configurable policy, in the latter the default policy

joakimeriksson:
Thanks for the comments - will make an update!

adamdunkels:
We did dig a bit deeper into this and found that there indeed was a real problem behind that one regression test that failed. It seems like a rebooting node may cause the network to lock itself up and not be able to add and delete routes. The problem seems to be related to that the sequence numbers wrap and begin again as a node reboots. I'll see if we can get the test running on the github contiki-os sources and make a pull request against your branch.

joakimeriksson:
Ok! That sounds like something we also need to fix! Is it the part of the network below that node, or more than that - that get locked?

adamdunkels:
Seems like that node cannot forward DAOs any more, effectively causing its children to be locked out of the network.

joakimeriksson:
Ok, yes so the rebooted now stops forwarding correctly. I will take a look at the code during next few days and see if I can find anything that could cause that type of behavior.

jonnteolsson:
any update on this one?

joakimeriksson:
I have not yet seen the problem yet - I am kind of hoping to get a regression test setup so that I can repeat the problems and debug. I am more than willing to spend time fixing it when I can repeat and understand the details of the problem!

core/net/rpl/rpl-mrhof.c
```diff
@@ -118,6 +120,23 @@ reset(rpl_dag_t *dag)
 }
 
 static void
+dao_ack_callback(rpl_parent_t *p, int status)
```
pablocorbalan: The `dao_ack_callback` will be only called if `RPL_WITH_DAO_ACK` is set. This function is not required for instance if we use the `RPL_MOP_NO_DOWNWARD_ROUTES`. Should we put it between `#if RPL_WITH_DAO_ACK` guards? You will also need to do the same in `rpl-of0.c`.

core/net/rpl/rpl.h
```diff
@@ -184,6 +184,7 @@ typedef struct rpl_instance rpl_instance_t;
 struct rpl_of {
   void (*reset)(struct rpl_dag *);
   void (*neighbor_link_callback)(rpl_parent_t *, int, int);
+  void (*dao_ack_callback)(rpl_parent_t *, int status);
```
pablocorbalan: `#if RPL_WITH_DAO_ACK` ... `#endif` guards?

examples/er-rest-example/project-conf.h
```diff
@@ -93,6 +96,13 @@
 #undef COAP_PROXY_OPTION_PROCESSING
 #define COAP_PROXY_OPTION_PROCESSING   0
 
+/* Turn of DAO ACK to make code smaller */
+#undef RPL_CONF_WITH_DAO_ACK
+#define RPL_CONF_WITH_DAO_ACK          0
```
simonduq: needed even now that we switched this to wismote?

core/net/rpl/rpl-nbr-policy.c
```diff
@@ -0,0 +1,252 @@
...
+     incoming packet */
+  switch(reason) {
+  case NBR_TABLE_REASON_RPL_DIO:
+    return find_removable_dio(&UIP_IP_BUF->srcipaddr, data);
+  case NBR_TABLE_REASON_RPL_DAO:
+    return find_removable_dao(&UIP_IP_BUF->srcipaddr, data);
+  case NBR_TABLE_REASON_RPL_DIS:
+    return find_removable_dis(&UIP_IP_BUF->srcipaddr);
+  default:
+    return NULL;
```
simonduq: this switch is not indented as per contiki code style is it?

core/net/ipv6/uip-ds6-nbr.c
```diff
@@ -79,9 +79,11 @@ uip_ds6_neighbors_init(void)
 /*---------------------------------------------------------------------------*/
 uip_ds6_nbr_t *
 uip_ds6_nbr_add(const uip_ipaddr_t *ipaddr, const uip_lladdr_t *lladdr,
-                uint8_t isrouter, uint8_t state)
+                uint8_t isrouter, uint8_t state, nbr_table_reason_t reason,
+                void *data)
 {
-  uip_ds6_nbr_t *nbr = nbr_table_add_lladdr(ds6_neighbors, (linkaddr_t*)lladdr);
+  uip_ds6_nbr_t *nbr = nbr_table_add_lladdr(ds6_neighbors, (linkaddr_t*)lladdr
+                                            ,reason, data);
```
simonduq: missing space before `reason`

core/net/ipv6/uip-ds6-route.c
```diff
@@ -297,11 +303,16 @@ uip_ds6_route_add(uip_ipaddr_t *ipaddr, uint8_t length,
        least recently used one we have. */
 
     if(uip_ds6_route_num_routes() == UIP_DS6_ROUTE_NB) {
+      uip_ds6_route_t *oldest;
+      oldest = NULL;
+#if UIP_DS6_ROUTE_REMOVE_LEAST_RECENTLY_USED
```
simonduq: I'd be ok with the whole code logic removed. It probably never is a good idea to have this module delete old routes but rather leave all responsibility to the routing protocol. I'd also add a comment

core/net/ipv6/uip-ds6-route.h
```diff
@@ -93,11 +93,45 @@ void uip_ds6_notification_rm(struct uip_ds6_notification *n);
...
+    (route)->state.state_flags &= ~RPL_ROUTE_ENTRY_DAO_PENDING;         \
+  } while(0)
+
+#define RPL_ROUTE_IS_DAO_NACKED(route)                                  \
+  ((route->state.state_flags & RPL_ROUTE_ENTRY_DAO_NACK) != 0)
+#define RPL_ROUTE_SET_DAO_NACKED(route) do {                            \
+    (route)->state.state_flags |= RPL_ROUTE_ENTRY_DAO_NACK;             \
+  } while(0)
+
+#define RPL_ROUTE_CLEAR_DAO(route) do {                                 \
```
simonduq: naming: `CLEAR_DAO_NACKED`? (consistency with the ones above)

core/net/ipv6/uip-ds6-route.h
```diff
@@ -162,7 +196,7 @@ uip_ipaddr_t *uip_ds6_route_nexthop(uip_ds6_route_t *);
 int uip_ds6_route_num_routes(void);
 uip_ds6_route_t *uip_ds6_route_head(void);
 uip_ds6_route_t *uip_ds6_route_next(uip_ds6_route_t *);
-
+int uip_ds6_route_is_nexthop(const uip_lladdr_t *lladdr);
```
simonduq: Any chance we can use ipaddr instead of lladdr as a parameter here? Would really increase consistency of the routing module interface. All other functions are ipaddr-based, including the similarly-named `uip_ds6_route_rm_by_nexthop`.

core/net/nbr-table.h
```diff
@@ -75,6 +75,17 @@ typedef struct nbr_table {
...
 
+typedef enum {
+	NBR_TABLE_REASON_UNDEFINED,
+	NBR_TABLE_REASON_RPL_DIO,
+	NBR_TABLE_REASON_RPL_DAO,
+	NBR_TABLE_REASON_RPL_DIS,
+	NBR_TABLE_REASON_ROUTE,
+	NBR_TABLE_REASON_IPV6_ND,
+	NBR_TABLE_REASON_MAC,
+        NBR_TABLE_REASON_LLSEC
```
simonduq: indentation needs fixing

core/net/rpl/rpl-mrhof.c
```diff
@@ -118,6 +120,23 @@ reset(rpl_dag_t *dag)
...
+{
+  if(status == RPL_DAO_ACK_UNABLE_TO_ACCEPT_ROOT) {
```
simonduq: Why treat this case differently? Shouldn't we try another parent if the root doesn't want to take our DAO? Would seem to make sense to me if the problem at the root is a neighbor table full, but space left for routing entries.

core/net/rpl/rpl-mrhof.c
```diff
@@ -118,6 +120,23 @@ reset(rpl_dag_t *dag)
...
+    return;
+  }
+  /* here we need to handle failed DAO's and other stuff */
+  PRINTF("RPL: MRHOF - DAO ACK received with status: %d\n", status);
+  if(status >= RPL_DAO_ACK_UNABLE_TO_ACCEPT) {
+    /* punish the ETX as if this was 10 packets lost */
+    neighbor_link_callback(p, MAC_TX_OK, 10);
```
simonduq: Ideally I think this shouldn't affect link metrics, but rather be recorded as a different flag. This way we can treat differently two different problems: broken links and parents unable to have us as a child. Could be addressed in a different PR.

core/net/rpl/rpl-dag.c
```diff
@@ -1142,6 +1143,9 @@ global_repair(uip_ipaddr_t *from, rpl_dag_t *dag, rpl_dio_t *dio)
 
   RPL_STAT(rpl_stats.global_repairs++);
 }
+
+void rpl_set_downward_link(uint8_t link);
```
simonduq: this should move to some header file

core/net/rpl/rpl-nbr-policy.c
```diff
@@ -0,0 +1,252 @@
...
+static void
+handle_periodic_timer(void *ptr)
+{
+  update_nbr();
+  ctimer_restart(&periodic_timer);
+}
+#endif /* DEBUG == DEBUG_FULL */
+/*---------------------------------------------------------------------------*/
+static void
+update_nbr(void)
```
simonduq: one comment explaining what this function is about would be nice

core/net/rpl/rpl-nbr-policy.c
```diff
@@ -0,0 +1,252 @@
...
+        if(uip_ds6_route_is_nexthop((uip_lladdr_t *)lladdr) == 0) {
+          worst_rank = rank;
+          worst_rank_nbr = lladdr;
+        } else {
+          printf("*** Can not use this as worst rank as it is a next hop\n");
+        }
+      }
+    }
+
+    /* Check if this neighbor is used as nexthop and therefor being a
```
simonduq: typo: therefore

core/net/rpl/rpl-nbr-policy.c
```diff
@@ -0,0 +1,252 @@
...
+/* Called whenever we get a unicast DIS - e.g. someone that already
+   have this node in its table - since it is a unicast */
+const linkaddr_t *
+find_removable_dis(uip_ipaddr_t *from)
+{
+
+  update_nbr();
+  if(num_free > 0) {
+    printf("num-free > 0 = %d", num_free);
+    printf("**** Should remove unused elements but can not... \n");
```
simonduq: confusing message

core/net/rpl/rpl-nbr-policy.c
```diff
@@ -0,0 +1,252 @@
...
+    /*    return 1; */
```
simonduq: ?

core/net/rpl/rpl-nbr-policy.c
```diff
@@ -0,0 +1,252 @@
...
+
+  update_nbr();
+
+  instance = rpl_get_instance(dio->instance_id);
+  if(instance == NULL || instance->current_dag == NULL) {
+    PRINTF("Did not find instance id: %d\n", dio->instance_id);
+    return NULL;
+  }
+
+  /* Add the new neighbor only if it is better than the preferred parent. */
```
simonduq: better than the preferred or better than the worst? (the code seems to compare against the worst)

core/net/rpl/rpl-icmp6.c
```diff
@@ -147,6 +196,30 @@ set16(uint8_t *buffer, int pos, uint16_t value)
...
+      PRINT6ADDR(from);
+      PRINTF(", ");
+      PRINTLLADDR((uip_lladdr_t *)packetbuf_addr(PACKETBUF_ADDR_SENDER));
+      PRINTF("\n");
+    }
+  }
+
+  if(nbr != NULL) {
+    /* set reachable timer if we added or found the nbr entry */
+    stimer_set(&nbr->reachable, UIP_ND6_REACHABLE_TIME / 1000);
```
simonduq: Not sure why we want to do this independent of the `reason`. What is the exact semantics of `rpl_icmp6_update_nbr_table`? Do we have a guarantee that, wherever it is called from, we know the neighbor is IPv6-reachable?

core/net/rpl/rpl-icmp6.c
```diff
@@ -914,41 +1093,114 @@ dao_output_target(rpl_parent_t *parent, uip_ipaddr_t *prefix, uint8_t lifetime)
...
+  if(instance == NULL) {
+    uip_clear_buf();
+    return;
+  }
+
+  parent = rpl_find_parent(instance->current_dag, &UIP_IP_BUF->srcipaddr);
+  if(parent == NULL) {
+    /* not a known instance - did we switch?? */
+    //    PRINTF("RPL: Received a DAO ACK from a not joined instance: %d",
+    //	   instance_id);
```
simonduq: I would remove dead code (or at least not use `//` style comments)

simonduq:
Very important feature and significant contribution, thanks!
Just went through the code, I left some comments but overall this very looks good to me. I plan to try it out very soon, will get back with more feedback ASAP.

core/net/rpl/rpl-icmp6.c
```diff
@@ -721,84 +783,121 @@ dao_input(void)
...
 #if RPL_CONF_MULTICAST
 fwd_dao:
 #endif
 
   if(learned_from == RPL_ROUTE_FROM_UNICAST_DAO) {
+    int should_ack = 0;
+
+    if(flags & RPL_DAO_K_FLAG) {
+      /* check if this route is already installed and we can ack now! */
+      /* not pending - and same seq-no means that we can ack.         */
```
simonduq: Does this mean we're already forwarding the same DAO and waiting for an DAO-ACK from up, before DAO-ACKing down? The comment could be made clearer

core/net/rpl/rpl-icmp6.c
```diff
@@ -86,6 +90,7 @@ void RPL_DEBUG_DAO_OUTPUT(rpl_parent_t *);
 #endif
 
 static uint8_t dao_sequence = RPL_LOLLIPOP_INIT;
+static uint8_t downward = 0;
```
simonduq: this variable and its getter and setter could easily have a clearer name

core/net/rpl/rpl-mrhof.c
joakimeriksson: This is mainly to handle the case then we are not connecting to ROOT. E.g. when root has no longer room for new routes, It is a disaster if all nodes try to switch if we are full at root and many new nodes are trying. They would be better off switching to another DODAG instead - if that is an option. Possibly we could rename it to RPL_DAO_ACK_UNABLE_TO_ADD_ROUTE_AT_ROOT?

core/net/rpl/rpl-icmp6.c
joakimeriksson: This is more or less refactoring of the "tricks" we have in Contiki RPL that adds and keeps neighbors in the ipv6 neighbor table. A bit incompliant with IPv6 ND RFCs - but have been around in Contiki RPL since start.

core/net/rpl/rpl-mrhof.c
simonduq: sounds like a better name yes

core/net/rpl/rpl-icmp6.c
simonduq: Right. From our discussion in https://github.com/contiki-os/contiki/issues/1380 I'm starting to think a reasonable thing to do would be to refresh reachability every time we get an incoming IPv6 datagram from a given neighbor. Means the neighbor is still using the address we know, and if it sends to us, it also believes we're reachable, i.e. it did receive from us lately. But probably not for this PR ;)

simonduq:
Have done some testing, works nicely including with TSCH. Have submitted a few minor fixes at https://github.com/joakimeriksson/contiki/pull/2

:+1: for me (after the various comments are addressed)

core/net/rpl/rpl-icmp6.c
```diff
@@ -721,84 +783,121 @@ dao_input(void)
...
     if(dag->preferred_parent != NULL &&
        rpl_get_parent_ipaddr(dag->preferred_parent) != NULL) {
+      uint8_t out_seq = 0;
+      /* if this is pending and we get the same seq no it is a retrans */
+      if(RPL_ROUTE_IS_DAO_PENDING(rep) &&
+         rep->state.dao_seqno_in == sequence) {
+        /* keep the same seq-no as before for parent also */
+        out_seq = rep->state.dao_seqno_out;
+      } else {
+        out_seq = prepare_for_dao_fwd(sequence, rep);
```
simonduq: I think we may have a problem here. If have a DAO pending, and then receive another DAO from the same child, we end up in the `else`, prepare the route for the new DAO and lose track of the previous one.

joakimeriksson:
I  did a full rebase of this PR now - and also did a few minor bug fixes that we have found lately at Yanzi.

alignan:
Oops, needs rebasing again

joakimeriksson:
Yes, but I will wait out a timeout in either direction. I can not spend more time on this at the moment. Timeout and remove or timeout and order a rebase. I will not rebase until I know where it is going. I have been waiting since August....

simonduq:
Guys, if we wait for a consensus where we're all confident enough that the PR is 100% correct, we will probably never merge this. Same goes for https://github.com/contiki-os/contiki/pull/1403 and https://github.com/contiki-os/contiki/pull/1508.

I'm concerned we're hindering/preventing progress here.

For this kind of major refactoring that brings new features (and more reliability!!) to a protocol, I'm in favor of merging now, fixing later. We will find out quick if things break. Waiting for the fix, users can stick to the last stable commit or even the release.

alignan:
@simonduq agree! as you have also tested this I'm keen to merge, :+1: as soon this is rebased and any pending comment is addressed

joakimeriksson:
Ok - I'll take some time during easter holidays here and do a rebase - hopefully it is not too bad since I did it fairly recently. Thanks!

simonduq:
BTW there are still a number of comments I left earlier, see above, please answer and/or address

joakimeriksson:
Yes, I will address these - the conflict seems small so I can spend the time addressing issues and comments!

core/net/rpl/rpl-icmp6.c
joakimeriksson: That is correct. Not sure if I would like to add more state to handle multiple pending DAOs? I am kind of assuming that the latest is the one to handle - but it might be a too simple assumption?

core/net/rpl/rpl-icmp6.c
joakimeriksson: Yes, and I will move it to the instance also as it in theory could be multiple active instances with only one of them having a downward route.

core/net/rpl/rpl-icmp6.c
joakimeriksson: This means that the route is already installed - e.g. the route have already been installed due to previous DAO / DAO_ACK interactions and therefore we know that it will be fine this time also - it is an optimisation to avoid waiting for the end-to-end ack this time. I will improve the comment a bit.

core/net/rpl/rpl-icmp6.c
joakimeriksson: removing code!

core/net/rpl/rpl-icmp6.c
joakimeriksson: I fixed this a bit also so that it is not only the reachable timer that is updated but also the state. I have sometime seen that this occurs just after the neighbor have timed-out (e.g. very stable RPL network and neighbor times out due to no DIO or other packets incoming  => no longer in state NBR_REACHABLE).

core/net/rpl/rpl-nbr-policy.c
joakimeriksson: Yes, you are right - it is comparing with the currently worst neighbor / potential parent.

core/net/rpl/rpl-nbr-policy.c
joakimeriksson: Will clean up the printout and add a comment (and remove dead code).

core/net/rpl/rpl-nbr-policy.c
joakimeriksson: will fix - even confusing to me now ;-) !

core/net/rpl/rpl-nbr-policy.c
joakimeriksson: will fix that!

core/net/rpl/rpl-dag.c
joakimeriksson: yes, but I removed it instead.

core/net/nbr-table.h
joakimeriksson: ok!

core/net/ipv6/uip-ds6-route.h
joakimeriksson: yes will fix!

core/net/ipv6/uip-ds6-route.h
joakimeriksson: ok!

core/net/rpl/rpl-icmp6.c
simonduq: I would add a comment to clearly put what the current limitations are

core/net/ipv6/uip-ds6-nbr.c
joakimeriksson: will fix!

core/net/rpl/rpl-nbr-policy.c
joakimeriksson: Hmmm. I think it is? - but I see that the function "head" is not.

examples/er-rest-example/project-conf.h
joakimeriksson: All examples are still sky based (e.g. simulations) - but I guess I can do a quick test and see if I can fix that also.

joakimeriksson: Ok. Not a good thing to do - these examples rely on a button press and wismote seems not to react to button press. So yes I will keep it at no-ack for now.

joakimeriksson:
Now I pushed a rebase and a set of fixes that address many (most I hope) of the above comments and issues.

regression-tests/12-rpl/04-rpl-large-network.csc
```diff
@@ -7057,5 +7057,4 @@ while(true) {&#xD;
     <location_x>603</location_x>
     <location_y>43</location_y>
   </plugin>
-</simconf>
-
+</simconf>
```
alignan: Newline missing?

alignan:
@simonduq are you ok with the changes?

simonduq:
yes

simonduq:
:+1:

alignan:
:+1:

PR #1215, opened 2015-08-26
Move locking of neighbors / children from RPL to uip-ds6-route module 

joakimeriksson:
This PR is spawned by the discussions on the PR #1213 . The PR moves the locking when DAO is received from its location in RPL (e.g. locking an RPL parent) to the corresponding location in uip-ds6-route module. The RPL locks was never unlocked which causes risk of ending up with a lot of locked entries in the nbr-table. Moving it to the uip-ds6-route module makes it easier to get a consistent locking / unlocking behaviour as this module removes the nbr-table entry for any neighbor that is no longer a nexthop. The locking is now made for each entry that is used as nexthop.

I also added some debug printouts for the nbr-table module that are useful when debugging locking and unlocking.

### DIFF
--- a/core/net/ipv6/uip-ds6-route.c
+++ b/core/net/ipv6/uip-ds6-route.c
@@ -367,6 +367,9 @@ uip_ds6_route_add(uip_ipaddr_t *ipaddr, uint8_t length,
     num_routes++;
 
     PRINTF("uip_ds6_route_add num %d\n", num_routes);
+
+    /* lock this entry so that nexthop is not removed */
+    nbr_table_lock(nbr_routes, routes);
   }
 
   uip_ipaddr_copy(&(r->ipaddr), ipaddr);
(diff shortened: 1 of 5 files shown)


simonduq:
:+1: for me
(one extra suggestion: how about also locking/unlocking default routes here, and getting rid of preferred parent lock from rpl?)

joakimeriksson:
Yes, I intended to lock the default routes but it turned out not to be a part of the neighbor table in the same way as the regular routes so I did not really see how I should lock them in an easy way.

simonduq:
I see, I agree, default routes are not in a nbr-table, so it's not as simple as I initially thought. I'd leave the PR as it is then.

adamdunkels:
This patch should also be run through this test, just to make sure we don't inadvertently lock too much: https://github.com/contiki-os/contiki/pull/1218

joakimeriksson:
Sure! Lets get the #1218 in and then I'll re-submit this with the new tests to run!

adamdunkels:
With #1218 merged, this can be rebased now

joakimeriksson:
Great - will do rebase and resubmit today.

joakimeriksson:
Ok, seems to work even with the new test!

adamdunkels:
:+1:

PR #1216, opened 2015-08-26
CoAP fixes and improvements

nfi:
This PR is a set of CoAP bug fixes and small feature improvements. They were developed during our work with a Contiki implementation of LWM2M and IPSO Objects that was interop tested during the IPSO Interop in May 2015.

Fixes and improvements include:
* support for observe on sub-resources
* fixed URI to always be NULL terminated for observe
* more defensive programming for bad arguments and buffer overruns

### DIFF
--- a/apps/er-coap/er-coap-observe.c
+++ b/apps/er-coap/er-coap-observe.c
@@ -58,9 +58,9 @@ LIST(observers_list);
 /*---------------------------------------------------------------------------*/
 /*- Internal API ------------------------------------------------------------*/
 /*---------------------------------------------------------------------------*/
-coap_observer_t *
-coap_add_observer(uip_ipaddr_t *addr, uint16_t port, const uint8_t *token,
-                  size_t token_len, const char *uri)
+static coap_observer_t *
+add_observer(uip_ipaddr_t *addr, uint16_t port, const uint8_t *token,
+             size_t token_len, const char *uri, int uri_len)
 {
   /* Remove existing observe relationship, if any. */
   coap_remove_observer_by_uri(addr, port, uri);
@@ -68,7 +68,12 @@ coap_add_observer(uip_ipaddr_t *addr, uint16_t port, const uint8_t *token,
   coap_observer_t *o = memb_alloc(&observers_memb);
 
   if(o) {
-    o->url = uri;
+    int max = sizeof(o->url) - 1;
+    if(max > uri_len) {
+      max = uri_len;
+    }
+    memcpy(o->url, uri, max);
+    o->url[max] = 0;
     uip_ipaddr_copy(&o->addr, addr);
     o->port = port;
     o->token_len = token_len;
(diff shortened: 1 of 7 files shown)


nfi:
I have updated the PR based on the comments and rebased on top of upstream master.

examples/er-rest-example/project-conf.h
```diff
@@ -93,6 +96,10 @@
 #undef COAP_PROXY_OPTION_PROCESSING
 #define COAP_PROXY_OPTION_PROCESSING   0
 
+/* Turn of DAO ACK to make code smaller */
+#undef RPL_CONF_WITH_DAO_ACK
+#define RPL_CONF_WITH_DAO_ACK          0
```
simonduq: this flag does not exist in upstream Contiki

examples/er-rest-example/project-conf.h
```diff
@@ -60,6 +60,9 @@
 #undef NETSTACK_CONF_RDC
 #define NETSTACK_CONF_RDC              nullrdc_driver
 
+#undef RPL_CONF_MAX_DAG_PER_INSTANCE
+#define RPL_CONF_MAX_DAG_PER_INSTANCE     1
```
simonduq: why?

apps/er-coap/er-coap.h
```diff
@@ -156,18 +156,18 @@ typedef struct {
...
 #define COAP_SERIALIZE_BLOCK_OPTION(number, field, text) \
   if(IS_OPTION(coap_pkt, number)) \
   { \
-    PRINTF(text " [%lu%s (%u B/blk)]\n", coap_pkt->field##_num, coap_pkt->field##_more ? "+" : "", coap_pkt->field##_size); \
+    PRINTF(text " [%lu%s (%u B/blk)]\n", (unsigned long)coap_pkt->field##_num, coap_pkt->field##_more ? "+" : "", coap_pkt->field##_size); \
     uint32_t block = coap_pkt->field##_num << 4; \
     if(coap_pkt->field##_more) { block |= 0x8; } \
     block |= 0xF & coap_log_2(coap_pkt->field##_size / 16); \
-    PRINTF(text " encoded: 0x%lX\n", block); \
+    PRINTF(text " encoded: 0x%lX\n", (unsigned long)block);		\
```
simonduq: isn't `%lX` signed?

simonduq:
Very nice cleanup, and this also happens to fix https://github.com/contiki-os/contiki/pull/1250.
Haven't tested the new observe for sub-resources but the feature was definitely missing (and I trust @nfi this works).
:+1:

examples/er-rest-example/project-conf.h
nfi: This is a mistake. It is a change from our work with DAO ACK for scalability and should not be included in this PR. I will remove it. Thanks!

examples/er-rest-example/project-conf.h
nfi: This was added to reduce the code size so it could be compiled on Tmote Sky. Since I think the Travis test now compiles for WiSMote it might no longer be needed. I will remove it.

apps/er-coap/er-coap.h
nfi: I think %x always prints the value as unsigned. The only difference between %x and %X is that one prints lowercase and the other uppercase.

nfi:
Thanks for your comments! I will remove the unnecessary code to reduce the example code size and do another pass through the casts in the debug output. I will let you know when it has been done.

apps/er-coap/er-coap.h
simonduq: right, my mistake!

nfi:
I have updated the PR based on the comments.

apps/er-coap/er-coap-observe.c
```diff
@@ -58,9 +58,9 @@ LIST(observers_list);
 /*---------------------------------------------------------------------------*/
 /*- Internal API ------------------------------------------------------------*/
 /*---------------------------------------------------------------------------*/
-coap_observer_t *
+static coap_observer_t *
```
nvt: When changing the signature of this function, the module prefix "coap_" should be removed. Furthermore, is there a need to not have the <code>const</code> qualifier set for the addr variable (and possibly doing the same for functions called from <code>coap_add_observer</code> that use <code>addr</code>?

nfi: I have renamed the function as suggested. It would not be a problem to change the ```addr``` argument to be ```const``` but then the other functions in the CoAP API should also be updated in the same manner. I completely agree that this would be nice but is this something we should do in this PR or leave it for another PR?

simonduq: I agree using `const` would be best but this seems pretty orthogonal to this PR, so I'd have it in another one. This PR is a nice cleanup and solved a known bug (https://github.com/contiki-os/contiki/pull/1250) so I'd focus on getting it in first.

nvt: Yes, the work on <code>const</code> is not needed for this PR to be merged.

nvt:
:+1:

PR #1217, opened 2015-08-26
Fixes compilation targeting c64

tiagoshibata:
Adds some casts, splits declaration and non constant initialization and changes a few prototypes. The cc65 compiler wasn't happy with implicit casts.

### DIFF
--- a/core/net/ipv6/uip-nd6.c
+++ b/core/net/ipv6/uip-nd6.c
@@ -662,7 +662,8 @@ rs_input(void)
         }
         if(memcmp(&nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET],
             lladdr, UIP_LLADDR_LEN) != 0) {
-          uip_ds6_nbr_t nbr_data = *nbr;
+          uip_ds6_nbr_t nbr_data;
+          nbr_data = *nbr;
           uip_ds6_nbr_rm(nbr);
           nbr = uip_ds6_nbr_add(&UIP_IP_BUF->srcipaddr, &lladdr_aligned,
                                 0, NBR_STALE, NBR_TABLE_REASON_IPV6_ND, NULL);
--- a/core/net/ipv6/uip6.c
+++ b/core/net/ipv6/uip6.c
@@ -73,6 +73,7 @@
 
 #include "sys/cc.h"
 #include "net/ip/uip.h"
+#include "net/ip/uip_arch.h"
 #include "net/ip/uipopt.h"
 #include "net/ipv6/uip-icmp6.h"
 #include "net/ipv6/uip-nd6.h"
@@ -1854,8 +1855,10 @@ uip_process(uint8_t flag)
       if((UIP_TCP_BUF->flags & TCP_SYN)) {
         if((uip_connr->tcpstateflags & UIP_TS_MASK) == UIP_SYN_RCVD) {
           goto tcp_send_synack;
+#if UIP_ACTIVE_OPEN
         } else if((uip_connr->tcpstateflags & UIP_TS_MASK) == UIP_SYN_SENT) {
           goto tcp_send_syn;
+#endif
         }
       }
       goto tcp_send_ack;
--- a/cpu/6502/net/ethernet-drv.c
+++ b/cpu/6502/net/ethernet-drv.c
@@ -34,17 +34,23 @@
 
 #include "contiki-net.h"
 #include "net/ethernet.h"
+#include "net/ip/tcpip.h"
 #include "net/ipv4/uip-neighbor.h"
 
 #include "net/ethernet-drv.h"
 
 #define BUF ((struct uip_eth_hdr *)&uip_buf[0])
+#define IPBUF ((struct uip_tcpip_hdr *)&uip_buf[UIP_LLH_LEN])
 
 PROCESS(ethernet_process, "Ethernet driver");
 
 /*---------------------------------------------------------------------------*/
 uint8_t
+#if NETSTACK_CONF_WITH_IPV6
+ethernet_output(const uip_lladdr_t *)
+#else
 ethernet_output(void)
+#endif
 {
   uip_arp_out();
   ethernet_send();
@@ -61,7 +67,7 @@ pollhandler(void)
   if(uip_len > 0) {
 #if NETSTACK_CONF_WITH_IPV6
     if(BUF->type == uip_htons(UIP_ETHTYPE_IPV6)) {
-      uip_neighbor_add(&IPBUF->srcipaddr, &BUF->src);
+      uip_neighbor_add(&IPBUF->srcipaddr, (struct uip_neighbor_addr *)&BUF->src);
       tcpip_input();
     } else
 #endif /* NETSTACK_CONF_WITH_IPV6 */
--- a/apps/webserver/httpd-cgi.c
+++ b/apps/webserver/httpd-cgi.c
@@ -155,10 +155,13 @@ make_tcp_stats(void *arg)
 {
   struct uip_conn *conn;
   struct httpd_state *s = (struct httpd_state *)arg;
+#if NETSTACK_CONF_WITH_IPV6
+  char buf[48];
+#endif
+
   conn = &uip_conns[s->u.count];
 
 #if NETSTACK_CONF_WITH_IPV6
-  char buf[48];
   httpd_sprint_ip6(conn->ripaddr, buf);
   return snprintf((char *)uip_appdata, uip_mss(),
          "<tr align=\"center\"><td>%d</td><td>%s:%u</td><td>%s</td><td>%u</td><td>%u</td><td>%c %c</td></tr>\r\n",
--- a/cpu/6502/net/ethernet-drv.h
+++ b/cpu/6502/net/ethernet-drv.h
@@ -42,6 +42,10 @@ struct ethernet_config {
 
 PROCESS_NAME(ethernet_process);
 
+#if NETSTACK_CONF_WITH_IPV6
+uint8_t ethernet_output(const uip_lladdr_t *);
+#else
 uint8_t ethernet_output(void);
+#endif
 
 #endif /* ETHERNET_DRV_H_ */


cpu/6502/net/ethernet-drv.c
```diff
@@ -61,7 +67,7 @@ pollhandler(void)
   if(uip_len > 0) {
 #if NETSTACK_CONF_WITH_IPV6
     if(BUF->type == uip_htons(UIP_ETHTYPE_IPV6)) {
-      uip_neighbor_add(&IPBUF->srcipaddr, &BUF->src);
+      uip_neighbor_add(&IPBUF->srcipaddr, (struct uip_neighbor_addr *) &BUF->src);
```
simonduq: there should be no space before `&BUF`

core/net/ipv6/uip-nd6.c
```diff
@@ -627,7 +627,8 @@ rs_input(void)
         /* If LL address changed, set neighbor state to stale */
         if(memcmp(&nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET],
             uip_ds6_nbr_get_ll(nbr), UIP_LLADDR_LEN) != 0) {
-          uip_ds6_nbr_t nbr_data = *nbr;
+          uip_ds6_nbr_t nbr_data;
+          nbr_data = *nbr;
```
simonduq: why is this needed? if this is only for a c64 warning I'd just revert this one change

tiagoshibata: It's a compilation error. I'm almost sure declaring a variable and assigning a non-const value to it is invalid in some old C standard:
../../core/net/ipv6/uip-nd6.c(632): Error: `{' expected
../../core/net/ipv6/uip-nd6.c(632): Error: `{' expected
../../core/net/ipv6/uip-nd6.c(632): Error: `{' expected
../../core/net/ipv6/uip-nd6.c(632): Error: Constant expression expected
../../core/net/ipv6/uip-nd6.c(632): Error: `}' expected
../../core/net/ipv6/uip-nd6.c(632): Error: `}' expected
../../core/net/ipv6/uip-nd6.c(632): Error: `}' expected

simonduq: ok, I must admit I'm not a huge fan of going by old standards

greg-king5: It's a cc65 limitation.  The compiler code that initiates local variables does the same thing for arrays, structs, and unions.  We cannot assign the contents of array variables to arrays -- directly [we must use memcpy() explicitly].  Therefore, cc65 won't let us assign a struct variable to a struct that is being defined -- we must define, then assign, separately.

simonduq:
Fine with merging this, but we need a rebase

simonduq:
(setting a timeout: merging in a month if rebased, else closing)

tiagoshibata:
Rebased.

core/net/ipv6/uip-nd6.c
```diff
@@ -662,7 +662,8 @@ rs_input(void)
         }
         if(memcmp(&nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET],
             lladdr, UIP_LLADDR_LEN) != 0) {
-          uip_ds6_nbr_t nbr_data = *nbr;
+          uip_ds6_nbr_t nbr_data;
+          nbr_data = *nbr;
```
oliverschmidt: Is this change really necessary. cc65 is supposed to support variable definitions with initialization.

tiagoshibata: On my machine, running [latest git  cc65](https://github.com/cc65/cc65):

```c
typedef struct {
	int a, b, c;
} test_type;

void test() {
	test_type *pointer;
	test_type value;
	value = *pointer;
}
```
Compiles fine, while:
```c
typedef struct {
	int a, b, c;
} test_type;

void test() {
	test_type *pointer;
	test_type value = *pointer;
}
```
Gives:
```
test.c(7): Error: `{' expected
test.c(7): Error: Too many local variables
test.c(7): Error: Constant expression expected
test.c(7): Error: `}' expected
```
Am I using an unsupported cc65? Which one should be used with Contiki?

simonduq: Ah ok then I guess the problem is not just variable initialization, but the fact the variable is a struct and we derefernece it. One thing though is that as long as we don't have a regression test to verify this, the problem might slip in again any time...

oliverschmidt: > On my machine, running latest git cc65:
> [...]
> Am I using an unsupported cc65? Which one should be used with Contiki?

You're using the one and only cc65 suitable for Contiki. It's the one used by Travis too: https://github.com/contiki-os/contiki/blob/master/.travis.yml#L74

Thanks for isolating the issue with a small test program. I'll post it on the cc65 mailing list...

simonduq: Could you add a cc65 compile test that covers the bug you fixed?

tiagoshibata: Yes, I'll get it done sometime before the weekend.

tiagoshibata: Done.

regression-tests/25-compile-c64/Makefile
```diff
@@ -0,0 +1,14 @@
+EXAMPLESDIR=../../examples
+TOOLSDIR=../../tools
+
+# build c64 tests. Many examples won't fit in memory or won't compile due to the
+# lack of int64_t in cc65, so only a subset of IPv4 is tested.
+EXAMPLES = irc/c64 \
+telnet-server/c64 \
+webbrowser/c64 \
+webserver/c64 \
+wget/c64
```
oliverschmidt: I must admit that this puzzles me. There are already test-builds for these very examples. What do we expect from building them here once more?

I thought we would want to test-build IPv6 with cc65?

simonduq: Right so if they were already being built with no error, how is the change in `uip-nd6.c` needed @tiagoshibata ?

oliverschmidt: The examples above are all configured to be built with IPv4 while this pull request is (only) about IPv6.

simonduq: Ah ok sorry. @tiagoshibata what we need is not this new file, but rather a new line in `16-compile-6502-ports/Makefile` to compile ipv6 on c64

oliverschmidt: Exactly.

oliverschmidt: As far as I understand building hello-world for c64 is something not possible in master but becoming possible with this pull request.

tiagoshibata: Ok, I didn't notice the already existent tests. Sorry for the mess, I should have paid more attention there.

hello-world currently compiles on my side when UIP_ACTIVE_OPEN = 1. When it is disabled, Contiki fails to build [on this line](https://github.com/contiki-os/contiki/blob/master/core/net/ipv6/uip6.c#L1858) since [tcp_send_syn is undefined](https://github.com/contiki-os/contiki/blob/master/core/net/ipv6/uip6.c#L1798). Enclosing #if/#endif are missing there, I believe surrounding lines 1857/1858 would solve the problem without breaking anything else, could anyone confirm that? I am adding this change to this pull request.

Linkage fails due to code size (overflows by 4849 bytes). Let me know if you have any suggestions to reduce code size.

oliverschmidt: > Enclosing #if/#endif are missing there, I believe surrounding lines 1857/1858 would solve the problem without breaking anything else, could anyone confirm that?

The code section in question is conceptually the same for IPv4 and IPv6. So I checked out the IPv4 code (which builds fine without UIP_ACTIVE_OPEN) for additional insights. The whole nested "if"'s are missing in IPv4, see https://github.com/contiki-os/contiki/blob/f9a30236c87741259542210fc53340c39a7829ca/core/net/ipv4/uip.c#L1473. Git blame for IPv6 shows https://github.com/contiki-os/contiki/commit/f31d6f271fe7b1989fb335a251cfa07a22f83f57 meaning that Adam provided a fix to IPv6 (which might likely make sense for IPv4 too) very likely without taking UIP_ACTIVE_OPEN into account. I consider this another indication that the #if/#endif in question is indeed the correct approach.

> Linkage fails due to code size (overflows by 4849 bytes).

Ooops, that's a lot :-( And this for hello-world which basically does nothing? I asn't aware that IPv6 is soo much bigger than IPv4. I can tell you for sure that you won't find any cc65 compiler/linker/setup magic to improve the situation. As space is always very tight for the cc65 targets this is all already tuned to the limit. The only thing you can do is find Contiki (CFG_...) macros that reduce/remove stuff. However https://github.com/contiki-os/contiki/blob/master/cpu/6502/6502def.h already turns off about everything by default. Maybe IPv6 has some more config options?

core/net/ipv6/uip6.c
```diff
@@ -1854,8 +1855,10 @@ uip_process(uint8_t flag)
       if((UIP_TCP_BUF->flags & TCP_SYN)) {
         if((uip_connr->tcpstateflags & UIP_TS_MASK) == UIP_SYN_RCVD) {
           goto tcp_send_synack;
+#if UIP_ACTIVE_OPEN
         } else if((uip_connr->tcpstateflags & UIP_TS_MASK) == UIP_SYN_SENT) {
           goto tcp_send_syn;
```
greg-king5: Based on the other `goto tcp_send_syn;` lines in this file, I think that there should be a `UIP_TCP_BUF->flags = 0;` line above this line.

greg-king5:
Disabling RPL will reduce the code size.
But, a lot more would need to be done if anyone wants to use IPv6 on a 6502 platform!

- This patch would need to be applied:
```diff
diff --git i/cpu/6502/Makefile.6502 w/cpu/6502/Makefile.6502
--- i/cpu/6502/Makefile.6502
+++ w/cpu/6502/Makefile.6502
@@ -49,6 +49,10 @@ CONTIKI_SOURCEFILES += $(CTK) ctk-conio.c petsciiconv.c cfs-posix-dir.c \
 
 MODULES += core/ctk
 
+ifeq ($(CONTIKI_WITH_IPV6),1)
+MODULES += core/net
+endif
+
 # Set target-specific variable values
 ${addprefix $(OBJECTDIR)/,${call oname, $(ETHERNET_SOURCEFILES)}}: ASFLAGS += -D DYN_DRV=0
 
```
- A `uip_icmp6chksum()` function would need to be written in `cpu/6502/net/uip_arch.c`.
- IPv6 code would need to be added to several files in `cpu/6502/{lib,net}/`.

simonduq:
removing the timeout as this still is ongoing work

core/net/ipv6/uip6.c
tiagoshibata: You seem to be right, there is no check if some flag is already set, so I should reset it before ORing with SYN.

tiagoshibata:
Adding `core/net` increases the overflow to 9980 bytes, quite a lot. Disabling RPL generates linkage errors because of missing functions (`uip_icmp6chksum` among them), I'll work on them.

simonduq:
Please work on this iff you actually need the feature. Else let's just merge as is. There seems to be no c65 IPv6 user around anyway.

tiagoshibata:
Fine, this can be merged. Is the change proposed by @greg-king5 to reset UIP_TCP_BUF->flags needed? It looks so, but the code was working without it.

simonduq:
:+1:

oliverschmidt:
- As far as I understand @greg-king5's remark on the reset of UIP_TCP_BUF->flags isn't related to this pull request.

- All changes in IPv6 specific code for cc65 targets are more or less ornament as we don't see IPv6 working on those targets for memory size restrictions anyway.

From that perspective this should be merged.

oliverschmidt:
@greg-king5 BTW: Maybe you want to take the the reset of UIP_TCP_BUF->flags as opportunity to start providing your own pull requests to Contiki...

simonduq:
merging after timeout

Issue #1106, opened 2015-06-09
icmpv6 messages are forwarded by error


laurentderu:
The code managing the icmpv6 packets in uip6.c requires that the handler set uip_len to 0 when it does not have to send a reply packet.

```
  /*
   * Search generic input handlers.
   * The handler is in charge of setting uip_len to 0
   */
```

However, the DIO and DAO handlers simply return when an invalid message or option is received (e.g. DAO with invalid rank or from an unknown DAG). As uip_len is not set, the stack tries to send the packet again, resulting of the sending of a NS message, with target set to the host own IP, to resolve the destination address.

Either we should add a call to uip_clear_buf() as proposed in #927 or change the behavior of the imp management code and force uip_len to zero when the handler returns. If a reply must be sent, then the handler should instead call uip_icmp6_send().

g-oikonomou:
You are right, this needs fixed. I have a slight preference to `uip_clear_buf()` because i) it puts less strain on the stack and ii) it's consistent with other parts of the code, but I understand it will be needed in many places and it could be error prone. `goto discard;` could also work, as per the handlers in `uip-ndc6.c`.

Irrespective of which one we end up choosing, I think we should do it consistently across the board. i.e. either we add `_clear_buf` everywhere, or we remove `uip_len = 0` from everywhere (e.g. ND code)

adamdunkels:
This should be fixed by #1218, so I'm closing this - @laurentderu, please reopen if you think more fixing is needed!

PR #1218, opened 2015-08-27
RPL large-scale test & bugfix

adamdunkels:
This patch contains a test with a too-large RPL network and one RPL  bugfix.

# Too large RPL test
The RPL mesh should be able to support more nodes than the number of neighbor table entries and routing table entries.

This patch adds a test where 50 nodes attempt to do an HTTP socket connection to download the Contiki web page. The network is configured to have 3 neighbor table entries and routing table entries, except for the gateway node which has 8 entries.

Once the HTTP socket has downloaded its data, it sets the node into RPL feather mode, in which the network drops all direct routes to the node.

To make this work, this patch also adds a NOPATH DAO transmission when a node transitions into feather mode.

# RPL bugfix
The RPL ICMPv6 input handling code should drop incoming packets that it cannot handle, such as DIOs and DAOs, but the code currently does not actually drop them. Instead, the packets are forwarded unchanged, which will cause problems for nodes that receive them. This is particularly bad for DAOs, as the receiving node will get a bad route (because the receiving node believes that the erroneously forwarding node is a valid route). The fix is to ensure that `uip_len` is set to `0` every time a packet should be dropped. This patch fixes that.


joakimeriksson:
Good - I have also seen that bug! It was introduced when we moved from the switch that switched on the different RPL messages over to the ICMP handlers.

laurentderu:
see https://github.com/contiki-os/contiki/issues/1106 ;) I was waiting for https://github.com/contiki-os/contiki/pull/1116 to fix it

adamdunkels:
@laurentderu nice! This should definitely fix #1106! And should be possible to simply this one after #1116.

adamdunkels:
Rebased on top of #1116

simonduq:
:+1: (if Travis is green)

adamdunkels:
Travis was green before rebasing - let's hope it still is :)

adamdunkels:
Travis is green!

fros4943:
:+1:

joakimeriksson:
Is it fairly easy to run this test in Instant Contiki? - I did a quick try but there are things failing (wpcap-slip location, etc)? I would prefer to debug locally rather than sending up to github for complete travis runs to speed up the development and to avoid spending travis CPU cycles when not needed.

adamdunkels:
Yes it runs nicely in Instant Contiki. One thing may need to be altered though is the `CMD_PATH` on line 12 in `testscript.js`. It should be either `"../../wpcapslip"` or `"../../../wpcapslip"` depending on whether the test is run from within Cooja or by running `make` in the test directory.

joakimeriksson:
Yes, I think I am getting it to run now (it gets further at least)! Thanks.

joakimeriksson:
Now I can run the simulation in COOJA and see that it is close but a bit slower than the test requires (I have made some minor modifications to the neighbor policy). It ends like this:
---
Data compelete 24, heard 28 in total
Data compelete 16, heard 29 in total
Data compelete 19, heard 30 in total
TEST TIMEOUT
TEST FAILED
---

PR #1222, opened 2015-08-28
RPL: removed pointer to next parent since it is not used anymore

joakimeriksson:
Removes an unused next-pointer in the RPL parent struct.


simonduq:
:+1:

PR #1226, opened 2015-09-01
Minor fixes to Z1 and Wismote

simonduq:
This PR initializes the random seed on Z1 and Wismote, and fixes CC2420 SFD timestamp configuration on Z1 (to now use the flag `CC2420_CONF_SFD_TIMESTAMPS`, as other cc2420 platforms do).


alignan:
:+1:

PR #1227, opened 2015-09-01
Travis: compile er-rest-example for the Wismote platform rather than Sky

simonduq:
Because we don't want to restrain ourselves with 700 bytes of new features!

### DIFF
--- a/regression-tests/01-compile-base/Makefile
+++ b/regression-tests/01-compile-base/Makefile
@@ -13,7 +13,7 @@ hello-world/wismote \
 hello-world/z1 \
 eeprom-test/native \
 collect/sky \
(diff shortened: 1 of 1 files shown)


cmorty:
Isn't being able to run more features without the need of more powerful hardware one of the awesome developments in the WSN domain? Yet more serious: What about a full version and a reduced example for sky? After all it is still one of the most popular platforms.

simonduq:
We still have plenty of examples compiled for Sky, enough IMO. The platform is more than 10 years old, I'd rather move on than be stuck on the magic number "48k".

g-oikonomou:
:+1:

nvt:
:+1:

PR #1228, opened 2015-09-01
Replaced function pointer cast in contikimac

bkozak-scanimetrics:
Replaced a cast of the function pointer to powercycle with a wrapper calling powercycle which has the correct signature.

The previous implementation was an instance of undefined behaviour according to the C standard.

### DIFF
--- a/core/net/mac/contikimac/contikimac.c
+++ b/core/net/mac/contikimac/contikimac.c
@@ -269,6 +269,7 @@ off(void)
 }
 /*---------------------------------------------------------------------------*/
 static volatile rtimer_clock_t cycle_start;
+static void powercycle_wrapper(struct rtimer *t, void *ptr);
 static char powercycle(struct rtimer *t, void *ptr);
 static void
 schedule_powercycle(struct rtimer *t, rtimer_clock_t time)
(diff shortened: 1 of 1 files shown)


core/net/mac/contikimac/contikimac.c
```diff
@@ -281,8 +282,7 @@ schedule_powercycle(struct rtimer *t, rtimer_clock_t time)
       time = RTIMER_NOW() - RTIMER_TIME(t) + 2;
     }
 
-    r = rtimer_set(t, RTIMER_TIME(t) + time, 1,
-                   (void (*)(struct rtimer *, void *))powercycle, NULL);
+    r = rtimer_set(t, RTIMER_TIME(t) + time, 1,powercycle_wrapper, NULL);
```
cmorty: There should be space after the comma.

cmorty:
As it's a static function that _always_ returns 1: Why not remove the return value?

bkozak-scanimetrics:
All style issues should be resolved.

tonnenpinguin:
Actually @cmorty got a point. powercycle() is only called from within the wrapper. Correct me if I'm wrong, but wouldn't that mean that we could ditch the wrapper function altogether and just declare the powercycle() function as void?

bkozak-scanimetrics:
powercycle is a proto-thread which means that it has hidden return statements in it's various protothread macros (such as ```PT_YIELD```).

cmorty:
@tonnenpinguin Well that's the problem with patches that look simple. ;) 

With the style-fixes it get :+1:  from me.

simonduq:
:+1:

simonduq:
timing out, merge in a month

simonduq:
merging after timeout

PR #1230, opened 2015-09-01
Fix Wismote randon_init for link-layer addresses shorter than 8 bytes

simonduq:
https://github.com/contiki-os/contiki/pull/1226 was wrongly assuming link-layer addresses of 8 bytes in `random_init` for Wismote (not an issue with the z1 code, where `node_mac` always is 8 bytes long).


simonduq:
I think this is a basic enough fix to yesterday's PR, just merging.

Issue #1229, opened 2015-09-01
CC2650 - Simple RPL Network Fails to Build


bkozak-scanimetrics:
I have a simple RPL network made up of two CC2650 (srf06 platform) nodes; one is a border router and the other a custom application that does nothing until a button is pressed. Up until recently this was working fine. After the major changes that took place last week however (such as #1212) the nodes are not even able to form an RPL network.

On my sniffer, I can see that RPL control packets are sent by both nodes but that no network is formed (the border router's web-page confirms this).

I have found, however, that everything mysteriously starts to work if I enable ```DEBUG``` in contikimac.c. This leads me to believe that this is some sort of a timing or LPM related issue as the debug prints from contikimac will act to slow down certain sections of code and delay sleeping.

bkozak-scanimetrics:
A brief update:

It would seem that a network will form eventually if the nodes are left long enough but it is *extremely* slow. 

After the network forms, over the air communication is very unreliable (~90% failure to reply to pings sent through the border router).

g-oikonomou:
Smells like timing of ACKs, but I have not observed anything like that. I must say my testing was mostly with slip radio + 6lbr

bkozak-scanimetrics:
> Smells like timing of ACKs

With correctly functioning devices the nodes will first try to establish an RPL network using broadcast packets. A unicast packet is only sent just before the network forms.

What I'm seeing now is that no unicast packet is sent at all (for a long time) so there is no chance for there to be a problem with the ACKs.

g-oikonomou:
Ok so the mesh node doesn't pick up the DIOs sent by the BR: either it does not wake up to detect it, or the detection fails (CCA settings?)

bkozak-scanimetrics:
> Ok so the mesh node doesn't pick up the DIOs sent by the BR

Yes, this seems to be what's happening.

> either it does not wake up to detect it, or the detection fails (CCA settings?)

Something like that seems probable to me.

tonnenpinguin:
+1 on that issue.
I had massive problems on my custom platform based on the cc2650 after merging in the changes from #1212. I did not get the chance to do any thorough tests until now though.

adamdunkels:
We found (and fixed) a number of issues with the radio driver and the clock code in #1212, which may very well cause problems like these. Fixes are coming!

bkozak-scanimetrics:
@g-oikonomou 

[...]

PR #1231, opened 2015-09-01
CC26xx - enable correct IRQ in rf_core_cmd_done_en

bkozak-scanimetrics:
Fix for #1229

rf_core_cmd_done_en() was enabling the wrong irq for detecting the completion of foreground operations. This was causing cc26xx devices to not wake-up on time when calling lpm_sleep() from transmit().

PR #1233, opened 2015-09-02
CC26xx/CC13xx fixes

adamdunkels:
This fixes a few bugs in the latest cc26xx/cc13xx code:

# Tickless clock

The tickless clock interrupt interrupt timing was not computed from the previous interrupt's trigger time, but from the current time (`RTIMER_NOW()`), which may have progressed significantly, causing the next timer to occur with a delay

Also, the `clock_time()` function should return an increasing time even if there hasn't been any clock interrupts. This patch recomputes the `count` variable inside the `clock_time()` call instead of on each interrupt.

# IEEE radio

A few minor fixes to the RF driver. We also have a workaround for the `lpm_sleep()` issue in #1231, which is not part of this pull request, but maybe that will be taken care of by #1231?

### DIFF
--- a/cpu/cc26xx-cc13xx/clock.c
+++ b/cpu/cc26xx-cc13xx/clock.c
@@ -120,31 +120,35 @@ clock_init(void)
         ((TIMER_CFG_B_ONE_SHOT >> 8) & 0xFF) | GPT_TBMR_TBPWMIE;
 }
 /*---------------------------------------------------------------------------*/
+static void
+update_clock_variable(void)
+{
+  uint32_t aon_rtc_secs_now;
+  uint32_t aon_rtc_secs_now2;
+  uint16_t aon_rtc_ticks_now;
+
+  do {
+    aon_rtc_secs_now = HWREG(AON_RTC_BASE + AON_RTC_O_SEC);
+    aon_rtc_ticks_now = HWREG(AON_RTC_BASE + AON_RTC_O_SUBSEC) >> 16;
+    aon_rtc_secs_now2 = HWREG(AON_RTC_BASE + AON_RTC_O_SEC);
+  } while(aon_rtc_secs_now != aon_rtc_secs_now2);
+
+  /* Convert AON RTC ticks to clock tick counter */
+  count = (aon_rtc_secs_now * CLOCK_SECOND) + (aon_rtc_ticks_now >> 9);
+}
+/*---------------------------------------------------------------------------*/
 CCIF clock_time_t
 clock_time(void)
 {
+  update_clock_variable();
+
   return (clock_time_t)(count & 0xFFFFFFFF);
 }
 /*---------------------------------------------------------------------------*/
 void
 clock_update(void)
 {
-  bool interrupts_disabled;
-  uint32_t aon_rtc_secs_now;
-  uint16_t aon_rtc_ticks_now;
-
-  interrupts_disabled = ti_lib_int_master_disable();
-
-  aon_rtc_secs_now = HWREG(AON_RTC_BASE + AON_RTC_O_SEC);
-  aon_rtc_ticks_now = HWREG(AON_RTC_BASE + AON_RTC_O_SUBSEC) >> 16;
-
-  /* Convert AON RTC ticks to clock tick counter */
-  count = (aon_rtc_secs_now * CLOCK_SECOND) + (aon_rtc_ticks_now >> 9);
-
-  /* Re-enable interrupts */
-  if(!interrupts_disabled) {
-    ti_lib_int_master_enable();
-  }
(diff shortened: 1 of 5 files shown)


g-oikonomou:
Keeping the RF on at the end of transmit() aims to avoid the OSC startup time overhead between consecutive strobes. At the end of a strobe sequence, off() is called explicitly to shut us down. Are you sure we want to change this logic?

fros4943:
@g-oikonomou Why not turn on() the radio before the first strobe, instead of assuming that all other radios also leaves the radio on?

adamdunkels:
I assume you're talking about ContikiMAC? Haven't tested this with ContikiMAC, just saw that the default behavior from other drivers was changed.

g-oikonomou:
I meant ContikiMAC, yes. It turns us on/off explicitly, so the idea was that we just linger in the on state in anticipation of next strobe. As you say, it is different logic. CPE startup overhead longer than other radios, since we have to power the domain, boot the thing, patch it and fire up the OSC - not sure if other drivers have to do so many things. By the way this logic was there pre-cc1310.

I also do not understand the change in lpm.c. Will it not schedule the wakeup earlier than desired?

g-oikonomou:
Does this RF logic not play nicely with TS RDC?

adamdunkels:
The radio driver semantics is to keep the radio in non-listening mode after transmitting, if it was in non-listening mode prior to the transmission. This is what duty cycling protocols should assume of the radio driver.

The problem in this case (and not only this case, other radios like the CC2420 also have this problem, but we have never really addressed it) is that the oscillator takes non-negligible time to start up. The real fix is to update the radio API to allow turning on and off the oscillator from the upper layer and rewriting the duty cycling protocols to take the startup time into account. Updating ContikiMAC to explicitly turn `on()` the radio before starting to send strobes like @fros4943 suggests is a much better workaround in this case. Otherwise other duty cycling protocols will misbehave (this is how we noticed it: the duty cycle was unusually high on the CC2650).

Regarding the timer issue:

The problem occurs if a c/etimer is set to expire at time _t_ and another is to expire at time _t_ + _d_. Assuming that the first timer takes time _c_ to run, the existing code would then schedule the second timer at _t_ + _c_ + _d_ instead of at _t_ + _d_ (because `RTIMER_NOW()` will be _t_ + _c_ after the timer code has been run). This patch will store the value of _t_ when the first timer expires (`last_isr_time`) so that it will be able to schedule the next interrupt at _t_ + _d_.

bkozak-scanimetrics:
I (still) don't quite understand the problem with the timer either but shouldn't you be updating ```last_isr_time``` only for ```RTC_CH1```? CH0 seems to be used for the rtimer only.

Also, couldn't you read the ```last_isr_time``` right out of the compare registers instead of calling RTIMER_NOW() (might be a little faster and more accurate I think).

adamdunkels:
The scheduling code used `RTIMER_NOW()`, so it felt safest to use the same timebase (not sure if `RTIMER_NOW()` uses a scaled version of the register value).

Yes, the `last_isr_time` could be moved into the non-rtimer handling code. As it currently stands, if the rtimer fires right in the middle of an etimer execution, `last_isr_time` would be (slightly) wrong.

g-oikonomou:
@adamdunkels wrt the timer: this is because `update_time()` in `etimer.c` is called _before_ the first process has received the event, therefore `_next_expiration_time()` does not (cannot) take into account processing time _c_. Therefore it returns "_d_ ticks ahead of _t + c_" instead of ahead of _t_ (which is what we want). Right?

g-oikonomou:
^:) I'm sorry but I'm still not getting it, `_next_expiration_time()` gives us absolute time, not count of ticks in the future. Therefore duration _c_ would be irrelevant.

What am I missing?

adamdunkels:
Yes, something like that. Or maybe the problem was that `clock_time()` originally wasn't updating itself until the next tick? Not 100% sure now that I look into the code again - I always get really confused about timing issues like these, so please sanity check what I say :P (I've written three or four tickless clock implementations for Contiki by now, each being a little different, and it always ends up messing with my head...)

g-oikonomou:
Let's form a club and print t-shirts

I am convinced by your argument to change `transmit()` and :+1: for the RF changes. How about we merged those with this pull and also fix #1229 and leave the clock changes for further investigation?

Regarding `clock_time()`: Even with the proposed changes, `count` should only ever increment by 1 at the same time as when the interrupt fires.

g-oikonomou:
@adamdunkels any chance of separating the RF changes from the clock changes? I'm mentally prepared to merge the RF stuff! :smile: 
Replicating them for CC13xx would be great too, but I can do that myself in a separate pull.

PR #1235, opened 2015-09-03
SICSLoWMAC does not handle RADIO_TX_COLLISION at all / Introduce radio_txresult_t to avoid bugs

cmorty:
NB: This is rather an issue with a path than an actual pull-request:
SICSLoWMAC does not handle RADIO_TX_COLLISION at all. By using enumerations instead of ints, bugs can be avoided, because the compiler shows a warning in some cases (here because a switch is used).

This DOES NOT fix the issue with SICSLoWMAC and it is not complete - I only fixed the radio drivers which are covered by the regression testing.
In some why this therefore also covers insufficient regression testing.

### DIFF
--- a/core/dev/nullradio.c
+++ b/core/dev/nullradio.c
@@ -11,16 +11,16 @@ init(void)
 static int
 prepare(const void *payload, unsigned short payload_len)
 {
-  return 1;
+  return RADIO_RESULT_OK;
 }
 /*---------------------------------------------------------------------------*/
-static int
+static radio_txresult_t
 transmit(unsigned short transmit_len)
 {
   return RADIO_TX_OK;
 }
 /*---------------------------------------------------------------------------*/
-static int
+static radio_txresult_t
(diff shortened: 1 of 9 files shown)


adamdunkels:
sicslowmac.c is extremely obsolete and should be removed.

cmorty:
@adamdunkels OK. So the RDC/MAC protocol options then are down to ContikiMAC, CX-MAC and nullrdc with csma or nullmac. Right? Seems like MAC-Research is dead. :)

simonduq:
@cmorty the idea of returning enums is interesting but the PR is far from being complete, and this will affect all platforms and require checking nothing breaks etc.
I also agree sicslowmac is really outdated... (would like to remove it but some older platforms still rely on it)

If you want to keep the idea alive you could maybe open an issue and close this?

simonduq:
Let me close this one.
@cmorty if you want more MAC layers you may want you look at https://github.com/contiki-os/contiki/pull/1285 ;)

PR #1240, opened 2015-09-04
Improve energest accuracy

atiselsts:
Each time when switching between modes, energest could lose a tick or more. This happened when rtimer tick counter advanced between the call of RTIMER_NOW() in ENERGEST_OFF() and the call of RTIMER_NOW() in ENERGEST_ON(), so that end of the old mode was registered at time t, but start of the new mode at time t+1 or greater.

Thanks to Carlo Boano for reporting this.

### DIFF
--- a/core/sys/energest.h
+++ b/core/sys/energest.h
@@ -105,6 +105,21 @@ extern energest_t energest_leveldevice_current_leveltime[ENERGEST_CONF_LEVELDEVI
 										energest_current_time[type]); \
 										energest_current_mode[type] = 0; \
                                        } while(0)
+
+#define ENERGEST_SWITCH(type_off, type_on) do { \
+                                             rtimer_clock_t energest_local_variable_now = RTIMER_NOW(); \
+                                             if(energest_current_mode[type_off] != 0) { \
+                                               if (energest_local_variable_now < energest_current_time[type_off]) { \
+                                                 energest_total_time[type_off].current += RTIMER_ARCH_SECOND; \
+                                               } \
+                                               energest_total_time[type_off].current += (rtimer_clock_t)(energest_local_variable_now - \
+                                                 energest_current_time[type_off]); \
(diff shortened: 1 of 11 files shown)


core/sys/energest.h
```diff
@@ -105,6 +105,21 @@ extern energest_t energest_leveldevice_current_leveltime[ENERGEST_CONF_LEVELDEVI
 										energest_current_time[type]); \
 										energest_current_mode[type] = 0; \
                                        } while(0)
+
+#define ENERGEST_SWITCH(type_off, type_on) do { \
+                                             rtimer_clock_t now = RTIMER_NOW(); \
```
adamdunkels: There is a chance this will collide with other local variables that happen to be called `now` - a better name might be `energest_local_variable_now` or something similar (which has a much lower risk of collision).

tools/cc2538-bsl
```diff
@@ -1 +1 @@
-Subproject commit 1223bfe03cdb31c439f1a51593808cdabc1939d2
+Subproject commit d6711e24ceeb1de09421166a3dc1b97378648af5
```
adamdunkels: This seems like it was unintended.

adamdunkels:
This seems like a good idea!

cmorty:
:+1: Two days ago I wrote a section on this, because I have over 100% radio-on time and quite a bit of jitter in my measurements without radio duty cycling.

simonduq:
Excellent! :+1:

core/sys/energest.h
cmorty: Well actually it can not. The life time of a variable is limited to a block. Also block-local variables hide any variable outside the block. Thus the only risks are confusing compile warnings or debug output.

adamdunkels: Yep, exactly my point.

nvt:
:+1:

PR #1242, opened 2015-09-04
Use volatile access for CMD_IEEE_TX

g-oikonomou:
As discussed very briefly in #1229, access to the data structure representing CMD_IEEE_TX should be volatile.

### DIFF
--- a/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
+++ b/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
@@ -766,7 +766,7 @@ transmit(unsigned short transmit_len)
   uint16_t stat;
   uint8_t tx_active = 0;
   rtimer_clock_t t0;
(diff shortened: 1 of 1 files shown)


g-oikonomou:
Gentle bump here: @bkozak-scanimetrics any thoughts on this one?

bkozak-scanimetrics:
It's certainly not the most pressing issue but, if the compiler ever managed to inline ```lpm_sleep() ```, ```rf_core_init_radio_op()``` or  ```rf_core_send_cmd()```  (maybe ```-flto``` gets turned or ```lpm_sleep()``` becomes a macro for example) then we can get some major problems. For example, the loop at line 818 might become an infinite loop or even get optimized out depending on how the compiler is feeling that day. 

Given the problems that can arise and the lack of any likely performance impact (at least, as far as I can tell), I think that this patch is worth applying. Also worth noting, however, is that there may be other similar issues in this driver.

PR #1243, opened 2015-09-05
CoAP: Fix handling of Token Length greater than 8

unmole:
According to the RFC, TKL greater than 8 MUST be processed as a message format error.


locomuco:
Good catch, should be merged

here the link to the RFC definition
https://tools.ietf.org/html/rfc7252#section-3

simonduq:
:+1:

simonduq:
merging in a month if there are no more comments

nvt:
:+1:

PR #1246, opened 2015-09-07
Minor fixes for the WisMote platform

nvt:
This PR changes the WisMote's default MAC and RDC layers to CSMA and ContikiMAC, respectively. Furthermore, it fixes a compilation error concerning an invalid header inclusion when the IPv4 stack is enabled.


simonduq:
:+1:

sumanpanchal:
+1

cmorty:
What about a compilation test-case? (There should be more of those!)

nvt:
Yes, good test cases are welcome contributions.

PR #1247, opened 2015-09-07
z1: sht25: Don't disable i2c

arthurfabre:
The sht25 driver disables i2c when `SENSORS_DEACTIVATE(sht25)` is used.
This is really annoying, as a bunch of other sensors rely on i2c, and if one is in the process of reading values from other sensors everything breaks.
(Also I can't seem to find any advantages of disabling i2c - it won't save power)
Other similar devices (such as the sht11) never disable i2c, they only enable it.

### DIFF
--- a/platform/z1/dev/sht25.c
+++ b/platform/z1/dev/sht25.c
@@ -51,8 +51,6 @@ configure(int type, int value)
   }
   if(value) {
     i2c_enable();
-  } else {
-    i2c_disable();
   }
   enabled = value;
   return 0;


simonduq:
looks ok for me but would like @alignan to confirm

alignan:
I2C is disabled when the sensor is deactivating mostly to force sensors to activate/configure the I2C bus previous any operation, mostly to ensure the lines/bus is OK.  The sht11 driver doesn't disable I2C as this sensor is not I2C but rather compatible bit-bang.

Ideally (for the RE-Mote it will be this way) when deactivating the I2C bus also the pins would have to be reconfigure to avoid any current draw, and save power this way.

If you need to keep the I2C bus enabled, then don't deactivate the sensor if not needed.

I'm closing the issue if there's no objection.

PR #1250, opened 2015-09-09
Fix compiler warning in er-coap.c

simonduq:
(for platforms where `size_t` is not `unisgned int`)

### DIFF
--- a/apps/er-coap/er-coap.c
+++ b/apps/er-coap/er-coap.c
@@ -110,6 +110,7 @@ static size_t
 coap_set_option_header(unsigned int delta, size_t length, uint8_t *buffer)
 {
   size_t written = 0;
+  unsigned int length_unsigned = (unsigned int)length;
 
   buffer[0] = coap_option_nibble(delta) << 4 | coap_option_nibble(length);
 
@@ -123,7 +124,7 @@ coap_set_option_header(unsigned int delta, size_t length, uint8_t *buffer)
     } else if(*x > 12) {
       buffer[++written] = (*x - 13);
     }
-  } while(x != &length && (x = &length));
+  } while(x != &length_unsigned && (x = &length_unsigned));
 
   PRINTF("WRITTEN %u B opt header\n", 1 + written);
 
@@ -478,7 +479,7 @@ coap_parse_message(void *packet, uint8_t *data, uint16_t data_len)
 
   unsigned int option_number = 0;
   unsigned int option_delta = 0;
-  size_t option_length = 0;
+  unsigned int option_length = 0;
 
   while(current_option < data + data_len) {
     /* payload marker 0xFF, currently only checking for 0xF* because rest is reserved */


apps/er-coap/er-coap.c
```diff
@@ -123,7 +123,7 @@ coap_set_option_header(unsigned int delta, size_t length, uint8_t *buffer)
     } else if(*x > 12) {
       buffer[++written] = (*x - 13);
     }
-  } while(x != &length && (x = &length));
+  } while(x != (unsigned int*)&length && (x = (unsigned int*)&length));
```
cmorty: I'm not sure how x and &length can become the same value.....

cmorty:
I think the cast is only a workaround for the real problem. `x` is unnecessary and `delta` should probably be size_t, too.

apps/er-coap/er-coap.c
simonduq: At the end of the same line I believe: `(x = (unsigned int*)&length)`.
I agree this piece of code is very obscure... seems to simply run something twice, first with `x` pointing to `delta`, second with `x` pointing to `length`.

simonduq:
You'right, there seem to be a real underlying problem when size_t is not of same size than unsigned.

apps/er-coap/er-coap.c
cmorty: Yea found that out by now, too. :( What about rewriting it to
```
    uint16_t ptr[] = {&size, $length)
    int pos;
    for(pos = 0; pos < 2; pos++){
       ptr[pos].....
```
The Compiler should be able to optimize it or even unroll.

simonduq: Sounds better, let the compiler optimize. Are you sure about the type
uint16_t instead of unsigned?

cmorty: Well I'm not sure. I wonder whether `delta` and `length` will ever be more than 2^16−269 = 65267.

g-oikonomou:
Just to add to the discussion, I would not have chosen `size_t` in the first place (both locations under discussion as well as in other places in this file).

For `_delta` and `_number`, I would have gone for `uint32_t`. Unless I am misinterpreting the RFC, the max theoretically allowed value is 65535 + 269

Let's not discuss the placement of variable definitions :)

g-oikonomou:
Mind you, [Table 4](https://tools.ietf.org/html/rfc7252#section-5.10) in the RFC specifies longest option to have a length of 1034, and [IANA option number allocations](http://www.iana.org/assignments/core-parameters/core-parameters.xhtml#option-numbers) stop at 65535.

simonduq:
OK I went for a simple solution, without the for loop because it did look that great on `coap_parse_message`.

cmorty:
Yet why not at least make `length` `unsigned int`? There is no advantage in the cast, besides hiding potential problems.

simonduq:
For consistency with all other functions and types in coap that have length as `size_t`

nfi:
There is a loop unrolled and bug fixed version of coap_set_option_header in #1216.

simonduq:
That looks perfect! (haven't had time to look at the rest of the PR yet though)

simonduq:
much nicer fix at https://github.com/contiki-os/contiki/pull/1216, closing this

PR #1251, opened 2015-09-09
Cooja: Track CFS file size

MattiasBuelens:
Previously, the Cooja mote assumed that its file was always initially empty (`file.endptr == 0`). Therefore, a file uploaded to a mote's CFS could never be read by the mote, as the mote would prevent reads from going past the EOF (indicated by `endptr`).

By tracking the file size and making it accessible to Cooja, the correct size of the uploaded file can be reported to the mote and allow it to read the uploaded file.


fros4943:
:muscle: :+1: Great!

PR #1255, opened 2015-09-11
Added Exception to symbols.h inside core/loader.

VGH05T:
symbols.h is available in source code of contiki-OS. Hence when we clone this repository, we are able to do a normal make inside any example. However, when we push our cloned folder to another repository, we do not push core/loader/symbols.h, which generates error.
Hence, i do request this pull to create an exception only to this file, allowing easy operations of contiki-OS, as in the case of avr compilations.
Thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ summary
 .DS_Store
 obj_*
 symbols.*
+!core/loader/symbols.h
 Makefile.target
 doc/html
 patches-*


simonduq:
I don't understand the problem you're describing.
What do you mean "when we push our cloned folder to another repository, we do not push core/loader/symbols.h"?
What do you mean "generates error", what error?

I don't see why we'd ignore this specific core file.

VGH05T:
First, thank you for your attention. I'm sorry for the brief description of the error, I will try to describe it better now.

### The situation ###
   I have a project using contikiOS with some integrated boards, which has Atmel microprocessor (atmega328p). This project is not supposed to be opensource until it is done, so, for now, I'm using bitbucket and a private repository.
   When I cloned the contikiOS and added some features, I did a push into my private bitbucket repository. After that, I went to another computer and cloned this private contikeOS of my repository.
   At this time, I had an error:
```
  CC        ../../core/net/llsec/ccm-star-packetbuf.c
  CC        ../../core/net/llsec/nullsec.c
cp ../../tools/empty-symbols.c symbols.c
cp ../../tools/empty-symbols.h symbols.h
  CC        symbols.c
In file included from symbols.c:1:0:
symbols.h:1:28: fatal error: loader/symbols.h: Arquivo ou diretório não encontrado
 #include "loader/symbols.h"
                            ^
compilation terminated.
```
### Fast solution ###
After some research, I found that people were having this problems too, and they did just copy the original symbols.h of contikiOS original repository into their "/core/loader/" folder.

After some study, I have seen that in each compilation, symbols.h and .c are created inside the application folder (like /example/hello-world), and for this reason, inside .gitignore of contikiOS repository we have a "symbols.*".

When we do first clone the orginal contikiOS repository, we get the file "/core/loader/symbols.h" in our folder, and everything works fine, but when we push it to another place and try to clone it again, this file isn't uploaded because of gitignore.

### Possible solution ###
Add an exception for this specific file inside .gitignore

VGH05T:
I clicked on the wrong button... I'm sorry.

simonduq:
I see, thanks for the detailed explanation. We currently ignore `symbols.*`, you're now adding an exception to not ignore the `symbols.h` from `core/loader`. Sounds reasonable.
:+1:

atiselsts:
I've met this problem before. However, I feel unconvinced about a few things in this patch.
1) Why fix this just for `symbols.*`? There are bunch of other Contiki files that will not be pushed because of `.gitignore` contents besides `symbols.*`, some of which are critically important for Contiki. For example, `Makefile.sky`, `Makefile.z1` and all other platform makefiles.
2) Can't this issue be avoided altogether with better git development practices, such as using git remotes instead of doing a clone + push?

simonduq:
OK I thought `core/loader/symbols.h` was now being ignored but that's obviously not the case as already tracked files are not ignored [1]. So I agree with @atiselsts, with good git practices there should be no issue.

[1] http://git-scm.com/docs/gitignore

simonduq:
@VGH05T, I propose to close this PR

VGH05T:
Thank you for this discussion, I truly agree with @atiselsts and agree closing this topic, but with some final words:
* The best solution for this problem is better git practices indeed;
* Still, this problem isn't that obviously solved, because symbols.h is also copied from empty_symbols.h during compilation time due to makefile rules. That causes a strange feeling about why isn't this new created symbols.h used by my application example instead of that in ```core/loader/symbols.h```?

Perhaps adding this discussion somewhere may prevent waste of time for new developers, about this matter.

Thank you!

PR #1256, opened 2015-09-14
Refactor tcpip.c packet input

vsaw:
The UIP_CONF_FORWARD is now handled in a better way that the general structure
of packet_input can be clearer seen.

### DIFF
--- a/core/net/ip/tcpip.c
+++ b/core/net/ip/tcpip.c
@@ -185,30 +185,17 @@ check_for_tcp_syn(void)
 static void
 packet_input(void)
 {
-#if UIP_CONF_IP_FORWARD
   if(uip_len > 0) {
+
+#if UIP_CONF_IP_FORWARD
     tcpip_is_forwarding = 1;
-    if(uip_fw_forward() == UIP_FW_LOCAL) {
+    if(uip_fw_forward() != UIP_FW_LOCAL) {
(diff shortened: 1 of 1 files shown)


core/net/ip/tcpip.c
```diff
@@ -185,46 +185,32 @@ check_for_tcp_syn(void)
...
+#endif /* UIP_CONF_IP_FORWARD */
+
     check_for_tcp_syn();
     uip_input();
     if(uip_len > 0) {
 #if UIP_CONF_TCP_SPLIT
       uip_split_output();
 #else /* UIP_CONF_TCP_SPLIT */
-#if NETSTACK_CONF_WITH_IPV6
+#if UIP_CONF_IPV6
```
simonduq: `UIP_CONF_IPV6` is deprecated, `NETSTACK_CONF_WITH_IPV6` is to be used as a replacement

vsaw: @simonduq thanks. good catch! It's fixed now

simonduq:
Excellent refactor, makes the code flow so much clearer! :+1: 
I wish we could also get rid of `tcpip_is_forwarding`, but it is used for aodv on the cooja platform apparently...

nvt:
:+1:

PR #1257, opened 2015-09-14
CC26xx - fully shutdown the rf-core & abort rx op

bkozak-scanimetrics:
Prior to this patch, the ieee radio driver did not explicitly abort the rx operation or power down the analog components of the radio prior to shutting down the rf-core.

The result of this was that the rf-core continued to use a lot of power even while "off".

This patch fixes this problem.


cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
```diff
@@ -1101,6 +1101,8 @@ off(void)
 
   while(transmitting());
 
+  /* stopping the rx explicitly results in lower sleep-mode power usage */
+  rx_off();
```
g-oikonomou: @bkozak-scanimetrics I'll be testing this over the weekend. Meanwhile, do we _really_ need this call here? Surely powering down analog in addition to the PD (which is what your changes to `rf_core_power_down()` should do under all circumstances) will result in the same current draw?

g-oikonomou:
I've been testing this and it looks good, plus it has all the positive comments from the community, so I'm ready to merge it.

The only pending question is the one I asked above. In case the answer is positive, then ideally we want the same change applied to the prop mode driver. I somehow suspect though that the changes to `rf-core.c` are all we need here.

cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
bkozak-scanimetrics: I did test this and I actually did find that the rx_off() is necessary to get the full power reduction. You're logic seems like it should be correct however so maybe I did something wrong in my tests. I can try it again when I have some more time.

g-oikonomou: I thought about this a bit more after posting. Perhaps `CMD_FS_POWERDOWN` can only be accepted if there are no running ops, I'll have a test myself when I get a chance to see what the command returns if we don't abort first.

bkozak-scanimetrics: Just tested this and it's exactly as you say. The FS power-down fails if ```rx_off()`` is not called first. The ```CMDSTA``` return code is 0x86 (SchedulingError).

I don't have any CC13xx hardware or knowledge so if you want to add anything for prop-mode in this PR you'll have to test yourself and tell me what to do.

bkozak-scanimetrics:
Merged in commits from @g-oikonomou to also fully shutdown the CC13xx rf-core fully.

bkozak-scanimetrics:
Not sure if it makes sense to keep the merge commit or not. Let me know.

g-oikonomou:
If you can be bothered with rebasing my two commits manually in your branch then that's fine, otherwise I'm happy to pull this one in with the merge commit.

bkozak-scanimetrics:
Should be good to go now.

g-oikonomou:
I am not 100% sure I understand this statement :) Did you mean to remove the merge commit?

bkozak-scanimetrics:
Yes, merge commit is removed.

g-oikonomou:
Many thanks!

PR #1266, opened 2015-09-21
AVR revival: new battery and temperature sensors and er-rest-example for the Raven

drugo72:


malvira:
very cool and related to #1267. +1

simonduq:
:+1:

PR #644, opened 2014-04-16
Network filter for contiki

lebrush:
Implements capabilities for filtering TCP/UDP packets before they reach the uip interface. Currently only for input packages.

PR #1268, opened 2015-09-22
Network filter for contiki

vsaw:
Offers capabilities for filtering TCP/UDP packets before they reach
the uip interface. Currently only for input packages.

This is a rebase of https://github.com/contiki-os/contiki/pull/644 as the original author did not respond within 25 days.

### DIFF
--- a/core/contiki-default-conf.h
+++ b/core/contiki-default-conf.h
@@ -178,13 +178,19 @@
 #define UIP_CONF_MAX_CONNECTIONS 8
 #endif /* UIP_CONF_MAX_CONNECTIONS */
 
-
 /* UIP_CONF_TCP_SPLIT enables a performance optimization hack, where
    each maximum-sized TCP segment is split into two, to avoid the
    performance degradation that is caused by delayed ACKs. */
 #ifndef UIP_CONF_TCP_SPLIT
 #define UIP_CONF_TCP_SPLIT 0
 #endif /* UIP_CONF_TCP_SPLIT */
+
+/* UIP_CONF_NETFILTER_ENABLE specifies if the netfilter is enabled
+   on the platform and therefore if the incoming packets should be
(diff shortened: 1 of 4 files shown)


simonduq:
Thanks for the rebase.
Two comments:
* Is "filter" really this right name? It feels more like a sniffer or callback interface, but doesn't allow to select which packets will get in or not.
* I'm not comfortable with unused new features. Would need at least one example using it (or even better, a regression test for it).

laurentderu:
It's rather a sniffer or a hook, so a name change would be appropriate. A bigger issue is that the function is called before packet validation and extension header processing done in uip6.c:uip_process(). 
Anyway, this feature is a nice tool for example to build traffic statistics without having to hack the Contiki stack. By making it into a proper filter (returning a flag indicating if the packet should be handled or dropped), it could be used to make proxy application.

vsaw:
It can be useful as a transparent proxy if used in the border router e.g.

**DNS Trimmer**
Assuming a highly constrained node with `UIP_CONF_BUFFER_SIZE` set to less than 576 bytes (the minimum IPv4 Datagram size a client must be able to receive) wants to a DNS resolution. The DNS Server might respond to the query in with packet containing loads of Authority RRs and Additional RRs. (See http://www.networksorcery.com/enp/protocol/dns.htm) these will often bloat the DNS Response so that the node can not process them anymore. Here the border router running the netfilter can easily strip the DNS responses or intercept requests and respond with data from it's own cache.

Other than that many more options are thinkable:

* Network statistics
* IPv6 NAT on the Border Router
* Traffic Redirection (e.g. create a (D)TLS endpoint on the border router to transparently protect outgoing packets for nodes that can not do transport security themselves)
* ...

simonduq:
There is no supporters of this PR, will close in a month if nobody speaks for it

simonduq:
closing after timeout

PR #222, opened 2013-05-15
More flexibility for the fallback interface

lebrush:
When there's a failure to send a packet in the fallback interface the
destination is unreachable and therefore the source should be notified.

tadoº GmbH (www.tado.com)

PR #1269, opened 2015-09-22
More flexibility for the fallback interface

vsaw:
When there's a failure to send a packet in the fallback interface the
destination is unreachable and therefore the source should be notified.

This is a rebase of #222 as the original author did not respond within 27 days.

### DIFF
--- a/core/net/ip/tcpip.c
+++ b/core/net/ip/tcpip.c
@@ -586,7 +586,16 @@ tcpip_ipv6_output(void)
 	    /* This should be copied from the ext header... */
 	    UIP_IP_BUF->proto = proto;
 	  }
-	  UIP_FALLBACK_INTERFACE.output();
+	  /* Inform the other end that the destination is not reachable. If it's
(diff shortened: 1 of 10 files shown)


simonduq:
Really shows someone cares about this new feature, that's a good thing!
From the comments in https://github.com/contiki-os/contiki/pull/222 this could get in now. I will merge in a little while to give others a last chance to review this. :+1:

laurentderu:
In the meantime IP64 has been introduced and it also uses uip_fallback_interface, internally for the actual physical interface but also to connect to netstack, it must be updated too.

vsaw:
@laurentderu Good point. I would have thought though that Travis would have detected it via a broken build because of the API change?

simonduq:
I'm guessing you only got a warning for this, and we don't have a test for warnings (we should!)

vsaw:
Ping :)

simonduq:
Right, merging!

PR #1271, opened 2015-09-22
RPL: fix clock_time_t calculations

drugo72:
This issue happened with the AVR Raven. When used with the Cetic 6lbr border router, the computation of the RPL DAO lifecycle timer was incorrect due to the limited clock_time_t width of the avr-raven and the default, relatively large values, of the DIO default lifecycle and lifecycle unit used by the Cetic 6lbr (30 and 256 respectively).

Do not assume clock_time_t is wide enough on all platforms but cast to the widest type avaliable when doing calculations.

### DIFF
--- a/core/net/link-stats.c
+++ b/core/net/link-stats.c
@@ -51,7 +51,7 @@
 /* Maximum value for the freshness counter */
 #define FRESHNESS_MAX                   16
 /* Statistics with no update in FRESHNESS_EXPIRATION_TIMEOUT is not fresh */
-#define FRESHNESS_EXPIRATION_TIME       (10 * 60 * CLOCK_SECOND)
+#define FRESHNESS_EXPIRATION_TIME       (10 * 60 * (clock_time_t)CLOCK_SECOND)
 
 /* EWMA (exponential moving average) used to maintain statistics over time */
 #define EWMA_SCALE            100
@@ -206,6 +206,6 @@ void
 link_stats_init(void)
 {
   nbr_table_register(link_stats, NULL);
-  ctimer_set(&periodic_timer, 60 * CLOCK_SECOND * FRESHNESS_HALF_LIFE,
+  ctimer_set(&periodic_timer, 60 * (clock_time_t)CLOCK_SECOND * FRESHNESS_HALF_LIFE,
       periodic, NULL);
 }


laurentderu:
You could still have an overflow when casting  expiration_time into clock_time_t when calling ctimer_set().
Maybe a safer approach would be to define clock_time_t as an unsigned long for AVR platforms too like is is done on all the other platforms ?

drugo72:
Thanks for reviewing it.
Correct. However it's better than before and it's good practice to cast to a wider type in intermediate multiplications.
About changing the type clock_time_t for the Raven, I've considered this and I'll take a closer look (I have a PR for AVR fixes too).
Ciao,
  Cristiano

joakimeriksson:
Looks ok even if I think the expiration time variable might as well be clock_time_t unless you make use of its size. You could for example compare expire == (clock_time_t) expire and see if there is a difference - and if so it is likely a wrap - then it might make sense to wait the maximum time for that size of clock_time_t instead of a wrapped time.

cdealti:
@joakimeriksson thanks for the feedback.
Something like this?
```
unsigned long expire;
...
if (expire !=  (clock_time_t)expire) {
    expire = (clock_time_t)ULONG_MAX;
}
...
```
I can do this but I wonder in how many other places we should do something similar.

nvt:
You don't need to cast every operand in the arithmetic expression since they are promoted to the largest type among the operands according to C's conversion rules. Also, the check for too large times for clock_time_t to hold is good too add, as already pointed out.

cdealti:
You are right about the implicit conversions but they are often difficult to understand and quite error prone. I found that explicit casts and strongly-typed constants are more easily understood.
I can remove the cast from all operands but the first if so desired and add the above saturation code to prevent rollover.

nvt:
Sounds good. The casting is not necessary to change if you prefer that more explicit style. One other observation, however: we may want to enclose the division <code>CLOCK_SECOND / 2UL</code> in parentheses to enforce an order of calculation that could reduce the risk of overflow further.

drugo72:
Hello, I've checked and the largest value of ```CLOCK_SECOND``` in the codebase is currently 1024.
The RPL Default Lifetime in an 8 bit quantity and the Lifetime Unit is a 16 bit quantity.

So it should be safe for now do the multiplication on an uint32_t.
I'm not comfortable on doing the multiplication on 64 bit integers because it might increase the code size.

Doing the division first causes precision loss.

My proposal is to cast only the first operand to uint32_t and let the compiler do the type promotions for the others.

The division by 2 can be replaced by a right shift by one.

The  saturation code for the expiration time will be the one in my previous post or maybe

```
uint32_t expire;
...
if (expire > (clock_time_t)~0) {
    expire = (clock_time_t)~0;
}
...
```
assuming that a pattern of all ones is the largest value representable by the underlying ```clock_time_t``` type integer (assuming it's an integer.

Than I guess a careful code inspection will be needed to address the same potential issue in the existing codebase ;)

drugo72:
I'm/was not comfortable on doing the multiplication on 64 bit integers because it increases the code size (+150 bytes on the Raven) and consume extra cycles. But of course this is the safest approach.

I've checked and the largest value of CLOCK_SECOND in the codebase is currently 1024.
The RPL Default Lifetime in an 8 bit quantity and the Lifetime Unit is a 16 bit quantity.
The multiplication could exceed 32 bits but this is unlikely if Default Lifetime and Lifetime Unit are reasonably small.

However, the first multiplication is between 32 bit integers (safe).
The second can exceed 32 bits so it's done with 64 bit integers.
The division has been replaced by a right shift

Then we have the test to saturate the result to the maximum clock_time_t value.

The ```%llu``` conversion in printf should be replaced by the ```PRIu64``` but I cannot make it to compile so don't be too strict here.

Than I guess a careful code inspection will be needed to address the same issue in the existing codebase ;)

cdealti:
Travis CI build failure:

It looks like ```uint64_t``` is not supported by SDCC:

http://sdcc.sourceforge.net/mediawiki/index.php/Standard_compliance

I'll try replacing it with ```unsigned long long int```

This PR is a pain for such a stupid thing

cmorty:
@cdealti: If understand things right, SDCC either does not support 64bit types or has a bug: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdint.h.html

simonduq:
This makes me realize `get_probing_target` in `rpl-timer.c` is also broken on platforms where `clock_time()` wraps within the network lifetime, because all time comparisons are done with `<` (rather than `CLOCK_LT`, which is not currently defined on all platforms)

drugo72:
Those ```CLOCK_LT``` macros seem broken too. For example if you compile the following program
```
#include <stdio.h>
#include <limits.h>

#define LT(a,b) ((int)((a)-(b)) < 0)

/*
  In the following it is assumed that a and b are ticks
  of a timer counter whose width is unsigned int.

  It is assumed that b is the counter value of a time tb which is
  in the future. In other words tb > ta.

  Also, it is assumed that (tb - ta) <= timer period 
  (the time it take to wrap starting from 0).

  Note that, under the above assumptions the difference (b - a) is always
  correct (modular arithmetic) while the "less then" comparison
  gives different results (it should always evaluate to 1).
*/
int main(int argc, char *argv) {
  unsigned int a, b;

  a = 0;
  b = UINT_MAX;

  printf("a: %u, b: %u, a - b: %u, b - a: %u, LT(a,b) == %d\n",
         a, b, a - b, b - a, LT(a,b));

 
  a = UINT_MAX/2;
  b = a + 1;

  printf("a: %u, b: %u, a - b: %u, b - a: %u, LT(a,b) == %d\n",
         a, b, a - b, b - a, LT(a,b));


  a = UINT_MAX;
  b = 0;

  printf("a: %u, b: %u, a - b: %u, b - a: %u, LT(a,b) == %d\n",
         a, b, a - b, b - a, LT(a,b));


  a = 1;
  b = UINT_MAX + 1;

  printf("a: %u, b: %u, a - b: %u, b - a: %u, LT(a,b) == %d\n",
         a, b, a - b, b - a, LT(a,b));

  return 0;
}
```

The output is

```
a: 0, b: 4294967295, a - b: 1, b - a: 4294967295, LT(a,b) == 0
a: 2147483647, b: 2147483648, a - b: 4294967295, b - a: 1, LT(a,b) == 1
a: 4294967295, b: 0, a - b: 4294967295, b - a: 1, LT(a,b) == 1
a: 1, b: 0, a - b: 1, b - a: 4294967295, LT(a,b) == 0
```

Did I miss something?

simonduq:
The output seems correct, `LT` is a signed comparison. The idea is to keep consistent time comparison even across `UINT_MAX` boundaries. For instance if `t0` is `UINT_MAX` and `t1` is 1 clock tick later, you have `t1 == 0` and `LT(t0,t1) == 1`.

simonduq:
Something else: I'm not sure about changing the definition of `RPL_PROBING_INTERVAL` and `RPL_PROBING_EXPIRATION_TIME`. They use `CLOCK_SECOND` and shouldn't have to cast to `clock_time_t` explicitly.

Instead, shouldn't we include the cast to the `CLOCK_SECOND` definition on AVR platforms? (and platforms with `clock_time_t` larger than `int` in general)

drugo72:
> The output seems correct

Actually, it is not. In the four cases above b is alway in the future.
In the first and fourth it is in the future of UINT_MAX ticks. ```LT``` should evaluate to 1.
You can try also the following cases:
```
  a = 0;
  b = UINT_MAX/2 + 1;
```
where ```LT``` evaluates to 1 (correct).

and 

```
  a = 0;
  b = UINT_MAX/2 + 2;
```
where ```LT``` evaluates to 0 (wrong).

So ```LT``` does not seem to work in all cases.

> They use CLOCK_SECOND and shouldn't have to cast to clock_time_t explicitly.

In the original code

```#define RPL_PROBING_EXPIRATION_TIME (10 * 60 * CLOCK_SECOND)```

With avr-gcc these literals are of type ```int```.
With this compiler ```int``` has size 2, hence the above operation causes a true overflow for signed integers.
Casting the ```CLOCK_SECOND``` to ```clock_time_t``` forces the arithmetic on some unsigned integer type.
It would be probably better to define every ```CLOCK_CONF_SECOND``` like this:

```#define CLOCK_CONF_SECOND (clock_time_t) some_value```
 
Please note that I merely wanted to remove the avr-gcc warning. The right fix requires that the result of the above macro and  results using the macro do not exceed the maximum representable value of ```clock_time_t```.

**Comment by simonduq** [comment]
What you describe is the intended behavior of `CLOCK_LT`. Guarantees correct comparison between two times that differ by no more than `UINT_MAX/2`. For anything with larger difference, it is impossible to tell if `b` is in the future (`b == a + delta`) or if `a` is in the future but has warped (`a == (b + delta) % UINT_MAX`).

As for your comment on the cast, seems like we're on the same page. Platforms where  `clock_time_t` is larger than `int` should make sure their `CLOCK_SECONDS` is cast to `clock_time_t`.

**Comment by drugo72** [comment]
@simonduq  There are many implicit assumptions in all these calculations.

For example we assume that ```clock_time_t``` and CLOCK_SECONDS are such that some (all?) time deltas are within half of the clock wrap period.

Well, all this may work if ```clock_time_t``` is ```uint32_t``` and ```CLOCK_SECONDS``` is 1000 as it is in most platforms.

If we allow that ```clock_time_t``` is a 16 bit unsigned integer and ```CLOCK_SECONDS``` is 128, like in the raven,  then we also need to check the result of every calculation and saturate it to the maximum value representable by ```clock_time_t```. This is error prone and increases code size.

 @laurentderu I'm now convinced you were right sice the beginning and it would be better to require a fixed size for ```clock_time_t``` (for all platforms). So the plan is to do that for the raven and delete this PR.

**Comment by simonduq** [comment]
I fully agree with that. Contiki needs to schedule timers tens of minutes
or hours in the future to use RPL. We should fix the clock of platforms
unable to support that rather than try workarounds in RPL.

**Comment by drugo72** [comment]
Would ```uint32_t``` do for ```clock_time_t```?
Please note that even with this time it's still possible that some intermediate RPL calculations exceed the maximum representable value of ```clock_time_t``` (just less likely to happen so we can ignore it).

```CLOCK_SECONDS``` should be defined such that it expands to a cast expression, for example ```((clock_time_t)128)```

Your observation about the broken time comparison above should be addressed.

Is this the plan?

**Comment by simonduq** [comment]
Sounds like a great plan yes!
Do you want to do this for Raven only or AVR in general?

As for the broken comparison in probing, feel free to fix but this can also be another PR. Requires `CLOCK_LT`, which currently is not defined on all platforms.

**Comment by drugo72** [comment]
Actually the change is quite simple because... well, apparently it's already supported. In every AVR-based platform there are these lines
```
/* The AVR tick interrupt usually is done with an 8 bit counter around 128 Hz.
 * 125 Hz needs slightly more overhead during the interrupt, as does a 32 bit
 * clock_time_t.
 */
/* Clock ticks per second */
#define CLOCK_CONF_SECOND 125
#if 1
/* 16 bit counter overflows every ~10 minutes */
typedef unsigned short clock_time_t;
#define CLOCK_LT(a,b)  ((signed short)((a)-(b)) < 0)
#define INFINITE_TIME 0xffff
#define RIME_CONF_BROADCAST_ANNOUNCEMENT_MAX_TIME INFINITE_TIME/CLOCK_CONF_SECOND /* Default uses 600 */
#define COLLECT_CONF_BROADCAST_ANNOUNCEMENT_MAX_TIME INFINITE_TIME/CLOCK_CONF_SECOND /* Default uses 600 */
#else
typedef unsigned long clock_time_t;
#define CLOCK_LT(a,b)  ((signed long)((a)-(b)) < 0)
#define INFINITE_TIME 0xffffffff
#endif
``` 

Where only the ```CLOCK_CONF_SECOND``` change among platforms (it's 125 or 128).

I can improve it a little bit like this, also fixing a couple of bugs:

```
/* The AVR tick interrupt usually is done with an 8 bit counter around 128 Hz.
 * 125 Hz needs slightly more overhead during the interrupt, as does a 32 bit
 * clock_time_t.
 */
/* Clock ticks per second */
#define CLOCK_CONF_SECOND 128

#if 0
/* 16 bit counter overflows every ~10 minutes */
typedef uint16_t clock_time_t;
#define CLOCK_LT(a,b)  ((int16_t)((a)-(b)) < 0)
#define INFINITE_TIME UINT16_MAX
#define RIME_CONF_BROADCAST_ANNOUNCEMENT_MAX_TIME INFINITE_TIME /* Default is 3600 s */
#define COLLECT_CONF_BROADCAST_ANNOUNCEMENT_MAX_TIME INFINITE_TIME /* Default is 600 s */
#else
typedef uint32_t clock_time_t;
#define CLOCK_LT(a,b)  ((int32_t)((a)-(b)) < 0)
#define INFINITE_TIME UINT32_MAX
#endif
```
When building er-rest-example-raven this results in +700 bytes of code size.

Note that I did not change the definition of ```CLOCK_SECONDS``` to

```#define CLOCK_CONF_SECOND ((clock_time_t)128)```

because this does not compile due to some usages of this macro in other preprocessor expressions.
I've also tried to define ```CLOCK_SECONDS``` to 128 (an int) or 128UL (an unsigned int) accordingly to the ```clock_time_t``` type but in the latter case it increases the code size of 40 bytes because this causes type promotions in every expression using the constant.

In the end, I think I'm ready to delete commit  80555ea but not f6e79ff which forces the multiplication on a more suitable type.

Sorry, but I cant't find a better solution.

simonduq:
Great! With your compelling argument I'm ok with https://github.com/contiki-os/contiki/commit/f6e79ff29eecfd733e8c9a26e1c6505d28e4f733
What would you think about selecting unsigned long clocks whenever RPL (or even IPv6) is enabled?

cdealti:
@simonduq is there any interest to merge this or should I close it?

simonduq:
Right, sorry, the PR ended up burried under new ones (and I had very little Contiki time lately).
Setting a timeout for merging, i.e. if nobody complains in the coming month I'll merge. :+1: 
Rebase needed. Thanks

cdealti:
It looks like the original issue has been resolved in master.
However a similar issue has been introduced in ``core/net/link-stats.c``.
I'll update this PR to correct the new issue.

alignan:
@simonduq want to merge this now?

simonduq:
if you also approve yes sure!!

PR #1274, opened 2015-09-23
Regression test for compiler warnings

simonduq:
The PR adds regression testing for the number of compiler warnings. It is enabled for all compilations in compile-base, the ARM and NXP ports. The makefile variable `EXAMPLES` reflects the current number of warnings for every build as follows:
`example-path:target[:warning-count]` (no `warning-count` means 0 warning)

* The test passes iff the warning count is unchanged;
* If the warning count decreases, the variable `EXAMPLES` must be update to reflect the progress;
* If the warning count increases, new warnings must be fixed.

### DIFF
--- a/regression-tests/01-compile-base/Makefile
+++ b/regression-tests/01-compile-base/Makefile
@@ -1,39 +1,41 @@
 EXAMPLESDIR=../../examples
 TOOLSDIR=../../tools
 
+ENFORCE_WARNING_COUNT = 1
+# Format: example-path:target[:warning-count]
 EXAMPLES = \
-hello-world/avr-raven \
-hello-world/exp5438 \
-hello-world/eval-adf7xxxmb4z \
-hello-world/micaz \
-hello-world/minimal-net \
-hello-world/native \
-hello-world/sky \
-hello-world/wismote \
-hello-world/z1 \
-eeprom-test/native \
-collect/sky \
-er-rest-example/wismote \
-example-shell/native \
-netperf/sky \
-powertrace/sky \
-rime/sky \
-rime/z1 \
-ravenusbstick/avr-ravenusb \
-servreg-hack/sky \
-sky/sky \
-sky-ip/sky \
-sky-shell/sky \
-sky-shell-exec/sky \
-sky-shell-webserver/sky \
-tcp-socket/minimal-net \
-telnet-server/minimal-net \
-webserver/minimal-net \
-webserver-ipv6/eval-adf7xxxmb4z \
-wget/minimal-net \
-z1/z1 \
-settings-example/avr-raven \
-ipv6/multicast/sky \
+hello-world:avr-raven:6 \
+hello-world:exp5438:8 \
+hello-world:eval-adf7xxxmb4z:1 \
+hello-world:micaz:6 \
+hello-world:minimal-net:1 \
+hello-world:native:5 \
+hello-world:sky:5 \
+hello-world:wismote:6 \
+hello-world:z1:7 \
+eeprom-test:native:6 \
+collect:sky:21 \
+er-rest-example:wismote:8 \
+example-shell:native:54 \
+netperf:sky:21 \
+powertrace:sky:7 \
+rime:sky:7 \
+rime:z1:9 \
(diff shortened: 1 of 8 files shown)


simonduq:
Reworked this for easier maintenance: now instead of one file per build, the warning count is specified directly in the makefile. The end goal being to achieve a 0-warning Contiki of course :)

joakimeriksson:
Very nice! I think this is a good test - and the aim of zero warnings will increase code quality in the long run! :+1:

sieben:
@simonduq Do you think you could include the scan-build one into this PR?

simonduq:
I didn't because I was unsure what scan-build exactly does, how it is used and if the "OK" and "FAIL" output was of any use for it. Can you say a little more on how scan-build can also benefit from warning counting? How do we check the current warning count and know if we have increased or decreased it?

sieben:
For reference go check out this PR: https://github.com/contiki-os/contiki/pull/698
Basically scan-build do some static analysis and find code path that will cause errors and failures.
As you can see in http://leone.iiens.net/scan-build-2014-05-28-175201-3054-1/ there is a quantity of warning for each category. You could imagine a count for every project/example and go from there.

Do you think this could be interesting and worthwhile doing?

adamdunkels:
How about simply turning on `-Werror` instead? Or are there any good reasons for allowing more than 0 warnings for certain platforms? (E.g. lack of a maintainer for some platforms, which makes it difficult to get all warnings fixed.) Compiling Contiki cleanly with `-Werror` certainly is possible with a bit of work.

adamdunkels:
@sieben: static analysis sounds like a really good idea. Would be great to be able to run that as part of the per-PR regression testing!

joakimeriksson:
I would be ready to help if we go all the way for zero warnings and I am sure there are more that would like that (I could do PRs for Sky and other things). -Werror should be what we aim for in the long run! (we did that in our "fork" project and sometimes it is a pain - but it is worth it since it will capture some actual bugs now and then).

simonduq:
@adamdunkels the goal is exactly this, enabling `-Werror` eventually. This new test is not more than a  tool (maybe a temporary one) to help us get there and avoid any regression in the process.

Some of the current warnings point to real bugs that sometimes require a non-trivial fix (e.g. in coap, where compiling for jn516x revealed a bug https://github.com/contiki-os/contiki/pull/1250, with a fix in the pending PR https://github.com/contiki-os/contiki/pull/1216, or in mtarch for msp430x https://github.com/contiki-os/contiki/pull/1169). Waiting for everything to be fixed, checking the warning count will avoid new warnings to get in (see https://github.com/contiki-os/contiki/pull/1269, which would have added a warning if @laurentderu didn't find the problem).

simonduq:
@sieben here you go :) (the warning count is not pretty though)

cmorty:
@simonduq No number for the scan-build path?

simonduq:
what do you mean @cmorty?

simonduq:
Ah ok that's a question to @sieben I guess :)
No the `scan_build` dir is not prefixed by a number, I guess simply because it does not run in Travis.

sieben:
@simonduq is right. Do we want this kind of static analysis deployed now? What about #771 in that case... ;)

adamdunkels:
I think it is way better to get a clean build quickly instead of adding mechanisms to count a (rather arbitrary) number of warnings that we allow. As you say, some of the warnings come from real problems with the code. If we allow _x_ warnings, then fix one the problems but unknowingly introduce another problem, we'll still have _x_ warnings and not know that we had a new problem. (Plus, it looks bad enough that we have warnings in the code. Adding mechanisms that say that we are _really_ ok with having warnings looks _really_ bad.)

I would rather fix all the warnings (and the real problems behind them), and then have a new PR that just turns on `-Werror` for all regression builds.

In any case, great that you brought this up! We should definitely have a warning-free build ASAP.

simonduq:
Fair enough.
I was about to close the PR, but do you think we should keep it open (and rebased from time to time) as a tool to find where there are warning left?

adamdunkels:
Let's close this one and open one with the true `-Werror` acid test once those errors are fixed instead.

PR #1275, opened 2015-09-23
Ignore various build and temporary files

bkozak-scanimetrics:
Ignore:
- .cooja (elf files generated by cooja builds)
- .swp (vim editor temporary files)
- doc/latex/ (generate with ```make -C doc pdf```)
- various build & temporary files generated in regression-tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ obj_*
 symbols.*
 Makefile.target
 doc/html
+doc/latex
 patches-*
 tools/tunslip
 tools/tunslip6
@@ -86,8 +87,14 @@ contiki-cc2530dk.lib
 *.d71
 *.d81
 
+# Cooja Build Artifacts
+*.cooja
+
 #regression tests artifacts
 *.testlog
+*.log.prog
+regression-tests/[0-9][0-9]-*/report
+regression-tests/[0-9][0-9]-*/org/
 
 # rl78 build artifacts
 *.eval-adf7xxxmb4z


sieben:
:+1: That said, editor spécific such as vim can be managed by a system wide .gitconfig

g-oikonomou:
Ignoring files generated by our own builds I agree with. .swp I'd keep in my local `.git/info/exclude`, which is what I also do for eclipse `.project`, `.cproject` and other similar files.

oliverschmidt:
Full ACK. Had the same discussion on other GH projects. IHO only files generated by the builds which are part of the GH project (here: Makefiles) belong in .gitignore files which are part of the GH project.

bkozak-scanimetrics:
> That said, editor spécific such as vim can be managed by a system wide .gitconfig 

Fair enough. Wasn't sure about that since there does seem to be an editor temporary pattern in there already.

Will change.

bkozak-scanimetrics:
All done

g-oikonomou:
This comes across as a no-brainer. Any reasons why not to pull it in?

.gitignore
```diff
@@ -86,8 +87,15 @@ contiki-cc2530dk.lib
 *.d71
 *.d81
 
+# Cooja Build Artifacts
+*.cooja
+!Makefile.cooja
```
simonduq: I don't think `!Makefile.cooja` is needed, as the file is already checked in, and no such file is generated (or am I missing something?). Other platforms don't have this line.

g-oikonomou:
```
$ l platform/cooja/Makefile.co*
-rw-r--r--  1 cexgo  staff  2843 24 Oct 19:04 platform/cooja/Makefile.cooja
```

so the file in question is already under version control (?)

simonduq: yes, right, and `.gitignore` doesn't apply to files already tracked

bkozak-scanimetrics: It really isn't needed. 

In my mind it makes sense to exclude it just to make the .gitignore match only what is needed and no more.  Presumably any file called Makefile.cooja is not a generated binary file but all other .cooja files are.

simonduq: Ready to merge after you update the PR!

bkozak-scanimetrics:
Removed specific exclusion of Makefile.cooja as requested by @simonduq

simonduq:
:+1:

PR #1280, opened 2015-09-27
Made CC1200 test mode and rx watchdog selectable

alignan:
Made `USE_RX_WATCHDOG` configurable at i.e `project-conf.h`.  Also changed `RF_TESTMODE` to `CC1200_RF_TESTMODE` to allow configuring per application.

### DIFF
--- a/dev/cc1200/cc1200-conf.h
+++ b/dev/cc1200/cc1200-conf.h
@@ -73,6 +73,17 @@
 #define CC1200_MAX_PAYLOAD_LEN           127
 #endif
 /*---------------------------------------------------------------------------*/
+/* 
+ * The RX watchdog is used to check whether the radio is in RX mode at regular
+ * intervals (once per second). Can be used to improve reliability especially
+ * if NullRDC is used. Turned of by default. 
+ */ 
+#ifdef CC1200_CONF_USE_RX_WATCHDOG
+#define CC1200_USE_RX_WATCHDOG CC1200_CONF_USE_RX_WATCHDOG
+#else
+#define CC1200_USE_RX_WATCHDOG 0 
+#endif
+/*---------------------------------------------------------------------------*/
 /*
  * Use 802.15.4g frame format? Supports frame lenghts up to 2047 bytes!
  */
--- a/dev/cc1200/cc1200.c
+++ b/dev/cc1200/cc1200.c
@@ -57,8 +57,7 @@
  * - 3: Print errors + warnings + information (what's going on...)
  */
 #define DEBUG_LEVEL                     2
-/* If set to 1, we use a timer to poll RX mode inside the cc1200 process */
-#define USE_RX_WATCHDOG                 1
+
 /*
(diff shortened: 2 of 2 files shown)


uknoblic:
Hi, 

i agree with this kind of modifications (in fact i already made the same changes regarding the watchdog as we are currently testin contikimac where the watchdog makes no sense). But i would propose to place these changes in cc1200-conf.h where all the other options are placed. If you don't mind, please use the following code to extend the cc1200-conf.h:

/*
 * The RX watchdog is used to check whether the radio is in RX mode at regular
 * intervals (once per second). Can be used to improve reliability especially
 * if NullRDC is used. Turned of by default.
 */
#ifdef CC1200_CONF_USE_RX_WATCHDOG
#define CC1200_USE_RX_WATCHDOG          CC1200_CONF_USE_RX_WATCHDOG
#else
#define CC1200_USE_RX_WATCHDOG          0
#endif

In addition, there is a bug with regard to the etimer used for the watchdog, should we combine these modifications or do you propose to split them in seperate PRs?

alignan:
I would make those changes in a separate PR, I will update my PR with your proposed changes.  Thanks!

tussock-cam:
@uknoblic Hi Ulf,

I'm working on a platform that uses your cc1200 driver, and am having a few issues with ContikiMAC.. Does the etimer issue you mentioned cause contikimac to behave oddly? I'm just about to start debugging it, but it seems to be waking up at weird times, etc.. Figure it was worth checking if this is related to the watchdog at all...

uknoblic:
Hi,
there are a few issues i found in the driver - the RX watchdog (etimer) is just one of it. Which processor do you use? I would propose to disable the RX watchdog (it makes no sense for Contikimac), but i don't believe that this causes your problems. We have contikimac running with the cc2538...

tussock-cam:
@uknoblic I'm also using the cc2538 and am noticing quite a few assertion failures. While debugging I've come across your `set_channel` function and notice that the chip enters idle fine, but after writing `CC1200_FREQ0` it seems to enter rx mode (`state()` returns `STATE_RX`) which is weird. I guess there could be issues with my arch file.

uknoblic:
Hi,

I can send you my arch file if you want. Please contact me if you want to have it.

Regards

Ulf

Von: tussock-cam [mailto:notifications@github.com]
Gesendet: Dienstag, 20. Oktober 2015 02:49
An: contiki-os/contiki
Cc: Ulf Knoblich
Betreff: Re: [contiki] Made CC1200 test mode and rx watchdog selectable (#1280)


@uknoblic<https://github.com/uknoblic> I'm also using the cc2538 and am noticing quite a few assertion failures. While debugging I've come across your set_channel function and notice that the chip enters idle fine, but after writing CC1200_FREQ0 it seems to enter rx mode (state() returns STATE_RX) which is weird. I guess there could be issues with my arch file.

alignan:
Rebased.

I haven't had any issue yet with the assertions, but I haven't tested thoughtfully.

Issue #1142, opened 2015-06-24
CSMA backoff logic and re-TX count ignore MAC_TX_COLLISION


g-oikonomou:
In `csma.c :: packet_sent()` ([L286](https://github.com/contiki-os/contiki/blob/master/core/net/mac/csma.c#L286)), we start with this:

```
switch(status) {
case MAC_TX_OK:
case MAC_TX_NOACK:
  n->transmissions += num_transmissions;
  break;
case MAC_TX_COLLISION:
  n->collisions += num_transmissions;
  break;
...
```
Thus, if the RDC signals `MAC_TX_COLLISION` we increment `n->collisions`, but _not_ `n->transmissions`

This results in two problems:

* The "abort if re-TX count exceeded" logic fails
```
if(n->transmissions < metadata->max_transmissions) {
  /* Do stuff */
} else {
  /* Drop the frame */
}
```
This logic ignores the number of times the RDC driver returned `MAC_TX_COLLISION`. This check will keep failing and CSMA will keep retrying until `MAC_TX_OK` or until we receive `MAC_TX_NOACK` for a number of times exceeding `metadata->max_transmissions`.

* The backoff logic does not seem to take into account `n->collisions` either:

```
num_tx = n->transmissions;
...
backoff_exponent = num_tx;
```

IMHO when RDC returns `MAC_TX_COLLISION`, the failed TX attempt should be taken into account for both parts of the CMSA logic

simonduq:
I agree that collisions need to be accounted for. My opinion is we should aim for making Csma (+Nullrdc) compliant with IEEE802.15.4 nonbeacon-enabled mode. That would mean:
* 1) perform CCA for a maximum of `macMaxCSMABackoffs` (default: 4), with exponential back-off between attempts;
* 2) transmit immediately after a successful CCA. In case of failure, iterate (reset backoff exponent, goto 1) until exceeding `macMaxFrameRetries` (default: 3).

laurentderu:
Is there someone working of this issue ? If not, I would propose a quick fix to avoid infinite retransmissions because of collisions.

alignan:
I have stumbled on to this as well, if there is a  constant channel interference it loops forever doing CCA checks and retransmission attempts.

A dirty fix is to check `n->collisions` with `metadata->max_transmissions` also, but having a separate `max_collisions` it would be better.

bkozak-scanimetrics:
[...]

PR #1281, opened 2015-09-28
sicslowpan should not set max mac transmits

tpetry:
In the actual implementation sicslowpan is setting the max mac transmission value of a packet to sent to a configurable value. But this is not needed because the mac layer has a default value which will be used if sicslowpan would not set this value.
In fact changing the max mac transmits setting at the mac layer will therefore not have any effect because of sicslowpan setting a value by itself. This is not a behaviour you would expect or find within a short time without thorough debugging the whole network stack.

Have been searching for this bug a long time because of network with an insanely high collision percentage and increasing the maximum mac transmit of csma did not have any effect.

### DIFF
--- a/core/net/ipv6/sicslowpan.c
+++ b/core/net/ipv6/sicslowpan.c
@@ -98,12 +98,6 @@ void uip_log(char *msg);
 #define UIP_LOG(m)
 #endif /* UIP_LOGGING == 1 */
 
-#ifdef SICSLOWPAN_CONF_MAX_MAC_TRANSMISSIONS
-#define SICSLOWPAN_MAX_MAC_TRANSMISSIONS SICSLOWPAN_CONF_MAX_MAC_TRANSMISSIONS
-#else
-#define SICSLOWPAN_MAX_MAC_TRANSMISSIONS 4
-#endif
-
 #ifndef SICSLOWPAN_COMPRESSION
 #ifdef SICSLOWPAN_CONF_COMPRESSION
 #define SICSLOWPAN_COMPRESSION SICSLOWPAN_CONF_COMPRESSION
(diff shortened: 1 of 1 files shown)


laurentderu:
About the amount of collision, it is a known issue, see #1142. A quick workaround is to count collisions as transmission.

tpetry:
The problem was that i wanted csma to allow a much higher collision count before dropping a packet because of a bluetooth disturbing the contiki communication. But because of this sicslowpan setting the higher allowed collision count has simply been ignored while sicslowpan used a much lower value.

simonduq:
I agree we don't need a max tx count in sicslowpan.
I think you should, in addition:
* in `csma.c`, set the default `CSMA_MAX_MAC_TRANSMISSIONS` to 4 (the current default of `SICSLOWPAN_MAX_MAC_TRANSMISSIONS`)
* remove all instances of `SICSLOWPAN_CONF_MAX_MAC_TRANSMISSIONS` in the tree

tpetry:
Is there any specific reason for having at maximum 4 transmits in sicslowpan?

simonduq:
No, my point is that this PR should focus on fixing the confusing config flags. It shouldn't also change the default behavior (decrease max re-transmissions from 4 to 3) for no good reason. This would have side effects.

tpetry:
Ok, wil do the cleanup at the weekend and ping you when it's done.

simonduq:
One more change you'd need to do to avoid side effects is to replace current definitions of `SICSLOWPAN_CONF_MAX_MAC_TRANSMISSIONS` by `CSMA_CONF_MAX_MAC_TRANSMISSIONS`. Several platforms define their own max, and we don't want this behavior changed. For instance, Cooja sets the max to 8, which is important for many regression test.

simonduq:
@tpetry any plan to address the above issues?

tpetry:
It's still on my plan but i had not yet the time for the required changes.

simonduq:
ping

tpetry:
My contiki project was finished shortly afterwards and i had no time for such a bug change and testin this all. Will not make the required changes i am no longer tied to contiki-os.

PR #1288, opened 2015-09-29
RPL: check if objective function is supported before creating dag.

joakimeriksson:
Currently if Contiki RPL tries to join a network that make use of a non-supported objective function it will try to both create a DAG and a parent before realising that it is not going to be possible to join. This fix checks the OCP against the currency supported objective function before trying to allocate or create anything and fails earlier. Without this fix some examples that use a border router with a non-supported OF do crash (er-coap examples, etc).


simonduq:
good catch, :+1: 
(extra suggestion: would be nice to also show the non-supported ocp in the `PRINTF`)

joakimeriksson:
Added the incoming DIOs OCP in the PRINTF now.

nvt:
:+1:

PR #1290, opened 2015-09-30
Add rtimer multiple access support

bkozak-scanimetrics:
This patch adds the ability to schedule multiple rtimers simultaneously and does some minor cleanup of the rtimer modules. 

To facilitate this feature, I've also added in an architecture agnostic API for disabling interrupts (essentially the same as #1117 but with some slight modifications). If this PR is merged, #1117 will be redundant. The multiple access support is not available for any platforms which don't implement this API (at the moment I've implemented CC26xx/CC13xx only).

I am proposing this patch as an alternative to #495 which I believe went to unnecessary lengths to avoid masking interrupts (and is overly complex as a result). I think that my implementation will be better performing and easier to test. Furthermore, this PR has the benefit of keeping the old single-access functionality around until such time as multiple-access is really needed for any given platform (the old implementation will perform better).

Note also that I've added some similar functionality to #1199 but this PR doesn't make #1199 redundant.


bkozak-scanimetrics:
Fixed a bug where it was possible for ```rtimer_set``` to change the time, func, and ptr of an rtimer that was currently queued.

bkozak-scanimetrics:
Defined ```RTIMER_CONF_MINIMAL_SAFE_SCHEDULE``` on CC26xx/CC13xx. The default value of 2 is not enough on this platform.

With this change made, the multi-access rtimer will work out of the box with CC26xxx/CC13xx (just set ```RTIMER_CONF_MULTIPLE_ACCESS``` true).

core/sys/rtimer.c
```diff
@@ -46,62 +46,293 @@
...
+/*---------------------------------------------------------------------------*/
+#if RTIMER_MULTIPLE_ACCESS
+#include "sys/isr-control.h"
+/*---------------------------------------------------------------------------*/
+static void
+schedule_rtimer_isr_safe(struct rtimer *rtimer)
 {
-  rtimer_arch_init();
+  rtimer_clock_t min = RTIMER_NOW() + RTIMER_MINIMAL_SAFE_SCHEDULE;
+  if(rtimer) {
```
tonnenpinguin: Couldn't min be calculated within the if branch?

tonnenpinguin:
I've tested this pr on my cc2650 based platform and did not run into any issues. On a sidenode coap is finally working again with my rtimer dependent code.
Thanks for the patch!

core/sys/rtimer.c
bkozak-scanimetrics: > Couldn't min be calculated within the if branch?

Good eye

bkozak-scanimetrics:
Moved the calculation of ```min``` in ```schedule_rtimer_isr_safe``` into the if block as per the comment by @tonnenpinguin on the outdated diff. This will prevent unnecessary reads of the RTC in some cases.

Also took the opportunity to rebase on the latest Contiki master.

tonnenpinguin:
Hi Billy,
I am trying to understand your code and I'm afraid I don't quite get it yet.
Could you please help me with the following points:
1.) in pop_expired_from_queue(): What happens when the next_rtimer was set to the past, e.g. the while loop is exited on the first run. Don't we schedule any other timers after that?
2.) in pop_expired_from_queue(): Why do we mark all future rtimers as running?
3.) in pop_expired_from_queue(): Why is only the very last rtimer scheduled?
4.) is func_call_list() really supposed to call all callbacks?
5.) What is the definition of the state RTIMER_RUNNING and why can't we cancel a running rtimer?

bkozak-scanimetrics:
> 1.) in pop_expired_from_queue(): What happens when the next_rtimer was set to the past, e.g. the while loop is exited on the first run. Don't we schedule any other timers after that?

I'm not really sure I understand where this question comes from. Maybe you just made a mistake about how ```RTIMER_CLOCK_LT``` works (I think you may have its functionality inverted in your head; I know I do that all the time)? Note also that the queue is always kept sorted so that the first item in the queue will be the next to expire. 

Some more detail...
The idea is that if the first item in the queue is in the past then the while loop will not exit. The while loop should only exit when ```now``` is less than ```iter->time``` and therefore  ```iter->time``` lies in the future. Unfortunately, due to the way that rtimers wrap, there is no general way to determine if a rtimer really lies in the past or if it actually just lies far in the future. Instead we have to rely on the calling code to  gaurantee that an rtimer is never scheduled "too far" into the future so that it might look like it is in the past. 

> 2.) in pop_expired_from_queue(): Why do we mark all future rtimers as running?

In ```pop_expired_from_queue()``` we are making a list of rtimers (**L**) which is separate from the main queue for use by ```func_call_list()```. After we've done this, if some other thread of execution or  an rtimer callback would call ```remove_from_queue()``` on an rtimer which is in **L** at the same time that ```func_call_list()``` is traversing **L** then we can have all sorts of weird problems. 

To prevent the above from happening, we mark all expired rtimers as ```RUNNING``` so that they cannot be cancelled from another thread until the rtimer is actually popped from **L**.  We could also solve this by popping each expired rtimer from the queue in sequence from within an atomic section then immediately calling the callback but we should prefer to keep the atomic sections as small as possible in case we have interrupt nesting on the platform or even if, somehow, ```rtimer_run_next()``` isn't called from an interrupt at all (maybe someone creates a proper multi-threading implementation for example). Of course we don't even need to use the atomic sections unless we can have this kind of multi-threading. 

> 3.) in pop_expired_from_queue(): Why is only the very last rtimer scheduled? In add_to_queue() it looks like the next rtimer that has to be called is the second in list, not the last one as far as I understood.

The first item in the queue is the next to expire (or the one which has already expired first). Therefore, what ```pop_expired_from_queue()``` is doing is scheduling the first rtimer it finds which is in the future (this will be the next rtimer to expire). 

> 4.) is func_call_list() really supposed to call all callbacks?

The list given to ```func_call_list()``` is a list of expired rtimers. Therefore they all need to have their callbacks called. 

 > 5.) What is the definition of the state RTIMER_RUNNING and why can't we cancel a running rtimer?

Might have been a poor choice of name on my part now that I think about it. What ```RTIMER_RUNNING``` is mean to convey is that it is now too late to stop the rtimer callback from running even if the callback hasn't started executing quite yet. Once an rtimer has state ```RTIMER_RUNNING``` it's callback is guaranteed to run. 



I hope that helps, please do let me know if there is any-more confusion at all and I will do my best to explain!

tonnenpinguin:
> The idea is that if the first item in the queue is in the past then the while loop will not exit. The while loop should only exit when now is less than iter->time and therefore iter->time lies in the future. 

Ahh...you were right, I got somewhat confused!

> In pop_expired_from_queue() we are making a list of rtimers (L) which is separate from the main queue for use by func_call_list()

Are we really building a new list or are we manipulating the original structs? In the latter case I would have the following concern: If there are two rtimers queued for the future we are detecting the first one and scheduling it, however, the second one (which used to have a reference in the first ones next pointer) is being overwritten.

> The first item in the queue is the next to expire (or the one which has already expired first). Therefore, what pop_expired_from_queue() is doing is scheduling the first rtimer it finds which is in the future (this will be the next rtimer to expire).

Got it!

Thanks for the detailled explanation! That helped a lot!

bkozak-scanimetrics:
> Are we really building a new list or are we manipulating the original structs? 

Well, we're doing both in a way. The top part of the original list is chopped off and becomes its own list. The start of the rtimer queue will then point to the next to expire rtimer (NULL if empty) so that we don't lost the reference to the unexpired rtimers. 

> In the latter case I would have the following concern: If there are two rtimers queued for the future we are detecting the first one and scheduling it, however, the second one (which used to have a reference in the first ones next pointer) is being overwritten.

That's correct. This is essentially why we have to set state to ```RTIMER_RUNNING``` on rtimers which are moved to the new list; the code needs to know that the rtimer is no longer part of the main structure so that certain operations on it (specifically, cancelling) are disallowed.

tonnenpinguin:
I finally got it!
Thanks so much for your patience!

bkozak-scanimetrics:
I've just recently performed a rebase and travis has failed on large-rpl. Even though this same test passed when run on my own travis account, this doesn't appear to be the same sort of failure that usually afflicts large-rpl however.

The error doesn't appear to be related in any way to this PR. It may be a good idea for someone with more Cooja experience than me to take a look at this one.

simonduq:
the test is a bit random, just restarted it

core/sys/rtimer.h
```diff
@@ -87,13 +109,20 @@ struct rtimer {
...
+  int state;
+#endif
 };
 
 enum {
   RTIMER_OK,
   RTIMER_ERR_FULL,
   RTIMER_ERR_TIME,
   RTIMER_ERR_ALREADY_SCHEDULED,
+  RTIMER_ERR_NOT_SCEDULED,
```
gebart: `RTIMER_ERR_NOT_SCEDULED` -> `RTIMER_ERR_NOT_SCHEDULED`

bkozak-scanimetrics:
@gebart 
| RTIMER_ERR_NOT_SCEDULED -> RTIMER_ERR_NOT_SCHEDULED

Got it, thanks.

cquijano:
You should merge this, I am implementing a dimmer light based on Contiki and this PR really saved my day.

Now my app is working fine. Thanks @bkozak-scanimetrics

PR #1296, opened 2015-10-06
fixed memory bug

vozi:
This line assumes that the plaintext directly follows the AddAuthData. But the plaintext is already at memory location m, so there is no location change needed.


simonduq:
I think this is correct, @kkrentz can you have a look and confirm?

kkrentz:
Indeed, this seems to be a leftover from the previous version.

kkrentz:
It is also fixed in #1146

simonduq:
:+1:

PR #1297, opened 2015-10-06
Change the default IPv6 prefix from aaaa::/64 to fd00::/64

simonduq:
This PR brings https://github.com/contiki-os/contiki/pull/712 back to life, now with `fd00` rather than `fc00`. As mentioned [here](https://github.com/contiki-os/contiki/pull/712#issuecomment-137351534) I (and some others) believe `fd00` to be a better alternative than the Sixxs prefix.

### DIFF
--- a/core/net/ipv6/sicslowpan.c
+++ b/core/net/ipv6/sicslowpan.c
@@ -1757,8 +1757,8 @@ sicslowpan_init(void)
 #ifdef SICSLOWPAN_CONF_ADDR_CONTEXT_0
   SICSLOWPAN_CONF_ADDR_CONTEXT_0;
 #else
-  addr_contexts[0].prefix[0] = 0xaa;
-  addr_contexts[0].prefix[1] = 0xaa;
+  addr_contexts[0].prefix[0] = UIP_DS6_DEFAULT_PREFIX_0;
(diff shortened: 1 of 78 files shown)


joakimeriksson:
This should just go in! It is much better than the aaaa:: prefix as we always have been using! :+1:

alignan:
:+1: after resolving conflicts

simonduq:
rebased, conflict resolved, thanks!

joakimeriksson:
Why is this not just getting in? :+1: from me again! We need to get rid of aaaa:: and have something that is a valid address for the default (local-global) prefix which fd00:: is!

simonduq:
Rebased again, conflicts solved. This is going to affect many external examples / tutorials etc, so I'd like to see some extra support before merging.
@laurentderu, any thoughts from a 6LBR point of view?
@adamdunkels @g-oikonomou?

g-oikonomou:
Certainly better than `aaaa::`! :+1:

laurentderu:
For 6lbr, I think there will be a surge in support request but no real issues . The network prefixes are either user configurable or auto-configured, and the non regression tests are already using different prefixes, so everything should be fine.

simonduq:
merging, then!

PR #1300, opened 2015-10-08
Fixes a possible unsafe memcpy in uip_udp_packet_send

vsaw:
If the buffers overlap, `memcpy` must not be used as it might have arbitrary
results. This commit checks if the buffers overlap and safely copies everything
into the `uip_buf`

### DIFF
--- a/core/net/ip/uip-udp-packet.c
+++ b/core/net/ip/uip-udp-packet.c
@@ -54,9 +54,9 @@ uip_udp_packet_send(struct uip_udp_conn *c, const void *data, int len)
   if(data != NULL) {
     uip_udp_conn = c;
     uip_slen = len;
-    memcpy(&uip_buf[UIP_LLH_LEN + UIP_IPUDPH_LEN], data,
-           len > UIP_BUFSIZE - UIP_LLH_LEN - UIP_IPUDPH_LEN?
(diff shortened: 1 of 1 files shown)


nvt:
Why aren't you using memmove() directly?

vsaw:
Good question. I guess I wanted to handle the special case of ovelapping buffers while sticking to the old memcpy behaviour as much possible. But you're right. I'll update the commit.

vsaw:
@nvt Please review again.

oliverschmidt:
Sorry for the stupid question but how would a scenario look like in which the source and destination of this memcpy() would overlap?

Apart from that the indentation of the actual parameters would need be adjusted.

adamdunkels:
@oliverschmidt: the problem occurs when the caller to `uip_udp_packet_send()` has constructed its data within the `uip_appdata` buffer. `uip_udp_packet_send()` then tries to copy this with `memcpy()`, which fails.

vsaw:
@oliverschmidt Indentation has been fixed. And to give an example of what @adamdunkels has described. See https://github.com/contiki-os/contiki/blob/master/core/net/ip/resolv.c

oliverschmidt:
Thanks for the explanation. I was always under the impression that apps constructing the data to be sent right in the uIP buffer would have them sent in a copy-free way. Still so much to learn after all those years ;-))

adamdunkels:
:+1:

PR #1304, opened 2015-10-13
Fix tunslip6

simonduq:
This PR reverts https://github.com/contiki-os/contiki/commit/6608b62b01953b8a263ffda65802c1bc9052c06c which seems to break tunslip6 at  least for `examples/ipv6/rpl-border-router`. It also disables by default the IP address inquiry option, not supported by the BR.

### DIFF
--- a/tools/tunslip6.c
+++ b/tools/tunslip6.c
@@ -238,21 +238,13 @@ serial_to_tun(FILE *inslip, int outfd)
 	  slip_send(slipfd, SLIP_END);
         }
 #define DEBUG_LINE_MARKER '\r'
-	if(uip.inbuf[0] == DEBUG_LINE_MARKER ||
-           is_sensible_string(uip.inbuf, inbufptr)) {
-          unsigned char *out = uip.inbuf;
-          unsigned int   len = inbufptr;
-          if(uip.inbuf[0] == DEBUG_LINE_MARKER) {
-            out++;
-            len--;
-          }
-          fprintf(stderr, "\n***");
-	  fwrite(out, len, 1, stderr);
-          fprintf(stderr, "***\n");
-	} else {
-	  fprintf(stderr,
-		  "serial_to_tun: drop packet len=%d\n", inbufptr);
-	}
+      } else if(uip.inbuf[0] == DEBUG_LINE_MARKER) {
+	fwrite(uip.inbuf + 1, inbufptr - 1, 1, stdout);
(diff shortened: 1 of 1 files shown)


mcr:
Could you please detail how rpl-border-router is broken?  It is precisely that example that needs this.

simonduq:
The symptom is that on linux you get 
```
tunslip6: serial_to_tun: write: Invalid argument
ifconfig tun0 down
netstat -nr | awk '{ if ($2 == "tun0") print "route delete -net "$1; }' | sh
make: *** [connect-router] Error 1
```
(was reported to me by two different people and I could reproduce).

I didn't dig enough to fully understand why this occurs but the problem I noticed is that with the new conditionals, https://github.com/contiki-os/contiki/blob/master/tools/tunslip6.c#L275 is reached even if `uip.inbuf[0] == DEBUG_LINE_MARKER` or `is_sensible_string(uip.inbuf, inbufptr)` (unwanted behavior I believe).

The other problem fixed here is that `tunslip6` kept sending `?IPA` to which the BR replied `Got request message of type I` without taking any further action

mcr:
I will retest things this week: I have been running the code for two years without those problems (on Linux).

simonduq:
OK, thanks! I hope you can reproduce the same issue.

fvdnabee:
I also experienced tunslip6 being broken in exactly the same way as Simon. I didn't manage to diagnose the problem (a work around I found was setting verbose > 2). 
So, I tested commit 30f69b0 (in cooja with one z1 rpl-border-router) and tunslip6 doesn't crash with "tunslip6: serial_to_tun: write: Invalid argument" and is working fine.

Also, +1 for commit a9a46b2 as per the reference to PR #676.

bkozak-scanimetrics:
I've also been seeing this exact same problem since 6608b62. Also, since this commit I'm seeing a lot of the following.
> *** dropping large 2000 byte packet

simonduq:
@mcr, any news? More and more people are reporting this issue, would you mind if we merge this now?

simonduq:
@mcr, merging as tunslip6 currently broken for several users, please open a new PR if you come up with a better fix.

mcr:
Simon Duquennoy <notifications@github.com> wrote:
    > @mcr, any news? More and more people are reporting this issue, would you mind
    > if we merge this now?

Go ahead. I have been ill for a week.
I'll work something out on top of the new code.

--

PR #1307, opened 2015-10-15
Cooja: Make sure motes are always removed from the list of unintialized motes

cmorty:
This is an old bug of mine. As mine was the only plug-in that actually needed this, and only because it always added nodes from the simulation-thread, this never popped up.


alignan:
Merge?

fros4943:
:+1:

PR #1308, opened 2015-10-15
Fix SafeRandom

cmorty:
I implemented SafeRandom to avoid potential problems with simulations not being reproducible. Seems like I didn't think things all the way through. This should fix things.

### DIFF
--- a/tools/cooja/java/org/contikios/cooja/SafeRandom.java
+++ b/tools/cooja/java/org/contikios/cooja/SafeRandom.java
@@ -45,17 +45,30 @@ public class SafeRandom extends Random {
   
   Simulation sim = null;
   Thread initThread = null;
+  Boolean simStarted = false;
   
   private void assertSimThread() {
-    // It we are in the simulation thread everything is fine (the default)
     // sim can be null, because setSeed is called by the super-constructor.
-    if(sim != null && !sim.isSimulationThread()) {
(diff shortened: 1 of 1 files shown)


fros4943:
Thanks for tglhe quick fix @cmorty! :+1:

tools/cooja/java/org/contikios/cooja/SafeRandom.java
```diff
@@ -36,26 +36,39 @@
...
+ * Rationale: By allowing another thread to use the random number
+ * generator concurrency is introduced, thus it can not be guaranteed
+ * that simulations are reproducible.
  *
  */
 public class SafeRandom extends Random {
   
   Simulation sim = null;
   Thread initThread = null;
+  Boolean simStarted = false;
```
nfi: Any reason to use a Boolean object instead of a boolean here?

cmorty: To much Scala coding. I'm not too familiar with Java optimizing. Isn't it converted to boolean? Otherwise I'll fix it.

nfi: The value will be stored as a Boolean object. One affect is that simStarted can not only be set to true or false but also null. But the overhead in this case is negligible. I was just curious.

PR #1309, opened 2015-10-15
Fix default tun/tap issue with multiple tunslip6

cladmi:
When running multiple tunslip6 instances, it collides with tun0 being already
used (cannot open file). However, system default is already to use "tun0" and "tap0".

By putting the default empty string as name, system automatically increment and
selects a free interface, "tun0", "tun1".

### DIFF
--- a/tools/tunslip6.c
+++ b/tools/tunslip6.c
@@ -852,14 +852,6 @@ exit(1);
     break;
   }
 
-  if(*tundev == '\0') {
-    /* Use default. */
-    if(tap) {
-      strcpy(tundev, "tap0");
-    } else {
-      strcpy(tundev, "tun0");
-    }
-  }
   if(host != NULL) {
     struct addrinfo hints, *servinfo, *p;
     int rv;


simonduq:
What is this tested on?
i tried on linux and this doesn't seem to help (i.e. needed -t to set an additional interface)

cladmi:
I tried on ubuntu 14.04 and on debian squeeze.

If I run 

`NUM=0; nc -l -p 2000${NUM} & sudo ./tunslip6 -a localhost -p 2000${NUM} aaa${NUM}::1/64`

In three terminals with NUM=0, 1, 2 to change the cort, I correctly get tun0, tun1 and tun2.

cladmi:
However, I see in https://www.kernel.org/doc/Documentation/networking/tuntap.txt that they don't document this behavior, I found it somewhere else but don't rememver, but using "tun%d" and "tap%d" should give the correct result too.

simonduq:
OK, was my mistake, now able to reproduce. Thanks for the fix! :+1:

saurabhbansal90:
hello all, i want to run tunsliip6 in my board, so i first tried in my system only (i.e. my PC). but i am not able to use tunslip6, everytime the connection is being refused.
--------------------------------------------------------------------------------------------------------------------------------
root@saurabh-3000-N200:/home/saurabh/B020-325A/Projects/EasyARM/6LowPAN/TunSLIP# ./tunslip6 -a localhost -p 2000 aaaa::1/64
client: connect: Connection refused
tunslip6: can't connect to ``localhost:2000'': Connection refused
--------------------------------------------------------------------------------------------------------------------------------


i ran another command with the tty also that i want to use, the SLIP got started, but not able to open tun in my system. similar problems i am getting in my board as well
--------------------------------------------------------------------------------------------------------------------------------
root@saurabh-3000-N200:/home/saurabh/B020-325A/Projects/EasyARM/6LowPAN/TunSLIP# ./tunslip6 aaaa::1/64 -s /dev/tty0
********SLIP started on ``/dev/tty0''
tunslip6: ioctl: Invalid argument
root@saurabh-3000-N200:/home/saurabh/B020-325A/Projects/EasyARM/6LowPAN/TunSLIP# 
--------------------------------------------------------------------------------------------------------------------------------

PR #1310, opened 2015-10-15
AVR platforms: use 32 bit clock_time_t

drugo72:
See conclusion of #1271

### DIFF
--- a/platform/avr-atmega128rfa1/contiki-conf.h
+++ b/platform/avr-atmega128rfa1/contiki-conf.h
@@ -56,18 +56,10 @@
  */
  /* Clock ticks per second */
 #define CLOCK_CONF_SECOND 128
-#if 1
-/* 16 bit counter overflows every ~10 minutes */
-typedef unsigned short clock_time_t;
-#define CLOCK_LT(a,b)  ((signed short)((a)-(b)) < 0)
-#define INFINITE_TIME 0xffff
-#define RIME_CONF_BROADCAST_ANNOUNCEMENT_MAX_TIME INFINITE_TIME/CLOCK_CONF_SECOND /* Default uses 600 */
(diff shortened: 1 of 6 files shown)


simonduq:
Looks good, but can you explain the changes to announcement timing?
Also, any reason to swap the declarations of the 16-bit and 32-bit versions? (makes the diff extra hard to read)

drugo72:
>Looks good, but can you explain the changes to announcement timing?

Sure, announcement timings are in ```clock_time_t``` units (not in seconds) and they both exceed the 16 bit clock period which is about 10 minutes. I've subtracted 10 ticks (80ms) because I'm not sure the etimer module correctly handles an interval that is exactly ```INFINITE_TIME``` in the future.

>any reason to swap the declarations of the 16-bit and 32-bit versions

Because I prefer ```#ifdef something``` instead of ```#ifdef !something```

simonduq:
ok, fine with me :+1:

nvt:
Why not always use 32-bit timers on this platform? Basing the selection of timer data types on the choice of routing protocol seems like a kludge.

cdealti:
As far as I know there is no real reason. The original code already supported both options with the 16 bit one used by default. I assume it's safe to unconditionally use the 32 bit one.
I think nobody is using the AVR platforms these days and I only push fixes for them while trying to learn Contiki.
Should I do the change, maybe disabling the 16 bit version through preprocessor ```#if 0```?

nvt:
Yes, enabling 32-bit clocks in all cases should be better than just having it enabled when RPL is activated. I think that it is best to simply remove unused code instead of leaving <code>#if 0</code> blocks, which make the file more difficult to read. The old code is available anyway through the repository's history.

simonduq:
@drugo72, any comment? If you implement @nvt's suggestion this is ready to merge

cdealti:
Done.

simonduq:
Thanks!

PR #1311, opened 2015-10-19
Add configuration of DIS start delay and interval in RPL conf.

arurke:
Created a configuration-define for DIS start delay and moved it + DIS interval out of rpl-private.h and into rpl-conf.h. It seems appropriate since DIO interval is already specified here, and the DIS parameters are similarly interesting to tweak.

### DIFF
--- a/core/net/rpl/rpl-conf.h
+++ b/core/net/rpl/rpl-conf.h
@@ -296,4 +296,22 @@
     + random_rand() % (RPL_PROBING_INTERVAL))
 #endif
 
+/*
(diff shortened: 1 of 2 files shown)


simonduq:
Looks good, but would be good to also fix the naming (was already off before this PR):
`RPL_FOO_BAR_CONF` should become `RPL_CONF_FOO_BAR`

arurke:
Agreed and fixed

nvt:
:+1:

PR #1317, opened 2015-10-21
AES-128 driver for the CC2538 SoC

kkrentz:
* rudimentary `aes_128_driver` for the CC2538 SoC
* deleted aes_128_padded_encrypt as suggested in #618 
* a fix for an error when including `aes-128.h`

### DIFF
--- a/core/lib/aes-128.c
+++ b/core/lib/aes-128.c
@@ -167,17 +167,6 @@ encrypt(uint8_t *state)
 }
 /*---------------------------------------------------------------------------*/
 void
-aes_128_padded_encrypt(uint8_t *plaintext_and_result, uint8_t plaintext_len)
-{
-  uint8_t block[AES_128_BLOCK_SIZE];
-  
-  memset(block, 0, AES_128_BLOCK_SIZE);
-  memcpy(block, plaintext_and_result, plaintext_len);
-  AES_128.encrypt(block);
-  memcpy(plaintext_and_result, block, plaintext_len);
-}
-/*---------------------------------------------------------------------------*/
-void
 aes_128_set_padded_key(uint8_t *key, uint8_t key_len)
 {
   uint8_t block[AES_128_BLOCK_SIZE];
--- a/core/lib/aes-128.h
+++ b/core/lib/aes-128.h
@@ -37,8 +37,8 @@
  *         Konrad Krentz <konrad.krentz@gmail.com>
  */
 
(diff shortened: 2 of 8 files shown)


bthebaudeau:
Isn't there a way of calling the existing functions to get the same result without reimplementing things? I have not checked yet, but I guess that it should be possible. If you have already tried, what kind of issues did you have?

kkrentz:
There must be a bug inside `aes_load_keys`. I tried to reuse it, but it did not work correctly.

bthebaudeau:
What does `it did not work correctly` mean precisely? Is the first call OK? Debug with `PRINTF()` if needed. It works with `examples/cc2538dk/crypto/ccm-test.c`. Can you try something like:
```
static void
set_key(const uint8_t *key)
{
  uint8_t ret;

  if(!is_initialized) {
    is_initialized = 1;
    crypto_init();
    /* TODO do not leave the crypto engine enabled */
  }

  ret = aes_load_keys(key, AES_KEY_STORE_SIZE_KEY_SIZE_128, 1, 0);
  if(ret != CRYPTO_SUCCESS) {
    PRINTF("cc2538-aes-128: aes_load_keys() failed: %u\n", ret);
    /* TODO handle error */
  }
}
```

Also, why don't you implement a hardware-based `ccm_star_driver` using the CC2538 `ccm.c`/`h` rather than an `aes_128_driver` with software CCM* (have to check whether the star thing is supported by `ccm.c`)?

kkrentz:
Today `aes_load_keys` worked - sorry for the confusion. I also need a bare AES-128 driver for generating cryptographic random numbers (to be included in another pull request).

bthebaudeau:
Then, it would be better to move the AES feature to `cpu/cc2538/dev/aes.c`/`h`, using APIs similar to the ones in `cpu/cc2538/dev/ccm.c`/`h`, and removing the 128-bit limitation. The added files would then just be wrappers.

And it would also be great to have a wrapper for CCM* besides the one for AES-128, of course provided that it can be supported by the existing CC2538 CCM driver.

bthebaudeau:
@kkrentz Don't worry: I'm not forgetting you. Just give me a few weeks to look into this thoroughly. I'm a bit busy. Thanks for your patience.

bthebaudeau:
Might be superseded by #1435.

bthebaudeau:
Superseded by #1435, which has now been merged.

PR #1318, opened 2015-10-22
Fix CSC & XML export to match .gitattributes

sieben:
This PR fix the XML export of CSC file in Cooja GUI that was with CRLF and not LF

http://permalink.gmane.org/gmane.comp.version-control.git/268715


cmorty:
Is this a good idea? Although they are xml, and therefore text, they are still application-generated, and should IMO be considered readable binaries.

simonduq:
I almost agree @cmorty but .csc files also contain javascript for regression testing and sometimes seeing the diff is actually useful (e.g. a PR modifying a test)

cmorty:
@simonduq: We're talking git, not svn. You can still diff them without crls-rules.

sieben:
I agree with Simon.

2015-10-22 15:25 GMT+02:00 Simon Duquennoy <notifications@github.com>:

sieben:
It also allows more uniformity in the file format and I think it's a good thing to have a consistant one accross all the repository for any text even if it's generated.

cmorty:
Ok, but then you need to ensure that Cooja also generates the correct format. Otherwise opening and saving a file will result is a huge diff.

sieben:
I've done it :) check out the Cooja.java ;-)

2015-10-22 15:31 GMT+02:00 Moritz Strübe <notifications@github.com>:

cmorty:
Ah ok, didn't see it between all those .csc. That's a nice example why making two commits and one PR is a good idea.  ;)  :+1:

simonduq:
@cmorty add `?w=1` to the github "files changed" URL to filter out space changes from diff:
https://github.com/contiki-os/contiki/pull/1318/files?w=1

simonduq:
:+1:

cmorty:
@simonduq Yea, I know. But I saw 1 Patch, and a long list of files and didn't run a diff. :)

sieben:
There was two binary files not taken care of by the .gitattributes. I've fixed that also

sieben:
All check pass :)

fros4943:
:+1:

vsaw:
@sieben My git is completely troubled by this. I can not get rid of the following warning

```
warning: CRLF will be replaced by LF in examples/antelope/netdb/netdb.csc.
The file will have its original line endings in your working directory.
```

Also it does not allow me to do any checkout or any other operation whatsoever on the repository? Am I the only one having trouble? How can this be fixed?

sieben:
I've you tried to rebase on top of upstream? Worst case scenario you can
add all this file change into a commit and rebase it around.

vsaw:
Luckily I did not have any local changes in my repo that were not saved elsewhere so I could simply delete the repository and do a fresh clone. But still weird....

Thanks for the quick reply. I just thought you should know in case it creates more problem for others as well.

sieben:
I've had a couple of problems. It's due I think to git that want to convert
all files to the new format and won't let you do otherwise. It can be a
little annoying I agree...

cmorty:
Ok, so what about excluding .csc from crnl-rules? As those added to the main repo are generated by the new version of Cooja things should be fine, but people adding old simulations to their fork don't get strange problems.

sieben:
Because you still want diff on those files. A simple dos2unix solve
trivially these problems anyway.

2015-10-27 9:20 GMT+01:00 Moritz Strübe <notifications@github.com>:

cmorty:
Yes you know that. :) But at least @vsaw hast some problems. Maybe the best Idea is to add a regression check that does a `find . -name "*.csc" -exec ....`

PR #1319, opened 2015-10-22
Live Traffic Capture and Sniffer for cc2420 

warner83:
Implemented Live Traffic Capture and Sniffer for cc2420 based on the
original sensniff project (https://github.com/g-oikonomou/sensniff).
    
See http://www.iet.unipi.it/c.vallati/z1_sniffer.shtml

### DIFF
--- a/dev/cc2420/cc2420.c
+++ b/dev/cc2420/cc2420.c
@@ -102,6 +102,11 @@ enum write_ram_order {
 #define LEDS_OFF(x)
 #endif
 
+#if CC2420_CONF_HEXDUMP
+#include "uart0.h"
+static const uint8_t magic[] = { 0x53, 0x6E, 0x69, 0x66 }; /* Snif */
+#endif
+
 /* Conversion map between PA_LEVEL and output power in dBm
    (from table 9 in CC2420 specification).
 */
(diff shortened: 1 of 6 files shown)


alignan:
Nice :-)

I have been using this since last year when we last chat about it (see https://github.com/alignan/contiki/commit/db6cb64e147fd61e9f9cfa4f68c666a0c01af88c), however after talking with @g-oikonomou about this at the time, I think the right way would be to try and have a generic sniffer application rather than one for each platform.

Also I think it would be best to add the `sensniff` as submodule rather than copying the application.

g-oikonomou:
A generic sniffer application would be pure great. I had some discussions a long time ago, but I just never got round to it.

Meanwhile, I definitely agree that there is no place for the sensniff script itself in the example folder. We may or may not ship it as a submodule under tools at some point, but that's a different story.

g-oikonomou:
@warner83 this PR cannot be merged as-is, mainly due to comments above. Are you willing/do you have time to address comments?

warner83:
@g-oikonomou @alignan I definitely agree that a generic sniffer application would be the best solution. I also agree that the sensniff tool should be place somewhere else (truth to be told I pushed it by mistake :) ).  
I will try to rework the PR to generalize its structure!

simonduq:
Hey,
This timed out a while ago.
@warner83 is that still on your list?

warner83:
Hi @simonduq it is still in my list, unfortunately, I didn't have time so far!

PR #1321, opened 2015-10-24
Remove redundant function calls in CC13xx/CC26xx RFC code

### LINKED ISSUE #1279: CC26xx / CC13xx - redundant and or wrong PDSTATN checking in rf-core
I've just had a look at ```rf_core_is_accessible``` in rf-core.c and I think there are some problems with the function.

If we look at the implementation for ```ti_lib_prcm_rf_ready``` and ```ti_lib_prcm_power_domain_status``` we can see that they both check the same bits (note that ```PDSTAT1RFC.ON``` "is an alias" for ```PDCTL1.RFC_ON``` according to the TRM). Also, ```ti_lib_prcm_power_domain_status``` will return a positive result if ```PDCTL0.RFC_ON``` **or** ```PDCTL1.RFC_ON``` are set.

I'm not sure if the ```PDCTL0.RFC_ON || PDCTL1.RFC_ON``` is sufficient; maybe there is some reason that this works but I can't see how. Regardless, there's a redundant read here.

Also , ```rf_core_power_up``` will call ```ti_lib_prcm_power_domain_status``` directly but will then again call it indirectly through ```rf_core_is_accessible``` soon after. This also seems to be redundant (and indeed I've never seen the ```PRINTF("rf_core_power_up: Not ready\n");``` get executed) .

g-oikonomou:
As discussed in #1279, the current CC13xx/CC26xx rf-core code makes some unnecessary power domain-related reads. This pull removes them.

Closes #1279

### DIFF
--- a/cpu/cc26xx-cc13xx/rf-core/rf-core.c
+++ b/cpu/cc26xx-cc13xx/rf-core/rf-core.c
@@ -114,9 +114,7 @@ PROCESS(rf_core_process, "CC13xx / CC26xx RF driver");
 uint8_t
 rf_core_is_accessible()
 {
-  if(ti_lib_prcm_rf_ready() &&
-     ti_lib_prcm_power_domain_status(PRCM_DOMAIN_RFCORE) ==
-     PRCM_DOMAIN_POWER_ON) {
+  if(ti_lib_prcm_rf_ready()) {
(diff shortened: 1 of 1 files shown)


jonnteolsson:
:+1:

PR #1322, opened 2015-10-24
Clear CC26xx/CC13xx interrupt flags selectively

### LINKED ISSUE #1294: CC26xx & CC13xx ISR Flag Clear Issue
In the radio driver ISR routines, the flag register is getting cleared as follows: ```HWREG(RFC_DBELL_NONBUF_BASE + RFC_DBELL_O_RFCPEIFG) = 0x0 ``` 

This is a problem because there are multiple flags which can trigger these routines. We could end up missing an event unnecessarily if the following were to occur:

1. Event A triggers an interrupt
2. The ISR runs, the flag register is read and event A is processed
3. Just before clearing the flag register event B occurs
4. the flag register is cleared and the event B flag is lost

g-oikonomou:
As discussed in #1294, every time a CC13xx/CC26xx RF interrupt fires, we clear all interrupt flags unconditionally. This may result in missed events. This patch fixes this bug by clearing only the flag that triggered the interrupt in the first place.

Fixes #1294

### DIFF
--- a/cpu/cc26xx-cc13xx/rf-core/rf-core.c
+++ b/cpu/cc26xx-cc13xx/rf-core/rf-core.c
@@ -478,8 +478,8 @@ cc26xx_rf_cpe1_isr(void)
     }
   }
 
-  /* Clear interrupt flags */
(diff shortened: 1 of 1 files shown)


jonnteolsson:
:+1:

PR #1324, opened 2015-10-26
Remove LED fade in lpm.c

msloth:
The LED fading is nice looking, but strange feature to have in the lpm-module. It consumes unnecessary power, and you even may want to have an LED on even in shutdown.

### DIFF
--- a/cpu/cc26xx-cc13xx/lpm.c
+++ b/cpu/cc26xx-cc13xx/lpm.c
@@ -84,7 +84,7 @@ void
 lpm_shutdown(uint32_t wakeup_pin, uint32_t io_pull, uint32_t wake_on)
 {
   lpm_registered_module_t *module;
-  int i, j;
+  int i;
   uint32_t io_cfg = (IOC_STD_INPUT & ~IOC_IOPULL_M) | io_pull |
     wake_on;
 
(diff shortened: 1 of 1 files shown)


g-oikonomou:
@msloth I agree that putting this in the CPU code in this fashion is probably too restrictive. At the same time, do you think we can get away without some sort of indication that the shutdown sequence was actually triggered? Perhaps turn this into platform code (e.g. in the button driver)?

In any case, `j` will also need undefined. Thus pull introduces a warning as it is right now.

msloth:
Ah, yes, I forgot about that one, thanks. Will update the PR asap.

Well, in my opinion any indication that we successfully entered a specific low-power state is very much up to an application and should in that case use eg the already in place `LPM_MODULE(...)` callbacks if you really want to do this. It doesn't feel like very useful information to convey to a user, and since it does consume some power (`2 LEDs * 0.5 avg DC * 1 mW * 1 second`, if that's much or not I'll leave to the wind) in an application already very concerned with power I feel like this should better reside in an application such as the low-power demo.

cmorty:
> 2 LEDs * 0.5 avg DC * 1 mW * 1 second, if that's much or not I'll leave to the wind
This also keeps the CPU form sleeping, thus the consumption is higher. :)

msloth:
Argh, messing up the commits a bit, let me fix that in a jiffy.... :{

msloth:
Now, so, finally, commits cleaned up and variable removed too.

PR #1327, opened 2015-10-27
Add option to specify make variables for regression compile tests

nfi:
Some platforms use make variables to specify different build options such as target boards. Adding the option to specify make variables in regression compile tests makes it possible to test compiling for different boards with same target.

```
EXAMPLES = \
hello-world/stm32nucleo-spirit1:BOARD=ids01a4:SENSORBOARD=iks01a1 \
hello-world/stm32nucleo-spirit1:BOARD=ids01a5:SENSORBOARD=iks01a1 \
hello-world/srf06-cc26xx:BOARD=srf06/cc26xx \
hello-world/srf06-cc26xx:BOARD=srf06/cc13xx \
```

### DIFF
--- a/regression-tests/Makefile.compile-test
+++ b/regression-tests/Makefile.compile-test
@@ -42,7 +42,6 @@ get_target_vars = $(wordlist 2,15,$(subst :, ,$1))
 define dooneexample
 @echo Building example $(3): $(1) for target $(2)
 @((cd $(EXAMPLESDIR)/$(1); \
- export STM32W_CPUREV=CC; \
  make $(4) TARGET=$(2) clean && make $(4) TARGET=$(2) WERROR=1) > \
       $(3)-$(subst /,-,$(1))$(2).report 2>&1 && \
  (echo $(1) $(2): OK | tee $(3)-$(subst /,-,$(1))$(2).summary) || \
--- a/regression-tests/Makefile.compile-test
+++ b/regression-tests/Makefile.compile-test
@@ -36,12 +36,14 @@ nine := x x x x x x x x x
 max = $(subst xx,x,$(join ${1},${2}))
 gt = $(filter-out $(words ${1}),$(words $(call max,${1},${2})))
 addzero = $(if $(call gt,${nine},$(1)),$(words ${1}),0$(words ${1}))
+get_target = $(firstword $(subst :, ,$1))
+get_target_vars = $(wordlist 2,15,$(subst :, ,$1))
 
 define dooneexample
 @echo Building example $(3): $(1) for target $(2)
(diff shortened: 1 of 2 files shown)


cmorty:
This is really nice. I'm especially looking forward to increasing the coverage. E.g. by also doing compilation testes with `DEBUG=1` and similar.

regression-tests/Makefile.compile-test
```diff
@@ -36,12 +36,13 @@ nine := x x x x x x x x x
...
 addzero = $(if $(call gt,${nine},$(1)),$(words ${1}),0$(words ${1}))
+get_target = $(firstword $(subst :, ,$1))
+get_target_vars = $(wordlist 2,15,$(subst :, ,$1))
 
 define dooneexample
 @echo Building example $(3): $(1) for target $(2)
 @((cd $(EXAMPLESDIR)/$(1); \
- export STM32W_CPUREV=CC; \
- make TARGET=$(2) clean && make TARGET=$(2)) > \
+ make $(4) TARGET=$(2) clean && make $(4) TARGET=$(2)) > \
```
simonduq: Not sure if something could really go wrong, but, does it make sense to clean with `$(4)` although the previous build was probably not with `$(4)`?

simonduq:
I agree with @cmorty this is very welcome, for increased coverage!

g-oikonomou:
Ditto, this is a very desirable feature

regression-tests/Makefile.compile-test
nfi: I included the ```$(4)``` also for clean in case it will activate some build rules that are needed for clean before compiling with ```$(4)```. Perhaps the safest way would be ```make TARGET=$(2) clean && make $(4) TARGET=$(2) clean && ...``` ?

simonduq: yes that sounds good to me :)

nfi: I tested this and it is not possible to call clean for target ```mbxxx``` without specifying ```STM32W_CPUREV``` (Travis failed). I think there will be more platforms that do not want to have default values to avoid building for the wrong hardware / board. I would suggest leaving it as it is now. We still need to assume that the platform will clean anything that is remaining from last compilation test for any other target. What do you think?

simonduq: ok with leaving it as it is!

simonduq:
:+1:

g-oikonomou:
Closes #972 too by the way

nfi:
Yes, it should be possible to use

nfi:
Rebased due to conflicts with updated master.

nvt:
:+1:

PR #1329, opened 2015-10-28
Fix serialdump.c warnings (clang)

tsparber:
Using clang (on osx) yields to some warnings
(conversion between signed/unsigned, unused variable)
which gcc would only produce when used with -Wall.
This commit fixes those.


simonduq:
:+1: (apparently no need to update the compiled versions -- at least on Linux this PR does not affect the binary)

nvt:
:+1:

PR #1332, opened 2015-10-28
collect-view gui: Add support for OS X serialdump and motelist

tsparber:
Select the tools for macos when checking for motes or opening a
serial dump connection on OS X.


nvt:
:+1:

PR #1336, opened 2015-10-28
Zolertia Z1: Direct memory access using UART.

sumanpanchal:


simonduq:
Great! From offline discussions with @sumanpanchal I know this improves serial reliability significantly, especially on a BR running TSCH (frequent interrupts).

I wonder if this shouldn't be disabled by default at first, so that people can try it without affecting others. Also, I know this was tested on H/W, but does mspsim also support DMA on f2xxx? If not, that's an extra reason to disable by default.

alignan:
@simonduq agree, this should be optional, happy to merge if so.
@sumanpanchal thanks for the contribution!

sumanpanchal:
@simonduq @alignan  Thanks for the feedback. 
As mspsim doesnt have DMA support on f2xxx. I have disabled the DMA for Z1 mote as per your suggestion.

simonduq:
:+1:

sumanpanchal:
@alignan Code style fixed as per your suggestion.

alignan:
:+1:

PR #1348, opened 2015-11-01
Wait for a valid RSSI reading in CC13xx/CC26xx RF drivers

### LINKED ISSUE #1341: cc2650 rssi measurement
There seems to be an issue with the get_rssi command in IEEE mode.

In particular, when I read the RSSI noise (i.e., in absence of packet transmissions) the first time, the module returns a value around -101 dBm, which is reasonable.

However, if the same command is repeated afterwards, for example if I include the NETSTACK_RADIO.get_value(RADIO_PARAM_RSSI, ...) command in a broadcast callback, the returned value is always -128 dBm.

It seems the issue can be corrected by adding a short delay before the rf_core_send_cmd.
I tested the script below and it seems to correct the issue.

   ...
   memset(&cmd, 0x00, sizeof(cmd));
   cmd.commandNo = CMD_GET_RSSI;
   rssi = RF_CMD_CCA_REQ_RSSI_UNKNOWN;

   clock_delay_usec(500);  <-----

   if(rf_core_send_cmd((uint32_t)&cmd, &cmd_status) == RF_CORE_CMD_OK) {
   ...

g-oikonomou:
As discussed in #1341, the current CC13xx/CC26xx IEEE mode driver sends `CMD_GET_RSSI` once and returns the RSSI reading uncondtionally. This happens within the `get_rssi()` function.

This logic is broken if `get_rssi()` is called with the radio off. The function will make sure to turn on the radio first, but it does not make sure the RSSI reading is valid, which only happens a number of symbol periods after the radio enters RX. The outcome is that `NETSTACK_RADIO.get_value(RADIO_PARAM_RSSI, ...)` will always return -128 (meaning that RSSI is unavailable) if the radio was off at the time of calling the function.

The same condition affects the prop mode driver.

This commit changes the logic of `get_rssi()`. If `CMD_GET_RSSI` returns an invalid RSSI, we send it again. For unknown reasons, `CMD_GET_RSSI` on occasion returns 0, so we ignore that value too.

To avoid waiting forever for an RSSI reading that may never become valid, we still return `RF_CMD_CCA_REQ_RSSI_UNKNOWN` after a number of failed attempts.

Fixes #1341

### DIFF
--- a/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
+++ b/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
@@ -384,9 +384,8 @@ static radio_value_t
 get_rssi(void)
 {
   uint32_t cmd_status;
-  int8_t rssi;
   uint8_t was_off = 0;
-  rfc_CMD_GET_RSSI_t cmd;
+  rfc_CMD_IEEE_CCA_REQ_t cmd;
 
   /* If we are off, turn on first */
   if(!rf_is_on()) {
@@ -398,13 +397,19 @@ get_rssi(void)
   }
 
   memset(&cmd, 0x00, sizeof(cmd));
-  cmd.commandNo = CMD_GET_RSSI;
+  cmd.ccaInfo.ccaEnergy = RF_CMD_CCA_REQ_CCA_STATE_INVALID;
 
-  rssi = RF_CMD_CCA_REQ_RSSI_UNKNOWN;
+  while(cmd.ccaInfo.ccaEnergy == RF_CMD_CCA_REQ_CCA_STATE_INVALID) {
+    memset(&cmd, 0x00, sizeof(cmd));
+    cmd.commandNo = CMD_IEEE_CCA_REQ;
 
-  if(rf_core_send_cmd((uint32_t)&cmd, &cmd_status) == RF_CORE_CMD_OK) {
-    /* Current RSSI in bits 23:16 of cmd_status */
-    rssi = (cmd_status >> 16) & 0xFF;
+    if(rf_core_send_cmd((uint32_t)&cmd, &cmd_status) == RF_CORE_CMD_ERROR) {
+      PRINTF("get_rssi: CMDSTA=0x%08lx\n", cmd_status);
+
+      /* Make sure to return RSSI unknown */
+      cmd.currentRssi = RF_CMD_CCA_REQ_RSSI_UNKNOWN;
+      break;
+    }
   }
 
   /* If we were off, turn back off */
@@ -412,7 +417,7 @@ get_rssi(void)
     off();
   }
 
-  return rssi;
+  return cmd.currentRssi;
 }
 /*---------------------------------------------------------------------------*/
 /* Returns the current TX power in dBm */
--- a/cpu/cc26xx-cc13xx/rf-core/prop-mode.c
+++ b/cpu/cc26xx-cc13xx/rf-core/prop-mode.c
@@ -271,6 +271,7 @@ get_rssi(void)
 {
   uint32_t cmd_status;
   int8_t rssi;
+  uint8_t attempts = 0;
   uint8_t was_off = 0;
   rfc_CMD_GET_RSSI_t cmd;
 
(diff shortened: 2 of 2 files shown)


jonnteolsson:
:+1:

fros4943:
Have limited experience with the prop-mode driver, but did observe the same behavior recently. As I understood it then, the rssi returned is a moving average starting from 0 (hence the 0 returned immediately after turning on the radio). I'm currently testing a version which enforces a fixed delay each `get_rssi()`, whereas a proper fix would be to monitor when the radio was turned on, and optionally enforce such a delay.

Did you observe too high readings with your patch? I.e., is the very first rssi value you get higher than the second if your turn on the radio and then quickly `get_rssi()` twice?

g-oikonomou:
I did not test this @fros4943, but I'll do so asap. I only noticed the 0 return value as per the OP. Were we to add a fixed delay as you recommend, how would we calculate a suitable value?

@jonnteolsson any thoughts about this?

fros4943:
I've only tested it very quickly, and then used a delay of 10/32768 seconds.

g-oikonomou:
OK I'll mess around over the weekend.

bkozak-scanimetrics:
I've been doing alot of work with a new radio driver for the 2650 and I've found that I need to wait for 18 rtimer ticks after turning the driver on before taking an RSSI reading. At 16 ticks, the contikiMAC CCAs will miss packets because the RSSI reading is too low even when there is a packet in the air. 

If the rssi reading is based on a moving average, however, then maybe you need to wait for a longer period if the radio signal is weak.

alignan:
Where are on on this one?

g-oikonomou:
Still needs looked into this one

g-oikonomou:
I've got more information on this one now.

* If `get_rssi()` is called with the radio already on, it will return a valid reading first time round.
* If it gets called with the radio off, it will first turn the radio on. With the logic originally proposed in this pull, the code then starts sending `CMD_GET_RSSI` to the CPE until it reads a valid number. `CMD_GET_RSSI` returns the following values in `CMDSTA`.

```
/* RSSI in bits 23:16. Command result in bits 7:0 */
0xff800001 <--- RSSI == -128 (0x80 - invalid), CMD status 0x01 (Done)
0xff800001
0xff800001
0xff800001
0xff800001
0x00000001 <--- This reading is absolute nonsense: RSSI == 0x00, CMD status 0x01 (Done).
0xff800001
0xff800001
0xff800001
0xff800001
0xff800001
0xff800001
0xff9c0001 <--- This is our proper reading: RSSI = -100 (0x9c), CMD status 0x01 (Done).
```
Hence the originally proposed logic to ignore RSSI values of 0.

I have now updated this pull and I am now proposing to use `CMD_IEEE_CCA_REQ` to retrieve RSSI (instead of using `CMD_GET_RSSI`). This command gives us an RSSI reading, but it also gives us access to the state of all CCA criteria (Energy scan, correlator peaks, sync detection).

Thus, we can now inspect the value of `ccaInfo.ccaEnergy` of the command structure to determine whether an RSSI reading is ready. This increases stability, we have direct access to RSSI status valid/invalid, instead of having to draw conclusions based on the actual RSSI reading. Plus, the returned values make a lot more sense and I am no longer seeing those RSSI readings of 0.

Here is a sample output:

```
/* RSSI [CMD_IEEE_CCA_REQ.ccaInfo] */
-128 [0x2a] <--- ccaInfo: ccaState=2 (Invalid), ccaEnergy=2 (Invalid), ccaCorr=2 (Invalid)
-128 [0x2a]
-128 [0x2a]
-128 [0x2a]
-128 [0x2a]
-128 [0x2a]
-128 [0x2a]
-128 [0x08] <--- ccaInfo.ccaEnergy=2 (Invalid)
-102 [0x00] <--- ccaInfo.ccaEnergy=0 (Idle)
```

I should say that this only affects `NETSTACK_RADIO.get_value(RADIO_PARAM_RSSI, ...)`, `.channel_clear()` remains uninfluenced.

@fros4943, I could not reproduce this moving average situation that you describe, but with this pull + other recent changes to the driver perhaps it is no longer observed?

g-oikonomou:
I'm setting a timeout here, I'll merge in a month unless someone disapproves.

PR #1350, opened 2015-11-01
Correctly set new CC26xx TX power when the RFC is powered off

### LINKED ISSUE #1340: cc2650 set TX power
There seems to be an issue with the set_tx_power function. By default, the transmission power in IEEE mode is output_power[0], which corresponds to +5 dBm. At a distance of 10cm, a CC2531 USB dongle receives packets with an RSSI of -39 dBm, which seems reasonable.

However, whenever the NETSTACK_RADIO.set_value(RADIO_PARAM_TXPOWER, ...) function is called (including when the transmission power is set again to +5 dBm), all packets seem to be sent with the lowest transmission power. I indeed receive an RSSI of -76 dBm with the same settings.

Interestingly, the transmission power seems only to change correctly when I change the default value of output_config_t *tx_power_current = &output_power[X].

g-oikonomou:
The radio driver has 2 problems:

* Firstly, detection of whether a command is a Radio Op is broken: Some immediate commands will get categorised as Radio Ops, when in reality they are not. One such command is `CMD_SET_TX_POWER`.
* Secondly: When `set_tx_power()` is called in IEEE mode, it will try to immediately apply the new setting by sending `CMD_SET_TX_POWER`. If this command fails, then the new TX power setting will not be applied at all. One reason of such failure is if the RFC is powered off.

Thus, this pull:

* Fixes the command type classification logic.
* Changes the logic in `set_tx_power()` for IEEE mode. If the new user setting is valid, it always gets saved. If the RFC happens to be powered up, the new setting gets applied immediately with `CMD_SET_TX_POWER`. If the RFC is powered off, the new power setting will get applied next time the RFC is turned on.

Fixes #1340

### DIFF
--- a/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
+++ b/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
@@ -401,25 +401,32 @@ set_tx_power(radio_value_t power)
   int i;
   rfc_CMD_SET_TX_POWER_t cmd;
 
-  /* Send a CMD_SET_TX_POWER command to the RF */
-  memset(&cmd, 0x00, sizeof(cmd));
-
-  cmd.commandNo = CMD_SET_TX_POWER;
-
+  /* First, find the correct setting and save it */
   for(i = OUTPUT_CONFIG_COUNT - 1; i >= 0; --i) {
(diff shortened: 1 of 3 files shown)


jonnteolsson:
(Y)

PR #1351, opened 2015-11-02
Update port.h

Simpleyyt:
Fix a bug.


tonnenpinguin:
Without having looked into the code that uses these macros:
Wouldn't a do{...}while(0) wrapper be appropriate?

Simpleyyt:
when the code without
PORT_CLEAR_X(port,pin) PORT_CLEAR_X(port,pin)
expand, it will be
do {...} whie(0)  do {...} while(0)
syntax error for c.

在2015年11月5，下午 9:55，Berni <notifications@github.com> 
写道：

msloth:
Without being too familiar with this specific file and what it does, it rather seems the bug is whoever wrote those lines accidently wrote `PORT_CLEAR_X()` twice. Is that necessary? Especially on the `TOGGLE` it will (should) just toggle back and forth, not what you would expect.

g-oikonomou:
Yes I think the original was an accidental additional paste.

Simpleyyt:
I agree.

在2015年11月5，下午 10:32，George Oikonomou 
<notifications@github.com> 写道：

Issue #1352, opened 2015-11-02
Reproducible builds


vsaw:
While this does not mean much to many of us it however is an non-deterministic build which can cause severe problems to some (see the [31C3 talk on reproducible builds](https://media.ccc.de/v/31c3_-_6240_-_en_-_saal_g_-_201412271400_-_reproducible_builds_-_mike_perry_-_seth_schoen_-_hans_steiner) for example).

One thing that happened to us was the following scenario:

1. Compile a binary in your ContikiOS repository
2. Flash Binary on Device
3. Forget which version you flashed onto the device ;-)
4. Take a memory dump of the device
5. Copy ContikiOS repository in `/tmp/contiki`
5. Compile binary again in `/tmp/contiki`
6. Compare memdump against the binary compiled in `/tmp/contiki` (Make sure to ~~compile~~ compare the `.bin` and not the ELF as the ELF might contain debug symbols, ~~(~~including the complete path of the compiled source)
7. Notice how binaries are different even though they are being compiled on the same machine by the exact same codebase. Therefore it will be impossible to determine which firmware version this file runs and if the memory is corrupted or not.

Searching for the source, we noticed that the the sourcery ARM compiler creates different binaries based on the order how the `.o` files are added to the static `.a` library. This is easy to fix.

However I'm not sure if this is enough to fix it for all platforms. Possible dependencies to `libc` and other external libraries exist for some platforms that would need to be resolved for truly deterministic builds

PR #1353, opened 2015-11-02
Sort ContikiOS sourcefiles before compilation to make ARM builds reproducible

vsaw:
See https://github.com/contiki-os/contiki/issues/1352

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -100,7 +100,7 @@ else
     CFLAGS += -DUIP_CONF_IPV6_RPL=0
 endif
 
-CONTIKI_SOURCEFILES += $(CONTIKIFILES)
+CONTIKI_SOURCEFILES += $(sort $(CONTIKIFILES))
 
 CONTIKIDIRS += ${addprefix $(CONTIKI)/core/,dev lib net net/llsec net/mac net/rime \
                  net/rpl sys cfs ctk lib/ctk loader . }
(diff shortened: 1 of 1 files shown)


g-oikonomou:
I think you may be chasing a ghost here. As you say in #1352, there are many more factors that may result in a different output image: Library versions, toolchain version to mention 2 of them. I don't think we can or should try to control this from within the build system. If I had your problem, I'd try to solve it by storing metadata either in some sort of database or in the bin itself (security considerations with the latter approach; if a malicious user can see what versions you used to build an image they can potentially exploit any known vulnerabilities).

Mind you, the way things are now and if you are building from a git working dir, it is trivial to extract the source code version from a bin file by exploiting `CONTIKI_VERSION_STRING`, which is stored verbatim in the image.

Here are some examples for c2538dk images. Using a hex viewer:

screen shot 2015-11-02 at 15 05 39

Or, for comedy value:

```
$ cat border-router.bin | LC_CTYPE=C tr -c -d '\11\12\15\40-\176' | grep -o -E "Contiki-[0-9a-z\.\-]+"
Contiki-3.x-1635-g0105454
```

And of course you can do the same from the hex file:
```
$ xxd -ps -r border-router.hex | LC_CTYPE=C tr -c -d '\11\12\15\40-\176' | grep -o -E "Contiki-[0-9a-z\.\-]+"
Contiki-3.x-1635-g0105454
```

vsaw:
We have the fix in our repository and it solves it. Regardless if there are different ways to determine the version of a binary, sorting the compilation units probably wont hurt.

simonduq:
I agree with @g-oikonomou. Also, I tried the patch and found the build order rather confusing. I propose to close this

simonduq:
closing after timeout

PR #1358, opened 2015-11-05
Implement the Wireshark ZEPv2 encapsualtion for the avr-ravenusb.

drugo72:
zep
With this encapsulation Wireshark will display the channel, RSSI,
LQI, frame sequence number the precise frame reception timestamp.
The timestamp is in the NTP format but due to the implementation
it will rollback after about one hour. It is meant for relative
time differences and clocked by the high precision reference clock
of the RF230 (1MHz).
I've also fixed, refactored and removed compiler warnings from
the radio driver.


cdealti:
@simonduq any interest in merging this? If ZEP is not interesting, this PR also contains several fixes to the radio driver.

PR #1361, opened 2015-11-09
cc2538-bsl update to 2.0, cc13xx and cc26xx support

JelmerT:
This PR updates the cc2538-bsl submodule to the latest version (2.0).

**Biggest change**: the script now supports upload of firmware to the **cc13xx and cc26xx chips**. This comes with a big Thank You to @g-oikonomou for adding support for these chips and improving the script in general!
Makefiles of the relevant platforms have been updated to support `.upload` for a 1 command compilation and uploading.

Other improvements:
- Fixed an annoying bug involving garbage characters left over in a serial buffer.
- Added support for uploading `hex` files.
- Improved abstraction in the script to support multiple chips or targets
- Added chip autodetection
- Improved chip info readout
- Improved PY3 support
- Improved read command
- Added basic testing
- Added a `force` option and a `disable bootloader` option catered to production line programming
- And several smaller fixes

Note that with this script now supporting more than only the cc2538-bsl, I intend to change the name of the script to **TI-bsl** in the future. I'll make a new PR once that change is made.

### DIFF
--- a/platform/remote/Makefile.remote
+++ b/platform/remote/Makefile.remote
@@ -5,7 +5,7 @@ ifndef CONTIKI
 endif
 
 PYTHON = python
-BSL_FLAGS += -e -w -v -b 115200
+BSL_FLAGS += -e -w -v
 
 ifdef PORT
   BSL_FLAGS += -p $(PORT)
--- a/platform/remote/Makefile.remote
+++ b/platform/remote/Makefile.remote
@@ -36,11 +36,18 @@ MODULES += core/net core/net/mac \
            core/net/mac/contikimac \
            core/net/llsec
 
+PYTHON = python
+BSL_FLAGS += -e -w -v -a $$($(OBJDUMP) -h $*.elf | grep -B1 LOAD | grep -Ev 'LOAD|\-\-' | awk '{print "0x" $$5}' | sort -g | head -1)
+
+ifdef PORT
+  BSL_FLAGS += -p $(PORT)
+endif
+
 BSL = $(CONTIKI)/tools/cc2538-bsl/cc2538-bsl.py
 
 %.upload: %.bin %.elf
 ifeq ($(wildcard $(BSL)), )
 	@echo "ERROR: Could not find the cc2538-bsl script. Did you run 'git submodule update --init' ?"
 else
-	$(PYTHON) $(BSL) $(BSL_FLAGS) -a $$($(OBJDUMP) -h $*.elf | sed -n '/\<LOAD\>/{g;1!p;};h' | awk '{print "0x" $$5}' | sort -g | head -1) $<
+	$(PYTHON) $(BSL) $(BSL_FLAGS) $<
 endif
--- a/platform/cc2538dk/Makefile.cc2538dk
+++ b/platform/cc2538dk/Makefile.cc2538dk
@@ -28,11 +28,18 @@ MODULES += core/net core/net/mac \
            core/net/mac/contikimac \
            core/net/llsec
 
+PYTHON = python
+BSL_FLAGS += -e -w -v -a $$($(OBJDUMP) -h $*.elf | grep -B1 LOAD | grep -Ev 'LOAD|\-\-' | awk '{print "0x" $$5}' | sort -g | head -1)
+
+ifdef PORT
+  BSL_FLAGS += -p $(PORT)
+endif
+
 BSL = $(CONTIKI)/tools/cc2538-bsl/cc2538-bsl.py
 
 %.upload: %.bin %.elf
 ifeq ($(wildcard $(BSL)), )
 	@echo "ERROR: Could not find the cc2538-bsl script. Did you run 'git submodule update --init' ?"
 else
-	python $(BSL) -e -w -v -a $$($(OBJDUMP) -h $*.elf | sed -n '/\<LOAD\>/{g;1!p;};h' | awk '{print "0x" $$5}' | sort -g | head -1) $<
+	$(PYTHON) $(BSL) $(BSL_FLAGS) $<
 endif
--- a/platform/srf06-cc26xx/README.md
+++ b/platform/srf06-cc26xx/README.md
@@ -102,7 +102,10 @@ To use the port you need:
   as a submodule when you clone Contiki.
 * TI's CC13xxware sources. The correct version will be installed automatically
   as a submodule when you clone Contiki.
-* Software to program the nodes. Use TI's SmartRF Flash Programmer
+* Contiki can automatically upload firmware to the nodes over serial with the
+  included [cc2538-bsl script](https://github.com/JelmerT/cc2538-bsl).
+  Note that uploading over serial doesn't work for the Sensortag, you can use
+  TI's SmartRF Flash Programmer in this case.
 * A toolchain to build firmware: The port has been developed and tested with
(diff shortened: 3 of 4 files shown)


jonnteolsson:
Very nice addition! I will test on cc13xx and cc26xx all supported platforms.

g-oikonomou:
I think `.upload` support should be simply added to the common makefile under srf06, in other words in `..` compared to where it is now.

I also really like the idea of the `PORT` make variable, as per [platform/remote](https://github.com/contiki-os/contiki/blob/master/platform/remote/Makefile.remote#L10), where we invoke with `make PORT=/dev/whatever example.upload`

Other than that, really really keen to get this one merged in.

JelmerT:
Ah yes, `Makefile.srf06` makes more sense. I wanted to stay clear of the sensortag, but went one too deep.

Let me update and add the `PORT`

alignan:
Cool!

I'm actively using this with the `RE-Mote` to flash without manually doing the bootloader dance, I will test this new commit.

JelmerT:
Updated, I also added `PORT` to the cc2538dk.

@alignan any reason why you force the speed to 115200 for the re-mote? It's CC2538 based right? That would make the upload speed quite slow.

alignan:
No reason at all, just a leftover from the first fork, I wasn't sure which was the default speed so we enforced the `115200`

JelmerT:
Default is 500000 and after the oscillator is started it goes up to 1000000.

I'll remove it and add it in this PR, since it's relevant. (You'll have to test it with a RE-mote of course ;) )

g-oikonomou:
I too was wondering last night about that default. I'll give it a whirl without -b myself too

JelmerT:
Also updated cc26xx readme. Should there be a note added on how to use the `.upload` and `PORT` make variables?

g-oikonomou:
Adding an example invocation using `.upload` and `PORT` would be great yes

JelmerT:
Rebased and added `.upload` and `PORT` examples. This turned into a restructure of the readme though, also added a mention of the sleepy node example and folded the srf06 vs sensortag explanation into the Example heading.

This might be getting out of the scope of this PR though, let me know if you want me to split these up.

I also still don't have any CC26xx EMs to test with (poor student and all, hint hint TI ;) ). So please make sure the examples section in the cc26xx readme is correct, since I haven't been able to actually test it.

g-oikonomou:
:+1:

g-oikonomou:
@bradjc, following the discussion in #1385, have you tested this version here? Any feedback?

I am really keen on getting this one in: It combines many new features and many bug fixes.

g-oikonomou:
I forgot to poke @alignan :stuck_out_tongue_closed_eyes:

alignan:
Works great without the `-b 115200` as well :-)

````
Opening port /dev/ttyUSB0, baud 500000
Reading data from zoul-demo.bin
Firmware file: Raw Binary
Connecting to target...
CC2538 PG2.0: 512KB Flash, 32KB SRAM, CCFG at 0x0027FFD4
Primary IEEE Address: 06:16:0E:74:00:12:4B:00
Erasing 524288 bytes starting at address 0x00200000
    Erase done
Writing 524288 bytes starting at address 0x00200000
Write 16 bytes at 0x0027FFF0F8
    Write done                                
Verifying by comparing CRC32 calculations.
    Verified (match: 0x8e273276)
````
Just one suggestion, and I know RTFM (or README for this case) is normmaly suggested, but in case an user stumble upon this:

````
Cannot auto-detect firmware filetype: Assuming .bin
Connecting to target...
ERROR: Timeout waiting for ACK/NACK after 'Get ChipID (0x28)'
````
Perhaps a suggestion about installing `python-magic` and `intelhex` would be appreciated.

**Comment by g-oikonomou** [comment]
> Perhaps a suggestion about installing python-magic and intelhex would be appreciated. 

* `-V` warns about python-magic. If a user knowingly doesn't want python-magic, we don't want to spam him/her with additional unnecessary messages.
* We longer upload a hex file as if it were a bin (which is what the current version in our master does), we fail with an error and inform about IntelHex.

Original discussion about the "whys" here: https://github.com/JelmerT/cc2538-bsl/pull/37

**Comment by bthebaudeau** [comment]
I've not yet looked into this thread, which seems very promising, but it now needs to be rebased because of the merger of #1162. If this new PR switched the Makefiles to .hex instead of .bin hacked with .elf for "upload", that would be great.

**Comment by JelmerT** [comment]
>If this new PR switched the Makefiles to .hex instead of .bin hacked with .elf for "upload", that would be great.

This would mean that you'd have to manually install [extra dependencies](https://github.com/JelmerT/cc2538-bsl#dependencies) for the script before you can upload anything to your chip. Which would make it harder for first time users to get going with Contiki. Is this preferred over the hacky solution with `.elf`?

**Comment by g-oikonomou** [comment]
My main concern is that the script always mass-erases, so if there is something saved in coffee, it will get deleted when re-programming anyway.

**Comment by JelmerT** [comment]
Rebased and fixed bug mentioned in #1162 for OSX. This does not address the `.hex` instead of `.bin` + `.elf` yet or the fact that we always mass-erase possible coffee files.

**Comment by bradjc** [comment]
Just tried this out and it seems to work fine for my use cases. I will update if that changes.

File: platform/remote/Makefile.remote [hunk 1]
```diff
@@ -36,11 +36,18 @@ MODULES += core/net core/net/mac \
            core/net/mac/contikimac \
            core/net/llsec
 
+PYTHON = python
+BSL_FLAGS += -e -w -v -a $$($(OBJDUMP) -h $*.elf | grep -B1 LOAD | grep -Ev 'LOAD|\-\-' | awk '{print "0x" $$5}' | sort -g | head -1)
+
+ifdef PORT
+  BSL_FLAGS += -p $(PORT)
+endif
+
```
g-oikonomou: There is already a block that does all that near line 7 of this file, so with this change we define `PYTHON`, `BSL_FLAGS` etc twice

g-oikonomou:
Here is how I would have written this:

```
PYTHON = python
BSL_FLAGS += -e -w -v

ifdef PORT
  BSL_FLAGS += -p $(PORT)
endif

%.upload: %.bin %.elf
ifeq ($(wildcard $(BSL)), )
	@echo "ERROR: Could not find the cc2538-bsl script. Did you run 'git submodule update --init' ?"
else
	$(eval BSL_ADDRESS_ARG := -a $(shell $(OBJDUMP) -h $*.elf | grep -B1 LOAD | \
	                               grep -Ev 'LOAD|\-\-' | awk '{print "0x" $$5}' | \
	                               sort -g | head -1))
	$(PYTHON) $(BSL) $(BSL_FLAGS) $(BSL_ADDRESS_ARG) $<
endif
```

The benefit of doing so is that one can now see what is being passed as the argument of `-a`, like so:

<pre>
make remote-demo.upload PORT=/dev/tty.usbserial-00001014B 
using saved target 'remote'
[...]
arm-none-eabi-objcopy -O binary --gap-fill 0xff remote-demo.elf remote-demo.bin
<b>python ../../tools/cc2538-bsl/cc2538-bsl.py -e -w -v -p /dev/tty.usbserial-00001014B -a 0x00200000 remote-demo.bin</b>
[...]
</pre>

I'm opening a pull on top of your branch.

Thanks for your efforts!

JelmerT:
Updated with much nicer solution from @g-oikonomou (also removes duplicate)

alignan:
:+1:

PR #1363, opened 2015-11-10
Resolved Contiki error: Reading outside memory

sumanpanchal:
Wismote with large memory model, rpl-border-router is generating error: "reading outside memory". Please look into attached screenshot for the same. Problem has been resolved after few changes in ctimer_set function in ctimer.c.
screenshot from 2015-11-10 13 16 02

### DIFF
--- a/core/sys/ctimer.c
+++ b/core/sys/ctimer.c
@@ -99,15 +99,8 @@ void
 ctimer_set(struct ctimer *c, clock_time_t t,
 	   void (*f)(void *), void *ptr)
 {
-  ctimer_set_with_process(c, t, f, ptr, PROCESS_CURRENT());
-}
-/*---------------------------------------------------------------------------*/
-void
-ctimer_set_with_process(struct ctimer *c, clock_time_t t,
-	   void (*f)(void *), void *ptr, struct process *p)
-{
   PRINTF("ctimer_set %p %u\n", c, (unsigned)t);
-  c->p = p;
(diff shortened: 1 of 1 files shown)


vsaw:
Have you tried to investigate why it's reading from illegal memory? I don't see no reason why the code should be wrong...

nfi:
Which compiler was used? I have seen similar errors when I used mspgcc 4.7.0 with large memory model but those problems were fixed in mspgcc 4.7.2.

sumanpanchal:
Thanks @nfi. I tried with mspgcc 4.7.2 now and works fine. Before I was using mspgcc 4.7.0.
@vsaw. Your PR is safe now!!

vsaw:
@sumanpanchal @nfi Where can I find mspgcc 4.7.2. http://sourceforge.net/projects/mspgcc/ seems to have removed all development releases :(

simonduq:
The binaries used in Travis are at https://github.com/contiki-os/contiki/blob/master/.travis.yml#L20
Have you tried this script? https://github.com/tecip-nes/contiki-tres/wiki/Building-the-latest-version-of-mspgcc

I hope the sources are not gone from sourceforge, else we should have someone upload it somewhere

sumanpanchal:
@vsaw I compiled mspgcc-4.7.2 manually.
@simonduq all the sources related to mspgcc are available on sourceforge.

PR #1365, opened 2015-11-11
sicslowpan.c: warning-free compilation even with fragmentation disabled

simonduq:
This PR fixes unused variable warnings in case 6lowpan fragmentation is disabled


nvt:
:+1:

Issue #1236, opened 2015-09-03
CC26xx / CC13xx - lpm sleep sequence flaws


bkozak-scanimetrics:
I've discovered a number of flaws in the way that the MCU enters sleep mode.

These are:
1. In ```lpm_drop()``` It is possible that an ISR occurs just after calling the lpm modules' request routines which causes a peripheral (such as the radio) to turn on just before the MCU goes to sleep. The subsequent deep sleep will abruptly shut-down the peripheral.
2. In ```lpm_drop()``` it is possible that just after ```process_nevents()``` is called, an ISR fires which then polls a process (like a button ISR calling the sensor process for example). The result will be that the event will not be processed until some indeterminate point in the future (when the next ISR fires).
3. In ```lpm_drop()``` it is possible an rtc interrupt is scheduled which lies in the past (which should indicate that an etimer has expired and requires processing).
4. In ```lpm_drop()``` no rtc interrupt is scheduled if ```next_event``` is zero. As far as I can tell, zero is a valid clock time which could be the expiry time of an actual valid etimer.
5. In the radio driver in the ```transmit()``` routine, it is possible that the last command done interrupt fires just before entering into ```lpm_sleep()``` but just after checking the value of the status register. This will cause the MCU to wake up at an indeterminate point in the future or be reset due to a watchdog expiry.
6. A maximum sleep time is required for sleep mode because the watchdog can expire in sleep mode  (deep sleep mode is ok, I think, because the watchdog is shut-down).
7. Currently, ```lpm_drop``` does not attempt to schedule an rtc interrupt for the etimer if it is going into sleep mode (rather than deep sleep mode).

I realize that there may already be people working on some or all of these issues so I am opening an issue instead of immediately starting work on a patch. I do, however, think that I have a pretty good idea of how to fix all of these problems and can start work on this if no-one else wants to.

g-oikonomou:
> 1.

I don't understand. Even if an interrupt does occur, the module will be subsequently notified before we drop to DS.

> 2.

Correct

> 3

I think I see what you mean

> 4

This is a problem with etimer semantics: If we don't have any scheduled etimers, `etimer_next_expiration_time()` will return 0, which - as you say - is also a valid expiration time. We need some brainstorming here, but I think getting rid of the `if(next_event)` check is probably the way forward, since my gut feeling says that we'll always have at least one etimer scheduled. This is definitely the case for IPv6 builds.

> 5.

[...]

PR #1371, opened 2015-11-14
Improve the stability of CC13xx/CC26xx low-power operation

### LINKED ISSUE #1220: Power consumption issue on CC2650STK very sleepy demo
I have test power consumption on CC2650STK very sleepy demo. I still see about 2 mA@3V in sleep mode. I think this is much higher than expectation. I also find another problem is that the current would lock at about 12 mA@3V after CC2650STK very sleepy demo runs a while and it never goes back to 2mA. Does anyone have test this and get some good results?

g-oikonomou:
This pull addresses a number of issues and applies a number of improvements to the CC13xx/CC26xx low-power logic.

_Firstly_, we make sure the AON RTC CH1 event handler always sets a valid time for the next compare event.

The current logic is to schedule CH1 events on each 512 AON RTC counter boundary while running (not when deep sleeping). However, under certain conditions the AON RTC interrupt handler will end up scheduling the next CH1 compare event too soon in the future, or on some extreme cases even in the past.

When this happens, the LPM logic gets confused thinking that there is not sufficient time to deep sleep and always chooses to put the SoC to sleep instead. The condition manifests itself by observing devices fully operational but otherwise stuck in a high power consumption state, as reported by @chenek in #1220.

AON RTC compare events cannot happen within 2 SCLK_LF cycles after a clearance (4 rtimer ticks == RTC ticks in the 16.16 format). Thus, if the next 512 boundary is too soon (5 rtimer ticks for margin), we skip it altogether. When this happens, etimers that would have expired on the skipped tick will expire 1 event later instead. Skipping a compare event has no negative impact on our s/w clock counter, since this is always read directly from the hardware counter.

_Secondly_, this pull applies a number of improvements to the logic used when trying to drop to a CC13xx/CC26xx low-power mode:
    
* We identify whether there are any pending etimers by using `etimer_pending()` instead of `etimer_next_expiration_time()`. This subsequently allows us to also identify whether an etimer is set to fire at time 0.
* We run a larger portion of the code with the global interrupt disabled. This prevents a number of messy conditions that can occur if an interrupt fires after we have started the low-power sequence.
* We check for pending events earlier in the sequence.
* We make sure to schedule a next wakeup event even when an LPM module prohibits deep sleep and forces sleep instead.
    
This fixes some of the issues discussed in #1236

Fixes #1220

### DIFF
--- a/cpu/cc26xx-cc13xx/dev/soc-rtc.c
+++ b/cpu/cc26xx-cc13xx/dev/soc-rtc.c
@@ -142,13 +142,11 @@ soc_rtc_last_isr_time(void)
 void
 soc_rtc_isr(void)
 {
-  uint32_t now, next;
+  uint32_t next;
 
   ENERGEST_ON(ENERGEST_TYPE_IRQ);
 
   last_isr_time = RTIMER_NOW();
-
-  now = ti_lib_aon_rtc_current_compare_value_get();
 
   /* Adjust the s/w tick counter irrespective of which event trigger this */
   clock_update();
@@ -161,7 +159,8 @@ soc_rtc_isr(void)
      * event on the next 512 tick boundary. If we drop to deep sleep before it
      * happens, lpm_drop() will reschedule us in the 'distant' future
      */
-    next = (now + COMPARE_INCREMENT) & MULTIPLE_512_MASK;
+    next = ((ti_lib_aon_rtc_current_compare_value_get() + 5) +
+            COMPARE_INCREMENT) & MULTIPLE_512_MASK;
     ti_lib_aon_rtc_compare_value_set(AON_RTC_CH1, next);
   }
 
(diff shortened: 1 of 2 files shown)


bkozak-scanimetrics:
Just a couple of points:

- if no etimer is set at all then we can sleep for at most 512 ticks.

 Is there any reason why you can't set a wakeup at some "distant" point (say half the watchdog period) in the future if no etimer is set?

- There is no check to make sure that we wakeup before the watchdog expires in the case where we enter lpm_sleep because a peripheral requested it, then set the wakeup based on an etimer. 

 Although, the peripheral that asked for lpm_sleep should also probably wake the MCU soon via an interrupt, this will not necessarily be possible for every peripheral. Furthermore it may not even be a peripheral which is forcing lpm_sleep, someone might have manually set ```LPM_MODE_MAX_SUPPORTED``` for debugging purposes.

Otherwise I think everything looks good. Looking forward to testing this.

jonnteolsson:
Have tested this over several days, bug appears to be solved. 

:+1:

g-oikonomou:
This fixes a bad bug and overall improves the situation on many fronts. The comments in the last post about sleep mode are valid, so I'm documenting them as open issues in #1236 and I'm merging this one as-is

PR #1372, opened 2015-11-16
RPL probing enhancement

simonduq:
This PR used relative clock diffs rather than absolute clock comparison, for consistent operation across clock_time_t wrap boundaries.


joakimeriksson:
Looks like a good fix - now it looks like many of the timer systems in Contiki (which are also working good with clock wraps).

simonduq:
I should have mentioned this comes from a discussion in https://github.com/contiki-os/contiki/pull/1271

simonduq:
obsoleted by https://github.com/contiki-os/contiki/pull/1403

PR #1375, opened 2015-11-17
off_t changed to signed to stay the same as POSIX

myrfy001:
On many other tool chains, like TI's new MSG430-GCC, the typedef will be conflict.


simonduq:
:+1:

simonduq:
merging after timeout

PR #1376, opened 2015-11-17
Wismote : Direct memory access using UART.

sumanpanchal:


simonduq:
great!
maybe this should be disabled by default (for now), like on the z1?

sumanpanchal:
@simonduq 
Thanks. I have disabled the DMA for Wismote as like Z1.

simonduq:
oh right, that was in the `contiki-conf.h`
:+1:

nvt:
:+1:

PR #1383, opened 2015-11-19
Simplified LLSEC

simonduq:
With https://github.com/contiki-os/contiki/pull/1146, link-layer security is now implemented as as a FRAMER, and the scope of LLSEC has reduced. Only its `name` and `init` function remain really needed, as all functionality has moved away from the `send` and `input` functions. This makes LLSEC not really a layer anymore. This PR removes LLSEC as a layer, but keeps LLSEC as a driver (for `name` and `init`) and of course the link-layer security framer. I find this to also better match what link-layer security is from an architectural point of view: a service rather than a layer.

This PR does not have any (intended!) functional impact on link-layer security; it really only is about architecture.

Am very open to suggestions and discussions. Let me solicit @kkrentz explicitly here.

### DIFF
--- a/core/net/ipv6/sicslowpan.c
+++ b/core/net/ipv6/sicslowpan.c
@@ -1338,7 +1338,7 @@ send_packet(linkaddr_t *dest)
 
   /* Provide a callback function to receive the result of
      a packet transmission. */
-  NETSTACK_LLSEC.send(&packet_sent, NULL);
+  NETSTACK_MAC.send(&packet_sent, NULL);
 
   /* If we are sending multiple packets in a row, we need to let the
      watchdog know that we are still alive. */
--- a/core/net/llsec/llsec.h
+++ b/core/net/llsec/llsec.h
@@ -41,12 +41,7 @@
  * \ingroup net
  * \defgroup llsec Link Layer Security
  * 
- * Layer for implementing link layer security.
- *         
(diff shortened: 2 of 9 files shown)


simonduq:
@kkrentz , any comment?

kkrentz:
As for TSCH and noncoresec, this simplification works, but probably not for my [adaptivesec_driver](https://github.com/kkrentz/contiki/blob/master/core/net/llsec/adaptivesec) if configured to use pairwise keys. The main obstacle there is that I am securing broadcast frames with pairwise keys, which requires sending an extra frame prior to the actual data frame. The cleanest solution to do this seems to be an extra layer ...

simonduq:
Isn't this something that could be done with a sicslowpan `rime_sniffer` (or maybe as an app)? If the current sniffer is not enough, we could maybe extend it. I don't really like having a layer that is unused in the upstream code.

kkrentz:
If you remove the LLSEC layer, I will probably put my stuff into a MAC driver. `rime_sniffer` looks also suitable, but I actually want to implement link layer security below 6LoWPAN.

simonduq:
I'm not merging this without your approval. If you think the PR is not reasonable let us close. The question is more whether you can live with a MAC extension / a sniffer or not.

simonduq:
closing as this didn't get any attention and https://github.com/contiki-os/contiki/pull/1425 restores the some features in the llsec send function

Issue #1380, opened 2015-11-18
New NBRs are not considered (even if table full of non-reachable NBRs)


pablocorbalan:
When the Neighbors Cache Table is full and new neighbours are detected, these new NBRs are dropped because they can not be added to the NBRs table. This happened even if the state of the entries available in the table is not `REACHABLE`, for instance, `DELAY` or `STALE`. With the default configuration with `NDP disabled`, this could happen (in a high network density scenario) as there is no mechanism in the whole RPL code that may remove a neighbor entry from the cache table. Also, neighbours that perhaps are not parents and neither children, may appear in the table with a state that is not `REACHABLE` and take the space of a possible decent parent or child. 

When this happens, we could check the state of the current neighbors and perhaps replace the most outdated neighbor with the new neighbor detected. I do not know if this is non-compliant with neither IPv6 NDP or 6LoWPAN-ND, but I think we could consider something like this. On the other hand, this could lead to the removal of Neighbor Entries that for some reason are not refreshed even when used (this could happen with the preferred RPL parent if its DIO has been suppressed and we do not consider Link Layer ACKs valid to refresh neighbor cache entries). 

Also, a probing mechanism (similar to the RPL probing) to do NUD could be considered, but instead of doing this perhaps we should just use the NDP protocol available.

I could implement the mechanism to replace an outdated neighbor with a new one, but I would prefer to hear some comments before, in case I am missing something or I am mistaken.

Notice that imilar issues are discussed and explained in [#1063](https://github.com/contiki-os/contiki/pull/1063)

joakimeriksson:
Quick question: Do you know if this is due to locking or the default deletion policy of the nbr-table module? From what I remember I usually have seen the opposite problem - throwing out perfectly valid neighbors when new possibly less interesting neighbors were detected. I know that the routing module locks any neighbor that is nexthop and I also the preferred parent is locked. So I guess maybe this is a case where all the neighbors are either the default parent or a nexthop? Should they then be thrown out? (if they get in a bad state - that should be fixed separately - by better interaction between NDP and RPL- as you mention in the other issue #1379 ).

simonduq:
> This happened even if the state of the entries available in the table is not REACHABLE, for instance, DELAY or STALE. With the default configuration with NDP disabled, this could happen 

[...]

PR #1387, opened 2015-11-20
NDP disabled: also disable uip_ds6_neighbor_periodic

simonduq:
Addresses one of the issues in https://github.com/contiki-os/contiki/issues/1380
Nodes running RPL with NDP disabled will now remain in REACHABLE state.
Removing some state in `uip-ds6-nbr` also results in about 20 bytes saved per neighbor.


pablocorbalan:
This gets my :+1: 

The only thing you could update, once you are on it, is to change the `rpl_ipv6_neighbor_callback `debug message and include as well the link metric in the printf messages. In [#1179](https://github.com/contiki-os/contiki/pull/1179) I did not realise sometimes to identify a neighbor is removed you also need to check out the link metric. This could be done in a separate PR though.

simonduq:
good suggestion, will do!

simonduq:
Was about to add the extra info in the printf as suggested by @pablocorbalan but realized this would conflict with https://github.com/contiki-os/contiki/pull/1403. I propose to do the modification later, in a different PR.

As for this one, it is a no-brainer IMO (and we reached a consensus in https://github.com/contiki-os/contiki/issues/1380 that it was needed) so I propose to time it out for merging in a month.

simonduq:
merging after timeout

PR #1388, opened 2015-11-20
Fix RPL use of `acceptable_rank`

simonduq:
In RFC6550 there are three sets of neighbors (see https://tools.ietf.org/html/rfc6550#section-8.2.1):
- **candidate neighbor set:** neighbors reachable with link-local multicast. May become parent, sibling or child;
- **parent set:** nodes above in the DODAG. With the following property: `A node's Rank MUST be greater than all elements of its DODAG parent set.`;
- **preferred parent:** selected one among the parent set, this is the parent currently in use.

Now Contiki's neighbor table `rpl_parents` has a misleading name; it is the candidate neighbor set, not the parent set. We have all sorts of nodes in the table, including with rank higher than us, and including children. That's actually ok, as we want to keep track of the rank and metric of all neighbors, in part as they might one day move to the parent set or as a preferred parent.

This PR removes the filtering out of nodes leading to a rank above `MaxRankIncrease`, which should be done on the parent set, not the candidate neighbor set.

The PR also includes the following fixes:
- adds a sanity check to `acceptable_rank` (needed because `min_rank` is set to infinite at every version update);
- clarifies `rpl_select_dag`, which actually operates on an instance and a dag, not an instance and a parent.

### DIFF
--- a/core/net/rpl/rpl-dag.c
+++ b/core/net/rpl/rpl-dag.c
@@ -1352,15 +1352,12 @@ rpl_recalculate_ranks(void)
 int
 rpl_process_parent_event(rpl_instance_t *instance, rpl_parent_t *p)
 {
-  int return_value;
   rpl_parent_t *last_parent = instance->current_dag->preferred_parent;
 
 #if DEBUG
   rpl_rank_t old_rank;
   old_rank = instance->current_dag->rank;
 #endif /* DEBUG */
-
-  return_value = 1;
 
   if(RPL_IS_STORING(instance)
       && uip_ds6_route_is_nexthop(rpl_get_parent_ipaddr(p))
@@ -1369,19 +1366,6 @@ rpl_process_parent_event(rpl_instance_t *instance, rpl_parent_t *p)
     PRINT6ADDR(rpl_get_parent_ipaddr(p));
     PRINTF("\n");
     rpl_remove_routes_by_nexthop(rpl_get_parent_ipaddr(p), p->dag);
-  }
-
-  if(!acceptable_rank(p->dag, p->rank)) {
-    /* The candidate parent is no longer valid: the rank increase resulting
-       from the choice of it as a parent would be too high. */
-    PRINTF("RPL: Unacceptable rank %u (Current min %u, MaxRankInc %u)\n", (unsigned)p->rank,
-        p->dag->min_rank, p->dag->instance->max_rankinc);
-    rpl_nullify_parent(p);
-    if(p != instance->current_dag->preferred_parent) {
-      return 0;
-    } else {
-      return_value = 0;
-    }
   }
 
   if(rpl_select_dag(instance, p->dag) == NULL) {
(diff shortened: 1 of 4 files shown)


pablocorbalan:
Should we then add a short description of the `rpl_parents` table and detail that the table actually represents the `candidate neighbor set` in line [rpl-dag.c#L77](https://github.com/contiki-os/contiki/blob/c1797e0ffefe15ef9b9278b3d7fb968ba4637938/core/net/rpl/rpl-dag.c#L77) ?. We could also change the name of the table itself I guess.

simonduq:
Good point, I've added a description. I agree we should ideally also rename the table (and many functions names and countless variables)

pablocorbalan:
Thanks for the update! Now, the table is properly explained in the code. The rest of the changes seem good to me, but I am not familiar with the RPL code to give a clear opinion...

core/net/rpl/rpl-dag.c
```diff
@@ -691,29 +696,31 @@ rpl_find_parent_any_dag(rpl_instance_t *instance, uip_ipaddr_t *addr)
...
 /*---------------------------------------------------------------------------*/
 rpl_dag_t *
-rpl_select_dag(rpl_instance_t *instance, rpl_parent_t *p)
+rpl_select_dag(rpl_instance_t *instance, rpl_dag_t *dag)
 {
+  /* This function (1) updates 'dag' (2) chooses a new current DAG in 'instance'
+   * and (3) updates the current DAG. */
   rpl_parent_t *last_parent;
-  rpl_dag_t *dag, *end, *best_dag;
+  rpl_dag_t *d, *end, *best_dag;
```
joakimeriksson: Is there a bit more informative name that we could use for d? Single character variables (beyond i,j) is not optimal when it comes to readability.

joakimeriksson:
Ok, I think the PR looks good - the main difference is that ContikiRPL is not nullifying parents (maybe not the optimal name of that function) when processing "parent events". It is a small change and it clean things up. Maybe just if we could improve the naming of the dag variable "d" (which I think is a bit short) I would be happy with the PR.

core/net/rpl/rpl-dag.c
simonduq: right, maybe we keep `dag` as the loop variable and come up with something else for the function argument

simonduq:
Thanks. I'll look into what you suggest and get back to you

simonduq:
I addressed your comment on naming of `d`.
As for `process_parent_event`, I couldn't agree more, but I feel this should be part of a general cleanup of `rpl-dag.c`, where we need to simplify the logic of e.g. `process_parent_event`, `rpl_select_dag`, `rpl_select_parent` etc. I'm not going to tackle this before https://github.com/contiki-os/contiki/pull/1403 and https://github.com/contiki-os/contiki/pull/1508 are merged.

simonduq:
setting a timeout for merging

Issue #1277, opened 2015-09-24
NDP bug


simonduq:
The function `uip_ds6_nbr_get_ll` returns `const uip_lladdr_t *`, a pointer to the read-only link-layer address used by `nbr-table` as index.

In `uip-nd6.c`, the return value is cast to non-constant pointer in several locations, with:
`uip_lladdr_t *lladdr = (uip_lladdr_t *)uip_ds6_nbr_get_ll(nbr);`
This is followed by writing to `lladdr`, which will corrupt the `nbr-table` module.

A fix would be to remove the old neighbor and add a new one rather than try to write the address in place.

joakimeriksson:
Yes, this is an issue. I think we have just been hit by that bug - it can cause multiple instances of the same lladdr in the nbr-table module which can cause anything including the inability to communicate at all with that neighbor (unrecoverable until global repair in RPL). The only time it is ok to change would be when the lladdr is NULL. But maybe it would be better to not allow for any NULL element in the nbr-module and handle the NULL element separately since it is only one other module that make use of it (uip-nd6). Then there will be no allocations in the nbr module when there is a INCOMPLETE entry installed, but only when it changes to a REACHABLE or STALE state and have both IPv6 address and lladdr.

simonduq:
Handling incomplete entries separately would probably be the best I agree.
But in a first time, I think allowing a single NULL entry (like in the current codebase), but making sure to always remove then add but never write in place should fix the bug.
Note: I do not think it is ever ok to write to the lladdr of a table entry from outside of the nbr-table module; This might result in duplicates, even when the address was NULL before or is becoming NULL.

joakimeriksson:
Yes totally agree that no other module can write the lladdr. If we keep the NULL entry and want to change the lladdr it can only be made by the module. But I guess the safest way is to remove and add. The only case so far when NULL is used is the ND when Incomplete. Maybe we just should never store an Incomplete entry? E.g. we just add a REACHABLE when we get a NA with is-solicited flag set (then we know that we sent the NS).

simonduq:
Both options would work for me (adding only reachable entries OR remove incomplete then add reachable)

joakimeriksson:
The nice thing with  not adding incomplete would be that we could slightly simplify nbr-module and avoid throwing out a possibly important neighbour just because we happen to get an incomplete entry.  For any cases where there are no NA coming back we just do not throw anyone else out.

joakimeriksson:
[...]

PR #1394, opened 2015-11-24
Removed memcpy that cause inconsistency in nbr-table when adding nd6 neighbors.

joakimeriksson:
This is a fix for ND6 that makes the nbr-table have two entries with the same link layer address. This cause permant disconnect from RPL root (until global repair or reboot). The reason it happens is that the nbr table allow adding a NULL entry for the neighbor discovery to handle the case when there is not yet any know link layer address. The NA code then just do a memcopy so that NULL suddenly is a valid link layer address and when that happen to exist in the tables already - things break down. This fix removes the NULL entry and re-installs it again to avoid the problem.

This bug was found by the Intel IoT team in Malaysia that have been testing the 6LoWPAN stack significantly so many thanks for your efforts! (@jeremy-ang, @usmansarwar and @pedquarkbld).

### DIFF
--- a/core/net/ipv6/uip-nd6.c
+++ b/core/net/ipv6/uip-nd6.c
@@ -492,10 +492,11 @@ na_input(void)
   } else {
     uip_lladdr_t *lladdr;
     nbr = uip_ds6_nbr_lookup(&UIP_ND6_NA_BUF->tgtipaddr);
-    lladdr = (uip_lladdr_t *)uip_ds6_nbr_get_ll(nbr);
     if(nbr == NULL) {
       goto discard;
     }
+    lladdr = (uip_lladdr_t *)uip_ds6_nbr_get_ll(nbr);
+
     if(nd6_opt_llao != 0) {
       is_llchange =
         memcmp(&nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET], (void *)lladdr,
@@ -505,8 +506,16 @@ na_input(void)
       if(nd6_opt_llao == NULL) {
         goto discard;
       }
-      memcpy(lladdr, &nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET],
-             UIP_LLADDR_LEN);
+
+      /* Remove this neighbor - since it has a NULL MAC address */
+      uip_ds6_nbr_rm(nbr);
+      /* Re-add this neighbor - now with a correct MAC address */
+      nbr = uip_ds6_nbr_add(&UIP_ND6_NA_BUF->tgtipaddr,
+                            (const uip_lladdr_t *) &nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET],
+                            is_router, NBR_STALE);
+      if(nbr == NULL) {
+        goto discard;
+      }
       if(is_solicited) {
         nbr->state = NBR_REACHABLE;
         nbr->nscount = 0;
--- a/core/net/ipv6/uip-nd6.c
+++ b/core/net/ipv6/uip-nd6.c
@@ -490,14 +490,16 @@ na_input(void)
     PRINTF("NA received is bad\n");
     goto discard;
   } else {
+    const uip_lladdr_t *lladdr;
     nbr = uip_ds6_nbr_lookup(&UIP_ND6_NA_BUF->tgtipaddr);
     if(nbr == NULL) {
       goto discard;
     }
 
+    lladdr = uip_ds6_nbr_get_ll(nbr);
     if(nd6_opt_llao != NULL) {
-      is_llchange =
-        memcmp(&nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET], uip_ds6_nbr_get_ll(nbr),
+      is_llchange = lladdr == NULL ||
+        memcmp(&nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET], lladdr,
                UIP_LLADDR_LEN);
     }
     if(nbr->state == NBR_INCOMPLETE) {
--- a/core/net/ipv6/uip-nd6.c
+++ b/core/net/ipv6/uip-nd6.c
@@ -135,11 +135,13 @@ static uip_ds6_prefix_t *prefix; /**  Pointer to a prefix list entry */
 #if UIP_ND6_SEND_NA || UIP_ND6_SEND_RA || !UIP_CONF_ROUTER
 /*------------------------------------------------------------------*/
 /* Copy link-layer address from LLAO option to a word-aligned uip_lladdr_t */
-static void
+static int
 extract_lladdr_from_llao_aligned(uip_lladdr_t *dest) {
   if(dest != NULL && nd6_opt_llao != NULL) {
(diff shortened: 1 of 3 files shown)


simonduq:
Great, I think this is the same issue as https://github.com/contiki-os/contiki/issues/1277
Two comments:
* I believe you can now remove the unsafe cast `uip_ds6_nbr_get_ll` to non const, i.e. declare `lladdr` as a const *, as discussed in https://github.com/contiki-os/contiki/issues/1277
* Please use `extract_lladdr_aligned` instead of casting `nd6_opt_llao` to `(const uip_lladdr_t *)`, see https://github.com/contiki-os/contiki/pull/1207

joakimeriksson:
It is exactly a fix for that issue and specifically the NULL case which is really happening in real-world cases (which maybe changing MAC address do typically not). I will try to fix the PR according to your comments and update it ASAP.

joakimeriksson:
Ok, that was a bit more changes than I expected - I changed to aligned now - which also ended up forcing me to change the other memcpy that might cause inconsistency (not that it ever happen in a normal case but - now it is also fixed in the same way as the NULL case).

simonduq:
:+1:

core/net/ipv6/uip-nd6.c
```diff
@@ -489,23 +489,32 @@ na_input(void)
...
     if(nbr == NULL) {
       goto discard;
     }
+    extract_lladdr_aligned(&lladdr_aligned);
+
     if(nd6_opt_llao != 0) {
       is_llchange =
-        memcmp(&nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET], (void *)lladdr,
+        memcmp(&nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET], (void *)&lladdr_aligned,
                UIP_LLADDR_LEN);
```
nfi: With the latest update, lladdr_aligned will always set to &nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET] and not NBR lladdr so the compare will always say no change.

simonduq: oops I missed that, needs to compare against `uip_ds6_nbr_get_ll`

joakimeriksson:
Good catch... I will fix now!

joakimeriksson:
It is now fixed and ready to go!

core/net/ipv6/uip-nd6.c
```diff
@@ -489,23 +489,32 @@ na_input(void)
...
-    lladdr = (uip_lladdr_t *)uip_ds6_nbr_get_ll(nbr);
     if(nbr == NULL) {
       goto discard;
     }
-    if(nd6_opt_llao != 0) {
+
+    if(nd6_opt_llao != NULL) {
       is_llchange =
-        memcmp(&nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET], (void *)lladdr,
+        memcmp(&nd6_opt_llao[UIP_ND6_OPT_DATA_OFFSET], uip_ds6_nbr_get_ll(nbr),
```
simonduq: I think you actually need to first get the lladdr, then check it is non-null before `memcmp`

joakimeriksson: I can add that so that if we have an invalid NBR it will even work (if the nbr is from the nor-module it will never have a null entry - yet).

simonduq: yes! (better than memcmp on NULL, which could fail on MCUs with memory protection)

joakimeriksson: Yes - which will soon happen when we remove the NULL MAC address thing in the nor-module ;-)

core/net/ipv6/uip-nd6.c
```diff
@@ -489,23 +489,34 @@ na_input(void)
...
+    const uip_lladdr_t *lladdr;
     nbr = uip_ds6_nbr_lookup(&UIP_ND6_NA_BUF->tgtipaddr);
-    lladdr = (uip_lladdr_t *)uip_ds6_nbr_get_ll(nbr);
     if(nbr == NULL) {
       goto discard;
     }
-    if(nd6_opt_llao != 0) {
+
+    lladdr = uip_ds6_nbr_get_ll(nbr);
+    if(nd6_opt_llao != NULL && lladdr != NULL) {
```
simonduq: don't we want to set `is_llchange` if `lladdr` is `NULL`?

joakimeriksson:
Ok - so to sum it up. I added the null check on lladdr base on @simonduq's comment and also fixed so that we now flush the NBR cache whenever the override flag is set in the NA packet. Also we do not flush the NBR cache when the override flag is not set and there is no change to the lladdr. Before there was a risk of updating the lladdr with a memcpy with the same lladdr - and with my fix it flushed the cache instead of doing nothing when getting a non changed lladdr and a non-override. Note: before we just updated lladdr and never flushed the NBR cache (which the RFC says should be done on override).

joakimeriksson:
I finally moved the change of lladdr into the nbr-module in order to get a consistent model of how to update the lladdr for a neighbor cache entry when the lladdr is changed (either from null to a real value at INCOMPLETE state or from one to another MAC address in other states).

joakimeriksson:
Ping!  Are we ok with having this bug still around? Or is it now solved by other fixes?

core/net/nbr-table.c
```diff
@@ -169,6 +169,26 @@ nbr_set_bit(uint8_t *bitmap, nbr_table_t *table, nbr_table_item_t *item, int val
...
+      if(nbr_get_bit(used_map, all_tables[i], removed_item) == 1) {
+        all_tables[i]->callback(removed_item);
+      }
+    }
+  }
+  /* Empty used map */
+  used_map[index_from_key(least_used_key)] = 0;
+  /* Remove neighbor from list */
+  list_remove(nbr_table_keys, least_used_key);
+  /* Return associated key */
```
simonduq: the above comment should be removed

core/net/nbr-table.c
```diff
@@ -169,6 +169,26 @@ nbr_set_bit(uint8_t *bitmap, nbr_table_t *table, nbr_table_item_t *item, int val
...
+}
+
 static nbr_table_key_t *
```
simonduq: for consistency I would add `/*---------------------------------------------------------------------------*/` here

simonduq:
looks good but I'll need to take a closer look, will do asap

simonduq:
Ready to be merged after my cosmetic comments are addressed (and after some timeout)

joakimeriksson:
Did a rebase - and I guess the timeout should be triggered already!

simonduq:
ah, forgot the timeout tag! let me add

simonduq:
as there was no timeout label, let's give this PR another couple days before merging

joakimeriksson:
Ok - I will be picky next time the tag is not there ;-)

nvt:
:+1: after @joakimeriksson makes some minor updates discusses offline.

simonduq:
are the updates done?

joakimeriksson:
I think that was exactly what I did last commit - so yes!

simonduq:
merging after timeout!

PR #1395, opened 2015-11-25
cc1200: Make it more suitable for contikimac + bugfix in rx restart routine

uknoblic:
With regard to contikimac, i modified the usage of the RX watchdog and calibration behavior. There was a bug in the usage of the etimer used for the RX watchdog and also in the rx_rx() routine which was meant to restart rx in every case of error. In fact, flushing the RX fifo while keeping the cc1200 in rx is not a good idea...

### DIFF
--- a/dev/cc1200/cc1200-conf.h
+++ b/dev/cc1200/cc1200-conf.h
@@ -73,15 +73,15 @@
 #define CC1200_MAX_PAYLOAD_LEN           127
 #endif
 /*---------------------------------------------------------------------------*/
-/* 
+/*
  * The RX watchdog is used to check whether the radio is in RX mode at regular
  * intervals (once per second). Can be used to improve reliability especially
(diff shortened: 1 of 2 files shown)


alignan:
:+1: 

I will wait for Travis to go green

dev/cc1200/cc1200-conf.h
```diff
@@ -159,9 +159,7 @@
 #define CC1200_DEFAULT_CHANNEL          CC1200_CONF_DEFAULT_CHANNEL
 #else
 /* 868.325 MHz */
-//#define CC1200_DEFAULT_CHANNEL      26
-/* 865.725 MHz */
-#define CC1200_DEFAULT_CHANNEL          13
+#define CC1200_DEFAULT_CHANNEL			26
```
alignan: Replace tabs by spaces

PR #1396, opened 2015-11-25
Print IPv6 addresses properly in ICMPv6 and NDP debug messages

pablocorbalan:
This PR simply fixes some debug log messages in which IPv6 addresses are printed together with some other text, instead of separated by white spaces.


simonduq:
:+1:

PR #1399, opened 2015-11-25
Added PWM driver for the Zolertia Zoul module and CC2538 platforms

alignan:
This PR adds support for PWM on CC2538-based platforms.

* Frequencies from 1 Hz to 160 KHz.
* Duty cycle from 1-99%.
* Allows to toggle duty cycle.
* The pin state when the PWM signal is stopped can be defined by the user (low/high)
* `GPT0 timer A` is reserved for `clock_delay_usec` so returns an error if selected


adamdunkels:
I have tested this on the RE-Mote hardware and can confirm it works nicely :+1:

One thing to note is that PWM requires `lpm_max_pm(0);` to avoid the PWM timers being stopped during low-power mode.

alignan:
@adamdunkels thanks! I will include a note in the driver

g-oikonomou:
Are we ready / happy with this?

alignan:
:+1: for me as expected

g-oikonomou:
Documentation added as requested.

An additional future improvement can be made by using `lpm_register_peripheral()` within `pwm.c`.

I'm merging this :+1:

PR #1400, opened 2015-11-26
Refresh nbr reachable state after received IPv6 unicast message

pablocorbalan:
This PR creates a function called `uip_ds6_nbr_refresh_reachable_state`in `uip-ds6-nbr.c` that refreshes the `Reachable` state of a neighbor. This function should be called (if NDP enabled, i.e., `UIP_ND6_SEND_NA`enabled) when the IPv6 layer or higher receives a message from a neighbor that confirms its reachability, for instance, DAO ACK or ECHO Reply messages. By refreshing the reachability of a neighbor with these messages, we may reduce the amount of traffic that NDP creates.

This PR addresses some of the comments and suggestions mentioned in [#1380](https://github.com/contiki-os/contiki/issues/1380). Let me know your suggestions/improvements.

### DIFF
--- a/core/net/ipv6/uip-ds6-nbr.c
+++ b/core/net/ipv6/uip-ds6-nbr.c
@@ -322,6 +322,20 @@ uip_ds6_neighbor_periodic(void)
   }
 }
 /*---------------------------------------------------------------------------*/
+#if UIP_ND6_SEND_NA
+void
+uip_ds6_nbr_refresh_reachable_state(const uip_ipaddr_t *ipaddr)
+{
+  uip_ds6_nbr_t *nbr;
+  nbr = uip_ds6_nbr_lookup(ipaddr);
+  if(nbr != NULL) {
+    nbr->state = NBR_REACHABLE;
+    nbr->nscount = 0;
+    stimer_set(&nbr->reachable, UIP_ND6_REACHABLE_TIME / 1000);
+  }
+}
+#endif /* UIP_ND6_SEND_NA */
+/*---------------------------------------------------------------------------*/
 uip_ds6_nbr_t *
 uip_ds6_get_least_lifetime_neighbor(void)
 {
--- a/core/net/ipv6/uip-ds6-nbr.c
+++ b/core/net/ipv6/uip-ds6-nbr.c
@@ -96,8 +96,12 @@ uip_ds6_nbr_add(const uip_ipaddr_t *ipaddr, const uip_lladdr_t *lladdr,
     uip_packetqueue_new(&nbr->packethandle);
 #endif /* UIP_CONF_IPV6_QUEUE_PKT */
 #if UIP_ND6_SEND_NA
-    /* timers are set separately, for now we put them in expired state */
-    stimer_set(&nbr->reachable, 0);
+    if(nbr->state == NBR_REACHABLE) {
+      stimer_set(&nbr->reachable, UIP_ND6_REACHABLE_TIME / 1000);
+    } else {
+      /* We set the timer in expired state */
+      stimer_set(&nbr->reachable, 0);
+    }
     stimer_set(&nbr->sendns, 0);
     nbr->nscount = 0;
 #endif /* UIP_ND6_SEND_NA */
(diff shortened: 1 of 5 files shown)


simonduq:
Great, especially with the new `uip_ds6_nbr_refresh_reachable_state`!
How about this: shouldn't we refresh the reachable state anytime we get an IP packet from a neighbor, rather than at many different specific places in RPL etc? After all, if we get an IP packet from a neighbor it means (1) the neighbor still runs its IP stacks and has the same IP address (2) the neighbor is reachable, else it would never hear from us, conclude we're not reachable (NUD involves receiving from the neighbor) and stop talking to us.

joakimeriksson:
Well as long as it is Unicast it is ok (e.g. not multicast DIO, DIS) - even if RPL today keeps neighbors in there on multicast. I guess the main thing with the rules in the RFC is to avoid keeping a neighbor alive because of link-layer activity.

simonduq:
Yes sorry, I did have unicast-only in mind. For refreshing at least. We might still need the initial addition  to the table on DIO.

pablocorbalan:
@simonduq I see your point as we would not probably receive any IPv6 unicast message from a neighbor unless this neighbor got successfully one of our packets before. I think this would be `almost` as efficient as link layer ACKs in a `typical` 6LoWPAN application. Not completely sure if this is completely `compliant` with the RFC, but it should be ok as @joakimeriksson says.

Should I change then the call to `uip_ds6_nbr_refresh_reachable_state` to [uip6.c#L1144](https://github.com/contiki-os/contiki/blob/master/core/net/ipv6/uip6.c#L1144)? Should we also check within the `uip_ds6_nbr_refresh_reachable_state` if the neighbor was `INCOMPLETE` or something else?

joakimeriksson:
The only case I can come up with is a radio-packet stuck in a faulty radio or lower layer that corresponds to a full IPv6 frame and that is sent repeatedly (now and then) even if there is no longer any functional IP stack. The RFC says something about forward progress. Not sure how we could check that on UDP packets and other things. From RFC: about reachability: " When possible, upper-layer protocols provide a positive confirmation that a connection is making "forward progress". I think it is better than link layer ACKs at least (easy to see that ACK might come without a working IP stack).

simonduq:
@joakimeriksson pretty corner case ;)
I think refreshing on any incoming IPv6 datagram is fine. Plus, this is something we're already doing anyway (on RPL ICMP input), but refactored and more general.
@pablocorbalan yes this looks like the right location in uip6. But please double-check the runtime behavior, and that it works nicely both with ICMP and other types.
Your current `uip_ds6_nbr_refresh_reachable_state` looks fine to me.

pablocorbalan:
I won't have time to work on this the next week, but I will run some simulations the week after to see if everything works as expected and I'll let you know my findings and then we can see how to improve this and merge it. Sorry for the delay.

simonduq:
Any update @pablocorbalan?
I feel we probably want to extend this a bit and refresh the state in more places as discussed above.

pablocorbalan:
Sorry, I've been busy and I left this PR. Let me try to update and test it during this week and I'll let you know my progress.

pablocorbalan:
@simonduq I have updated the PR with the changes we talked about. Now, the state should be refreshed after every incoming IPv6 unicast message. I changed the name of the PR to reflect the changes and the commit message as well. 

I haven't done any extensive evaluation of this aspect but I have seen that the function to refresh the state is called appropriately when NA enabled and after every incoming unicast IPv6 packet.

I am sorry I took that long to just update the PR. I have been quite busy (and also sick) these days. Let me know if you want me to do some more testing or updates.

simonduq:
This looks good. Very useful. One thing: there are a couple left-over state refresh in core that seem not needed anymore. How about simply taking these out?

pablocorbalan:
@simonduq I have rebased the PR and added a couple of commits to it. 

Firstly, I made sure that `uip_ds6_nbr_refresh_reachable_state()` is only called upon receiving unicast messages.

Secondly, I updated `uip_ds6_nbr_add()`to set the reachable timer to `UIP_ND6_REACHABLE_TIME / 1000` if the neighbor being added is `NBR_REACHABLE`. This is necessary to avoid calling `uip_ds6_nbr_refresh_reachable_state` or similar after adding a reachable neighbor. Then, I removed the unneeded left-over state refresh in `uip-nd6.c` and `rpl-icmp6.c`. 

Let me know if you want me to do any further modifications.

simonduq:
Great, :+1:

simonduq:
setting a timeout for merging

simonduq:
merging after timeout, thanks!!

PR #1402, opened 2015-11-29
Travis: improved output for failed Cooja simulation

simonduq:
Simply print out the standard output from java instead of the Cooja log.
stdout includes the full Cooja log, but also error messages in cases the JVM crashes (e.g. native node triggers segfault).


simonduq:
will merge that one in a month if nobody argues

alignan:
:+1:

I'm OK with merging now, if you want to hit the button

simonduq:
Thanks

PR #1403, opened 2015-11-29
Compliance fixes with RFC 6550, 6551, 6552, 6719

simonduq:
This PR fixes a number of discrepancies between the RoLL RFCs and our Contiki implementation, with a focus is on the OFs and metrics. This also moves link statistics from the MRHOF OF to an more general module, `link-stats`, which can work with any RPL OF or other protocols than RPL.

Overview of fixes and features:
* **MRHOF:** fixed terminology, implemented MAX_LINK_METRIC and MAX_PATH_COST as per RFC6719. Implement MinHopRankInc and fix ETX divisor to 128 as per RFC6551, while keeping a configurable MinHopRankInc (as discussed in section 6.1 of RFC6719).
* **OF0:** complete rewrite to match RFC6552, now with rank stretch, rank factor, step of rank. OF0 can work with different link metrics, and it is now possible to configure which metric to use. One of the configuration options is ETX-based, and matches the work going on in the 6TiSCH WG. The other implementation has a constant step of rank and results in hop count.
* **Metric Containers:** now the MC is not configured at compile time, but a run-time when joining. There is a compile-time config flag to disable MC altogether to save memory.
* **OFs:** now all (two) OFs are embedded in the code, and selected on a per-instance basis when joining.
* **Link statistics:** new `link-stats` module keeping track of ETX, link freshness and RSSI, independently from RPL. Improved ETX maintenance with (1) two-speed EWMA for fast bootstraping and (2) optional RSSI-based ETX initialization.
* **RPL:** various more minor fixes on RPL to match RFC6550. Now checking IPv6 reachability of candidate preferred parents and children. Added a check, whenever a new preferred parent is found, that we have up-to-date link estimates to that node. If node, schedule urgent probing and switch only once the statistics are refreshed, if the node is still appealing. As for children un-reachable children, their routing entry is now automatically removed and a no-path DAO sent upwards.
* **RPL probing:** now based on link statistics freshness, now supporting "urgent probing".


simonduq:
Added one more fix, which consists in removing routes via unreachable children and sending a no-path up. "Unreachable" is defined as here "ND-unreachable" or "present in table and with fresh link statistics and with bad link metric". This is in compliance with RFC6550 section 8.5.1 ```Routes through unreachable neighbor MUST be removed from the routing table.```. Fixes Travis flaky test 12-rpl/07-rpl-random-rearrangement.csc.

Upstream:
```
Running test 07-rpl-random-rearrangement with random Seed 1: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 2: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 3: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 4: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 5: ............................. OK
Running test 07-rpl-random-rearrangement with random Seed 6: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 7: ............................FAIL
Running test 07-rpl-random-rearrangement with random Seed 8: .......................... OK
Running test 07-rpl-random-rearrangement with random Seed 9: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 10: ........................... OK
Running test 07-rpl-random-rearrangement with random Seed 11: ............................FAIL
Running test 07-rpl-random-rearrangement with random Seed 12: .............................FAIL
Running test 07-rpl-random-rearrangement with random Seed 13: .............................FAIL
Running test 07-rpl-random-rearrangement with random Seed 14: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 15: ........................... OK
Running test 07-rpl-random-rearrangement with random Seed 16: ............................. OK
Running test 07-rpl-random-rearrangement with random Seed 17: .............................. OK
```

WIth the fix
```
Running test 07-rpl-random-rearrangement with random Seed 1: ........................... OK
Running test 07-rpl-random-rearrangement with random Seed 2: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 3: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 4: ............................. OK
Running test 07-rpl-random-rearrangement with random Seed 5: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 6: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 7: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 8: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 9: ............................. OK
Running test 07-rpl-random-rearrangement with random Seed 10: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 11: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 12: ........................... OK
Running test 07-rpl-random-rearrangement with random Seed 13: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 14: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 15: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 16: ............................ OK
Running test 07-rpl-random-rearrangement with random Seed 17: ........................... OK
```

pablocorbalan:
I like the idea of the `link-stats`module a lot. It makes the neighbours table more comprehensive, and probably it also moves some complexity out of RPL. 

I see you are also including the RSSI ETX initialisation and Squared ETX configuration that you proposed in Orchestra. If these get merged, then all the proposed Orchestra RPL changes will be included or available in the main repo, am I right?

The only problem I see with this PR is that by allowing runtime reconfiguration of the RPL MC and OF, we may use some extra-memory that could be used for other purposes in very constrained platforms (TelosB). Is it possible to change this with any `#define`?

This seems to need a rebase though, and there are some indentation problems (specially in the first commit), but very nice to see all these changes/updates coming to main repo. I will probably consider using the Squared ETX in the close future.

pablocorbalan:
@simonduq Would you also consider including `LQI` in the `link-stats`module? I just read a poster in which they can characterise good and bad links (i.e., high or low PRR) in a fast way only based on `LQI` variance for instance, so it might be interesting to have `LQI` information as well in the `link-stats`.

simonduq:
> all the proposed Orchestra RPL changes will be included or available in the main repo, am I right?

That is correct, but note that none of the enhancement are Orchestra or even TSCH-specific. We found them to also help with ContikiMAC or NullRDC, as long as the main objective is reliability.

> The only problem I see with this PR is that by allowing runtime reconfiguration of the RPL MC and OF, we may use some extra-memory that could be used for other purposes in very constrained platforms (TelosB). Is it possible to change this with any #define?

Yes it is, see `RPL_SUPPORTED_OFS`

One could also add LQI but I'd say maybe not until this is used somewhere in the codebase.

joakimeriksson:
I think this looks very good - but I have not tested it yet. But many of the fixes and improvements are very very good to get into the Contiki RPL implementation!

core/net/link-stats.c
```diff
@@ -0,0 +1,212 @@
...
+
+  stats = nbr_table_get_from_lladdr(link_stats, lladdr);
+  if(stats == NULL) {
+    /* Add the neighbor */
+    stats = nbr_table_add_lladdr(link_stats, lladdr);
+    if(stats != NULL) {
+      /* Initialize */
+      stats->rssi = packet_rssi;
+      stats->etx = LINK_STATS_INIT_ETX(stats);
+    } else {
```
pablocorbalan: I think that you should `return`whether you manage to add the neighbor or not. If you don't return after `stats->etx = LINK_STATS_INIT_ETX(stats);`, then you will execute the RSSI EWMA even if the neighbor was just added. In that case, the EWMA does not change the value of `stats->rssi` because  `stats->rssi` and `packet_rsssi`are equal. Then, the updated function would look like:
```
void
link_stats_input_callback(const linkaddr_t *lladdr)
{
  struct link_stats *stats;
  int16_t packet_rssi = packetbuf_attr(PACKETBUF_ATTR_RSSI);

  stats = nbr_table_get_from_lladdr(link_stats, lladdr);
  if(stats == NULL) {
    /* Add the neighbor */
    stats = nbr_table_add_lladdr(link_stats, lladdr, NBR_TABLE_REASON_LINK_STATS, NULL);
    if(stats != NULL) {
      /* Initialize */
      stats->rssi = packet_rssi;
      stats->etx = LINK_STATS_INIT_ETX(stats);
    }
    return;
  }

  /* Update RSSI EWMA */
  stats->rssi = ((int32_t)stats->rssi * (EWMA_SCALE - EWMA_ALPHA) +
      (int32_t)packet_rssi * EWMA_ALPHA) / EWMA_SCALE;
}
```

simonduq: agree, will do, thanks!

simonduq:
@alignan this PR is ready! It is a subset of the non-storing PR https://github.com/contiki-os/contiki/pull/1508, so should be a no-brainer :)
I'd like to get this in before rebasing the other.

core/net/link-stats.c
simonduq: done

joakimeriksson:
This is a good one - half of it has been tested significantly at Yanzi in deployments for a very long time (months) now - and we are very happy with it! 👍

alignan:
Seems great to me :+1: 
Just one question, by how much does this change increases RAM usage? asking because the  `NBR_TABLE_CONF_MAX_NEIGHBORS` changes

simonduq:
On cc2538, this PR takes an extra 656 bytes ROM and 24 bytes RAM. The space per neighbor decreases from 61 to 59 bytes RAM.

I did just re-add a commit I had initially removed, which I think actually is needed https://github.com/contiki-os/contiki/pull/1403/commits/797bead5e54fd07a87cd46d0b9afe0cd38f516e6 (it is what made `07-rpl-random-rearrangement` more reliable).

simonduq:
Travis green, merging!

PR #1404, opened 2015-11-30
Add route_discovery_explicit_open()

amitgeron:
For route discovery connections, 2 logical channels are used:
One channel for netflood messages, and one for unicast.

When opening a route discovery connection using
route_discovery_open(), only one channel number is specified
and used for netflood, and the unicast channel number is
hard-coded to be the next channel (+1).

This commit adds an alternative function for opening route
discovery connections, allowing the user to specify both channels
independently, to gain improved control, flexibility and readability.

The existing function (route_discovery_open()) was modified to
make use of the new function, to prevent code duplication,
without breaking the expected behavior.


simonduq:
:+1:

PR #1408, opened 2015-12-02
JN516x: tickless clock, power saving update, 32 kHz rtimer

simonduq:


simonduq:
Not merging yet, there is known bug in the tickless clock.

simonduq:
Now with bugfix

PR #1409, opened 2015-12-02
Implementation of OMA LWM2M Engine / IPSO Objects

nfi:
This is an implementation of Open Mobile Alliance Lightweight M2M (OMA LWM2M) and IPSO Objects for Contiki.

It is based on the Erbium CoAP implementation and consists of two apps: `lwm2m-engine` and `ipso-objects`. The `lwm2m-engine` handle the specifics of LWM2M including bootstrapping and how read/writes of objects and resources are handled. The `ipso-objects` contains implementations of some of the IPSO Smart Objects.

This implementation was used during the IPSO Interop in May 2015, Kista, Sweden, and was successfully tested with other implementations. We have Github wiki documentation that will be uploaded if/when the PR is accepted.

There are some examples for controlling LEDs and reading out temperature. See `examples/ipso-objects` and `apps/ipso-objects`.

Supports:
* OMA TLV and plain text format
* Bootstrapping and registration on LWM2M server
* LWM2M read/write/execute operations

Limitations:
* no support for DTLS (yet)
* partial support for JSON and different formats

Future work:
* update Erbium CoAP to allow more dynamic setup of LWM2M objects
* add templates for a variety of objects to make it easier to implement IPSO objects and LWM2M management objects
* add implementation of more IPSO Smart Objects
* add DTLS support
* better error handling
* add support for full firmware upgrade over LWM2M

### DIFF
--- a/apps/ipso-objects/Makefile.ipso-objects
+++ b/apps/ipso-objects/Makefile.ipso-objects
@@ -1,0 +1,3 @@
+ipso-objects_src = ipso-temperature.c ipso-button.c ipso-leds-control.c \
+	ipso-light-control.c ipso-objects.c
+CFLAGS += -DWITH_IPSO=1
--- a/apps/ipso-objects/ipso-button.c
+++ b/apps/ipso-objects/ipso-button.c
@@ -1,0 +1,163 @@
+/*
+ * Copyright (c) 2015, Yanzi Networks AB.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDER AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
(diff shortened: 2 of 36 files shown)


g-oikonomou:
!!

Can you briefly summarize what needs to be done before a platform can support this?

alignan:
Nice one!

I was testing [this implementation](https://github.com/jenseliasson/iot-apps) as well, so it is great to have a supported Contiki implementation.

Instructions to get platforms running this would be appreciated, easier to test myself and review.

nfi:
Nothing really. For example you can compile the IPSO Objects example for
different platforms:

  cd examples/ipso-objects
  make TARGET=wismote
  make TARGET=cc2538dk

If the platform has leds, the example will setup IPSO Light control to
control the leds. However if you want to expose platform specific sensors
and actuators, you need to add corresponding IPSO objects.
Den 2 dec 2015 16:03 skrev "George Oikonomou" <notifications@github.com>:

jonnteolsson:
this is great to see!! I will test it on our CC13xx and CC26xx platforms

simonduq:
Great!! My only suggestion would be to add some non-regression testing, at least compilation, with full coverage of the new features if possible.

joakimeriksson:
Yes, that we can do. We already have some simulations in COOJA so it should be easy to turn it into a regression test.

simonduq:
Yes, even better with simulations! (but I'd like compilation anyway, because this is where we have -Werror :))

joakimeriksson:
Sure, we will also add to compilation! Maybe that first!

joakimeriksson:
I just fixed a minor issue with the light-control - now it works with any platform that supports red, blue and green leds. Will add a wiki page on how to test it also.

dbjorklu:
Hi,
It seems your are returning responses in content type LWM2M_TEXT_PLAIN 
to GET requests to resources of string type, and content type LWM2M_TLV 
for others (float,int etc), am I right?
This regardless of the requested content format.

In the standard I find e.g. this sentence:

"The Object specification defines the data format that a Resource 
supports, either plain text or opaque for singular Resources or TLV for 
multiple instance Resources."

So my interpretation of this is was that if the "Instances" column in 
the "Resource definitions" of an object says "Single" one should use 
plain text or opaque; if it says "Multiple" TLV should be used.

So TLV would not usually be used for the representation of a single 
resource.

Anyway if a do a GET Accept: LWM2M_TEXT_PLAIN to e.g. 3303/0/5700 I 
would expect to get the value as an ASCII decimal.

Could explain how you reasoned on this?

Thanks,
     Dag Björklund

joakimeriksson:
This is an initial implementation and it is from our IPSO Interop test in May. It is not yet complete and in some cases might be not 100% compliant. I do think that ok to respond with TLV for single-valued resources but I agree that there is a misfeature that it currently responds with TLV when you request in LWM2M_TEXT_PLAIN. We do have support for TLV and plain-text in our readers/writers so it should not be too hard to improve on this.

dbjorklu:
Ok good, just wanted to know if this was a design decision or missing 
features.

joakimeriksson:
I added some test instructions in a readme now:
https://github.com/nfi/contiki/blob/oma-lwm2m-ipso/examples/ipso-objects/README.md

dbjorklu:
Another similar comment..

Doing a GET Accept:application/link-format to /3303/0/ on the 
example-ipso-objects node results in a 5.03. This is because the result 
does not fit into 64 B. I would say it happens quite easily that the 
result of a discovery does not fit into one packet and blockwise 
transfer would be required.

So.. maybe the implementation of block-wise could be added to the 
limitations or future work list?

Regards,
    Dag

alignan:
Tested and working OK, just as suggested by @nfi if using the `rpl-boder-router` one needs to increate the `UIP_CONF_BUFFER_SIZE` up to 240 or so.

:+1:

alignan:
@simonduq are you OK with merging this one?

simonduq:
:+1:

PR #1411, opened 2015-12-04
Rime Chameleon fix for 32-bit platforms

simonduq:
use BITOPT_HDR_SIZE rather than sizeof(struct bitopt_hdr)

### DIFF
--- a/core/net/rime/chameleon-bitopt.c
+++ b/core/net/rime/chameleon-bitopt.c
@@ -57,6 +57,8 @@ struct bitopt_hdr {
   uint8_t channel[2];
 };
 
+#define BITOPT_HDR_SIZE 2
+
 static const uint8_t bitmask[9] = { 0x00, 0x80, 0xc0, 0xe0, 0xf0,
 				 0xf8, 0xfc, 0xfe, 0xff };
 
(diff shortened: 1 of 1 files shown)


tonnenpinguin:
Out of curiosity: Why is using the preprocessor define in this case better than using sizeof?

simonduq:
Because ` sizeof(struct bitopt_hdr))` is 4 on 32-bit platforms (rounds up to word size), while the real, intended bitopt header size is 2.

simonduq:
A consequence is that a 32-bit node and a 16-bit node won't be able to communicate as they will work on different header sizes.

tonnenpinguin:
I see, I see..
Thanks for the explanation!

tonnenpinguin:
One more question though: Is this sizeof behaviour compiler specific?
I just tried it on my CC2650 (which is 32 Bit AFAIK) and sizeof returns 2..

simonduq:
I think it is compiler-specific yes, what it returns is the size the struct will actually take in memory.
Interesting that you get 2 on Cortex M3, I would have guessed 4.
In our case the problem arose on JN516x, where it was 4.

tonnenpinguin:
I've only seen this varying code size when the compiler needs to do some padding...really interesting indeed! Learn something new every day I guess!

PR #1416, opened 2015-12-11
Added support for the RE-Mote on-board Real Time Clock Calendar (RTCC)

alignan:
This PR adds support for the RE-Mote's on-board Real Time Clock Calendar.
More information about the RTCC (Abracon AB080x) can be found in our [Resource](https://github.com/Zolertia/Resources/tree/master/RE-Mote/Hardware/Revision%20A/Datasheets/Components) repository

The port is missing a function to select the oscillator source, but it is hardcoded to use the XT (external crystall).  A pending improvement would be also to wake up the device from PM3.


alignan:
Merging already as it is a platform-dependant feature and I'm the platform maintainer

PR #1417, opened 2015-12-11
We can now set BUILDDIR to build in seperate dir

bkozak-scanimetrics:
The make variable BUILDDIR can now be used to specify a separate build directory for project builds.

By default everything is still built in the project's root but if we set BUILDDIR, all of the build files (including output executables) are moved to BUILDDIR. This will help when building a single project with multiple configurations and will also help to keep the project tree clean.

I am not able to test on every platform so I will appreciate it if anyone takes the time to confirm that it works for them.

Test with 
```make BUILDDIR=build <other make options> <target>```


alignan:
This is a handy feature, works well for the `zoul` and `z1`.  However when cleaning up:

`make BUILDDIR=test clean`

Deletes all files in the directory but the empty directory, it would be better to remove as well.

bkozak-scanimetrics:
> Deletes all files in the directory but the empty directory, it would be better to remove as well.

Agreed but I didn't do this just in case the BUILDDIR happens to point to the root of the source tree (as it will by default). I suppose I could add a specific check to work around this if people really want the build directory gone on make clean.

bkozak-scanimetrics:
I've done some maintenance on this branch

- Performed a rebase
- Fixed the merge conflict caused by the deletion of sensinode 
- Modified ```Makefile.customrules-galileo``` to work with my scheme

alexrayne:
works fine. could i some how vote to accept it into master?

PR #1419, opened 2015-12-13
cc2538: pwm: Fix clock and automatically disable PM1+ if running

bthebaudeau:
This is a follow-up of #1399.

The peripheral core clock of the general-purpose timers used by the PWM driver is the system clock, not the I/O clock (see the CC2538 user's guide).

The peripheral core clocks of the PWM timers are gated in PM1+, so these power modes must be disabled if a PWM timer is running. Use lpm_register_peripheral() to handle this automatically and dynamically (suggested by @g-oikonomou in #1399).

### DIFF
--- a/cpu/cc2538/dev/pwm.c
+++ b/cpu/cc2538/dev/pwm.c
@@ -46,6 +46,7 @@
 #include "dev/gpio.h"
 #include "dev/sys-ctrl.h"
 #include "dev/pwm.h"
+#include "lpm.h"
 #include <stdio.h>
 #include <stdlib.h>
 /*---------------------------------------------------------------------------*/
@@ -73,6 +74,21 @@ pwm_configured(uint8_t timer, uint8_t ab)
   return 0;
 }
 /*---------------------------------------------------------------------------*/
+static bool
+permit_pm1(void)
+{
+  uint8_t timer, ab;
+
+  for(timer = PWM_TIMER_0; timer <= PWM_TIMER_3; timer++)
+    for(ab = PWM_TIMER_A; ab <= PWM_TIMER_B; ab++)
+      if(pwm_configured(timer, ab) &&
+         REG(PWM_GPTIMER_NUM_TO_BASE(timer) + GPTIMER_CTL) &
+           (ab == PWM_TIMER_A ? GPTIMER_CTL_TAEN : GPTIMER_CTL_TBEN))
+        return false;
+
+  return true;
+}
+/*---------------------------------------------------------------------------*/
 int8_t
 pwm_enable(uint32_t freq, uint8_t duty, uint8_t timer, uint8_t ab)
 {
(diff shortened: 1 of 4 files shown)


bthebaudeau:
@alignan

As further improvements to the PWM driver, I think that it would be good to simplify the API to use only the following functions:
 - `pwm_start(timer, a_b, port, pin, freq, polarity)`, which would be the equivalent of `pwm_enable(0%)` + `pwm_set_direction(polarity)` + `pwm_stop()` with the existing API,
 - `pwm_set_duty_cycle(timer, a_b, duty_cycle)`, which would dynamically adjust the duty cycle without stopping the timer, except for 0% and 100%, which would be the equivalent of `pwm_stop()` with the existing API,
 - `pwm_stop(timer, a_b)`, which would be the equivalent of `pwm_disable()` with the existing API, the used port and pin having been saved by `pwm_start()` in order not to have to pass them twice, or they could also be derived from the MUX + GPIO registers.

The rationale for this is that there should be a PWM session between `pwm_start()` and `pwm_stop()` during which the chosen pin is always generating a PWM signal between 0% and 100%. Apart from the duty cycle, all the PWM settings should be constant during a session. During this PWM session, the regular usage of a PWM requires to be able to change the duty cycle without stopping the timer, so that this operation is simple, quick and glitch-free. `pwm_set_duty_cycle()` could automatically start and stop the PWM timer when switching between 0%/100% and the other duty cycle values, in order to optimize power consumption. After `pwm_stop()`, the end of the PWM session should revert the timer and the pin to their reset states.

alignan:
:+1: for the fixes, thanks!

About the PWM API, I'm OK with the suggestions but I would keep the `pwm_toggle_direction()` as this is later useful to control servo motors.

bthebaudeau:
@alignan

> I would keep the `pwm_toggle_direction()` as this is later useful to control servo motors.

Why? With the API that I suggest, the signal polarity is set once for the PWM session with `pwm_start()`. The `PWML` bit is just a convenience to have duty cycles matching the hardware instead of having to complement them in software, so I don't see any case where it would make sense to toggle this bit right in the middle of a PWM session, since the hardware does not change. Moreover, toggling this bit would have the exact same effect on the signal as passing `100 - dc` to `pwm_set_duty_cycle()` if `dc` is the current duty cycle, so we don't need an extra function here.

Also, it would probably be more efficient to use a fractional data type rather than a percentage to express the duty cycle passed to this API, because:
 - % are very human while the duty cycle values passed to this API will probably be software-generated rather than human-generated, especially if the duty cycle has to change in real time,
 - % are inefficient for the CPU, requiring a division, while binary-oriented representations would only require a bit shift instead,
 - the accuracy would be better and could better benefit from the timer resolution if the frequency is well chosen.

bthebaudeau:
@alignan

It would also be great to have the following options:
 - hardware synchronization of timers,
 - rounding the start counter value to a power of 2, minus 1, preferably to get full 8- or 16-bit resolution for the duty cycle (using the 8-bit prescaler can help here, and the accuracy of the frequency is often unimportant for PWM signals), in order to ease the use of duty cycle values resulting from computations,
 - updating the duty cycle through µDMA.

alignan:
Thanks for the suggestions, there's still a lot of room for improvement :+1: 
I have some other stuff on the to-do list currently so I will not work on this for the moment, however I'm happy to review and double check any enhancement proposal.  Should we keep the comments here, or create an issue to keep the conversation pinned-up?

bthebaudeau:
You're right: It's better not to hijack existing PRs, so I've created #1505.

PR #1162, opened 2015-07-12
cc2538: Add support for Coffee

bthebaudeau:
This pull request adds support for Coffee. The first commits improve the CC2538 infrastructure in order to make it easier to support Coffee in the last commit, but the CC2538 support also generally benefits from these improvements.

This PR has been build-tested. The Coffee port has also been runtime-tested in another context, but not with this PR. This PR can be runtime-tested using https://github.com/contiki-os/contiki/wiki/Coffee-filesystem-example#coffeeexamplec, after having defined `COFFEE_CONF_SIZE` (e.g. `(4 * COFFEE_SECTOR_SIZE)`) and maybe (but this does not seem to be necessary for this example) some other `COFFEE_CONF_*` to something useful in `project-conf.h`.

PR #1420, opened 2015-12-13
cc2538dk: Add Coffee filesystem example

bthebaudeau:
Add the Coffee filesystem example from the Wiki as suggested by @alignan in #1162.

### DIFF
--- a/examples/cc2538dk/coffee/Makefile
+++ b/examples/cc2538dk/coffee/Makefile
@@ -1,0 +1,7 @@
+DEFINES+=PROJECT_CONF_H=\"project-conf.h\"
+CONTIKI_PROJECT = coffee-example
+
+all: $(CONTIKI_PROJECT)
+
+CONTIKI = ../../..
+include $(CONTIKI)/Makefile.include
--- a/examples/cc2538dk/coffee/Makefile.target
+++ b/examples/cc2538dk/coffee/Makefile.target
@@ -1,0 +1,1 @@
+TARGET = cc2538dk
--- a/examples/cc2538dk/coffee/coffee-example.c
+++ b/examples/cc2538dk/coffee/coffee-example.c
@@ -1,0 +1,129 @@
+/*
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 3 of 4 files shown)


alignan:
After merging #1162 I noticed there is an existing [CFS example](https://github.com/contiki-os/contiki/blob/master/examples/sky/example-coffee.c) for the `sky` platform.  I think it would be better to have a single coffee example for "most" platforms, in the same fashion we moved the `rssi-scanner` to avoid code repetition.

I also have been thinking about creating a `cc2538` example directory, and move examples like the `mqtt`, `pwm`, etc., to this location instead.  At least for users it would be less confusing having to search i.e for `remote` examples while working with the `cc2538dk`, `openmote`, etc, and generally it would improve the cc2538 support.

bthebaudeau:
Do you think that it would be worth explicitly formatting the filesystem at the beginning of this example rather than to rely on the flash erase when programming the example?

> After merging #1162 I noticed there is an existing CFS example for the `sky` platform.

Yes, and for several other platforms too.

> I think it would be better to have a single coffee example for "most" platforms, in the same fashion we moved the `rssi-scanner` to avoid code repetition.

Yes, that would be great, but:
 - How would you merge all the existing examples, some of which seem to have some specifics?
 - How would you ensure that this common example is automatically built (Travis...) and tested for all the platforms supporting it?
 - How would users know where to find this common example and that it is supported by a given platform?

Some additional Makefile settings and documentation would probably do.

> I also have been thinking about creating a `cc2538` example directory, and move examples like the `mqtt`, `pwm`, etc., to this location instead. At least for users it would be less confusing having to search i.e for `remote` examples while working with the `cc2538dk`, `openmote`, etc, and generally it would improve the cc2538 support.

I agree, but here too, we have to find a way of at least documenting and automatically building these common examples for all the supported platforms, `cc2538dk` being the default.

alignan:
I was thinking on something like [this](https://github.com/alignan/contiki/tree/pull/cfs-example-unified), what do you think?

nvt:
@alignan I agree that the examples and tests should be placed in a new directory. It's been a long time since the Sky was the only platform with support for Coffee.

bthebaudeau:
@alignan I have not tested your suggestion, but it looks great.

Besides my comments on the diff, my `project-conf.h` is missing for the `cc2538` targets. Without it, the examples would not run properly on these platforms. Some conditional mechanism has to be created here, as discussed in #1447, e.g. a `TARGET_<TARGET_NAME>` or `SOC_<SOC_NAME>` C preprocessor definition that could be automatically generated by `Makefile.include` from the make variables.

bthebaudeau:
Some function pointers suggested by @g-oikonomou in #1447 could not fix this issue, but they could help if something else needs to be adjusted for a given platform.

g-oikonomou:
Makefile.include generates this: `-DCONTIKI_TARGET_$(TARGET_UPPERCASE)=1`. Are you referring to something different?

bthebaudeau:
This definition can do the trick. I did not know it, but this is exactly the kind of definition that I was suggesting. But here, we would have to:
```
#if CONTIKI_TARGET_CC2538DK || CONTIKI_TARGET_ZOUL
#define ...
#endif
```
while a CPU or SoC definition would simplify things to:
```
#if CONTIKI_CPU_CC2538
#define ...
#endif
```

But `CONTIKI_TARGET_<TARGET_NAME>` is just fine enough for now.

Devesh24:
Hello Guys,
                       After flashing coffee example there is nothing coming on UART, any idea ?

alignan:
Should we close this in favour of #1490 ?

bthebaudeau:
Superseded by #1490.

PR #1422, opened 2015-12-16
packetbuf cleanup

kkrentz:
This pull request removes unnecessary `packetbuf` functions.

### DIFF
--- a/core/net/packetbuf.c
+++ b/core/net/packetbuf.c
@@ -47,6 +47,7 @@
 #include "contiki-net.h"
 #include "net/packetbuf.h"
 #include "net/rime/rime.h"
+#include "sys/cc.h"
 
 struct packetbuf_attr packetbuf_attrs[PACKETBUF_NUM_ATTRS];
 struct packetbuf_addr packetbuf_addrs[PACKETBUF_NUM_ADDRS];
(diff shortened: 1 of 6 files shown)


simonduq:
Removing `packetbuf_compact`while `packetbuf_hdrreduce` is still in use seems wrong to me. Am I missing something?

kkrentz:
Maybe it is wrong. Anyway, what is strange is that `packetbuf_compact` is called very inconsistently. For example, `packetbuf_compact` is called by `cxmac`, but not by `nullrdc`. Apparently, the most appropriate place to call `packetbuf_compact` is then a packet changes its direction because, at this point, the hdrreduce calls should be done and hdralloc calls should begin. The reason why everything works even without `packetbuf_compact` seems to be that `packetbuf_copyto` does the job of `packetbuf_compact` as a side effect, which is also strange. Perhaps you have a better idea to clean up these things.

simonduq:
I would also like a cleanup but am equally confused with this feature of packetbuf...

kkrentz:
I restored `packetbuf_compact` - now these changes should do no harm

simonduq:
:+1:

simonduq:
merging after timeout

PR #1424, opened 2015-12-17
RE-Mote power management support

alignan:
This PR provides support for the Power management block on board the RE-Mote (revision A), composed of a nano-watt gating timer, a low-power external MCU and the battery manager.

The external MCU governs the power management block, allowing to enable/disable the battery, nano timer, configure events from the RTCC, etc.  When enabling the nano timer, it waits for a period (namely the` shutdown period`, given by an external resistor, default is 1 minute) before entering to the `shutdown mode`.

The `shutdown mode` disables the external battery and powers off the RE-Mote (and all connected sensors and peripherals), but the power management block, effectively reducing its power consumption down to 150nA.  After a new `shutdown period`, the RE-Mote is powered back.

The application at any given time can enter the `shutdown mode` by invoking a macro.

Alternatively it can be used with the on-board RTCC (see #1416), to awake the system from the shutdown mode.

Below are some current consumption measurements in each state using the `test-power-mgmt` application included in this PR.

re-mote power management overall

re-mote power management active mode

re-mote power management shutdown mode

### DIFF
--- a/examples/zolertia/zoul/Makefile
+++ b/examples/zolertia/zoul/Makefile
@@ -1,5 +1,5 @@
 DEFINES+=PROJECT_CONF_H=\"project-conf.h\"
-CONTIKI_PROJECT = zoul-demo test-tsl2563 test-sht25 test-pwm
+CONTIKI_PROJECT = zoul-demo test-tsl2563 test-sht25 test-pwm test-power-mgmt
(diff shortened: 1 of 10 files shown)


alignan:
The 170nA consumption is due to the RTCC module not configured for low power operation, as it is not merged yet, but anyways, only 20nA :smile:

alignan:
Merging as it is platform-specific

PR #1425, opened 2015-12-17
random LLSEC fixes

kkrentz:
* READMEs describing `llsec` and `noncoresec`
* It is no longer necessary to use a static network-wide security level in LLSEC drivers
* Support for using a custom FRAMER in `noncoresec`
* If LLSEC is off, `packetbuf.h` currently defines "pseudo-attritubtes". This can cause hard-to-find bugs if LLSEC is configured improperly. Therefore, I removed this "feature".

Apart from that, I also overcame various limitations of `noncoresec`, but that is going to be part of another pull request - see https://github.com/kkrentz/contiki/tree/master/core/net/llsec/adaptivesec


core/net/mac/tsch/tsch-security.h
```diff
@@ -39,22 +39,36 @@
...
-#endif /* TSCH_SECURITY_ENABLED */
-#if TSCH_SECURITY_ENABLED && LLSEC802154_CONF_USES_FRAME_COUNTER
-#error TSCH_SECURITY_ENABLED set but LLSEC802154_CONF_USES_FRAME_COUNTER set
-#endif /* TSCH_SECURITY_ENABLED */
+
+#ifdef TSCH_SECURITY_CONF_LEVEL
+#define TSCH_SECURITY_LEVEL TSCH_SECURITY_CONF_LEVEL
+#else /* TSCH_SECURITY_CONF_LEVEL */
+#define TSCH_SECURITY_LEVEL FRAME802154_SECURITY_LEVEL_NONE
+#endif /* TSCH_SECURITY_CONF_LEVEL */
```
simonduq: I don't think we need such a flag at all: TSCH security levels (and keys and key indexes) are defined in https://github.com/contiki-os/contiki/blob/master/core/net/mac/tsch/tsch-security.h

We used `LLSEC802154_CONF_SECURITY_LEVEL` so far only in a binary way (sec enabled or not), I guess `LLSEC802154_ENABLED` is now perfectly suited for the task.

kkrentz: TSCH_SECURITY_LEVEL is gone

simonduq:
Looks great to me! :+1: (except I still don't understand why link-layer security should sit above MAC, see https://github.com/contiki-os/contiki/pull/1383)

kkrentz:
I have added a fix for Issue #1537

simonduq:
Let me add a timeout and schedule for merging.

simonduq:
merging after timeout

tonnenpinguin:
Sorry to bring this up after merging, but I'm getting a build error caused by `LLSEC802154_CONF_ENABLED` not being defined.

Shouldn't [this line](https://github.com/contiki-os/contiki/blob/master/core/net/mac/rdc.h#L54) check for `LLSEC802154_ENABLED` rather than `LLSEC802154_CONF_ENABLED`?

kkrentz:
Somehow, my compiler does not throw any error, while it should. Can you check whether #1625 solves your problem.

tonnenpinguin:
For me that does fix the issue, yes! (Also, IMHO it does make sense to stick with common practice)

PR #1427, opened 2015-12-19
fixed make login for Zoul on MAC

joakimeriksson:
Added a MAC OS X serial-dump binary and fixed so that the Zoul platform(s) now support make login on OS X.


alignan:
:+1:

PR #1429, opened 2015-12-20
Allow for using a fixed frame header length in sicslowpan.c

wbober:
When SICSLOWPAN_FRAMER_HDRLEN is defined its value is used as a frame header length. This allows for using sicslowpan without calling a framer. This is usefull if framer is not used on a given platform or when header length is always the same. 

This PR doesn't change the default behaviour.

In addition an unused variable warning, which shows up when fragmentation isn't used, is fixed.

### DIFF
--- a/core/net/ipv6/sicslowpan.c
+++ b/core/net/ipv6/sicslowpan.c
@@ -169,6 +169,14 @@ void uip_log(char *msg);
 #define COMPRESSION_THRESHOLD 0
 #endif
 
+/** \brief Fixed size of a frame header. This value is
+ * used in case framer returns an error or if SICSLOWPAN_USE_FIXED_HDRLEN
+ * is defined.
+ */
+#ifndef SICSLOWPAN_FIXED_HDRLEN
+#define SICSLOWPAN_FIXED_HDRLEN 21
+#endif
+
 /** \name General variables
  *  @{
  */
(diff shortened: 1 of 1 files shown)


joakimeriksson:
Looks like an improvement! But maybe we should also add SICSLOWPAN_USE_FIX_HDRLEN and have 21 as s default value for the SICSLOWPAN_HDRLEN so that it can be used in the framer failure case also. Then it would be very obvious that we intend to use FIX header length when the SICSLOWPAN_USE_FIX_HDRLEN is set and that it otherwise make use of the framer.

wbober:
I've added SICSLOWPAN_USE_FIXED_HDRLEN which controls if framer should be used and SICSLOWPAN_FIXED_HDRLEN which defines the length. Please take a look.

joakimeriksson:
Yes, now I think it is easier to understand what to configure and also nicer with no hardcoded HDRLEN = 21. For me it looks good to merge now!

simonduq:
this looks good but seems unneeded after the latest sicslowpan changes. Can you please confirm @wbober ?

wbober:
@simonduq I've checked this PR against latest master and it looks to me that it's still needed. Could you please tell me why you think otherwise? Maybe I've missed something.

simonduq:
I don't recall the pre-rebase diff, and maybe I read it wrong, but the PR as it is now does seem useful to me yes! :+1: if Travis is green

simonduq:
merging after timeout

PR #1432, opened 2015-12-21
ARM Cortex-M3 specific implementation of contiki's mt-architecture

drandreas:
This pull request adds platform specific functions for the CC2538 CPU. I've verified the compatibility and stability using ```examples/multi-threading```. The pull request features all functions as specified in https://github.com/contiki-os/contiki/wiki/Multithreading#architecture-support. I assume the code is compatibly with any ARM Cortex-M3 based platform.

Note: Currently mt.c does not expose  ```current``` for this reason mtarch.c holds a second variable with similar meaning. Both variables ought to be in sync in all cases.

I could also provide code for preemptive-multi-threading. However this requires repurposing the SysTick handler ```clock_isr()``` as far as I understand the CPU specification.


cpu/cc2538/mtarch.c
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * The schedule switches back and forth between
+ * main stack and thread stack(s). NULL pointers in
+ * next_thread and current_thread are used to indicate
+ * main stack.
+ *
+ * \param sp current stack pointer
+ * \return next stack pointer
+ */
+void*
```
alignan: void *

alignan:
Tested with the `RE-Mote` and works as expected, :+1: once the conflicts are resolved

drandreas:
Should I add a patch for ```mt.c```?

drandreas:
Ah I'didn't notice the actual conflicts I'll rebase my soon.

drandreas:
@alignan I resolved the merge conflict

alignan:
Any comment @g-oikonomou @bthebaudeau?

drandreas:
should I rename ```current```in `mt.h/.c```to ```mt_current```and define it as extern?

bthebaudeau:
I will have some comments, but I'd like to finish reviewing this first.

cpu/cc2538/dev/scb.h
```diff
@@ -74,6 +74,18 @@
 #define SCB_SYSCTRL_SLEEPEXIT  0x00000002  /**< Sleep on ISR exit */
 /** @} */
 /*---------------------------------------------------------------------------*/
+/** \name SCB_INTCTRL / NVIC_INT_CTRL register bits
```
bthebaudeau: This is a register of the SCB, so remove the NVIC part.

cpu/cc2538/dev/scb.h
```diff
@@ -74,6 +74,18 @@
...
+ * @{
+ */
+#define SCB_INTCTRL_NMI_SET   0x80000000   /**< NMI Set Pending */
+#define SCB_INTCTRL_PEND_SV   0x10000000   /**< PendSV Set Pending */
+#define SCB_INTCTRL_UNPEND_SV 0x08000000   /**< PendSV Clear Pending */
+#define SCB_INTCTRL_PENDSTSET 0x04000000   /**< SysTick Set Pending */
+#define SCB_INTCTRL_PENDSTCLR 0x02000000   /**< SysTick Clear Pending */
+#define SCB_INTCTRL_ISR_PRE   0x00800000   /**< Debug Interrupt Handling */
+#define SCB_INTCTRL_ISR_PEND  0x00400000   /**< Interrupt Pending */
```
bthebaudeau: Use the bit names from the CC2538 User's Guide: no underscores, so it should be `SCB_INTCTRL_NMISET`, not `SCB_INTCTRL_NMI_SET`, etc.

cpu/cc2538/dev/scb.h
```diff
@@ -74,6 +74,18 @@
...
+/** @} */
+/*---------------------------------------------------------------------------*/
```
bthebaudeau: Move this block before the block for `VTABLE` so that these blocks are sorted according to the register addresses.

cpu/cc2538/mtarch.c
```diff
@@ -0,0 +1,164 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
```
bthebaudeau: Please use the [current version of the BSD 3-clause license](https://opensource.org/licenses/BSD-3-Clause), i.e. use `copyright holder` everywhere, and not customized `ORGANIZATION` and `REGENTS AND CONTRIBUTORS` from the original version.

cpu/cc2538/mtarch.c
```diff
@@ -0,0 +1,164 @@
...
+#include "mtarch.h"
+#include "reg.h"
+#include "scb.h"
```
bthebaudeau: It should be `"dev/scb.h"`.

cpu/cc2538/mtarch.c
```diff
@@ -0,0 +1,164 @@
...
+#include "mt.h"
```
bthebaudeau: It should be `"sys/mt.h"`.

cpu/cc2538/mtarch.h
```diff
@@ -1,48 +1,72 @@
...
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
```
bthebaudeau: Please use the [current version of the BSD 3-clause license](https://opensource.org/licenses/BSD-3-Clause), i.e. use `copyright holder` everywhere, and not customized `ORGANIZATION` and `REGENTS AND CONTRIBUTORS` from the original version.

cpu/cc2538/mtarch.h
```diff
@@ -1,48 +1,72 @@
...
  */
 #ifndef MTARCH_H_
 #define MTARCH_H_
 
+#include "contiki.h"
```
bthebaudeau: Any reason not to simply use `"contiki-conf.h"`?

cpu/cc2538/mtarch.h
```diff
@@ -1,48 +1,72 @@
...
+#include "stdint.h"
```
bthebaudeau: It should be `<stdint.h>`, but you don't even use it.

cpu/cc2538/dev/scb.h
```diff
@@ -74,6 +74,18 @@
...
 
 #endif /* SCB_H_ */
```
bthebaudeau: These definitions are not CC2538-specific, so this file should be moved to `cpu/arm/cortex-m3/` in a separate commit, preferably before this one, and it should be changed to use the official ARM naming from http://infocenter.arm.com/help/topic/com.arm.doc.dui0552a/CIHFDJCA.html.

cpu/cc2538/mtarch.c
```diff
@@ -0,0 +1,164 @@
...
+                "bl schedule\n"                    /* Call scheduler */
+                "mov lr, r4\n");                   /* restore lr from callee saved register */
+
+  /* Restore next thread context */
+  asm volatile ("ldmfd r0!, {r4-r11} \n"           /* Pop registers from stack */
+                "msr msp, r0         \n"           /* Update stack pointer */
+                "cpsie i");                        /* Re-Enable interrupts */
+
+  asm volatile ("bx lr\n");                        /* Restore state and return  */
+}
```
bthebaudeau: As you mentioned, this implementation applies to all Cortex-M3 platforms, so it should be moved to `cpu/arm/cortex-m3/`.

cpu/cc2538/mtarch.h
```diff
@@ -1,48 +1,72 @@
...
+  void *psr;  /**< Reset value: 0x01000000 */
+} mtarch_stack_t;
+
 struct mtarch_thread {
-  unsigned char *sp;
+  void *stack[MTARCH_STACKSIZE];
+  void *sp;
 };
 
 #endif /* MTARCH_H_ */
```
bthebaudeau: As you mentioned, this implementation applies to all Cortex-M3 platforms, so this file should be moved to `cpu/arm/cortex-m3/`.

cpu/cc2538/dev/scb.h
bthebaudeau: But after moving to `cpu/arm/cortex-m3/`, this should even become `SCB_ICSR_NMIPENDSET`, etc.

cpu/cc2538/mtarch.h
```diff
@@ -1,48 +1,72 @@
...
+  void *r10;
+  void *r11;
+  void *r0;   /**< Ptr to argument (*data) */
+  void *r1;
+  void *r2;
+  void *r3;
+  void *r12;
+  void *lr;   /**< Ptr to cleanup function */
+  void *pc;   /**< Ptr to function (*function) */
+  void *psr;  /**< Reset value: 0x01000000 */
```
bthebaudeau: `uint32_t` would be more appropriate than `void *` for all these registers, except maybe for `lr` and `pc` (and `sp` below), but even for them, it's better to view them as 32-bit registers that may contain anything, not necessarily pointers.

cpu/cc2538/mtarch.h
```diff
@@ -1,48 +1,72 @@
...
+#include "contiki.h"
+#include "stdint.h"
+
+#ifdef MTARCH_CONF_STACKSIZE
+#define MTARCH_STACKSIZE MTARCH_CONF_STACKSIZE
+#else
+#define MTARCH_STACKSIZE 256
+#endif
+
+#define MTARCH_REGISTER_COUNT 16
```
bthebaudeau: `(sizeof(mtarch_stack_t) / sizeof(uint32_t))`

cpu/cc2538/mtarch.h
```diff
@@ -1,48 +1,72 @@
...
+  void *r3;
+  void *r12;
+  void *lr;   /**< Ptr to cleanup function */
+  void *pc;   /**< Ptr to function (*function) */
+  void *psr;  /**< Reset value: 0x01000000 */
+} mtarch_stack_t;
+
 struct mtarch_thread {
-  unsigned char *sp;
+  void *stack[MTARCH_STACKSIZE];
```
bthebaudeau: According to the [Procedure Call Standard for the ARM Architecture ABI r2.10](http://infocenter.arm.com/help/topic/com.arm.doc.ihi0042f/IHI0042F_aapcs.pdf), the stack must be double-word aligned at a public interface, so the whole thread stack must be too. Therefore, you have to use `uint64_t`, or `uint32_t` with `__attribute__((__aligned__(8)))` in order to use `uint32_t` like for registers.

cpu/cc2538/mtarch.h
```diff
@@ -1,48 +1,72 @@
...
+  void *sp;
```
bthebaudeau: `uint32_t`

cpu/cc2538/mtarch.c
```diff
@@ -0,0 +1,164 @@
...
+
+void
+mtarch_pstop(void)
+{
+}
+
+void
+mtarch_stop(struct mtarch_thread *thread)
+{
+}
```
bthebaudeau: You should also define `mtarch_pstart()` and `mtarch_remove()`.

cpu/cc2538/mtarch.c
```diff
@@ -0,0 +1,164 @@
...
+  if(current_thread) {
+    sp = current_thread->sp;
+  } else {
+    sp = msp;
+  }
+  return sp;
+}
+
+/* Disable prologue and epilogue of ISR */
+void pend_sv_isr(void)  __attribute__ ((isr, naked));
```
bthebaudeau: - `isr` contradicts `naked`, so only use `naked`.
 -  There is no need for a separate declaration, so merge this into the definition.

cpu/cc2538/mtarch.c
```diff
@@ -0,0 +1,164 @@
...
+  asm volatile ("mov r4, lr\n"                     /* backup lr into callee saved register */
+                "bl schedule\n"                    /* Call scheduler */
+                "mov lr, r4\n");                   /* restore lr from callee saved register */
+
+  /* Restore next thread context */
+  asm volatile ("ldmfd r0!, {r4-r11} \n"           /* Pop registers from stack */
+                "msr msp, r0         \n"           /* Update stack pointer */
+                "cpsie i");                        /* Re-Enable interrupts */
+
+  asm volatile ("bx lr\n");                        /* Restore state and return  */
```
bthebaudeau: All this could be put into a single `asm volatile`.

cpu/cc2538/mtarch.c
```diff
@@ -0,0 +1,164 @@
...
+mtarch_start(struct mtarch_thread *thread,
+             void (*function)(void *data),
+             void *data)
+{
+  /* Calculate SP */
+  mtarch_stack_t *sp = (mtarch_stack_t *)&thread->stack[MTARCH_STACKSIZE - MTARCH_REGISTER_COUNT];
+  thread->sp = sp;
+
+  /* Zero stack */
+  memset(thread->stack, '\0', sizeof(void *) * MTARCH_STACKSIZE);
```
bthebaudeau: `0, sizeof(thread->stack)`

cpu/cc2538/mtarch.h
bthebaudeau: `CCR.STKALIGN` should also be set because of that.

cpu/cc2538/mtarch.c
```diff
@@ -0,0 +1,164 @@
...
+  sp->lr = mt_exit;
+  sp->pc = function;
+  sp->psr = (void *)0x01000000;
+}
+
+void
+mtarch_yield(void)
+{
+  /* Trigger PendSV the handler will switch to main */
+  REG(SCB_INTCTRL) = SCB_INTCTRL_PEND_SV;
```
bthebaudeau: You should rather use `svc` here. The `SVCall` handler would always switch to the main thread (kernel + protothreads) using `msp`, and the `PendSV` handler would always switch to the process thread using `psp`. This would also make it possible to use the `mt` stuff in unprivileged mode.

In that way, you could remove most of your state variables, which would also remove the duplication of the current thread pointer with `mt.c`.

Using the banked versions of `sp` also gets rid of your `msp` variable.

cpu/cc2538/mtarch.c
```diff
@@ -0,0 +1,164 @@
...
+void
+pend_sv_isr(void)
+{
+  /* Backup current thread context */
+  asm volatile ("cpsid i             \n"           /* Disable interrupts */
+                "dsb                 \n"           /* Data synchronization barrier */
+                "isb                 \n"           /* Instruction synchronization barrier */
+                "mrs r0, msp         \n"           /* Read SP to r0 */
+                "stmdb r0!, {r4-r11} \n"           /* Push registers onto stack */
+                "msr msp, r0");                    /* Update stack pointer */
```
bthebaudeau: You're in an exception handler, so `msp` and `sp` are the same, so the 3 instructions above are equivalent to a simple `push {r4-r11}` followed by `mov r0, sp`.

drandreas:
Thanks for the in-depth feedback. 
I'll address them during the next week with dedicated commits. I'll squash and rebase at the end.
I contacted legal about "changing" the license. I hope to hear back soon.

bthebaudeau:
Hi @drandreas. Since my review, I have prepared locally a series that goes even further than the suggestions made here. It's almost ready. I will probably submit it next week. Maybe you want to discuss it before spending time reworking yours. My series:
 - updates the CMSIS code,
 - switches the whole CC2538 code base to CMSIS for the Cortex-M3,
 - adds a common mtarch implementation for all the Cortex-M platforms, using SVCall for exec and yield + SVPend for system preemption from exception handlers, without using any variable.

drandreas:
@bthebaudeau you really don't like variables :smirk:. I'm curious how you pass 
``mtarch_thread *thread`` from ``mtarch_start`` and ``mtarch_exec`` to your interrupt service routines. I'm fine with dropping my approach in favor of yours. However, I'm little concerned about running mt-threads in unprivileged mode. 

When I initially wrote this code I thought about having mt-threads on the process stack. I came to the conclusion that running mt-threads unprivileged would break some of the library calls. After all, most libraries are implemented without privilege separation in mind.

My main concern is the restricted access to NVIC from unprivileged mode. NVIC is often accessed in the cryptoprocessor drivers (AES, ECC, SHA...). Note, moving TinyDTLS to an alternate stack for running all crypto algorithms in background, was my initial intention.

drandreas:
Nevermind, a privileged thread can also run on the process stack.

bthebaudeau:
@drandreas #1816.

drandreas:
closing pull request in favor of #1816

PR #1196, opened 2015-08-17
New platform: Intel Galileo Board

aguedes:
Hi all,

This PR is a follow-up from the Galileo port effort announced a couple of weeks ago in Contiki's mailing list (for further information check this [link](http://sourceforge.net/p/contiki/mailman/message/34350483/)).

So this PR adds support for both Intel Quark X1000 SoC (cpu layer) and the Intel Galileo Board (platform layer). In order to assist the review process, below follows some comments that summarize the patches within this PR.

* Cpu layer patches are prefixed with "x86:" and implement the following features:
  * IA initialization (GDT, IDT and IRQs)
  * Programmable Interrupt Controller (PIC) driver
  * Real-time Clock (RTC) driver
  * Programmable Interval Timer (PIT) driver
  * UART driver
  * PCI driver
* Platform layer patches are prefixed with "galileo:" and implement the following features:
  * Clock module
  * Etimer, Ctimer and Rtimer libraries
  * Stdio library (stdout and stderr)
* The patch "examples: Add all-timers example" introduces a Contiki sample application to validate the timers implementation. This sample is not Galileo-specific and it can be used by any port.
* BSP-related files are located in platform/galileo/bsp/ directory. We provide shell scripts that downloads and build the BSP components.
* platform/galileo/README.md provides detailed instructions on how to build, run and debug application for the platform.

This PR doesn't provide Travis integration. Apart from that, it addresses all others requirements described in the [New Platforms section](https://github.com/contiki-os/contiki/wiki/Code-Contributions#new-platforms) from the contribution guide.

Looking forward to get your feedback!

Best regards,

Andre

PR #1433, opened 2015-12-23
X86 memory management

mdlemay:
Depends on #1481.

This extends the Intel Galileo platform port (#1196) with support for memory management based on paging, hardware-switched segmentation, or software-switched segmentation, plus Isolated Memory Regions (IMRs).  Memory management can help to detect invalid memory accesses with causes such as stack overflows, misconfigured DMA-capable devices, or software bugs that result in stray reads and writes.

The basic principle is to divide the software into multiple "protection domains," each of which is allocated specific memory permissions.  See cpu/x86/mm/README.md for more details.

Memory management support is disabled by default and can be enabled at build time using command-line flags.

As noted in cpu/x86/mm/README.md, a limitation of this particular patchset is that interrupts are blocked while any protection domain other than the default one is active.  This may delay interrupt delivery.  However, this is only the case when memory management support is enabled at build time.


jeez:
@mdlemay can you please confirm in which order I should review these commits, please?
Github changes their order to chronological, so I want to make sure I follow your original flow on it.

mdlemay:
@jeez: The chronological order is the same as the original flow in this case.

mdlemay:
I just pushed a new version with two improvements compared to the previous patchset:
 - Makes the segment_desc_t structure less convoluted.
 - Refactors the IMR initialization procedure to be in a separate file than the basic IMR driver.
 - Refactors the IMR print routines to be in an example program.

mdlemay:
Thank you for merging #1444, @nvt.  I rebased this PR, so it should now be ready to be merged as well.

nvt:
@mdlemay Thanks for submitting this major feature, but why is it disabled by default?

mdlemay:
Thank you for reviewing this, @nvt. Enabling it increases memory overhead, increases the time required to switch into drivers that are assigned separate protection domains, and can increase interrupt latency due to interrupts only being enabled in the default application protection domain.  These patches make this an opt-in feature set so that developers can evaluate these tradeoffs before enabling it.

I should also note that these patches can be a foundation for future extensions.  For example, I submitted a set of patches to the LLVM and Clang compiler projects to harden SafeStack (http://clang.llvm.org/docs/SafeStack.html) with segmentation.  I'm planning to submit patches to enable that on top of these Contiki patches if and when the compiler patches are merged.

alignan:
Ready to merge?

mdlemay:
Yes, it's ready to merge.  Thanks!

alignan:
:+1:

PR #1435, opened 2015-12-26
cc2538: Add support for hardware-accelerated llsec

bthebaudeau:
This pull request makes the CC2538 ready for hardware-accelerated llsec:
 - The CC2538 AES/CCM drivers have been reworked in order to get a core AES engine driver shared by CCM and all other future modes instead of duplicating most of the code each time.
 - Support has been added for AES-ECB on CC2538.
 - Deleted `aes_128_padded_encrypt()` as suggested in #618.
 - A fix for an error when including `aes-128.h`.
 - Added `aes_128_driver` and `ccm_star_driver` for the CC2538 SoC.

This PR is supposed to supersede #1317, and it is partially based on @kkrentz's commits in this PR.

It has been tested with the following examples on Zolertia Firefly (Zoul):
 - `examples/cc2538dk/crypto/ccm-test.c`,
 - `examples/cc2538dk/crypto/ecb-test.c`,
 - `examples/llsec/ccm-star-tests/encryption/tests.c`,
 - `examples/llsec/ccm-star-tests/verification/tests.c`.

It has not yet been tested with llsec enabled.


bthebaudeau:
@kkrentz This PR is what I actually meant in https://github.com/contiki-os/contiki/pull/1317#issuecomment-152871840. This would have been complicated to detail, and this would have required a lot of work for you, beyond the original purpose of your PR, so I did it here.

Are you fine with this new PR? You seem to be leading on llsec, so can you test it in this context? Thanks in advance.

kkrentz:
:+1: ```cc2538_ccm_star_driver ``` and ```cc2538_aes_128_driver``` work on OpenMotes.

Error handling is left as a follow-up ...

BTW do you know what issues arise when switching to PM2? I noticed that ```crypto_init``` sets ```lpm_register_peripheral(permit_pm1)```.

bthebaudeau:
@kkrentz

> Error handling is left as a follow-up ...

Yes. It would require a change of API in `struct aes_128_driver` and `struct ccm_star_driver`, so that the functions can at least return a status code, because the drivers can't do much here apart from `#define DEBUG 1`.

> BTW do you know what issues arise when switching to PM2? I noticed that ```crypto_init``` sets ```lpm_register_peripheral(permit_pm1)```.

Yes: in the PM1+ modes, the clock of the cryptoprocessor IP is disabled (see the section 7.3 in the CC2538 User's Guide), so these modes would mess with any on-going crypto operations, which can occur only if the calling protothread yields.

In PM2, besides the cryptoprocessor IP clock being disabled, there is also no state retention for this IP (§7.6), except for the key store RAM and registers (§8.9). That's why the `AES_CTRL_INT_CFG` register and some others are set by the functions initiating an operation in `aes.c`.

PR #1438, opened 2015-12-28
jsonparse: multiple improvements

jkent:
* input string now can be any json value type
* syntax handling completed
* stack is now used more efficiently for objects
* implemented atomic values true, false, and null
* jsonparse_copy_value now handles escapes

NOTE: jsonparse_next no longer returns ':' since this information is redundant


simonduq:
That one is a bit trickier from just reading the code. How much is this tested?
@joakimeriksson I believe you wrote the original code, would be great if you could take a look!

jkent:
I'm using this code in my [Wireless-Sol](https://github.com/jkent/wireless-sol) project.  I've got some test code I used here in a [gist](https://gist.github.com/jkent/6431d067b0e862d5aecf) as well.  Hope that helps!

joakimeriksson:
I will take a look at this next week and get back with some feedback. But from a quick look I think it looks promising! Nice to see users of the library!

joakimeriksson:
I think this looks good - a few minor code style issues (there is a space after if and while in some cases). So :+1:  from me.

simonduq:
:+1:

PR #1423, opened 2015-12-17
Add examples for NXP JN516x using TSCH

TeVeDe:
Examples for the JN516x using TSCH have been transferred to the main line.
Make file for the regression tests is updated as well.

PR #1441, opened 2016-01-04
Contact

TeVeDe:


cmorty:
This should probably go into two PRs. The contacts-thing is a no-brainer, while the examples might need a review. Having a very quick glance, it seems like there might be some redundancy that can be removed from the examples. But I might well be wrong.

simonduq:
Yes, the first commit is from https://github.com/contiki-os/contiki/pull/1423, only the second one should be in this PR.

TeVeDe:
I will close this PR and make a separate one with the NXP contacts as suggested

cmorty:
There's documentation on the wiki on how to update an PR.

PR #1443, opened 2016-01-06
cc2538: Complete support of hardware AES modes of operation

bthebaudeau:
This pull request adds drivers for the CC2538 hardware AES modes of operation that were not supported yet, i.e. CBC, CTR, CBC-MAC, and GCM.

It has been tested on Zolertia Firefly (Zoul). The test vectors are some of the NIST's, except for CBC-MAC (the NIST provides nothing for it, but instead for CMAC, because CBC-MAC is not secure for variable-length messages as stated in a doxygen warning in my `cbc-mac.h`), which uses vectors that I have derived from the last ciphertext blocks output by the CBC test customized with only encryption and 0 as IV.

I have noticed surprisingly long (i.e. a few ms) running times for messages of 160 bytes with some modes. I have not investigated this because it works fine, and maybe it's not directly linked to the crypto operations. The user's guide and the errata sheet say nothing about such a behavior.


bthebaudeau:
Self-contained platform-specific changes; all tests passed: merging already.

bthebaudeau:
For the record, the longer-than-expected execution times for some of the 160-byte messages are only caused by operations performed while the test protothread is yielding, so unrelated to the cryptoprocessor. If the test is modified to yield only between test vectors, and not in the middle of tests while waiting for results, then the execution times are always normal.

PR #1446, opened 2016-01-07
Enhanced TSL2563 driver for the Zoul module (RE-Mote/Firefly)

alignan:
This PR fixes and enhances the current `tsl2563` driver implementation with the following:

* Interruption when a value is over/below a given threshold
* Gain and timing selectable (16X, 13.7ms up to 402ms)
* Lux calculation checked for different configurations
* Partnumber and version check

This has been tested on the `firefly` and the `RE-Mote`, for the later (revision A) as the `PD1` pin is shared with the `power-mgmt` communication pin (with a pull-down) the interruption feature is broken, a quick workaround is to use a different pin (change in `board.h`).

### DIFF
--- a/examples/zolertia/zoul/project-conf.h
+++ b/examples/zolertia/zoul/project-conf.h
@@ -39,6 +39,7 @@
 #define PROJECT_CONF_H_
 
 #define BROADCAST_CHANNEL     129
+#define NETSTACK_CONF_RDC     nullrdc_driver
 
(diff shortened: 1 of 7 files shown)


g-oikonomou:
You are switching to NullRDC for all examples. Intentional?

alignan:
Yes, for the `test-tsl2563` example I had some UART garbage while printing

PR #556, opened 2014-02-03
Added the OpenMote platform and examples

peretuset:
I have copied the CC2538DK platform and examples and adapted it to the OpenMote platform, which is also based on the CC2538 SoC. The code compiles fine and I am able to upload it and run it on the target platform. Comments in the source code will need to be ammended to match the target platform, but this can be done in the future.

PR #1447, opened 2016-01-08
Added sensniff support for the CC1200

alignan:
This PR allows the CC1200 to run the [Sensniff](https://github.com/g-oikonomou/sensniff) Sniffer application.  This has been tested with the `RE-Mote` platform extensively.

As it is heavily based on the `cc2538dk` sniffer application (well, copy/pasted) I'm open for suggestions to have a single sniffer application, instead of instances per platform (there's was also a previous PR with support for the Z1 mote).

### DIFF
--- a/dev/cc1200/cc1200.c
+++ b/dev/cc1200/cc1200.c
@@ -374,6 +374,17 @@ extern const cc1200_rf_cfg_t CC1200_RF_CFG;
   } while(0)
 #endif
 /*---------------------------------------------------------------------------*/
+/* Sniffer configuration */
+#if CC1200_SNIFFER
+static const uint8_t magic[] = { 0x53, 0x6E, 0x69, 0x66 };
+#include "dev/uart.h"
+#define write_byte(b) uart_write_byte(CC1200_RF_CONF_SNIFFER_UART, b)
+#define flush()
+#else /* CC1200_SNIFFER */
+#define write_byte(b)
+#define flush()
+#endif /* CC1200_SNIFFER */
+/*---------------------------------------------------------------------------*/
 /* Variables */
 /*---------------------------------------------------------------------------*/
 /* Flag indicating whether non-interrupt routines are using SPI */
@@ -870,6 +881,10 @@ read(void *buf, unsigned short buf_len)
 
   int len = 0;
 
+  #if CC1200_SNIFFER
+    uint8_t i;
+  #endif
+
   if(rx_pkt_len > 0) {
 
     int8_t rssi = rx_pkt[rx_pkt_len - 2];
@@ -896,8 +911,22 @@ read(void *buf, unsigned short buf_len)
       packetbuf_set_attr(PACKETBUF_ATTR_LINK_QUALITY,
                          crc_lqi & ~(1 << 7));
 
-      RIMESTATS_ADD(llrx);
 
+      #if CC1200_SNIFFER
+        write_byte(magic[0]);
+        write_byte(magic[1]);
+        write_byte(magic[2]);
+        write_byte(magic[3]);
+        write_byte(len + 2);
+        for(i = 0; i < len; ++i) {
+          write_byte(((unsigned char *)(buf))[i]);
+        }
+        write_byte(rssi);
+        write_byte(crc_lqi);
+        flush();
+      #endif
+
+      RIMESTATS_ADD(llrx);
     }
 
   }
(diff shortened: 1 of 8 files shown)


bthebaudeau:
So one more instance of duplicate example. There are already the `cc2538dk` vs. `zoul` cases, and also my pending PR with a Coffee example for `cc2538dk` vs. the existing one for `sky`. See #556 too.

For the examples that could theoretically run on any platform for any SoC (e.g. sniffer, Coffee), we could create an `examples/common` folder. For the examples that are common to several platforms, but tied to a specific SoC (e.g. the CC2538 crypto and PKA examples), we could create an `examples/<soc>-common` folder. In both cases, the defined `TARGET` could be used in `Makefile` and in `project-conf.h` for platform/SoC-specific things required for the example to work (I have not checked whether or not the `TARGET` make variable is currently exported as a C preprocessor definition, but this could be easily added if not). The `README` of the example and/or of the platform should say somewhere that this example is supported for a given platform. As to Travis, in order to make sure that all shared examples are tested (at least built) by it for all supported platforms, we can use `/<TARGET>` in the make files under `regression-tests`, and we can also benefit from #1327 to customize things like `BOARD=`.

alignan:
I agree completely with the approach, `examples/cc2538-common` is a good naming and having a proper `README` file with the platforms for which the examples has been tested is nice.

In the case of the `sniffer` application is more tricky, as other platforms (like the ones based on the `cc2420` like the `Z1` and `sky`) can also be supported.  I will try to make a first stab at this and refactor the PR.

bthebaudeau:
> In the case of the `sniffer` application is more tricky, as other platforms (like the ones based on the `cc2420` like the `Z1` and `sky`) can also be supported.  I will try to make a first stab at this and refactor the PR.

If the example is really the same, then they should just go under `examples/common/`. The `TARGET=<target-name>` and `BOARD=<board-name>` make variable definitions should be exported to the C preprocessor by `Makefile.include` (possibly by reusing the `DEFINES` variable) as `TARGET_<TARGET_NAME>` and `BOARD_<BOARD_NAME>`. Then, both the `Makefile` and `project-conf.h` files, and even the example itself can have conditional parts resolving the support of multiple targets.

We can keep a default `TARGET` thanks to `Makefile.target` despite the `common` naming, e.g. `cc2538dk` for `cc2538-common`, as this would be handier. Or, we can keep all the common examples under `cc2538dk`, and just use the `README`s to point where needed, but that would be less clear.

g-oikonomou:
The main stumbling points towards making this really platform-independent are:

* Setting UART speed: 115200 really won't work well, 460800 is the bare minimum that I would recommend and we don't have a uniform way of setting this for the various CPUs
* Selecting between USB/UART and setting the correct UART instance
* Turning on sniffing at the RD driver

The example under `examples/cc2538dk` is one I've been running with a multitude of platforms, including CC26xx and CC1310 in sub-ghz mode. In all those cases, the example's `.c` files are used 100% as-is. The only thing that needs changed is i) RF driver support and ii) configuration in order to sort out the above points.

There were some attempts at a platform-independent example, and the first important change would be to move sniffing functionality away from the RF driver up to the RDC layer, as per #360. Unfortunately, we'd still need a good way to select the output interface and platform checks using `#if`s is something I'd like to avoid. I'd also like to avoid yet another `_arch` API.

Thinking aloud, perhaps it would be more suitable to put the sniffer under `apps/`, setting the output interface with an FP-based API. Platform-specific examples would then simply configure the platform accordingly and initialise the sniffer with the correct FPs.

g-oikonomou:
Something [like that](https://github.com/g-oikonomou/contiki/commits/contrib/sensniff) (work-in-progress - I don't expect it to work but you get the idea)

With that in place, we can get rid of all sniffer blocks inside RF drivers. The only requirement is that the driver supports the extended RF API so we can set promiscuous mode and set/get the channel.

Another little problem is that `netstack.c` in the app directory does not override the one in core, so we'd need a workaround.

I can't immediately think of how it could become even more platform-independent.

bthebaudeau:
@g-oikonomou This idea looks good. It seems to be the easiest way of abstracting things without having to make large changes in Contiki.

The set of function pointers passed to `sensniff_init()` are basically a serial driver API, so maybe this could be defined somewhere as a structure of function pointers to be passed to `sensniff_init()` instead of each individual pointer.

alignan:
So what do we do here? merge and then wait for a new PR? or refactor and just include the `cc1200` changes, leaving to the user to edit the `cc2538-common/sniffer` application?

alignan:
I'm adding a timeout, in my opinion is better to have a new feature than none while waiting for a new integration proposal

alignan:
:+1: merging after timeout

PR #1449, opened 2016-01-10
BMP085 barometric sensor for the RE-Mote platform

alignan:
This PR adds support for the BMP085 digital barometric sensor (atmospheric pressure and temperature, soon altitude as well).  It depends on #1446.


alignan:
Merging as it is platform-dependent

PR #1450, opened 2016-01-10
Obsolete the Sensinode Platform

g-oikonomou:
With this pull I am proposing removal of the Sensinode platform, for a number of reasons:

* The hardware is way too old (c. 2009). Contiki outgrew the platform quite a while ago. By keeping the platform in the main tree we give a false sense of support, when in reality the platform is not supported. If one is to use Sensinode, they have to use it with my cc-ports changes that are maintained externally.
* The hardware is no longer commercially available.
* Over the past years I have only seen a very small number of teams using the kit. I don't think the number of users justifies keeping the platform in the main tree.
* I no longer use the hardware and I do not have the bandwidth, nor willingness to maintain it
* The CPU (CC2430 SoC) is also way too old and no longer relevant, therefore removal of the platform means we can also remove the cpu files.

One could argue that the N601 NanoSensor could be used very nicely as a sniffer, but my counter argument is that the CC2531 USB dongle is a much much better solution and is also available to buy. This is also the reason why I am not obsoleting the CC2530 port yet.

This pull therefore removes `platform/sensinode`, `cpu/cc2430`, `examples/sensinode` and `tools/sensinode`. We also adjust travis to stop testing removed flies, we delete relevant lines from `.gitignore` and we remove references to the platform from various places in documentation and comments.


adamdunkels:
:+1:

PR #1452, opened 2016-01-10
Compile-test the CC2650 Sensortag and Srf06+CC1310EM code

g-oikonomou:
This adds three travis compile jobs in order to test:

* Drivers for the CC2650 Sensortag
* Drivers for CC1310 RF prop mode
* The cc2650 web demo and the very sleepy demo for the CC2650 Sensortag

This is what it looks like:
```
[...]
Building example 07: cc26xx/very-sleepy-demo/ BOARD=sensortag/cc2650 for target srf06-cc26xx
cc26xx/very-sleepy-demo/ srf06-cc26xx: OK
Building example 08: cc26xx/cc26xx-web-demo/ BOARD=sensortag/cc2650 for target srf06-cc26xx
cc26xx/cc26xx-web-demo/ srf06-cc26xx: OK
Building example 09: cc26xx/cc26xx-web-demo/ BOARD=srf06/cc13xx for target srf06-cc26xx
cc26xx/cc26xx-web-demo/ srf06-cc26xx: OK
[...]
```

PR #1451, opened 2016-01-10
Rename select() in various sensortag sensor drivers

g-oikonomou:
This is to prevent a compile error caused by function prototype conflict between a function called `select()` (present in a number of sensortag peripheral drivers) and the one in `sys/select.h`. The error manifests itself with gcc-arm-embedded version 5.2.1 20151202 (2015-q4).

To fix the error we simply rename our function.

This is simply a compile fix. I have not tested whether the resulting binaries work correctly.

Fixes #1434

PR #1453, opened 2016-01-10
Change travis arm tests to use gcc-arm-embedded v5

g-oikonomou:
Do we want to switch travis to arm-gcc v5.2.1?

We'll first need to make sure that resulting binaries work well for the affected platforms.

If #1452 gets merged, then merging this one here will cause travis errors unless #1451 also gets merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,9 +44,9 @@ before_script:
   ## in Ubuntu >= 14.04, but this external PPA is needed for 12.04.
   ## Install srecord
   - if [ ${BUILD_ARCH:-0} = arm-aapcs ] ; then
-      sudo add-apt-repository -y ppa:terry.guo/gcc-arm-embedded &&
+      sudo add-apt-repository -y ppa:team-gcc-arm-embedded/ppa &&
       sudo apt-get -qq update &&
-      sudo apt-get -qq install gcc-arm-none-eabi srecord &&
+      sudo apt-get -qq install gcc-arm-embedded srecord &&
       arm-none-eabi-gcc --version ;
     fi
 
--- a/platform/cc2538dk/README.md
+++ b/platform/cc2538dk/README.md
@@ -66,21 +66,15 @@ The platform has been developed and tested under Windows XP, Mac OS X 10.9.1 and
 
 Install a Toolchain
 -------------------
-The toolchain used to build contiki is arm-gcc, also used by other arm-based Contiki ports. If you are using Instant Contiki, you will have a version pre-installed in your system. To find out if this is the case, try this:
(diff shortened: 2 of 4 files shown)


bthebaudeau:
> Do we want to switch travis to arm-gcc v5.2.1?

It is the current GCC ARM Embedded version, so yes, unless there is a very good reason not to. Some people are already beginning to use this version. Also, if people follow some `README`s like https://github.com/contiki-os/contiki/blob/master/platform/cc2538dk/README.md, they'll end up with this version.

> We'll first need to make sure that resulting binaries work well for the affected platforms.

Sure. I have already tested the build of all the CC2538 examples with this version on CC2538DK and Zoul. I have also runtime-tested a few CC2538 examples with this version, but not all.

alignan:
I have tested V5 and works OK, but as you are updating travis, perhaps it would be worth to also update the `README.md` files of the `cc2538dk` and `zoul` platforms, and include instructions.

g-oikonomou:
I have updated the respective READMEs as requested and I have tested srf06-cc26xx platform with v5 without any obvious problems.

Please consider.

alignan:
:+1:

bthebaudeau:
There is just a slight issue with the new READMEs vs. `.travis.yml`: the latest version provided by the PPA will very likely be upgraded in the future. If/when this happens, we won't be notified, and the READMEs will no longer match Travis tests. So either the version should not be indicated in the READMEs, or `.travis.yml` should pick up this specific version explicitly. Having a controlled infrastructure and informing users is better, so I prefer the latter, which can be implemented with something like `sudo apt-get -qq install gcc-arm-embedded=5-2015q4-1~wily1 srecord &&` (I don't know which version of Ubuntu is used by Travis, and perhaps it is not even controlled, which might be an issue here, but `wget` could then be used instead of `apt-get`).

bthebaudeau:
According to the [documentation of Travis CI's build environment](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments), Contiki uses Ubuntu 12.04 LTS Server Edition 64 bit, so the choices here for consistency between `.travis.yml` and the READMEs are:
 - make `.travis.yml` use `sudo apt-get -qq install gcc-arm-embedded=5-2015q4-1~precise1 srecord &&`,
 - change the READMEs to indicate that the reference used for testing is the latest version of the GNU ARM Embedded Toolchain.

@g-oikonomou @alignan What do you prefer (if you agree that something needs to be changed)?

g-oikonomou:
We should probably force a version, lest travis decides to suddenly start failing due to an upgrade of the version provided in the PPA.

Changing a number of READMEs each time we bump to a new version is unnecessary PITA though, but I cannot think of a great solution. READMEs could simply state "Check travis.yml for the current version", but that would not be too friendly to the beginners (who are users that need READMEs the most)

PR #1454, opened 2016-01-10
Fixes and enhancements for the SHT25 sensor

alignan:
This PR fixes the SHT25 temperature and humidity driver and extends its available features, such as changing the resolution, checking for a low-voltage alarm, etc.

The PR depends on #1449.


alignan:
Merging as fixes a broken driver and has enhancements for the platform-dependent implementation

PR #1456, opened 2016-01-11
Fixed filter value preventing from initializing the ADC

alignan:
This PR fixed the ADC initialization as the filter condition was wrong.


alignan:
Merging as it is platform-dependent and fixes a broken implementation

PR #1459, opened 2016-01-12
ADC wrapper for the Zoul-based platforms

alignan:
This PR enables support to use an ADC wrapper and reuse its implementation for many different analogue sensors, using a simplified interface on top of the ADC interface.

As an example the following sensors are included, but more will be included in the next weeks.

* [Phidget rotation sensor 1109](http://www.phidgets.com/products.php?product_id=1109)
* [Grove loudness sensor](http://www.seeedstudio.com/wiki/Grove_-_Loudness_Sensor)
* [Grove light sensor](http://www.seeedstudio.com/wiki/Grove_-_Light_Sensor)


alignan:
Platform dependent, merging already :+1:

PR #1462, opened 2016-01-16
Fixed loading SerialSocketServer plugin in Cooja using FQDN

MichaelKlemm:
Loading examples using the SerialSocketServer plugin in Cooja failed. This was because of using SerialSocketServer instead of the full name org.contikios.cooja.serialsocket.SerialSocketServer. This patch fixed the problem on my side.

To reproduce the failure:
1.) Launch cooja (ant run)
2.) Load e.g. examples/ipv6/sky-websense/example-sky-websense.csc
3.) On the command prompt you should see an error on loading SerialSocketServer
4.) Setup tunslip (e.g. make connect-router-cooja) will fail without this patch

### DIFF
--- a/examples/er-rest-example/server-client-native.csc
+++ b/examples/er-rest-example/server-client-native.csc
@@ -219,7 +219,7 @@
     <location_y>182</location_y>
   </plugin>
   <plugin>
-    SerialSocketServer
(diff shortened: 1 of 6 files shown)


fros4943:
:+1:!

PR #1463, opened 2016-01-16
Fixed using target AVR ATmega128RFA1

MichaelKlemm:
Set operating speed to 16MHz and added missing dependency to llsec.
This fix was tested with sparkfun ATmega128RFA1 Dev Board.


MichaelKlemm:
Retrigger

herjulf:
Yes weird that the AtMegaXXRFXX platforms are using 8Mhz probably the default internal  RC-oscillator.  The AtMegaXXRFXX's are required to have 16MHz xtal for the radio tranceiver so it's possible to set the fuse bits to use the same xtal for the MCU as well. What is probably what you do. So patch seems  very understandable to me. Also what I do for the AtMega256RFR2 port. It's also a pending pull request.

nvt:
:+1:

PR #1466, opened 2016-01-19
Add driver for the Sparkfun Weather meter station

alignan:
This PR adds support for the [Sparkfun's weather meter](https://www.sparkfun.com/products/8942) station, composed of a wind vane, rain gauge and anemometer.  The implementation is based on the RE-Mote platform, extended to provide average values and interrupt events.


alignan:
:+1:

PR #1467, opened 2016-01-19
Removed harsh restriction on mask values and disabled debug flags

alignan:
Fixes mask OR'ed values being treated as invalid due to restrictive check, leftover from a previous commit


alignan:
:+1:

PR #1470, opened 2016-01-22
Ported Grove LCD with RGB backlight for the Zoul platforms

alignan:
This PR adds support for the [Grove's LCD display with RGB backlight](http://www.seeedstudio.com/wiki/Grove_-_LCD_RGB_Backlight), based on the [JHD1214](http://www.seeedstudio.com/wiki/images/0/03/JHD1214Y_YG_1.0.pdf) LCD and the [PCA9633 I2C LED driver](http://www.seeedstudio.com/wiki/images/1/1c/PCA9633.pdf)

re-mote and grove lcd


alignan:
:+1:

PR #1471, opened 2016-01-22
Unified BMP180 and BMP085 drivers

alignan:
This PR unifies the driver to support both BMP085 and the BMP180 atmospheric pressure sensors, as both reuse the same logic, but the BMP180 has been advertised to provide better results.


alignan:
No brainer, just renaming :+1:

PR #1472, opened 2016-01-22
Add support for the ITG-3200 3-axis gyroscope

alignan:
This PR adds support for the [Grove's 3-axis gyroscope](http://www.seeedstudio.com/wiki/Grove_-_3-Axis_Digital_Gyro), including a zero-point calibration function, return values in degrees/seconds and data-ready interrupts to maximize the data acquisition.


alignan:
Merging this before #1470

PR #1473, opened 2016-01-23
Update broadcast-announcement.c to prevent memory over-read

zq301:
when the third parameter of "memcpy" is a constant, it is necessary to have a length check before the copy operation to make sure "memory over-read" won't happen.


greg-king5:
You need to rethink your assumptions.  Those memcpy() calls are not copying strings; they are copying structures.  The code isn't buggy; there is **no** possibility of over-runs.

nvt:
Thanks for your efforts, but I'm closing this PR because it is invalid. The packet buffer is not a string, as pointed out above. Furthermore, there should not be a runtime check that silently fails. The subsequent code assumes that <code>adata</code> has been set.

PR #1475, opened 2016-01-23
(Update shell-vars.c) replacing the "sprintf" with a special arithmetic to improve performance

zq301:
use "sprintf" to convert data format  sometime is inefficiently, we can use a special arithmetic to replace it, so as to execution speed of program.

### DIFF
--- a/apps/shell/shell-vars.c
+++ b/apps/shell/shell-vars.c
@@ -104,16 +104,17 @@ PROCESS_THREAD(shell_var_process, ev, data)
 	sprintf(numbuf, " %d", *((int *)symbols[i].value));
 	shell_output_str(&var_command, (char *)symbols[i].name, numbuf);
 
+	static char hexmap[]={'0','1','2','3','4','5','6','7','8','9','a','b','c','d','e','f'};
 	for(j = 0; j < 8 * 8; j += 8) {
-	  sprintf(numbuf, "0x%02x 0x%02x 0x%02x 0x%02x 0x%02x 0x%02x 0x%02x 0x%02x",
-		  ((unsigned char *)symbols[i].value)[j],
-		  ((unsigned char *)symbols[i].value)[j + 1],
-		  ((unsigned char *)symbols[i].value)[j + 2],
-		  ((unsigned char *)symbols[i].value)[j + 3],
-		  ((unsigned char *)symbols[i].value)[j + 4],
-		  ((unsigned char *)symbols[i].value)[j + 5],
-		  ((unsigned char *)symbols[i].value)[j + 6],
-		  ((unsigned char *)symbols[i].value)[j + 7]);
+		char *p=numbuf;
+		int cnti;
+		for(cnti=j;cnti<j+8;cnti++){
+			(*p++)='0';
+           		(*p++)='x';
+           		(*p++)=hexmap[(unsigned char *)symbols[i].value)[cnti]>>4];
+        		(*p++)=hexmap[(unsigned char *)symbols[i].value)[cnti]&15];
+           		(*p++)=' ';
+		}
 	  shell_output_str(&var_command, numbuf, "");
 	}
 	PROCESS_EXIT();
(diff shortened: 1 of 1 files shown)


apps/shell/shell-vars.c
```diff
@@ -104,16 +104,17 @@ PROCESS_THREAD(shell_var_process, ev, data)
...
-		  ((unsigned char *)symbols[i].value)[j + 5],
-		  ((unsigned char *)symbols[i].value)[j + 6],
-		  ((unsigned char *)symbols[i].value)[j + 7]);
+		char *p=numbuf;
+		int cnti;
+		for(cnti=j;cnti<j+8;cnti++){
+			(*p++)='0';
+           		(*p++)='x';
+           		(*p++)=hexmap[(unsigned char *)symbols[i].value)[cnti]>>4];
+        		(*p++)=hexmap[(unsigned char *)symbols[i].value)[cnti]&15];
```
greg-king5: A left parenthesis is missing in the above two lines.  It is needed to balance the right parenthesis in "`.value)`".

zq301:
@greg-king5 the defination of "numbuf" in these codes: char numbug[32]; 
but the format dada wrote into it is"0x%02x 0x%02x 0x%02x 0x%02x 0x%02x 0x%02x 0x%02x 0x%02x". the length is 39. there is a memory overflow.

apps/shell/shell-vars.c
```diff
@@ -104,16 +103,18 @@ PROCESS_THREAD(shell_var_process, ev, data)
...
+		p=numbuf;
+		int cnti;
+		for(cnti=j;cnti<j+8;cnti++){
+			(*p++)='0';
+           		(*p++)='x';
+           		(*p++)=hexmap[((unsigned char *)symbols[i].value)[cnti]>>4];
+        		(*p++)=hexmap[((unsigned char *)symbols[i].value)[cnti]&15];
+           		(*p++)=' ';
+		}
 	  shell_output_str(&var_command, numbuf, "");
```
greg-king5: The string must be "terminated" (`'\0'`) before it is output.

zq301: yes, I reported it at https://github.com/contiki-os/contiki/pull/1478

zq301:
I have a simple test on the previous codes and current codes. the cost time of previous code used "sprintf" to transform 64 bytes is 64820ns. the current codes is 51898ns.
If the codes uesd frequently, I think it will save more CPU time to make a progress to contiki. but now, I don't know why the check failed?

vsaw:
@zq301 It's a compilation error complaining about an undefined symbol [`hexmap`](https://github.com/contiki-os/contiki/pull/1475/files#diff-cf8a6b015d2f8c067e1808eaff5dec2eR106) on the following environment:

```
$ export BUILD_TYPE='compile-6502-ports'
$ export BUILD_CATEGORY='compile'
$ export BUILD_ARCH='6502'
```

See https://travis-ci.org/contiki-os/contiki/jobs/104402464#L3160-L3170

apps/shell/shell-vars.c
greg-king5: Pull Request 1478 is a different issue.  I am talking about the output of your new code.  It needs to make a valid C string.  It needs to put a NUL character at the end of the array; or else, `shell_output_str()` will read past the end of `numbuf`, and show garbage.

greg-king5:
That function is a shell command. It is a part of the user interface.  Therefore, size is more important there than speed is.

The check "failed" because your code obeys the C99 standard.  But, `BUILD_ARCH='6502'` uses a compiler that wants to see C90 code; it doesn't like the location of your `hexmap[]` and `p` definitions (in between executable lines).

Even if that wasn't a problem, I recommend that you move all of your variables definitions to the beginning of that function; it makes slightly smaller binaries.  Also, `hexmap[]` should be declared as `static const`.

nvt:
It is not a good trade-off to reduce the execution time a bit for a seldomly executed code segment if it results in less readable and portable code. In addition, it has incorrect coding style and the tests fail. Closing the PR.

PR #1476, opened 2016-01-23
Updated to new adc-zoul renaming

alignan:
Updated as previously named `adc-sensors` has been renamed to `adc-zoul`


alignan:
:+1:

PR #1477, opened 2016-01-24
code  redundancy

zq301:
there is code redundancy in Contiki. I added three functions to process.h.
1.shaves off any leading  space :char*  process_space_begin(char*)
2.Find first space in a string, then replace the space with zero:void nullterminate(char *)
3.static int is_broadcast_addr(uint_t , uint_t *)


zq301:
I don't know if it is a right place to put the three functions in process.h?? And I think write the first function in assembly code will be more efficient. but I'm very sory I don't how to write assembly code.

apps/shell/shell.c
```diff
@@ -243,11 +243,7 @@ start_command(char *commandline, struct shell_command *child)
   int command_len;
   struct shell_command *c;
 
-  /* Shave off any leading spaces. */
-  while(*commandline == ' ') {
-    commandline++;
-  }
-
+commandline=process_space_begin(commandline);
```
oliverschmidt: Not properly indented.

core/net/mac/sicslowmac/sicslowmac.c
```diff
@@ -80,18 +80,7 @@ static uint16_t mac_dst_pan_id = IEEE802154_PANID;
...
-{
-  int i = mode == FRAME802154_SHORTADDRMODE ? 2 : 8;
-  while(i-- > 0) {
-    if(addr[i] != 0xff) {
-      return 0;
-    }
-  }
-  return 1;
-}
+
```
oliverschmidt: Spurious empty line.

apps/netconf/netconf.c
```diff
@@ -119,16 +119,7 @@ makestrings(void)
...
-static void
-nullterminate(char *cptr)
-{
-  /* Find the first space character in the ipaddr and put a zero there
-     to end the string. */
-  for(; *cptr != ' ' && *cptr != 0; ++cptr);
-  *cptr = 0;
-}
-/*-----------------------------------------------------------------------------------*/
+
```
oliverschmidt: Empty line instead of horizontal line.

core/sys/process.h
```diff
@@ -529,3 +529,26 @@ CCIF extern struct process *process_list;
...
+
+static int is_broadcast_addr(uint_t mode, uint_t *addr){
+	int i=mode==FRAME802154_SHORTADDRMODE ? 2 : 8;
+	while(i-->0){
+		if(addr[i] != oxff){
+			return 0;
+		}
+	}
+	return 1;
+}
```
oliverschmidt: The whole section above:
- Code in header files ?!?
- Inconsistent comments (Indented lower case, non-indented upper case, no comment at all)
- Inconsistent code indentation
- Not following Contiki code style guide

oliverschmidt:
> ... write the first function in assembly code ...

Are you aware that Contiki is "massive multi platform" project ?

zq301:
@oliverschmidt I'm sorry that I haven't used Contiki. I just do some static analysis to the source code. I think this code may be reduced to save storage. And I just want to improve performance of Contiki.

core/sys/process.h
zq301: I'm very sorry that I'm not a Contiki user, so I don't know the code style. I do some works on software static analysis to detect bugs. the same code of function "process_space_begin" is used 11 times in contiki. "nullterminate" is used 4 times. "is_broadcast_addr" is used 3 times.May be these codes can be write as functions in a header file or using some better ways to save storage and improve efficiency.

greg-king5: The file https://github.com/contiki-os/contiki/blob/master/doc/code-style.c describes some of the code style.

nvt:
The proposed functions do not belong in sys/process.h. If you try to refactor some common code segments into functions, it would be good if you could provide numbers on how much you reduce the size. Several of the listed modules are seldomly used together, and thus the savings may be small. The code size may even be increased, depending on how much the function calls add.

Code size and execution times are least relevant on the native platform, so the measurements should preferably be on one or more of the resource-constrained platforms supported by Contiki.

PR #1478, opened 2016-01-24
a memory overflow in shell-vars.c

zq301:
" sprintf(numbuf, "0x%02x 0x%02x 0x%02x 0x%02x 0x%02x 0x%02x 0x%02x 0x%02x",...)" . the formatted data wrote to "numbuf" is 39 bytes, but numbuf is 32 bytes.

### DIFF
--- a/apps/shell/shell-vars.c
+++ b/apps/shell/shell-vars.c
@@ -91,7 +91,7 @@ PROCESS_THREAD(shell_var_process, ev, data)
   int i;
   int j;
   
-  char numbuf[39];
+  char numbuf[40];
   
   PROCESS_BEGIN();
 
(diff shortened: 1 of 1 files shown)


atiselsts:
This one looks legit, except that the buffer size should be 40 instead of 39, as buffer space is also required for the terminating zero character.

zq301:
@atiselsts yes, I agree with you. and I have modified the size to 40.

simonduq:
how about using `snprintf`? (in addition to the buf len fix)

greg-king5:
The overhead of `snprintf()` isn't needed in this case because the length of the output string is guaranteed to be always 40 bytes.

simonduq:
yes but the day someone modifies the printf format or buffer length we get the overflow back for free

nvt:
Good fix, but I agree with @simonduq that using snprintf also would be nice.

nvt:
Let's merge this since the PR contains a bug fix, and additional imporovements could be put into a separate PR. :+1:

PR #1479, opened 2016-01-24
Allow the caller to access the entire content of CMDSTA

### LINKED ISSUE #1465: CC1310DK keep rebooting runing cc26xx-demo with latest Contiki version.
Just sync to latest Contiki version and build cc26xx-demo for CC1310DK with "make TARGET=srf06-cc26xx BOARD=srf06/cc13xx cc26xx-demo.bin". The binary will make my CC1310 keep rebooting.

g-oikonomou:
When sending a command to the CC13xx/CC25xx RF core, we wait for command completion by checking the LSB of CMDSTA (correctly so). However, in doing so we also zero out the 3 CMDSTA return bytes. For some commands, those bytes contain useful information (e.g. an RSSI value) and are required by the caller.

This problem manifests itself e.g. in PROP mode `channel_clear()`, whereby the caller will always see an RSSI value of 0.

This pull therefore fixes the logic in `rf_core_send_cmd()` to check for command completion by blocking on the CMDSTA result byte without zeroing out the 3 return bytes.

Fixes #1465

### DIFF
--- a/cpu/cc26xx-cc13xx/rf-core/rf-core.c
+++ b/cpu/cc26xx-cc13xx/rf-core/rf-core.c
@@ -160,7 +160,7 @@ rf_core_send_cmd(uint32_t cmd, uint32_t *status)
 
   HWREG(RFC_DBELL_BASE + RFC_DBELL_O_CMDR) = cmd;
   do {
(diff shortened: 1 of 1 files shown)


g-oikonomou:
Confirmed fix.

PR #1480, opened 2016-01-25
galileo: Revise include flags for LLVM Clang

mdlemay:
This patch revises the compiler flags when LLVM Clang is in use to
cause Clang to automatically use its built-in header files rather than
those for GCC.  This avoids compiler errors that can otherwise
be generated by including certain built-in headers.


mdlemay:
@aguedes, @jeez: Can you please review this?

nvt:
:+1:

PR #1486, opened 2016-01-29
Unified cc2538 examples into single cc2538-common placeholder

alignan:
This PR follows the conversation at #1420 about having a single `cc2538-common` directory for cc2538 examples.

* Simplifies the support of the cc2538
* Avoids repeated examples per platform
* Easier to look for cc2538 specific examples, instead of having to browse through platforms directories


g-oikonomou:
I didn't look into it in great detail, but this looks like a very reasonable change!

examples/cc2538-common/Makefile
```diff
@@ -0,0 +1,11 @@
+CONTIKI_PROJECT = test-pwm timer-test
```
bthebaudeau: `timer-test` uses rtimers, so it needs the original `project-conf.h` to avoid using the default conflicting `NETSTACK_CONF_RDC`.

examples/cc2538-common/Makefile
```diff
@@ -0,0 +1,11 @@
...
+
+ifndef TARGET
+TARGET = cc2538dk
+endif
```
bthebaudeau: Why don't you create a `Makefile.target` in the usual way instead?

examples/cc2538-common/crypto/Makefile
```diff
@@ -1,6 +1,10 @@
 CONTIKI_PROJECT = ecb-test cbc-test ctr-test cbc-mac-test gcm-test ccm-test
 CONTIKI_PROJECT += sha256-test
 
+ifndef TARGET
+TARGET = cc2538dk
+endif
```
bthebaudeau: Ditto: `Makefile.target`?

examples/cc2538-common/mqtt-demo/Makefile
```diff
@@ -1,10 +1,14 @@
 DEFINES+=PROJECT_CONF_H=\"project-conf.h\"
 
+ifndef TARGET
+TARGET = cc2538dk
+endif
```
bthebaudeau: Ditto: `Makefile.target`?

examples/cc2538-common/pka/Makefile
```diff
@@ -1,5 +1,9 @@
 CONTIKI_PROJECT = ecc-ecdh ecc-sign ecc-verify
 
+ifndef TARGET
+TARGET = cc2538dk
+endif
```
bthebaudeau: Ditto: `Makefile.target`?

examples/cc2538-common/sniffer/Makefile
```diff
@@ -1,6 +1,10 @@
 DEFINES+=PROJECT_CONF_H=\"project-conf.h\"
 PROJECT_SOURCEFILES += stub-rdc.c
 
+ifndef TARGET
+TARGET = cc2538dk
+endif
```
bthebaudeau: Ditto: `Makefile.target`?

regression-tests/18-compile-arm-ports/Makefile
```diff
@@ -17,19 +17,23 @@ ipv6/rpl-border-router/cc2538dk \
 er-rest-example/cc2538dk \
 ipso-objects/cc2538dk \
 webserver-ipv6/cc2538dk \
-cc2538dk/cc2538dk \
```
bthebaudeau: Why? There is still `cc2538-demo` left there.

examples/cc2538-common/test-pwm.c
```diff
@@ -0,0 +1,203 @@
+/*
```
bthebaudeau: Looks like this one has been copied instead of moved.

bthebaudeau:
This is based on the discussion in #1447, besides #1420.

Thanks. This is a very good change. There are just a few things to fix as commented on the diff.

examples/cc2538-common/Makefile
alignan: No particular reason, less files only.  I can change this if you prefer so.

bthebaudeau: Yes, please. I would not care much apart from the fact that it's unusual in Contiki. But I've just tested it, and it's broken:
```
make TARGET=zoul savetarget
make BOARD=firefly
```
builds for `cc2538dk` instead of `zoul` because the `TARGET` defined in `Makefile` takes precedence over the one defined in `Makefile.target`, because `Makefile.include`, which includes `Makefile.target`, is included at the end of `Makefile`, after your definition of `TARGET` to `cc2538dk`. But `make TARGET=zoul BOARD=firefly` works as expected because in this case `TARGET` is defined to `zoul` right from the beginning, so your definition does not override it.

examples/cc2538-common/test-pwm.c
bthebaudeau: This still has to be fixed:
```
git rm examples/zolertia/zoul/test-pwm.c
```

bthebaudeau:
So you still have to fix `Makefile.target` and `test-pwm.c` as discussed above. Also, please squash your commits. Hopefully, Travis will be happy next time (the current failure is unrelated to your changes).

alignan:
Done!

bthebaudeau:
You're all set. Thanks!
:+1:

PR #1487, opened 2016-01-31
Fixed casts and calculations for the wind direction circular averaging

alignan:
This PR fixes the wind direction average calculations, as using unsigned variables was provoking erratic values and missing some checks.  This has been tested in a personal project over the past days, and for different averaging values (validated [mitsuta](http://abelian.org/vlf/bearings.html) vs [mean of circular quantities](https://en.wikipedia.org/wiki/Mean_of_circular_quantities) with small variations)

fixed wind direction average int

### DIFF
--- a/platform/zoul/dev/weather-meter.c
+++ b/platform/zoul/dev/weather-meter.c
@@ -102,9 +102,9 @@ typedef struct {
 } weather_meter_sensors;
 
 typedef struct {
-  uint32_t value_buf_xm;
-  uint16_t value_prev;
-  uint16_t value_avg_xm;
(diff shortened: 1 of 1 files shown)


alignan:
:+1:

greg-king5:
Calling abs() in line 239 is overkill.  That expression is executed only when the value is negative; therefore, you can get the positive value simply by negating the negative.

alignan:
Yes, you are right, this is a leftover from testing before moving everything into separate conditions for negative/positive values.  I have more changes and testing pending, I will include this in the next commit.  Thanks!

PR #1490, opened 2016-02-02
Coffee filesystem example unified

alignan:
This follows the conversation in #1420 in order to have a single example for platforms supporting CFS.

### DIFF
--- a/examples/coffee/Makefile
+++ b/examples/coffee/Makefile
@@ -1,0 +1,7 @@
+CONTIKI = ../..
+
+all: test-cfs example-coffee test-coffee
+
+CONTIKI_WITH_RIME = 1
+
+include $(CONTIKI)/Makefile.include
--- a/examples/coffee/project-conf.h
+++ b/examples/coffee/project-conf.h
@@ -1,0 +1,42 @@
+/*
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+/*---------------------------------------------------------------------------*/
+#ifndef PROJECT_CONF_H_
+#define PROJECT_CONF_H_
+/*---------------------------------------------------------------------------*/
+#if CONTIKI_TARGET_CC2538DK || CONTIKI_TARGET_ZOUL
+#define COFFEE_CONF_SIZE              (CC2538_DEV_FLASH_SIZE / 2)
+#define COFFEE_CONF_LOG_TABLE_LIMIT   16
+#define COFFEE_CONF_LOG_SIZE          128
+#define COFFEE_CONF_MICRO_LOGS        1
+#define COFFEE_CONF_APPEND_ONLY       0
+#endif /* CONTIKI_TARGET_CC2538DK || CONTIKI_TARGET_ZOUL */
+
+#endif /* PROJECT_CONF_H_ */
+/*---------------------------------------------------------------------------*/
+/** @} */
--- a/core/cfs/cfs-coffee.c
+++ b/core/cfs/cfs-coffee.c
@@ -196,7 +196,7 @@ struct file_header {
 /* This is needed because of a buggy compiler. */
 struct log_param {
   cfs_offset_t offset;
-  const char *buf;
+  char *buf;
   uint16_t size;
 };
 
@@ -1189,7 +1189,7 @@ cfs_write(int fd, const void *buf, unsigned size)
     need_dummy_write = 0;
     for(bytes_left = size; bytes_left > 0;) {
       lp.offset = fdp->offset;
-      lp.buf = buf;
+      lp.buf = (void *)buf;
       lp.size = bytes_left;
       i = write_log_page(file, &lp);
       if(i < 0) {
--- a/cpu/cc2538/Makefile.cc2538
+++ b/cpu/cc2538/Makefile.cc2538
@@ -115,8 +115,7 @@ CUSTOM_RULE_LINK=1
 	cp $< $@
 
 ### This rule is used to generate the correct linker script
-LDGENFLAGS += $(addprefix -D,$(subst $(COMMA), ,$(DEFINES)))
-LDGENFLAGS += $(addprefix -I,$(SOURCEDIRS))
+LDGENFLAGS += $(CFLAGS)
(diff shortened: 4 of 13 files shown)


bthebaudeau:
- My `project-conf.h` from #1420 with `\addtogroup` removed, `#if CONTIKI_TARGET_CC2538DK || CONTIKI_TARGET_ZOUL` added, and `Makefile` updated, is still missing, causing runtime failure for all the CC2538 targets.
 - The definition of `COFFEE_CONF_SIZE` should be removed from `platform/zoul/contiki-conf.h` in order not to waste flash space when Coffee is unused. Or, if you always need Coffee on this platform, which would be surprising for an application-specific feature, you should at least add `#ifndef COFFEE_CONF_SIZE`.
 - Please squash your commits.

You will also need the following hunk:
```
diff --git a/cpu/cc2538/Makefile.cc2538 b/cpu/cc2538/Makefile.cc2538
index a11dc2b..f28a77d 100644
--- a/cpu/cc2538/Makefile.cc2538
+++ b/cpu/cc2538/Makefile.cc2538
@@ -115,8 +115,7 @@ CUSTOM_RULE_LINK=1
 	cp $< $@
 
 ### This rule is used to generate the correct linker script
-LDGENFLAGS += $(addprefix -D,$(subst $(COMMA), ,$(DEFINES)))
-LDGENFLAGS += $(addprefix -I,$(SOURCEDIRS))
+LDGENFLAGS += $(CFLAGS)
 LDGENFLAGS += -imacros "contiki-conf.h" -imacros "dev/cc2538-dev.h"
 LDGENFLAGS += -imacros "dev/flash.h" -imacros "cfs-coffee-arch.h"
 LDGENFLAGS += -x c -P -E
```
Without it, the conditional definitions in `project-conf.h` depending on `CONTIKI_TARGET_<TARGET_NAME>` are ignored at link time, which breaks the linker script, so the flashed applications crash.

You will also have to add a few adjustments of the Coffee settings to `project-conf.h` for the CC2538 targets in order to enable file modifications (and not only appending) used by these tests, something like:
```
#define COFFEE_CONF_LOG_TABLE_LIMIT 16
#define COFFEE_CONF_LOG_SIZE 128
#define COFFEE_CONF_MICRO_LOGS 1
#define COFFEE_CONF_APPEND_ONLY 0
```
But some fine tuning is still needed to get all three examples working.

You will also need this in order to fix a warning caused by Coffee:
```
diff --git a/core/cfs/cfs-coffee.c b/core/cfs/cfs-coffee.c
index 4b08d45..000368b 100644
--- a/core/cfs/cfs-coffee.c
+++ b/core/cfs/cfs-coffee.c
@@ -196,7 +196,7 @@ struct file_header {
 /* This is needed because of a buggy compiler. */
 struct log_param {
   cfs_offset_t offset;
-  const char *buf;
+  char *buf;
   uint16_t size;
 };
 
@@ -1189,7 +1189,7 @@ cfs_write(int fd, const void *buf, unsigned size)
     need_dummy_write = 0;
     for(bytes_left = size; bytes_left > 0;) {
       lp.offset = fdp->offset;
-      lp.buf = buf;
+      lp.buf = (void *)buf;
       lp.size = bytes_left;
       i = write_log_page(file, &lp);
       if(i < 0) {
```

alignan:
I have made the changes suggested above, also fixed a warning when compiling for the `z1` platform, and excluded `sky` from the regression tests, as currently it fails to compile due to an overflow.  I haven't tested the `avr-raven` target.

alignan:
It seems like the `sky` overflow cause is IPv6 being pulled-in as default, I just pushed a change, if travis doesn't complain about `sky` I will activate its regression test again.  For the `avr` target however I will need to check something else.

examples/coffee/project-conf.h
```diff
@@ -0,0 +1,42 @@
...
+#define COFFEE_CONF_SIZE              (CC2538_DEV_FLASH_SIZE / 2)
+#define COFFEE_CONF_LOG_TABLE_LIMIT   16
+#define COFFEE_CONF_LOG_SIZE          128
+#define COFFEE_CONF_MICRO_LOGS        1
+#define COFFEE_CONF_APPEND_ONLY       0
+#endif /* CONTIKI_TARGET_CC2538DK || CONTIKI_TARGET_ZOUL */
+
+#endif /* PROJECT_CONF_H_ */
+/*---------------------------------------------------------------------------*/
+/** @} */
```
bthebaudeau: This line no longer closes any doxygen block, so it should be removed.

examples/coffee/Makefile
```diff
@@ -0,0 +1,7 @@
+CONTIKI = ../..
```
bthebaudeau: This `Makefile` should be updated to use `project-conf.h`.

bthebaudeau:
Thanks for your changes. See what I've just added to the diff view. Don't forget that these examples are not yet stable on CC2538, probably because of some tuning needed in the Coffee settings as I mentioned previously, so this should also be fixed before this PR can be merged.

bthebaudeau:
I've started debugging this on Zoul Firefly. I feel like there is a bug in Coffee, or some undocumented constraints required for the microlog settings.

`example-coffee.c` works fine. `test-cfs.c` works fine the 1st time (i.e. with a formatted filesystem), but it hangs the following times.

With `test-coffee.c`, the execution hangs in `coffee_test_basic()`, during the 4th call to `cfs_write()`. This is caused by `read_header()` in `cfs-coffee.c` accessing a negative page.

Actually, the 1st call to `cfs_write()` writes some of the 8 pages that have initially been reserved from page 0. Then, the 3rd call to `cfs_write()`, which aims at writing 256 bytes after having rewound the file, reserves only one more page (page 8) and writes it with a header, the file data and a log, so it overflows to page 9, but there is no header written to page 9, and the header of page 8 is written with `max_pages` set to 1. Consequently, the 4th call to `cfs_write()` reads the beginning of page 9 as if it contained a header, but it is actually not, which results in a negative `max_pages`, then in a negative page being read.

If I set the log table limit to 256 and the log size to 1024 like on Sky, then this part works, but the test hangs later in `coffee_test_gc()`. If I also set the dyn size to 4096, then `coffee_test_gc()` returns an error in test 2 instead of hanging.

It's the same kind of issue for `test-cfs.c`.

So Coffee seems to only support some very specific sets of parameters, and Sky is lucky here. That's probably too specific and undocumented to be expected, and the behavior described above seems to be buggy. I still have to dig deeper.

@nvt You seem to be the main contributor to `cfs-coffee.c`. Do you have any idea before I investigate more? If you want to test, you can use the current branch of this PR on CC2538DK or Zoul with the following patch added:
```
diff --git a/examples/coffee/Makefile b/examples/coffee/Makefile
index df951c1..8be4650 100644
--- a/examples/coffee/Makefile
+++ b/examples/coffee/Makefile
@@ -1,3 +1,4 @@
+DEFINES+=PROJECT_CONF_H=\"project-conf.h\"
 CONTIKI = ../..
 
 all: test-cfs example-coffee test-coffee
diff --git a/examples/coffee/project-conf.h b/examples/coffee/project-conf.h
index 8751878..30429bf 100644
--- a/examples/coffee/project-conf.h
+++ b/examples/coffee/project-conf.h
@@ -39,4 +39,3 @@
 
 #endif /* PROJECT_CONF_H_ */
-/** @} */
```

nvt:
@bthebaudeau Thanks for investigating this problem -- it would be great to get Coffee running on the Zoul Firefly. I haven't yet tested this on hardware as I'm spending most of my time on a paper currently, but from your description, it appears that Coffee is unable to read exactly the same data that it expects to have written in the headers earlier. If the headers are read correctly, file I/O shouldn't overwrite headers. A few possible problems:

* The I/O functions in cfs-coffee-arch.* have a bug,
* There is some implicit integer size assumption in Coffee.
* The configuration in cfs-coffee-conf.h contains some unsuitable setting for the platform's flash chip.
* The particular flash I/O semantics are even more restrictive than that of the ST M25P80, which is a NOR flash memory used in the Tmote Sky. For platforms with page-programmable flash memories, it should work.

If you have some output logs from this test with DEBUG set to 1 in cfs-coffee.c, it would be great if you could send them to me so I might get some more hints as to what is going on.

bthebaudeau:
@nvt Thanks for your comments.

> @bthebaudeau Thanks for investigating this problem -- it would be great to get Coffee running on the Zoul Firefly. I haven't yet tested this on hardware as I'm spending most of my time on a paper currently, but from your description, it appears that Coffee is unable to read exactly the same data that it expects to have written in the headers earlier. If the headers are read correctly, file I/O shouldn't overwrite headers.

I've added a lot of traces to see what's going on. Appending always works fine. Only the micrologs are broken when overwriting the existing contents of a file. When the Coffee headers are read back, their contents seem to be identical to what Coffee had previously written to them. The issue seems to be that Coffee has not written things according to how it then interprets them when reading back, because of some reservation issue.

> A few possible problems:

> * The I/O functions in cfs-coffee-arch.* have a bug,

They have been tested a lot. Appending works fine, and file deletion too, so that's not very likely.

> * There is some implicit integer size assumption in Coffee.

Maybe, though I have checked the type rules given by the Wiki, and the CC2538 port is compliant.

> * The configuration in cfs-coffee-conf.h contains some unsuitable setting for the platform's flash chip.

No, they're compatible, but they may however be incompatible with some implicit assumptions in Coffee.

> * The particular flash I/O semantics are even more restrictive than that of the ST M25P80, which is a NOR flash memory used in the Tmote Sky. For platforms with page-programmable flash memories, it should work.

Yes.

> If you have some output logs from this test with DEBUG set to 1 in cfs-coffee.c, it would be great if you could send them to me so I might get some more hints as to what is going on.

Sure. I have even more detailed logs because `DEBUG` is not verbose enough for this issue, but they're a bit messy for now, so I will have to make them clearer first.

bthebaudeau:
@nvt Here you go:
https://gist.github.com/bthebaudeau/9bd5705bda6eba3f8c75

bthebaudeau:
Superseded by #1686.

The issues were caused by several bugs in the examples, and by an undocumented requirement in Coffee's configuration: COFFEE_LOG_SIZE has to be a multiple of COFFEE_PAGE_SIZE (or at least greater than or equal to it) if `cfs_coffee_configure_log()` is not used. I have fixed everything in #1686, and tested it on Zolertia Zoul Firefly.

PR #1493, opened 2016-02-03
Insert RPL ext header before sending ICMP, UDP and TCP messages

simonduq:
This PR adds missing calls to `rpl_insert_header` right after building Echo Request messages, before calling `tcpip_ipv6_output`. This results in a missing hop-by-hop option at the first hop. In this PR, `rpl_insert_header` is called consistently, for ICMP, UDP, and TCP. We add a check that the packet is not link-local before adding the header, as only routable unicast MUST include RPL headers.

### DIFF
--- a/core/net/ip/tcpip.c
+++ b/core/net/ip/tcpip.c
@@ -549,7 +549,12 @@ tcpip_ipv6_output(void)
   }
 
 #if UIP_CONF_IPV6_RPL
-  rpl_insert_header();
+  if(!rpl_update_header()) {
+    /* Packet can not be forwarded */
+    PRINTF("tcpip_ipv6_output: RPL header update error\n");
+    uip_clear_buf();
+    return;
+  }
 #endif /* UIP_CONF_IPV6_RPL */
 
   if(!uip_is_addr_mcast(&UIP_IP_BUF->destipaddr)) {
@@ -655,12 +660,6 @@ tcpip_ipv6_output(void)
 
     /* End of next hop determination */
 
-#if UIP_CONF_IPV6_RPL
-    if(!rpl_finalize_header(nexthop)) {
-      uip_clear_buf();
-      return;
-    }
-#endif /* UIP_CONF_IPV6_RPL */
     nbr = uip_ds6_nbr_lookup(nexthop);
     if(nbr == NULL) {
 #if UIP_ND6_SEND_NA
--- a/core/net/ip/tcpip.c
+++ b/core/net/ip/tcpip.c
@@ -548,6 +548,10 @@ tcpip_ipv6_output(void)
     return;
   }
 
+#if UIP_CONF_IPV6_RPL
+  rpl_insert_header();
+#endif /* UIP_CONF_IPV6_RPL */
+
   if(!uip_is_addr_mcast(&UIP_IP_BUF->destipaddr)) {
     /* Next hop determination */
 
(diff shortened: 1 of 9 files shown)


platform/avr-raven/apps/raven-lcd-interface/raven-lcd.c
```diff
@@ -146,6 +146,9 @@ raven_ping6(void)
     UIP_ICMP_BUF->icmpchksum = 0;
     UIP_ICMP_BUF->icmpchksum = ~uip_icmp6chksum();
 
+#if UIP_CONF_IPV6_RPL
+    rpl_insert_header();
+#endif /* UIP_CONF_IPV6_RPL */
 
     tcpip_ipv6_output();
```
joakimeriksson: It seems like all these calls are just before tcpip_ipv6_output() - is there any reason why we should not move the multiple calls into that function?

simonduq: Sounds like a good idea but I'm not 100% confident this will not break something else. If you look at tcpip.c this function is called from a lot of different contexts.

simonduq: The more I look at it the more I think you're right, but `rpl_insert_header` might need modifying to handle all different cases.

joakimeriksson: Yes, but I guess we should try to do that to make the code look nicer! It might also be the case that we need a few checks before the call - but then it might make sense to have them in the function to avoid cluttering the caller(s) code.

simonduq: yes I'll give this a try. From browing tcpip.c it seems we'll be missing some ext headers until we do as you suggest. For instance TCP traffic etc.

joakimeriksson: Sounds great - I'll be happy to take a look when you have something!

laurentderu: When Sebastien and I fixed the HBH option, we did only for UDP as a proof of concept, so some refactoring is indeed needed

simonduq:
I did exactly as suggested in the discussion above: move the insertion of the RPL ext header to `tcpip_ipv6_output` for consistency (now the header should be present regardless of where the packet was created)

joakimeriksson:
Looks good - I think this should make all current and future protocols be handled - which is nice!

simonduq:
setting timeout for merging

simonduq:
note to self: will need updating if RPL non-storing is merged

alignan:
Better to have a reminder label, although probably `needs rebasing` should also be added

simonduq:
Rebased now that non-storing is in. This PR has become quite minimal now :p

simonduq:
OK, things got broken and I ended up reworking ext-header handling quite a bit. The interface to `rpl-ext-header` is now consistent (return value 0 means failure) and simpler (basically only update, remove and verify). The logic on inclusion of HBH option or SRH is at a single place now, in `rpl_update_header`.

Removing timeout as the changes are significant.

simonduq:
This is ready again

simonduq:
merging after timeout

PR #1497, opened 2016-02-03
Fix compiler warnings in shell

simonduq:


nvt:
:+1:

cmorty:
@simonduq Having some time to look through the backlog (procrastinating). As Contiki is supposedly warning-free: Doesn't this show that there is a missing test-case compiling this code?

simonduq:
That is true @cmorty. FYI I triggered the warnings when adding a few shell commands to a jn516x code. I don't have a clean example to add at hand but will think about it, thanks for pointing this out

PR #1501, opened 2016-02-04
fix segfault in nbr_table.c

pscholl:
```->next``` pointers in nbr_table_keys may be corrupted after a few neighbor removals/insertions, which leads to segfaults at random times.


simonduq:
The patch is clearly incorrect, `nbr_table_remove` is supposed to remove a neighbor from a particular module only, not globally. Calling `list_remove` will lead to an inconsistent state and possible illegal accesses etc.

pscholl:
Sorry for the line noise. It's actually a bug in GCC 4.9.1, which causes the nbr_table_key_t to be too small, thus overwriting the ->next pointer in subsequent entries when copying in the lladdr.

PR #1502, opened 2016-02-04
Fix timestamps of packets recorded in pcap files

yatch:
This PR makes PcapExporter write frames with timestamps in the simulated time.

Currently, PcapExporter records timestamps when they are written into a pcap file. Because of that, by looking at the pcap file, you cannot know if packets are sent at your expected timings unless the simulation runs at 100% speed.

Here is an example which shows packets generated and recorded in the simulation of `regression-tests/11-ipv6/01-cooja-ipv6-udp.csc`:

before

After applying the changes, you will get this:

after

In both cases above,  timestamps are represented as "Seconds Since Beginning of Capture."  If you select "Date and Time of Day", you will see a result like this:

after-excerpt

### DIFF
--- a/tools/cooja/java/org/contikios/cooja/Simulation.java
+++ b/tools/cooja/java/org/contikios/cooja/Simulation.java
@@ -71,6 +71,7 @@ public class Simulation extends Observable implements Runnable {
   private long speedLimitLastSimtime;
   private long speedLimitLastRealtime;
 
+  private long lastStartTime;
   private long currentSimulationTime = 0;
 
   private String title = null;
@@ -248,7 +249,7 @@ public class Simulation extends Observable implements Runnable {
   }
 
   public void run() {
-    long lastStartTime = System.currentTimeMillis();
+    lastStartTime = System.currentTimeMillis();
     logger.info("Simulation main loop started, system time: " + lastStartTime);
     isRunning = true;
     speedLimitLastRealtime = System.currentTimeMillis();
@@ -1072,6 +1073,16 @@ public class Simulation extends Observable implements Runnable {
    */
   public long getSimulationTimeMillis() {
     return currentSimulationTime / MILLISECOND;
+  }
+
+  /**
+   * Return the actual time value corresponding to an argument which
+   * is a simulation time value in microseconds.
+   *
+   * @return Actual time (microseconds)
+   */
+  public long convertSimTimeToActualTime(long simTime) {
+    return simTime + lastStartTime * 1000;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


joakimeriksson:
Is there a reason to not use the method:  `long getSimulationTime()` in Simulator to get the time in simulated time instead och changing variables and adding extra functions in Simulator?

yatch:
Yes, I thought it would be better to make timestamps aligned to actual time somewhat, specially for the case where Cooja is used as an emulator which interacts with the outer world.

A value returned by `getSimulationTime()` represents time since the simulation starts. This value is interpreted by Wireshark as time since January 1, 1970 00:00:00 UTC. 
https://wiki.wireshark.org/Development/LibpcapFileFormat#Record_.28Packet.29_Header

Here is an example of using `getSimulationTime()` to record timestamps. In this example. "UTC Date and Time of Day" is selected in the "Time Display Format" menu of Wireshark.

relative

The newly introduced variable holds the actual time value, in milliseconds since January 1, 1970 00:00:00 UTC, when the simulation starts. The extra method returns the value that is used as the base time to make the actual time values for the timestamps of captured frames.

The reason why `conn.startTime` is used instead of `getSimulationTime()` in the change of RadioLogger.java is that I saw `conn.startTime` is used for the similar purpose at line 224 in the Java file.

joakimeriksson:
Ok! Maybe we should then introduce something in Simulator that corresponds to System.currentTimeMillis() that would make use of the start time variable and return the startTime + simulation time? - That would make it a more generic feature that could be re-used for other purposes. Otherwise your fix is better than the current PCAP timestamp generator and this might be done in later cleanups. So anyway you choose do it I would say it is an ok fix!

yatch:
Your idea sounds good to me; I tried making a few changes:

- introduce a more general method
- use the new method to generate a frame timestamp in actual time
- remove the old method which just returns `lastStartTime`

yatch:
Squashed a series of commits into one commit...

joakimeriksson:
Great! - Looks good now! :+1: from me! (but there is need for some merger to merge it - hopefully someone can do that quite quickly since it is not a large fix but a rather small and easy to review one!).

yatch:
@joakimeriksson Thank you for your review and comments! 

I amended the commit log message hopefully to make more sense, by the way...

simonduq:
I know the problem indeed exists as it was already reported to me before. But am not familiar with Cooja source. @fros4943, much appreciated if you could take a look!

yatch:
Could anyone review this PR, please...?

joakimeriksson:
Since all is green and I did review (one of the COOJA experts I would say ;-) I think it should be fine merging.

yatch:
@joakimeriksson Thanks for the reaction :-)

alignan:
:+1: from me, but another thumb is required

simonduq:
👍

yatch:
Thank you all for the quick responses in spite of weekend!!

PR #1508, opened 2016-02-10
RPL non storing mode

simonduq:
This PR implements the RPL storing mode of operation, as per RFC 6550 and RFC 6554 (compressed source routing header). It is based on https://github.com/contiki-os/contiki/pull/1403, hence the very large diff. The actual non storing mode boils down to https://github.com/contiki-os/contiki/commit/670e259c9ef7a22f3aadf85642622ea84e635576.

A nice thing with non-storing is that it is immune to distributed inconsistencies in the routing state (storing mode suffers a lot form such inconsistencies in dynamic settings).

All the RPL regression tests are now also executed in non storing mode (see new test suite `rpl-non-storing`). The code in this branch is tested in a 25 node (jn516x) testbed, with TSCH (minimal schedule) and a simple downwards traffic scenario (the root sends to a random node every 2 seconds) in a 48h run, got 99.6% delivery.


yatch:
This is an interesting PR.

I'm also implementing the Non-Storing mode on Contiki; an interesting thing is that there are some behavioral differences between yours and mine :-) 

For example, I use Address Registration Option defined in RFC 6775 in order to make parents learn their children's global addresses. Parent nodes need the addresses to process routing headers. On the other hand, you don't use the option. It looks that parent nodes convert the next-hop global address of a forwarding packet to an inferred link-local address by modifying its prefix part. Each implementer makes a different decision for the same thing.

I'm going to play with your implementation; I'll let you know if I find something.

simonduq:
Interesting! I think what you are implementing with RFC 6775 is the correct behavior. We should keep the routing module enabled even with non-storing, because the RH hops are global addresses and not guaranteed to be on link. We should at every RH hop lookup the routing table and find a next hop. The Address Registration Option looks to me like the right way to populate the table with our direct children.

What I do is a simplifying assumption, which works without 6LoWPAN-ND, and allows more space savings as it doesn't require any routing entry and even disables the routing module altogether. Note that this implementation could *very* easily be extended with your mechanism.

Are you using your own implementation of 6LoWPAN-ND btw or is it @laurentderu's?

yatch:
In my implementation, a parent node holds its children's global addresses in its neighbor table. Then, the parent looks up its neighbor table to determine if the next-hop address of a forwarding packet is on-link. This means, we consume memory for neighbor entries other than routing entries. And, the neighbor entries learned by NS with ARO should not be removed until their expiration.

I'm using my own 6LoWPAN-ND implementation. In fact, I've just added only the Address Registration part of it.

simonduq:
I see, makes sense!

mcr:
Yasuyuki Tanaka <notifications@github.com> wrote:
    > I'm also implementing the Non-Storing mode on Contiki; an interesting
    > thing is that there are some behavioral differences between yours and
    > mine :-)

    > For example, I use Address Registration Option defined in RFC 6775 in
    > order to make parents learn their children's global addresses. Parent

Are you saying that your leafs do not participate in RPL, and do not send
DAOs with their addresses?

--

simonduq:
The thing is that in non-storing, nodes send unicast DAOs directly to the root's Global IPv6, not to their parent's Link-Local address. Nodes end up *not* knowing the Global address of their children.

yatch:
@mcr Leaf nodes participate in a RPL network and send their DAOs. However, in Non-Storing mode, the DAOs are not destined for their parents. As @simonduq said, instead, they are destined for the root node of the network. The parents have no chance to know global addresses of their children through DAOs; they just forward DAOs received from their children at the L3 level.

> Are you saying that your leafs do not participate in RPL, and do not send
> DAOs with their addresses?

mcr:
Simon Duquennoy <notifications@github.com> wrote:
    > The thing is that in non-storing, nodes send unicast DAOs directly to
    > the root's Global IPv6, not to their parent's Link-Local address.
    > Nodes end up not knowing the Global address of their children.

I forgot that part. Blush.
Is the issue that the node needs to map the GUA to the LL address?

--

simonduq:
@mcr yes, exactly.

yatch:
I noticed the PIO in a DIO had not a full IPv6 address but a 64-byte prefix. As per RFC 6550, a PIO under the Non-Storing mode is supposed to have the full IPv6 address of a DIO sender with the 'R' flag set on.

> (In 6.7.10 of RFC 6550)
> A non-storing node SHOULD refrain from advertising a prefix till it owns an address of that prefix,
> and then it SHOULD advertise its full address in this field, with the 'R' flag set.

May I try to write a patch for this? Or, is it planned to make it fixed?

simonduq:
That is a good point. However, Contiki currently doesn't store neighbor global addresses, so I'm afraid the info in the PIO will be discarded anyway. I'd as long as we don't discover global addresses, we don't need the change you suggest (especially as this is a SHOULD, not MUST).

Once we have RFC 6775 this will be more needed IMO.

yatch:
Aha, I understood your thoughts behind the implementation. That sounds reasonable.

joakimeriksson:
@simonduq do we have plans for RFC 6775? I have not really seen any significant implementations in use yet?! (Thread skipped it completely - but they skipped other things also).

simonduq:
I think @laurentderu has an implementation he was planning on contributing some day

alignan:
:+1: once conflicts have been solved

simonduq:
Great. Will incorporate minor improvements I have on my private branch and resolve conflicts. Let me point out this now extensively tested through 100s experiments in testbeds up to 350 nodes :)

alignan:
Cool! looking forward!

simonduq:
This one is ready too now!

joakimeriksson:
Very happy to see this coming in! I look forward to test in some Yanzi systems in the future!

core/net/rpl/rpl-icmp6.c
```diff
@@ -885,13 +887,148 @@ dao_input(void)
...
+           instance_id);
+    goto discard;
+  }
+
+  if(instance->mop != RPL_MOP_NON_STORING) {
+    if(RPL_IS_STORING(instance)) {
+      dao_input_storing();
+    }
+  } else {
+    if(RPL_IS_NON_STORING(instance)) {
```
nvt: Do you need to check instance->mop both in the top-level <code>if</code> statement and in the nested if statement? Perhaps it would also be good to add a debug statement if we receive an unsupported MOP that is discarded.

core/net/rpl/rpl-ns.c
```diff
@@ -0,0 +1,211 @@
...
+/* Total number of nodes */
+static int num_nodes;
+
+/* Every known node in the network */
+LIST(nodelist);
+MEMB(nodememb, rpl_ns_node_t, RPL_NS_LINK_NUM);
+
+/*---------------------------------------------------------------------------*/
+int
+rpl_ns_num_nodes()
```
nvt: <code>void</code> is missing in the parameter part.

core/net/rpl/rpl-ns.c
```diff
@@ -0,0 +1,211 @@
...
+/*---------------------------------------------------------------------------*/
+void
+rpl_ns_init()
+{
+  num_nodes = 0;
+  memb_init(&nodememb);
+  list_init(nodelist);
+}
+/*---------------------------------------------------------------------------*/
+rpl_ns_node_t *rpl_ns_node_head()
```
nvt: Missing line break between return type and function name here and in the following function.

examples/ipv6/rpl-udp/project-conf.h
```diff
@@ -1,4 +1,5 @@
 /*
+<<<<<<< HEAD
```
nvt: This line was accidentally included.

regression-tests/23-rpl-non-storing/code/receiver-node.c
```diff
@@ -0,0 +1,131 @@
...
+}
+/*---------------------------------------------------------------------------*/
+static uip_ipaddr_t *
+set_global_address(void)
+{
+  static uip_ipaddr_t ipaddr;
+  int i;
+  uint8_t state;
+
+  uip_ip6addr(&ipaddr, 0xaaaa, 0, 0, 0, 0, 0, 0, 0);
```
nvt: It would be nice to use <code>UIP_DS6_DEFAULT_PREFIX</code> also in the tests instead of 0xaaaa.

nvt:
Nice contribution! I just had a few minor comments about some specific source lines.

simonduq:
Fantastic, thanks, will address!

core/net/rpl/rpl-icmp6.c
simonduq: Good point, there was one too many if.
As for checking for unsupported mop, this is done in `rpl_join_instance`

simonduq:
Comments addressed @nvt

simonduq:
With also an additional feature https://github.com/contiki-os/contiki/pull/1508/commits/535ff25da2a38fcc63b8e4513c5ec09e492b5c36

nvt:
:+1:

PR #1512, opened 2016-02-12
Fix z1-websense example

alignan:
This PR fix the broken `z1-websense` example and its cooja simulation


alignan:
:+1:

greg-king5:
How is the makefile in this example able to work?  The recipe in its last rule is on the line with the prerequisites.

PR #1513, opened 2016-02-12
galileo: build_newlib.sh: Exit on configuration error

mdlemay:
This patch adds a check for the result of the newlib configure command
and immediately exits the build_newlib.sh script if configure fails.

### DIFF
--- a/platform/galileo/bsp/libc/build_newlib.sh
+++ b/platform/galileo/bsp/libc/build_newlib.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 JOBS=5
 TARGET=i586-elf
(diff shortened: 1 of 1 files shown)


mdlemay:
@jeez, @aguedes: Can you please review this?

aguedes:
Hi @mdlemay, I think we should run 'set -e' at the beginning of build_newlib.sh so the script immediately exits in case any command fails, not only the configure command.

This is what we do in build_grub.sh, by the way.

mdlemay:
Good idea, @aguedes.  How does this new version look?

aguedes:
Yep, :+1: 

Thanks, Michael!

nvt:
:+1:

PR #1516, opened 2016-02-13
Weather meter: replaced rtimer by ctimer as it was breaking contikimac

alignan:
This PR includes a simple change from `rtimer` to `ctimer` as it was breaking up `ContikiMAC` when enabled, as `rtimer` is reserved.


alignan:
:+1:

PR #1518, opened 2016-02-15
Removed PM10 code lines from adc-sensors driver

ajlozano:
This pull request removes code lines related to PM10 from `adc-sensors` driver. The PM10 driver is implemented with her own files and only access to `adc-zoul` driver.


alignan:
:+1:

PR #1519, opened 2016-02-16
CoAP: fix possible concurrency of multiple responses

locomuco:
This PR fixes a concurrecyy problem, when two CoAP responses are arriving at the same time.
In the current version of the code, the yielded thread for a blocking request is polled through a callback.

The error scenario is as below:
1. CoAP request (1)
2. CoAP request (2)
3. CoAP response (1)
4. coap_blocking_request_callback (1)
5. CoAP response (2) -> overrides uip buf, where CoAP payload is stored
6. invoke yielded thread
7. call CoAP response (1) handler with wrong payload 

(in case of block 2 transfers, the block numbers are also affected and were moved to the state structure)

### DIFF
--- a/apps/er-coap/er-coap-engine.c
+++ b/apps/er-coap/er-coap-engine.c
@@ -361,9 +361,30 @@ PROCESS_THREAD(coap_engine, ev, data)
 void
 coap_blocking_request_callback(void *callback_data, void *response)
 {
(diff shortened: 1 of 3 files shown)


simonduq:
@mkovatsc , much appreciated if you could take a look!

PR #1520, opened 2016-02-18
Added enabling up to 6 ADC channels and fixed adc-zoul driver problem. 

ajlozano:
Added enabling up to 6 ADC channels for RE-Mote and Firefly platforms.
Solved problem with `enabled_channels` variable in adc-zoul driver.

### DIFF
--- a/platform/zoul/dev/adc-sensors.c
+++ b/platform/zoul/dev/adc-sensors.c
@@ -175,7 +175,7 @@ configure(int type, int value)
     return ADC_WRAPPER_ERROR;
   }
 
-  if((value < 0x01) || (value > 0x07) || (value == BUTTON_USER_PIN)) {
+  if((value < 0x01) || (value > 0x07) || ((value == BUTTON_USER_PIN) && (ADC_SENSORS_ADC6_PIN < 0))) {
     PRINTF("ADC sensors: invalid pin value, (PA0-PA1, PA3) are reserved\n");
     return ADC_WRAPPER_ERROR;
   }
--- a/platform/zoul/dev/zoul-sensors.c
+++ b/platform/zoul/dev/zoul-sensors.c
@@ -47,7 +47,12 @@
 #include <string.h>
 /*---------------------------------------------------------------------------*/
 /** \brief Exports global symbols for the sensor API */
-SENSORS(&button_sensor, &vdd3_sensor, &cc2538_temp_sensor);
+SENSORS(&vdd3_sensor,
+#if PLATFORM_HAS_BUTTON
+  &button_sensor,
+#endif
+  &cc2538_temp_sensor
+);
 /*---------------------------------------------------------------------------*/
 /**
  * @}
--- a/platform/zoul/contiki-main.c
+++ b/platform/zoul/contiki-main.c
@@ -218,9 +218,9 @@ main(void)
 #endif /* NETSTACK_CONF_WITH_IPV6 */
 
   process_start(&sensors_process, NULL);
-
+#if PLATFOM_HAS_BUTTON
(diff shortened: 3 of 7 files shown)


examples/zolertia/zoul/rtcc/test-rtcc.c
```diff
@@ -84,17 +84,16 @@ rtcc_interrupt_callback(uint8_t value)
...
-
+ 
   /* Alternatively for test only, undefine DATE and define on your own as
    * #define DATE "07 06 12 15 16 00 00"
    * Also note that if you restart the node at a given time, it will use the
    * already defined DATE, so if you want to update the device date/time you
    * need to reflash the node
    */
-
+
```
alignan: This is not related to the PR, revert

alignan:
Mmm there are changes I don't know why are being shown as deleted, let's take a look tomorrow at this.

g-oikonomou:
I think 6x ADC is more circumstancial than the user button, so in that context I would prefer assuming button and explicitly enabling ADC6, than explicitly enabling the button which seems to be what we're proposing here.

alignan:
Yup, we are refactoring

alignan:
:+1:

platform/zoul/dev/adc-sensors.c
```diff
@@ -175,7 +175,7 @@ configure(int type, int value)
     return ADC_WRAPPER_ERROR;
   }
 
-  if((value < 0x01) || (value > 0x07) || (value == BUTTON_USER_PIN)) {
+  if((value < 0x01) || (value > 0x07) || ((value == BUTTON_USER_PIN)&&(ADC_SENSORS_ADC6_PIN < 0))) {
```
g-oikonomou: Code style

platform/zoul/dev/zoul-sensors.c
```diff
@@ -47,7 +47,12 @@
 #include <string.h>
 /*---------------------------------------------------------------------------*/
 /** \brief Exports global symbols for the sensor API */
-SENSORS(&button_sensor, &vdd3_sensor, &cc2538_temp_sensor);
+SENSORS(&vdd3_sensor, &cc2538_temp_sensor
+
+#if PLATFORM_HAS_BUTTON
+, &button_sensor
```
g-oikonomou: That's a little ugly :( How about

```
SENSORS(&vdd3_sensor,
#if PLATFORM_HAS_BUTTON
  &button_sensor,
#endif
  &cc2538_temp_sensor
);
```

g-oikonomou:
I see the point (as discussed already offline). But then again, I think you are overcomplicating things for yourselves by providing a user configuration for the button. Why not simply just assume button is enabled (not configurable), except if ADC6 gets enabled (in which case you automatically disable it). 

Then you can simply document somewhere "You can have ADC6, but if you do then no button". This would perhaps simplify the `#if` logic.

Just a thought.

platform/zoul/dev/zoul-sensors.c
alignan: Agree

alignan:
At first I thought about keeping it simple in the case of the `RE-Mote`, but we also have other boards based on the `zoul` without an user button, so we preferred to generalize.

tools/cc2538-bsl
```diff
@@ -1 +1 @@
-Subproject commit 4340542370d3e319d692f9f969c5683231fce438
+Subproject commit 6185d8b16f10047053fd57000a552340273d7d85
```
alignan: @ajlozano  Remove this leftover, remember to make a `git diff` before pushing

alignan:
:+1:

PR #1531, opened 2016-02-23
Fixed numerous bugs in CC26xx-CC13xx lpm

### LINKED ISSUE #1509: cc2650 lpm.c update from January 11th causes sensortag to crash repeatedly
Hi,
The update to the cc26xx-cc13xx cpu files in commit number 707353f made changes to the lpm.c file in cpu/cc26xx-cc13xx. These changes have causes my chip to power cycle when powered on (i.e. the code that I had running fine on it no longer runs as expected). I reverted to commit 7b95309 and my code works fine, (as lpm.c has not been altered in that commit). The alterations to rf-core.c/.h ieee-mode.c and soc-rtc.c do not affect my code.
I tried a simpler program with the updated lpm.c file, one that just responded to button pushes and that works fine with the new lpm.c file. But my current program does not and I think that is because I am making use of the RTC on the cc2650 in my current code. I narrowed down the source of the error to lines 275-281 in lpm.c - where the next event is scheduled by calling soc_rtc_schedule_one_shot(...).
So I guess the issue is: is this change to lpm.c absolutely necessary? and has anyone else checked the latest lpm.c file and tried using it with an app that makes use of the RTC on the SensorTag?
Thanks for any thoughts/assistance provided,
Michael

bkozak-scanimetrics:
Bug fixes include:
- keep interrupts disabled during lpm_sleep() so that we don't miss any interrupts we may have been expecting.
- check that the pending etimer isn't already expired (and don't sleep at all if it is)
- check that the about-to-be scheduled rtimer wakeup is neither too far into  the future nor too close into the future (or even in the past) before actually setting the interrupt (should fix #1509); If the time is out of bounds we use a default min or max value instead.
- Correctly handle LPM_MODE_MAX_SUPPORTED set to zero (and added a macro for the zero value) so that sleeping can be disabled altogether
- If no etimer is set, we specify a wakeup time which is reasonably far into the future instead of setting none at all (this will save on power consumption whenever no etimers are set).

Also did a bit of re-factoring in that some long functions were broken into multiple functions.

### DIFF
--- a/cpu/cc26xx-cc13xx/lpm.c
+++ b/cpu/cc26xx-cc13xx/lpm.c
@@ -75,7 +75,12 @@ LIST(modules_list);
  * Don't consider standby mode if the next AON RTC event is scheduled to fire
  * in less than STANDBY_MIN_DURATION rtimer ticks
  */
-#define STANDBY_MIN_DURATION (RTIMER_SECOND >> 11)
+#define STANDBY_MIN_DURATION  (RTIMER_SECOND >> 11)
+#define MINIMAL_SAFE_SCHEDUAL 8u
+#define MAX_SLEEP_TIME        RTIMER_SECOND
+#define DEFAULT_SLEEP_TIME    RTIMER_SECOND
+/*---------------------------------------------------------------------------*/
+#define CLK_TO_RT(c) ((c) * (RTIMER_SECOND / CLOCK_SECOND))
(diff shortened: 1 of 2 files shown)


g-oikonomou:
These look like reasonable changes at first glance, need to do some testing first.

g-oikonomou:
Thoroughly tested over here with @atiselsts 

Thanks!

Issue #1533, opened 2016-02-25
[Zoul Remote] Upload failed: Timeout waiting for ACK/NACK after 'Get ChipID (0x28)'


PalmaITEM:
Hi,
I'm using Instant Contiki but I've cloned the contiki-os repo since I am following the tutorial "IoT in five days".

I have no problems with Z1 but with Remote I cannot upload any code, not even the demo. Heres the output:
```
make zoul-demo.upload
using saved target 'soul'
  CC        ../../../cpu/cc2538/./ieee-addr.c
  CC        ../../../cpu/cc2538/cc2538.lds
  CC        ../../../cpu/cc2538/./startup-gcc.c
  CC        zoul-demo.c
  LD        zoul-demo.elf
arm-none-eabi-objcopy -O binary --gap-fill 0xff zoul-demo.elf zoul-demo.bin
python ../../../tools/cc2538-bsl/cc2538-bsl.py -e -w -v -a 0x00202000 zoul-demo.bin
Opening port /dev/ttyUSB0, baud 500000
Reading data from zoul-demo.bin
Cannot auto-detect firmware filetype: Assuming .bin
Connecting to target...
ERROR: Timeout waiting for ACK/NACK after 'Get ChipID (0x28)'
make: *** [zoul-demo.upload] Error 1
rm obj_zoul/startup-gcc.o zoul-demo.co
```
Since I had been able to upload code before, and after trying to understand the erro, I tried an older version of _cc2538-bsl.py_ (version 1.1) and I was able to upload the code. However, when I connect to the serial line to debug nothing really happens.

If I use the old Contiki repo that comes with the Instant Contiki VM, the _remote-demo_ works just fine.

I'm a beginner in Contiki but please let me know if I can help further.

alignan:
Have you tested the instructions from this link?

https://github.com/Zolertia/Resources/wiki/Program-the-Zolertia-platforms

PalmaITEM:
Hi,
Thank you for the pointer! I was missing intelhex.

g-oikonomou:
I find it very difficult to understand how lack of IntelHex could cause the error you were getting. Can we try to clarify this, since if it was indeed the cause then the BSL could have a bug.

PalmaITEM:
As I said I not very experienced with Contiki. I'll try to explain better what happened and hopefully it'll be helpful.

[...]

PR #1534, opened 2016-02-29
Update of cc2538-bsl to 4340542

JelmerT:
This PR updates the CC2538-bsl submodule.

Main changes are fix for contiki-os/contiki#1533 and an added option to invert the RTS and DTR lines when auto starting the bootloader.


alignan:
:+1:

PR #1535, opened 2016-03-01
Fix warning about missing strcasecmp definition

bkozak-scanimetrics:
On platforms which do not define the macro ```__SDCC``` the ```strcasecmp``` function may not have a definition (this happens at least on CC26xx). The prototype for this function is now defined if ```__SDCC``` is false.

### DIFF
--- a/core/net/ip/resolv.c
+++ b/core/net/ip/resolv.c
@@ -111,6 +111,9 @@ strcasecmp(const char *s1, const char *s2)
   /* TODO: Add case support! */
   return strcmp(s1, s2);
 }
+#else
+int strcasecmp(const char *s1, const char *s2);
+int strncasecmp(const char *s1, const char *s2, size_t n);
 #endif /* __SDCC */
 
 #define UIP_UDP_BUF ((struct uip_udpip_hdr *)&uip_buf[UIP_LLH_LEN])


simonduq:
What would you think about including `#include <strings.h>` (plural) instead (along with other includes on top of the file). The existing inclusion of `string.h` might even be possible to remove?

bkozak-scanimetrics:
> What would you think about including #include <strings.h> (plural) instead

I had tried it already, It fails on the Travis build. Some platforms, even some platforms for which ```__SDCC``` is false, don't have a strings.h

g-oikonomou:
What version of arm-gcc are you using? I seem to remember having some problems with something similar when I upgraded between versions (one version was throwing a warning/error, whereas a different version was not).

bkozak-scanimetrics:
```bash
$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (4.8.2-14ubuntu1+6) 4.8.2
```

g-oikonomou:
I'm just playing around now. Inclusion of string.h is sufficient with arm-gcc v4.9 and v5.2.

bkozak-scanimetrics:
In gcc 4.8 string.h will not define strcasecmp if ```__STRICT_ANSI__``` is defined (and it is defined for ``-std=c99``).

simonduq:
then this looks like a reasonable fix/workaround to me :+1:

alignan:
Ready to merge @simonduq @g-oikonomou ?

simonduq:
yes for me

g-oikonomou:
Pass

PR #1539, opened 2016-03-03
added output function in NETSTACK network to enable IP packet filters, etc.

joakimeriksson:
Adding an output function to NETSTACK network adapted for IPv6 networking. It enables a full replacement of the network layer so that packet logging, packet filtering, firewalls, dual-radio trick can be easily implemented.

This is a yet failing pull request (some issues with Z1 builds) and will need some fixes in native platform (for the hooks into the linux / windows stack). I will take a look at fixing these things - but puts the PR in to get some comments on the idea of adding the output function.

### DIFF
--- a/core/net/ip/tcpip.c
+++ b/core/net/ip/tcpip.c
@@ -109,25 +109,18 @@ enum {
 /* Called on IP packet output. */
 #if NETSTACK_CONF_WITH_IPV6
 
-static uint8_t (* outputfunc)(const uip_lladdr_t *a);
-
 uint8_t
 tcpip_output(const uip_lladdr_t *a)
 {
   int ret;
-  if(outputfunc != NULL) {
-    ret = outputfunc(a);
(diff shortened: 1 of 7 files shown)


joakimeriksson:
Did a quick attempt at fixing the minimal-net tapdev and wpcapdev code but only for IPv6. I think NETSTACK network should also be defined for IPv4 fully - but possibly with different signatures on the output function. Or could we use the same signature? (e.g. a lladdr)?

simonduq:
you might want to rebase now with https://github.com/contiki-os/contiki/pull/1543 in

joakimeriksson:
Yes, will try that - but I am also a bit interested if this is a good way forward or if this is non-interesting. If this is interesting I would be ready to try to fix all the issues and maybe also do the same for the IPv4 part of the stack to avoid getting the two different (the output func is still in IPv4 even after this PR).

PR #1540, opened 2016-03-04
Minor TSCH fixes

simonduq:
Do not merge yet, this is not tested on all platforms.

Explicitly turn off the radio after ACK Tx, may be needed on some platforms (most platform would turn it off anyway as it was off before Tx). Also fixes `TSCH_PACKET_MAX_LEN` for platforms with smaller `PACKETBUF_SIZE`.

Added another fix: we now flush all queues and reset backoff exponents whenever re-associating.

### DIFF
--- a/core/net/mac/tsch/tsch-packet.h
+++ b/core/net/mac/tsch/tsch-packet.h
@@ -36,6 +36,7 @@
 /********** Includes **********/
 
 #include "contiki.h"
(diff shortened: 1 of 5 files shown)


thomas-ha:
Works perfect here on CC2538 platform.

simonduq:
Got more testing on my side too, merging

PR #1548, opened 2016-03-10
uip6, icmp6 and ext header handling fixes: traceroute now working!

simonduq:
This is a collection of fixes required to get traceroute to work.
There were several connected problems:
* icmp error messages should return the original incoming packet, but they returned an altered version due to extension header handling and buffer length bugs
* the last hop never returned a "port unreachable" message


joakimeriksson:
I think this is a great fix - I have also wondered quite a bunch of times why trace route did not work!
:+1:

alignan:
:+1:

nvt:
:+1:

PR #1549, opened 2016-03-10
Asyncronous / Interrupt Driven ContikiMAC

bkozak-scanimetrics:
This pull request adds support for a new version of ContikiMAC which can perform strobe sending and packet reception in the background using interrupts to drive the state transitions. The new driver also contains several other improvements which I will discuss below. The code currently only works on the CC2650 since it relies on the use of a new asynchronous radio driver; platform support can be added by implementing the asynchronous radio driver interface.

In order for this code to operate correctly, the pull-requests #1531 and #1290 must be merged first. For the sake of convenience I have created a branch which can be used as is at https://github.com/bkozak-scanimetrics/contiki/tree/contikimac_async_test
I apologize to the maintainers for making a pull request before this is ready to be merged but I want to start getting some eyeballs on this now since this is a fairly major change. If you think it is appropriate, I could merge all of these PRs into a single request.

To build a project with asynchronous ContikiMAC support, define the macros  ```NETSTACK_CONF_USE_RFASYNC``` and ```RTIMER_CONF_MULTIPLE_ACCESS``` to true.

I have been using the code for a couple months now and everything has been completely stable. I have not found any major issues.

Major improvements from using the asynchronous driver include:
- Strobes can be sent in the background, this makes it much more feasible to operate ContikiMAC with a lower channel check rate since we don't have to block until all strobes are sent.
- Lower transmit and receive power consumption through better management of the transceiver
- ContikiMAC phase optimization now works well with more accurate traffic timing
- Higher maximum TX throughput
- Much better performance for burst packet traffic patterns.
- Better handling of low level radio / driver errors; the new ContikiMAC driver will generally check all error codes and try to keep operating even in the face of errors.

This new driver does, however, use more memory than the original.

I believe that there is also lots of room for improvement on this code and this it would be an excellent jumping off point for further improvement of ContikiMAC phase optimization, advanced hardware features (such as DMA) etc.

Thank you all in advance for your feedback.


g-oikonomou:
I haven't even started looking at it, but the subject did catch my attention! High prio IMHO

bkozak-scanimetrics:
Fixed some minor errors causing the Travis build to fail on platforms other than the CC2650:
- Forgot to include <stdbool.h> in radio-async.h
- Misnamed one of the arguments in the doxygen comment for ```PT_YIELD_ON``` in pt.h

greg-king5:
The commit message and the doxygen description of `PT_YIELD_ON()` are backwards.  That macro executes first, yields second.

core/net/mac/contikimac/contikimac-async.c
```diff
@@ -0,0 +1,1438 @@
...
+#define CYCLE_MIN_SETUP RTIMER_MINIMAL_SAFE_SCHEDULE
+
+#define RTIMER_INTER_PACKET_DEADLINE (RTIMER_ARCH_SECOND / 32)
+
+#define YIELD_ON_OR_GOTO(pt, stmnt, label) \
+  PT_YIELD_ON(pt, if((stmnt)) { goto label; })
+#define STROBE_YIELD_ON_OR_GOTO(stmnt, label) \
+  YIELD_ON_OR_GOTO(&strobe_state.pt, stmnt, label)
+#define CCA_YIELD_ON_OR_GOTO(stmnt, label) \
+  YIELD_ON_OR_GOTO(&cca_state.pt, stmnt, label)
```
greg-king5: Those macro names might be confusing.  Those macroes actually implement `GOTO_ON_OR_YIELD()`.

core/sys/pt.h
```diff
@@ -296,6 +296,34 @@ struct pt {
...
+ */
+#define PT_YIELD_ON(pt,stmnt)                         \
+  do{                                                 \
+    PT_YIELD_FLAG = 0;                                \
+    LC_SET((pt)->lc);                                 \
+    if(PT_YIELD_FLAG == 0) {                          \
+      { stmnt ;}                                      \
+      return PT_YIELDED;                              \
+    }                                                 \
+  }while(0)
```
greg-king5: I'm confused.  How is that macro better than writing simply:
```C
stmnt;
PT_YIELD(pt);
```

Why should the code have lines such as
```C
PT_YIELD_ON(&cca_state.pt, call_off(cmd_cb_cca_sm));
```
instead of
```C
call_off(cmd_cb_cca_sm);
PT_YIELD(&cca_state.pt);
```
I don't see any significant difference between them -- even if `stmnt;` is recursive.

In other words, why should that macro be added to Contiki?  (Its name implies something that isn't true; there is no conditional relationship between the statement and the yield.)

bkozak-scanimetrics: The difference is when the "stmnt" causes re-entry back into the proto-thread. With PT_YIELD_ON the protothread will continue execution after the macro on re-entry and "stmnt" will not be run twice.  If we instead execute "stmnt" before a regular PT_YIElD then on re-entry the protothread starts execution at the point where it was previously suspended and stmnt executes again.

bkozak-scanimetrics: Also note that functions such as "call_off", although not recursive, will cause re-entry to the protothread indirectly by triggering an interrupt which calls into the protothread.

greg-king5: Now, I understand it.

I still don't like the name.  `PT_YIELD_FOR()`, while not perfect, seems, to me, to be more appropriate.

core/net/mac/contikimac/contikimac-async.c
greg-king5: Nevermind.  I realize, now, that the functions, on which those macroes are used, return an error code (0 means OK, 1 means failure), not a boolean value.  So, those names are good.

core/sys/pt.h
bkozak-scanimetrics: > I still don't like the name. ```PT_YIELD_FOR()```, while not perfect, seems, to me, to be more appropriate.

```PT_YIELD_ON``` makes sense to me because the macro yields 'on' the given statement such that the proto-thread's execution 'pointer' is made to point just after the statement.

**Comment by bkozak-scanimetrics** [comment]
I've been working on debugging the latest Travis failure on this branch.

I've found that the memory overflow on Sky builds can be avoided by wrapping the entire ```phase.c``` source file in a ```#if WITH_PHASE_OPTIMIZATION```. This is the only source file which could have significantly different code generation on any platform other than the 2650 so this makes sense.

The other build failure is caused because the compiler for the 8051 can't seem to handle directly passing a structure to a function or maybe just not in this specific case for some reason.

Correct me if I'm wrong, but I believe that passing a struct to a function is not a violation of any C standard in which case this should be considered a compiler bug. 

Given the circumstances, I believe that the best way to fix this would be to just prevent the 8051 compiler from seeing these function definitions at all (probably using a macro). I would certainly prefer this solution over amending all of the affected functions so that they don't take structs directly as arguments.

**Comment by bkozak-scanimetrics** [comment]
I've fixed all of the issues now.

Travis should be green when it completes.

**Comment by kkrentz** [comment]
Thanks for sharing this code. I am in the course of implementing an asynchronous radio driver for CC2538 SoCs.

**Comment by bkozak-scanimetrics** [comment]
That's great to hear @kkrentz. Good luck to you.

**Comment by bkozak-scanimetrics** [comment]
Added a commit to increase the watchdog timeout. The watchdog will occasionally trigger if using asynchronous ContikiMAC while sending and receiving alot of packets at once. With a higher timeout this doesn't happen any-more.

Also rebased on the latest master and tested all of the changes.

**Comment by kkrentz** [comment]
Unfortunately, this implementation turned out to be overly complex for my purposes. I just need some clean implementation of ContikiMAC for evaluating some security enhancements. So I actually now have an own implementation of ContikiMAC, too. Probably I can also make it open-source.

File: cpu/cc26xx-cc13xx/radio-async-arch.h [hunk 3]
```diff
@@ -0,0 +1,38 @@
...
+ */
+#ifndef RADIO_ASYNC_ARCH_H
+#define RADIO_ASYNC_ARCH_H
+
+/* Values determined experimentally */
+#define RFASYNC_RX_STABLE_TIME 18
+#define RFASYNC_ON_TIME        36
+#define RFASYNC_OFF_TIME       12
+
+#endif /* RADIO_ASYNC_ARCH_H */
```
alignan: Newline here

cpu/cc26xx-cc13xx/rf-core/rf-async.h
```diff
@@ -0,0 +1,188 @@
...
+#define rf_async_is_accessible rf_core_is_accessible
+#define rf_async_set_modesel   rf_core_set_modesel
+#define rf_async_init_radio_op rf_core_init_radio_op
+/*---------------------------------------------------------------------------*/
+#endif /* RF_ASYNC_H_ */
+/*---------------------------------------------------------------------------*/
+/**
+ * @}
+ * @}
+ */
```
alignan: Newline here also

cpu/cc26xx-cc13xx/radio-async-arch.h
bkozak-scanimetrics: Thanks, will fix soon.

alignan:
I think this is a great contribution, generally speaking I thought this was a new MAC one could use alternatively i.e to `ContikiMAC`, but it seems it touches other modules as well.  Perhaps it would be easier to review and merge otherwise.

bkozak-scanimetrics:
> Perhaps it would be easier to review and merge otherwise.

Sorry, not entirely clear on what you mean by that. Anything you need me to do here?

alignan:
Just thinking aloud, for me is difficult to review as it seems to modify cc2650 specifics, rather than being a separate MAC implementation, but I understand is a WIP and probably most of the proposed changes are required.

bkozak-scanimetrics:
Newlines now added and I've rebased on latest master.

bkozak-scanimetrics:
I've now also updated the branch at https://github.com/bkozak-scanimetrics/contiki/tree/contikimac_async_test to base on latest master and removed some of the temporary hacks I had added to the CC2650 lpm.c while waiting for #1531 to be merged.

In order to do this I also had to update #1290.

Also, just a reminder that this PR can't actually do anything without #1290.

kkrentz:
I have uploaded my reimplementation of ContikiMAC [here](https://github.com/kkrentz/contiki/blob/potr/core/net/mac/contikimac/secmac.c). It also works asynchronously, but lacks support for bursts and phase-lock at the moment. On the other hand, it includes a mechanism for mitigating denial-of-sleep attacks. This is part of a paper we are going to present at ARES 2016.

bkozak-scanimetrics:
> ... it includes a mechanism for mitigating denial-of-sleep attacks.

Now that's __very__ interesting. I'll certainly try to look at this when I can find some free time.

core/net/mac/phase.c
```diff
@@ -67,6 +67,12 @@ struct phase_queueitem {
   struct rdc_buf_list *buf_list;
 };
 
+#if PHASE_CONF_DRIFT_CORRECT
+#define PHASE_DRIFT(e) ((e)->drift))
```
niklas88: There's a lone unpaired ")" here

bkozak-scanimetrics:
Fixed unpaired bracket in ```PHASE_DRIFT()``` thanks to @niklas88 .

bkozak-scanimetrics:
I've completed a rebase of this branch onto the latest master. The branch at https://github.com/bkozak-scanimetrics/contiki/tree/contikimac_async_test (includes all of the necessary PRs for this feature to function) has also been updated and seems to be working.

Considering the recent TSCH related additions to the cc26xx/cc13xx radio driver, this was very much a non-trivial operation. I've tested my changes and everything appears to be working but I'm not setup to test TSCH so I'd appreciate if someone who uses TSCH could do a quick check to see if anything is obviously broken.

niklas88:
I have tested both your contiki_async_test branch from yesterday as well as the newly rebased one on the cc2650 SensorTag using a sensor collection application based on Rime's collect module (with phase opt, fast sleep and ContikiMAC headers, 2 Hz RDC). First let me say that I'm seeing a pretty big improvement in power used for listening when compared with mainline (from 150 µW to 90 µW average). I also saw huge improvements in phase optimization stability as stock drops so many packets that it's pretty much not useable. However it looks like there is a power consumption regression with the newest changes. I haven't done a git bisect yet because I just noticed it doing power meassurements again. I have attached two plots of the power over time when sending a sensor state every 5 seconds. The third plot is a zoom in of the high power consumption after sending part that isn't happening in yesterdays version.
contiki_async_first
contiki_async_second
contiki_async_second_zoom

bkozak-scanimetrics:
Thanks for the test data @niklas88!

I hadn't found the time to test power consumption on these recent changes yet so it's not entirely surprising that there is a problem here.

That period of high current after the TX looks to me like it may be the CPU turning on. I hope that this is the case since that's probably an easier problem than anything to-do with the radio. 

So my best guess is that this is the CPU coming out of sleep constantly (perhaps due to a rogue rtimer or etimer) or if not that, then maybe the radio isn't properly shut-down after a TX (but then why would there be that gap between the second set of strobes and the subsequent power pulse train) . I also suspect that, in the case of the first set of strobes, the pulse train is actually intersecting with the stobes (and so maybe this has nothing to do with the TX at all).

Unfortunately, I think that this may be a difficult problem to fix and I haven't even tried to reproduce the issue yet. It could take some time before I have this solved.

bkozak-scanimetrics:
I've spent a little bit of time now to try and reproduce your issue but I haven't seen any problems so far.

I haven't exactly reproduced your setup, however. I'm using a network with only two nodes (one BR and one remote node), IPv6, 8Hz channel check rate and using ICMP requests to force the remote node to transmit. I've tried both with phase optimization and without.  t'll take me some time to try and reproduce your network configuration more closely. 

I've attached my test results for comparison purposes. As you can see, I'm not observing anything like what you've produced here. 

Note:
I believe the final graph below is a broadcast transmission (thus explaining why it wasn't properly phase optimized). 

no-duty-cycle
duty-cycle
phase-opt-1
phase-opt-2

niklas88:
Thinking about it a bit more and especially what you're saying about rogue timers I'm not entirely sure that there couldn't also be some code changes in my code triggering this. I did a bigger refactoring before the first test but there might have been some small change in-between. I will look into this some more either on Monday or Tuesday.
On Jun 16, 2016 21:05, Billy Kozak <notifications@github.com> wrote:I've spent a little bit of time now to try and reproduce your issue but I haven't seen any problems so far.

I haven't exactly reproduced your setup, however. I'm using a network with only two nodes (one BR and one remote node), IPv6, 8Hz channel check rate and using ICMP requests to force the remote node to transmit. I've tried both with phase optimization and without.  t'll take me some time to try and reproduce your network configuration more closely. 

I've attached my test results for comparison purposes. As you can see, I'm not observing anything like what you've produced here. 

Note:
I believe the final graph below is a broadcast transmission (thus explaining why it wasn't properly phase optimized).

niklas88:
I've looked into this again the behavior now seems to persists with the master branch so I think it's actually unrelated.

niklas88:
@bkozak-scanimetrics did you already try combining this with the recent lpm fix? I Realize that hasn't been merged yet but I guess people are just away? I've seen really good results with this code and would love to see this getting merged.

bkozak-scanimetrics:
> @bkozak-scanimetrics did you try combining this with the recent lpm fix? I Realize that hasn't been merged yet but I guess people are just away? I've seen really good results with this code and would love to see this getting merged.

I've been busy with other projects so I haven't had the time to try it.

I've just done a quick rebase. There was a single merge conflict but everything auto-resolved. I haven't tested it other than to compile check it but I think there's a pretty good chance that everything just works.

As usual, I've also updated https://github.com/bkozak-scanimetrics/contiki/tree/contikimac_async_test which includes all of the dependencies that haven't yet been merged into mainline.

bkozak-scanimetrics:
I hadn't realized that the lpm fix wasn't merged yet. I've just done another rebase now that the lpm changes are included. Quite a few merge conflicts this time but each one resolved automatically. 

The lpm fix is now included in this branch and on https://github.com/bkozak-scanimetrics/contiki/tree/contikimac_async_test 

I don't have enough time to test right at this moment but there's a reasonably good chance that nothing has broken.

niklas88:
@bkozak-scanimetrics ok I've just run a very very limited test. It seems to work in principle but with both PHASE_CONF_DRIFT_CORRECT and CONTIKIMAC_CONF_WITH_PHASE_OPTIMIZATION I'm seeing significant packet loss.
I'll investigate further, but I wanted to check with you what your recommended settings are at NETSTACK_CONF_RDC_CHANNEL_CHECK_RATE = 2
maybe I'm simply missing something. Also I'm seeing similar packet loss on master when running with those settings but basically none when running without PHASE_OPTIMIZATION

dak664:
Have you looked at 
https://github.com/contiki-os/contiki/wiki/RDC-Phase-optimization? Look 
at the number of retries as you reduce the guard time. With proper phase 
correction that should drop to zero or one.

niklas88:
@dak664 Thanks for the link I'll take a look. I have removed my earlier comment already because it now works with phase optimization and ContikiMAC Async, I had a typo in the project-conf.h so it was still using normal ContikiMAC which in the standard configuration isn't working well with phase optimization for me.

bkozak-scanimetrics:
Perhaps I should mention now that, unlike with original ContikiMAC implementation, this driver doesn't limit the STROBE_TIME when it has a phase lock. This means that it shouldn't ever drop packets just because the phase optimization algorithm isn't working (instead it just ends up sending more strobes than it otherwise would).

I've fond that, in practice, the phase optimization algorithm is too imprecise and the clock drift between nodes is too quick for it to make sense to limit strobes. Once the phase lock has drifted too much there is a very severe penalty to correct the situation because of the way in which the CSMA algorithm interacts with the logic which detects loss of phase lock. 

In my tests, with strobe time limited (to ```MAX_PHASE_STROBE_TIME``` as in the original driver) most CC2650 devices will start dropping packets after only 10 minutes of dead air because the clocks have already drifted too far apart. 

Unfortunately, I think a much better phase lock algorithm is required before it will make sense to try limiting strobe time.

core/net/mac/contikimac/contikimac-async.c
```diff
@@ -0,0 +1,1438 @@
...
+static void schedule_phase(void);
+static void phase_wait_done(struct rtimer *t, void *ptr);
+
+/* Driver Process */
+PROCESS(contikmac_async_process, "cm async process");
+/*---------------------------------------------------------------------------*/
+/*-------------------------------API FUNCTIONS-------------------------------*/
+/*---------------------------------------------------------------------------*/
+static void
+init(void)
```
niklas88: This is missing phase_init() which leads to the nbr_phase table being used uninitialized which in turn can confuse other users of nbr_tables. For example adding it fixes the WIP Contiki Collect with Neighbor Table problem I was seeing.

niklas88:
Seeing as this PR has been inactive since July and that @bkozak-scanimetrics hasn't been active on GitHub in the last few months it looks like this PR needs to be assumed abandoned.

I'm using ContikiMAC in an internal project and have seen real improvements over mainline ContikiMAC especially concerning packet loss so I'm asking what the chance of merging this would be if I were to take over? I'm assuming the biggest hurdle would be CCxxxx specificity and a general lack of interest in ContikiMAC as of late?

kkrentz:
Meanwhile, our reimplementation of ContikiMAC is also ready to use. The code is [there ](https://github.com/kkrentz/contiki/blob/master/core/net/mac/contikimac/secrdc.c), but only tested on OpenMotes so far. Like in this PR, our reimplementation of ContikiMAC runs asynchronously. In addition, we added security enhancements, some of which will be presented at EWSN 2017. If my existing PRs got merged, I could try to create a PR for this.

alignan:
Thanks for bringing this up.  This PR was quite interesting, specially as ContikiMAC is Contiki's default option to NullRDC.  As the holidays are over it would be a good time to kickstart this once again.

PR #1550, opened 2016-03-11
Port of TSCH for CC2538

thomas-ha:
This pull request is a port of TSCH for the CC2538 cpu.
The cc2538-rf.c driver was extended with the poll mode. Incoming packets are timestamped with the MAC timer. In the off() function the FIFO is flushed only when there is no complete packet in it otherwise the FIFO is flushed before TSCH reads out the packet. The FIFO is flushed at the end of the read() function.

We tested it with 6TiSCH Minimal Schedule and Orchestra with two and up to 17 nodes.


alignan:
Cool! I will test this as well.

Can you please add the `contiki-conf.h` additions to the `zoul` platform as well?

core/net/mac/tsch/tsch-conf.h
```diff
@@ -175,4 +175,11 @@
 #define TSCH_ADAPTIVE_TIMESYNC 0
 #endif
 
+/* HW frame filtering enabled */
+#ifdef TSCH_CONF_HW_FRAME_FILTERING
+#define TSCH_HW_FRAME_FILTERING	TSCH_CONF_HW_FRAME_FILTERING
+#else /* TSCH_CONF_HW_FRAME_FILTERING */
+#define TSCH_HW_FRAME_FILTERING	1
```
simonduq: there are spaces missing here and above

core/net/mac/tsch/tsch-slot-operation.c
```diff
@@ -514,9 +514,11 @@ PT_THREAD(tsch_tx_slot(struct pt *pt, struct rtimer *t))
               TSCH_DEBUG_TX_EVENT();
               NETSTACK_RADIO.off();
 
+#if TSCH_HW_FRAME_FILTERING
               /* Leaving promiscuous mode */
               NETSTACK_RADIO.get_value(RADIO_PARAM_RX_MODE, &radio_rx_mode);
               NETSTACK_RADIO.set_value(RADIO_PARAM_RX_MODE, radio_rx_mode | RADIO_RX_MODE_ADDRESS_FILTER);
+#endif /* TSCH_HW_FRAME_FILTERING */
```
simonduq: could you also add the guards a few lines above (no need to leave promiscuous mode anymore!)

cpu/cc2538/dev/cc2538-rf.c
```diff
@@ -120,6 +120,17 @@ static const uint8_t magic[] = { 0x53, 0x6E, 0x69, 0x66 };      /** Snif */
...
 #endif
 /*---------------------------------------------------------------------------*/
+#if CC2538_RF_CONF_SFD_TIMESTAMPS
+static rtimer_clock_t volatile cc2538_sfd_rtime;
+#endif
+
+/* Are we currently in poll mode? Disabled by default */
+static uint8_t volatile poll_mode = 0;
+/* Do we perform a CCA before sending? Enabled by default. */
+static uint8_t send_on_cca = 1;
```
simonduq: `send_on_cca` seems unused, i.e. always disabled. That's fine for TSCH. But then no need for the variable. Simply always return 0 in `get_option`, and success for `set_option` if set to 0, failure if set to 1.

platform/cc2538dk/contiki-conf.h
```diff
@@ -40,6 +40,13 @@ typedef uint32_t rtimer_clock_t;
 #define RTIMER_CLOCK_LT(a,b)     ((int32_t)((a)-(b)) < 0)
 /** @} */
 /*---------------------------------------------------------------------------*/
+#define TSCH_CONF_HW_FRAME_FILTERING	0
+
+/* TODO: measure the delays, this are only estimations */
+#define RADIO_DELAY_BEFORE_TX     ((unsigned)US_TO_RTIMERTICKS(300))
+#define RADIO_DELAY_BEFORE_RX     ((unsigned)US_TO_RTIMERTICKS(100))
+#define RADIO_DELAY_BEFORE_DETECT 0
```
simonduq: Would like to have actual measurements for these before merging :)

simonduq:
Looks excellent! A few warnings make Travis unhappy.
Please also update the TSCH readme file to add the new supported platform(s)!
Thanks :)

simonduq:
Looking forward to trying on Zoul (I have no cc2538dk)

cpu/cc2538/dev/cc2538-rf.c
```diff
@@ -700,15 +732,15 @@ read(void *buf, unsigned short bufsize)
...
-  PRINTF("%02x%02x\n", (uint8_t)rssi, crc_corr);
+  PRINTF("%02x%02x\n", (uint8_t)cc2538_last_rssi, cc2538_last_crc_corr_lqi);
 
   /* MS bit CRC OK/Not OK, 7 LS Bits, Correlation value */
-  if(crc_corr & CRC_BIT_MASK) {
-    packetbuf_set_attr(PACKETBUF_ATTR_RSSI, rssi);
-    packetbuf_set_attr(PACKETBUF_ATTR_LINK_QUALITY, crc_corr & LQI_BIT_MASK);
+  if(cc2538_last_crc_corr_lqi & CRC_BIT_MASK) {
+    packetbuf_set_attr(PACKETBUF_ATTR_RSSI, cc2538_last_rssi);
+    packetbuf_set_attr(PACKETBUF_ATTR_LINK_QUALITY, cc2538_last_crc_corr_lqi & LQI_BIT_MASK);
```
g-oikonomou: Is this variable renaming strictly required? You could reduce the diff by using the current variable names.

cpu/cc2538/dev/cc2538-rf.c
```diff
@@ -995,10 +1070,20 @@ void
...
 {
   ENERGEST_ON(ENERGEST_TYPE_IRQ);
+  
+#if CC2538_RF_CONF_SFD_TIMESTAMPS
+  if(poll_mode) {
+    if(REG(RFCORE_XREG_FSMSTAT1) & RFCORE_XREG_FSMSTAT1_RX_ACTIVE) {
+      cc2538_sfd_rtime = RTIMER_NOW();
+    }
+  }
+#endif
```
g-oikonomou: Have you looked at the MAC timer (section 12) of the UG? It can be used to automatically capture SFD, so you don't have to enable the SFD interrupt. You would probably need to convert this capture value to an rtimer representation.

The SFD interrupt will also get raised after TX of an SFD. I cannot see any attempt at determining whether it was an RX or a TX that generated the SFD interrupt. Are you handling this anywhere that I cannot see?

platform/cc2538dk/contiki-conf.h
g-oikonomou: > Would like to have actual measurements for these before merging :)

:+1:

g-oikonomou:
This is a very desirable feature. Code style fixes required here and there, please uncrustify this. It would be nice to do some interop testing too! :)

simonduq:
Just wanted to say that I'm currently trying on Zoul.
Worked on the very first attempt!
Now going to take a closer look at timings and interop with jn516x.
Very well done guys!!

alignan:
Same here, working nicely.

alignan:
I would also rebase please

cpu/cc2538/dev/cc2538-rf.c
alignan: :+1:

alignan:
Two additional things, sorry pile things up:

The `examples/ipv6/rpl-tsch` test needs to be fixed:

````
node.co: In function `process_thread_node_process':
node.c:(.text.process_thread_node_process+0x23c): undefined reference to `node_id'
collect2: error: ld returned 1 exit status
````

Also adding a regression test would be OK?

**Comment by simonduq** [comment]
Yes absolutely, node_id is there mostly to enable easy cooja simulations (used in Travis), I need to come up with a better way of identifying the root

**Comment by alignan** [comment]
I can make a PR on top to add the `zoul` config, if I can help out with anything give me a task :-)  When I get back to the office I will try to have time measurements

**Comment by simonduq** [comment]
I got it to talk to jn516x!! One thing is that with a 32768 Hz timer, the timeslot duration does not divide a 10000usec slot. The slot duration is estimated to 328 ticks which actually is 10010 usec. Enough to disable interop with nodes running at true 10000usec. The sky platform has the same problem btw.

A hack is to force jn516x on 10010usec slots and this works.

But a nicer solution is to use @atiselsts' adaptive time compensation. Simply enable `TSCH_CONF_ADAPTIVE_TIMESYNC` and set this (copied from jn516x, used in the non-default case where a low-frequency oscillator is enabled):
```
/* Drift calculated using this formula:
*    ((US_TO_TICKS(10000) * 100) - RTIMER_SECOND) * 1e6 = 976.5625 ppm
*/
#define TSCH_CONF_BASE_DRIFT_PPM -977
```

This way, the node will compensate its own internal drift of -977 ppm and run true 10000usec slots (well, with a max error one tick, but no accumulating error).

As far as I can tell (assuming jn516x is a ground truth -- it was validated with a logic analyzer) the `DELAY_BEFORE` constants are not completely off. But real measurements are needed to find the exact numbers. Happy to help out on how to proceed.

One more note: by default, we use SFD timestamps for initial synchronization but afterwards, during TSCH operation, where we switch to rtimer timestamping of whenever NETSTACK_RADIO.pending() becomes true (more accurate on high-frequency rtimer). I tried enabling SFD timestamping even during slot operation, with `TSCH_RESYNC_WITH_SFD_TIMESTAMPS`. It failed, my guess being that SFD timestamping does not currently work from a rtimer interrupt on cc2538.

**Comment by thomas-ha** [comment]
I have the feeling that there could be a problem with the SFD interrupt. We wanted to measure the time between the beginning of the transmit() function until the SFD interrupt was raised but it was raised at the end of the transmission of the complete frame.

Von meinem iPhone gesendet

Am 11.03.2016 um 17:41 schrieb "Simon Duquennoy" <notifications@github.com<mailto:notifications@github.com>>:


One more note: by default, we use SFD timestamps for initial synchronization but afterwards, during TSCH operation, where we switch to rtimer timestamping of whenever NETSTACK_RADIO.pending() becomes true (more accurate on high-frequency rtimer). I tried enabling SFD timestamping even during slot operation, with TSCH_RESYNC_WITH_SFD_TIMESTAMPS. It failed, my guess being that SFD timestamping does not currently work from a rtimer interrupt on cc2538.

—
Reply to this email directly or view it on GitHub<https://github.com/contiki-os/contiki/pull/1550#issuecomment-195443906>.

Wichtiger Hinweis: Diese E-Mail kann Betriebs- oder Geschäftsgeheimnisse oder sonstige vertrauliche Informationen enthalten. Sollten Sie diese E-Mail irrtümlich erhalten haben, ist Ihnen eine Kenntnisnahme des Inhalts, eine Vervielfältigung oder Weitergabe der E-Mail ausdrücklich untersagt. Bitte benachrichtigen Sie uns und vernichten Sie die empfangene E-Mail. Vielen Dank. Important Note: This e-mail may contain trade secrets or privileged, undisclosed or otherwise confidential information. If you have received this e-mail in error, you are hereby notified that any review, copying or distribution of it is strictly prohibited. Please inform us immediately and destroy the original transmittal. Thank you for your cooperation.

**Comment by simonduq** [comment]
Yes and the problem occurs when receiving from an rtimer interrupt. Almost surely a matter of interrupt priority. At joining time, the reception is *not* done within an interrupt, that is why timestamping works.

Is there a way on cc2538 to capture SFD timestamp without an interrupt?

**Comment by g-oikonomou** [comment]
Simon, Yes. You can timestamp with the MAC timer: see my inline comment above

**Comment by simonduq** [comment]
I missed your comment. That's perfect, exactly what I hoped for!
Le 11 mars 2016 21:02, "George Oikonomou" <notifications@github.com> a
écrit :

**Comment by thomas-ha** [comment]
Using the MAC Timer was planned as my next step. I didn't figure out how to synchronize it with rtimer that's why I went with the easier way.

Von meinem iPhone gesendet

Am 11.03.2016 um 21:49 schrieb "Simon Duquennoy" <notifications@github.com<mailto:notifications@github.com>>:

I missed your comment. That's perfect, exactly what I hoped for!
Le 11 mars 2016 21:02, "George Oikonomou" <notifications@github.com<mailto:notifications@github.com>> a
�crit :

> Simon, Yes. You can timestamp with the MAC timer: see my inline comment
> above
>
> �
> <https://github.com/contiki-os/contiki/pull/1550#issuecomment-195524514>.
>

�
Reply to this email directly or view it on GitHub<https://github.com/contiki-os/contiki/pull/1550#issuecomment-195545250>.

Wichtiger Hinweis: Diese E-Mail kann Betriebs- oder Gesch�ftsgeheimnisse oder sonstige vertrauliche Informationen enthalten. Sollten Sie diese E-Mail irrt�mlich erhalten haben, ist Ihnen eine Kenntnisnahme des Inhalts, eine Vervielf�ltigung oder Weitergabe der E-Mail ausdr�cklich untersagt. Bitte benachrichtigen Sie uns und vernichten Sie die empfangene E-Mail. Vielen Dank. Important Note: This e-mail may contain trade secrets or privileged, undisclosed or otherwise confidential information. If you have received this e-mail in error, you are hereby notified that any review, copying or distribution of it is strictly prohibited. Please inform us immediately and destroy the original transmittal. Thank you for your cooperation.

**Comment by thomas-ha** [comment]
The rtimer interrupt can delay the SFD interrupt in the range of milliseconds?

Von meinem iPhone gesendet

Am 11.03.2016 um 20:33 schrieb "Simon Duquennoy" <notifications@github.com<mailto:notifications@github.com>>:


Yes and the problem occurs when receiving from an rtimer interrupt. Almost surely a matter of interrupt priority. At joining time, the reception is not done within an interrupt, that is why timestamping works.

Is there a way on cc2538 to capture SFD timestamp without an interrupt?

�
Reply to this email directly or view it on GitHub<https://github.com/contiki-os/contiki/pull/1550#issuecomment-195513500>.

Wichtiger Hinweis: Diese E-Mail kann Betriebs- oder Gesch�ftsgeheimnisse oder sonstige vertrauliche Informationen enthalten. Sollten Sie diese E-Mail irrt�mlich erhalten haben, ist Ihnen eine Kenntnisnahme des Inhalts, eine Vervielf�ltigung oder Weitergabe der E-Mail ausdr�cklich untersagt. Bitte benachrichtigen Sie uns und vernichten Sie die empfangene E-Mail. Vielen Dank. Important Note: This e-mail may contain trade secrets or privileged, undisclosed or otherwise confidential information. If you have received this e-mail in error, you are hereby notified that any review, copying or distribution of it is strictly prohibited. Please inform us immediately and destroy the original transmittal. Thank you for your cooperation.

**Comment by simonduq** [comment]
Absolutely: it will delay it by about one packet duration as TSCH
(currently) busy waits from rtimer interrupt until end of Rx. It will then
sleep and wakeup to TX an ACK if requested, then sleep until the next
scheduled slot.
The rtimer interrupt can delay the SFD interrupt in the range of
milliseconds?

Von meinem iPhone gesendet

Am 11.03.2016 um 20:33 schrieb "Simon Duquennoy" <notifications@github.com
<mailto:notifications@github.com>>:


Yes and the problem occurs when receiving from an rtimer interrupt. Almost
surely a matter of interrupt priority. At joining time, the reception is
not done within an interrupt, that is why timestamping works.

Is there a way on cc2538 to capture SFD timestamp without an interrupt?

�
Reply to this email directly or view it on GitHub<
https://github.com/contiki-os/contiki/pull/1550#issuecomment-195513500>.

Wichtiger Hinweis: Diese E-Mail kann Betriebs- oder Gesch�ftsgeheimnisse
oder sonstige vertrauliche Informationen enthalten. Sollten Sie diese
E-Mail irrt�mlich erhalten haben, ist Ihnen eine Kenntnisnahme des Inhalts,
eine Vervielf�ltigung oder Weitergabe der E-Mail ausdr�cklich untersagt.
Bitte benachrichtigen Sie uns und vernichten Sie die empfangene E-Mail.
Vielen Dank. Important Note: This e-mail may contain trade secrets or
privileged, undisclosed or otherwise confidential information. If you have
received this e-mail in error, you are hereby notified that any review,
copying or distribution of it is strictly prohibited. Please inform us
immediately and destroy the original transmittal. Thank you for your
cooperation.

—
Reply to this email directly or view it on GitHub
<https://github.com/contiki-os/contiki/pull/1550#issuecomment-195685993>.

File: core/net/mac/tsch/tsch-slot-operation.c (hunk 2 above)
Comment (thomas-ha) [inline core/net/mac/tsch/tsch-slot-operation.c]: Done, thank you

File: cpu/cc2538/dev/cc2538-rf.c (hunk 5 above)
Comment (thomas-ha) [inline cpu/cc2538/dev/cc2538-rf.c]: I renamed them because I had to make them global and wanted to make sure there is no local variable with the same name. Of course it's not strictly required, will rename them back.

Comment (g-oikonomou) [inline cpu/cc2538/dev/cc2538-rf.c]: Yes, had you made them global you'd need to follow naming convention, but you have simply moved them to file scope (but still static), hence my request to undo the rename. Thank you!

**Comment by thomas-ha** [comment]
I measured RADIO_DELAY_BEFORE_TX and it is 353us. I don't know how to measure RADIO_DELAY_BEFORE_RX on CC2538 but the CC2538 Users Guide says 192us (page 672, 23.9.2 RX State Timing of the document).
If I use this values I get some "TSCH:! drop dup" messages (especially when I reduce RX_CONF_WAIT). That's not the case with the estimated values of 300 and 100us.

**Comment by thomas-ha** [comment]
Does anyone know how to sync the MAC timer with the rtimer? We didn't manage to get the MAC timer working yet.

**Comment by simonduq** [comment]
The drop duplicate messages are very confusing..
A way to roughly validate RADIO_DELAY_BEFORE_TX (and RADIO_DELAY_BEFORE_DETECT if it needs to be non-zero) is to look for sync asymmetry, i.e. the node and its time source don't agree on time and perceive a different time error when they receive from one another. Look at the drift observed at the node on Rx (dr, in ticks) and on Unicast Tx, with ACK-based sync (also dr, in ticks). With incorrect timing you typically have the node jump by a few ticks (plus or minus) after a Unicast Tx and back on the next Rx.

**Comment by thomas-ha** [comment]
Thank you, I haven’t seen any jumps of dr ticks.

I tested TSCH on CC2538 with a RADIO_DELAY_BEFORE_TX value of 300us and RADIO_DELAY_BEFORE_RX value of 100us extensive and it was very successful (with 17 nodes).
I also tested successfully with RX_CONF_WAIT as small as 500us (only with 2 nodes so far).

I tested the configuration with the RADIO_DELAY_BEFORE_TX with the measured 352us and the RADIO_DELAY_BEFORE_RX with the 192us from datasheet. I sometimes get drop dup messages at the receiver side and they come very infrequent and not always. They are caused because the sender sends the message repeatedly (although there are ACKs sent by the receiver, but I have also seen that there were no ACKs sent sometimes).

It seems like the configuration with 300us and 100us is working a lot better, there I haven’t seen any drop dups which is very strange I think … Could there be still some timing problem so that the sender doesn't receive the ACKs?

**Comment by simonduq** [comment]
Yes, that sounds likely. To find out if ACKs are lost, in the TSCH logs, look for Unicast transmission:
* uc-1 means Unicast-data (bc is broadcast, -0 is beacon)
* tx
* look at the status: e.g. "st 0 1" means OK ("0": OK, "2": NOACK, see mac.h), "1" here is the tx attempt number, will increase when there are retransmissions

**Comment by atiselsts** [comment]
You can measure the error between the expected time and the actual time you start receiving ACK, it will give some indication to tune the `RADIO_DELAY_BEFORE_RX` parameter:

```
int32_t delta_ack = ack_start_time - (current_slot_start + tsch_timing[tsch_ts_tx_offset] + tx_duration + tsch_timing[tsch_ts_tx_ack_delay]);
TSCH_LOG_ADD(tsch_log_message,  snprintf(log->message, sizeof(log->message),  "ack error: %d", delta_ack));
```

**Comment by thomas-ha** [comment]
We are using the MAC timer instead of the SFD interrupt now.

**Comment by thomas-ha** [comment]
I updated RADIO_DELAY_BEFORE_TX to the measured value of 352 micro-seconds. RADIO_DELAY_BEFORE_RX should be 192 micro-seconds as in data sheet but then ACKs are not always received. I adjusted RADIO_DELAY_BEFORE_RX to 250 micro-seconds and it works fine (tested with up to 16 nodes and TSCH_CONF_RX_WAIT as small as 600us).
I couldn't figure out why it doesn't work as well with a RADIO_DELAY_BEFORE_RX of 192us.

File: core/net/mac/tsch/tsch-conf.h (hunk 1 above)
Comment (thomas-ha) [inline core/net/mac/tsch/tsch-conf.h]: Done, thank you.

File: cpu/cc2538/dev/cc2538-rf.c [hunk 7]
```diff
@@ -395,7 +427,9 @@ off(void)
   /* Wait for ongoing TX to complete (e.g. this could be an outgoing ACK) */
   while(REG(RFCORE_XREG_FSMSTAT1) & RFCORE_XREG_FSMSTAT1_TX_ACTIVE);
 
-  CC2538_RF_CSP_ISFLUSHRX();
+  if (!(REG(RFCORE_XREG_FSMSTAT1) & RFCORE_XREG_FSMSTAT1_FIFOP)) {
```
g-oikonomou: Code style

cpu/cc2538/dev/cc2538-rf.c
```diff
@@ -995,10 +1066,12 @@ void
...
   ENERGEST_ON(ENERGEST_TYPE_IRQ);
+  
+  if(!poll_mode) {
+    process_poll(&cc2538_rf_process);
+  }
 
-  process_poll(&cc2538_rf_process);
-
-  /* We only acknowledge FIFOP so we can safely wipe out the entire SFR */
+  /* We only acknowledge FIFOP or SFD so we can safely wipe out the entire SFR */
```
g-oikonomou: Does this comment actually need updated? In other words, are we still acknowledging the SFD interrupt?

cpu/cc2538/dev/cc2538-rf.h
```diff
@@ -129,6 +129,15 @@
...
 } while(0)
+/*---------------------------------------------------------------------------
+ * MAC timer
+ *---------------------------------------------------------------------------*/
+/* Timer conversion */
+#define RADIO_TO_RTIMER(X) ((X) * RTIMER_ARCH_SECOND / SYS_CTRL_32MHZ)
+
+#define CLOCK_STABLE() do {															\
+			while ( !(REG(SYS_CTRL_CLOCK_STA) & (SYS_CTRL_CLOCK_STA_XOSC_STB)));	\
+		} while(0)
```
g-oikonomou: Why are we putting those 2 macros in the header file? They should i) either be moved to the C file if that's the only place that uses them, or ii) Rename them to CC2538_RF_BLAH if they are staying in the header.

g-oikonomou:
Thomas, where are we on this one? This will be a great addition once completed, we will need to do some interop testing of course across platforms.

cpu/cc2538/rtimer-arch.h
```diff
@@ -65,6 +65,20 @@
...
+                               (((int32_t)(US) * (RTIMER_ARCH_SECOND) + 500000) / 1000000L) :      \
+                               ((int32_t)(US) * (RTIMER_ARCH_SECOND) - 500000) / 1000000L)
+
+#define RTIMERTICKS_TO_US(T)   ((T) >= 0 ?                     \
+                               (((int32_t)(T) * 1000000L + ((RTIMER_ARCH_SECOND) / 2)) / (RTIMER_ARCH_SECOND)) : \
+                               ((int32_t)(T) * 1000000L - ((RTIMER_ARCH_SECOND) / 2)) / (RTIMER_ARCH_SECOND))
+
+/* A 64-bit version because the 32-bit one cannot handle T >= 4295 ticks.
+   Intended only for positive values of T. */
+#define RTIMERTICKS_TO_US_64(T)  ((uint32_t)(((uint64_t)(T) * 1000000 + ((RTIMER_ARCH_SECOND) / 2)) / (RTIMER_ARCH_SECOND)))
```
g-oikonomou: I know it's PITA, but those macros need renamed...

thomas-ha: The name of these macros were defined by @simonduq in the TSCH implementation.

g-oikonomou: Hmmm OK oops. So TSCH expects those to exist somewhere, right?

thomas-ha: Exactly. In the other platforms the macros are also defined in rtimer-arch.h

g-oikonomou: Right yep, I see it now. Thanks

thomas-ha:
Hello,

The status of TSCH for CC2538:

I tested this port with up to 16 nodes with 6TiSCH minimal schedule and orchestra. This also includes long term testing for 3 days. UDP packets were sent from nodes to the root and from root to nodes with reply from the node. The loss ratio was about 0% to 0,2%.
I only tested with CC2538 nodes and no interoperability testing has been performed yet. @simonduq got it talking to JN516x he said.

What I struggled with was the RADIO_DELAY_BEFORE_TX and RADIO_DELAY_BEFORE_RX values. I measured 352us for RADIO_DELAY_BEFORE_TX and the CC2538 Users Guide says 192us for RADIO_DELAY_BEFORE_RX (page 672, 23.9.2 RX State Timing). I had to increase RADIO_DELAY_BEFORE_RX to 250us to not get "drop duplicate" messages by TSCH because of not received ACKs.

Recently I also discovered a problem using the rpl-border-router with TSCH. As soon as TSCH is started on the border-router by the NETSTACK_MAC.on() command the slip_input_callback() isn't called anymore on CC2538. I couldn't really figure out the reason for that yet ...

g-oikonomou:
> Recently I also discovered a problem using the rpl-border-router with TSCH. As soon as TSCH is started on the border-router by the NETSTACK_MAC.on() command the slip_input_callback() isn't called anymore on CC2538. I couldn't really figure out the reason for that yet ...

This could be a problem with LPM. Can you try disabling LPM2 on the border router?

thomas-ha:
Only LPM1 was active.

g-oikonomou:
What device are you using as a BR? If it's a CC2538EM, can you try over native USB?

thomas-ha:
I'm using a CC2538DK and our own CC2538 based hardware platform. I used the SmartRF06EB, I wasn't able to connect the CC2538EM directly to my PC, it wasn't recognized.

If I comment out the command to turn off the rf module in the off() function of the radio driver ( https://github.com/thomas-ha/contiki/blob/tsch-cc2538-port/cpu/cc2538/dev/cc2538-rf.c#L436 ) the slip input callback gets called. I have absolutely no idea what this has to do with the slip input callback but it really is like that. BUT then the slip input callback doesn't allways receive all of the bytes sent over SLIP, so it doesn't work reliable anyway. That's all pretty strange to me.
I have seen that the JN516x platform has implemented its own slip.c Maybe some of these changes are needed for CC2538 too?

cpu/cc2538/dev/cc2538-rf.c
thomas-ha: Done, thanks!

cpu/cc2538/dev/cc2538-rf.h
thomas-ha: Done, thanks

g-oikonomou:
If the radio is off, the LPM module will drop to some low power mode. This will normally gate the clock to the UART and for that reason the callback won't get called (basically bytes won't get received, the UART RX interrupt won't get raised). I am not familiar with how TSCH behaves on BRs, but the logic with ContikiMAC is to keep the radio always on. As a side-effect of that, the chip will never drop to deep sleep and the UART stays active and clocked at all times.

simonduq:
OK then this might be the problem @g-oikonomou 
In TSCH the BR behaves just like other nodes: it follows a schedule and sleeps both between active slots and within an active slot (between Rx/Tx etc).
One more thing: do we have DMA for UART input? If not we might also get drops, as TSCH spends significant time in interrupt.

sieben:
@simonduq Would you consider systematic interoperability testing for TSCH (for instance as regression tests)?

thomas-ha:
> If the radio is off, the LPM module will drop to some low power mode. This will normally gate the clock to the UART and for that reason the callback won't get called (basically bytes won't get received, the UART RX interrupt won't get raised). I am not familiar with how TSCH behaves on BRs, but the logic with ContikiMAC is to keep the radio always on. As a side-effect of that, the chip will never drop to deep sleep and the UART stays active and clocked at all times.

Ok, thank you very much @g-oikonomou , I didn't know that. I checked the CC2538 manual and have seen the registers for gating the UART clock. Disabling LPM completely also helps of course.

Then there is still the problem that @simonduq mentioned. Sometimes there are missing bytes in the packet received over the slip interface of the packet isn't received at all. I guess because the UART interrupt is raised too late because of TSCH busy waiting from an rtimer interrupt.

simonduq:
Again: without DMA, UART is not going to work with TSCH. TSCH may spend up to a couple millis in interrupt (e.g. when busy waiting for a Tx to terminate).

g-oikonomou:
It should be possible to allow PM0 but leave the UART clocked and operational.

A quick (but perhaps dirty) workaround would be to set UART RX interrupt at a higher priority. We have DMA support for the radio and for USB, so the fundamental code is there. The UART driver would need changes of course.

thomas-ha:
The border router works with the CC2538EM and USB instead of UART of course because then I can use slip over USB with DMA.
I will take a look at the UART and DMA.

g-oikonomou:
Just to clarify (or to simply state the obvious). When we enable native USB on the CC2538 (e.g. SLIP over USB), the USB PLL is always on. When the PLL is active, the USB driver prohibits LPM1 or lower. So in our context here, as soon as we enable USB, the Border-Router problem goes away.

simonduq:
Except for the DMA part @g-oikonomou right?

g-oikonomou:
There is _some_ DMA support in the USB driver, but it's not exactly what we'd be looking for in this context here. However, my gut feeling is that we won't miss bytes over USB purely down to speed.

simonduq:
@thomas-ha could you please update the TSCH readme file to mention the new supported platform?
https://github.com/contiki-os/contiki/blob/master/core/net/mac/tsch/README.md

Other than that this is ready to merge AFAIC

thomas-ha:
Yes, done! Thank you for the reminder!

simonduq:
@g-oikonomou you had some comments on enabling xtal on cc2538, can you say a little more so we can have it enabled by default at least in the tsch example?

simonduq:
@thomas-ha anything else you want to put in before we merge?

thomas-ha:
XTAL is enabled for the TSCH example in the project-conf.h:
#undef SYS_CTRL_CONF_OSC32K_USE_XTAL
#define SYS_CTRL_CONF_OSC32K_USE_XTAL 1

No, I don't have anything to add,

simonduq:
Ah ok. Good to merge then! Let's wait for Travis to be green

simonduq:
In the readme there is one missing mention of cc2538dk after "he easiest is probably to start from one of the existing port"

g-oikonomou:
Yes we were discussing on skype about how this needs the LF XTAL for the CC2538 or it would not work and we agreed that I'd comment about it, only to realise that you have already applied this change in the project-conf.h. Sorry if I confused!

joakimeriksson:
Any list of platforms that are know to have the XTAL on the CC2538? Re-mote?

g-oikonomou:
Pretty certain CC2538DK has it. Will need to double-check, but almost 100%

joakimeriksson:
I know that I have a few devices without it and the 32768 Hz sleep timer is really not running 32768 but rather 32753 and usually up to 1% off that also. At cold-start it can be as much as 5% off.

simonduq:
That is annoying. TSCH typically expects clocks with 20-40 ppm. Playing with the guard time and sync interval one can try to handle perhaps a couple hundred ppms.
1% off is 1000 ppm.

joakimeriksson:
I guess for non deep-sleep devices that can keep the 32MHz (crystal) clock always ticking that could also be used instead of the 32KHz one.

simonduq:
If you have this available then yes it is an option!

alignan:
The `zoul` and all based boards like the `RE-Mote` and `Firefly` have a 32.768KHz Crystal with +/-20ppm

thomas-ha:
Ok, I don't know why the Travis check has failed. All I did was changing the Readme

alignan:
It is unrelated, Travis is currently broken (fixing now!)

simonduq:
merging now, thanks for the contribution!

thomas-ha:
Thank you for all you help @simonduq , very appreciated!

cpu/cc2538/dev/cc2538-rf.c
```diff
@@ -731,14 +772,18 @@ read(void *buf, unsigned short bufsize)
...
+    if(REG(RFCORE_XREG_FSMSTAT1) & RFCORE_XREG_FSMSTAT1_FIFOP) {
+      if(REG(RFCORE_XREG_FSMSTAT1) & RFCORE_XREG_FSMSTAT1_FIFO) {
+        process_poll(&cc2538_rf_process);
+      } else {
+        CC2538_RF_CSP_ISFLUSHRX();
+      }
     }
   }
+  
+  CC2538_RF_CSP_ISFLUSHRX();
```
g-oikonomou: @thomas-ha Why are you flushing the RX FIFO unconditionally here?

The original logic was:
* If FIFO overflow, flush it
* If it holds more frames, poll the process again.

If there are more frames in the FIFO here, the process will get polled but by the time it gets called the FIFO will have been emptied.

What am I missing?

PR #1553, opened 2016-03-14
Fix a number of leftover mentions of aaaa as a prefix (now fd00)

simonduq:
A number of new "aaaa" slipped in between the original PR and its merge. This replaces them with the new local prefix "fd00"


nvt:
:+1:

PR #1575, opened 2016-03-31
Cooja: add Msp802154BitErrorRadio interface

atiselsts:
This Cooja interface more accurately models bit errors as observed in IEEE 802.15.4-compatible radios, such as CC2420 and CC2520.

Only errors due to signal fading are supported, as opposed to errors to interference. Interesting and quite realistic performance can be observed if this is used together with a dynamic channel fading model.

A usage example [is available here](https://github.com/atiselsts/contiki/commit/758202de6c2dfd28d2fc3199b336d2b15a8df5f0). The example includes two `.csc` files, one using the existing Msp802154Radio and one the new Msp802154BitErrorRadio, but otherwise identical. The BitError example shows corrupted packets detected by the Contiki `.c` code and occasionally packets being completely lost (due to corruption in the header).

This pull request is based on work done in 2014 but not continued afterwards; maybe someone else finds this useful for their research.

### DIFF
--- a/tools/cooja/apps/mspsim/src/org/contikios/cooja/mspmote/interfaces/Msp802154BitErrorRadio.java
+++ b/tools/cooja/apps/mspsim/src/org/contikios/cooja/mspmote/interfaces/Msp802154BitErrorRadio.java
@@ -1,0 +1,370 @@
+/*
+ * Copyright (c) 2014, Uppsala University
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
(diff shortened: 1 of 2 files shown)


alignan:
Seems like a good contribution and adds a new feature rather than modifying existing ones, is a nice to have.

It would be great to have this double-checked by someone more involved in Cooja. 

:+1:

fros4943:
Looks great! Could be extremely useful for radio-near simulations.

To make this new bit error radio more easily available in simulation, the class name may be added to mote types' `getAllMoteInterfaceClasses()` method.

:+1: Merging

PR #1576, opened 2016-03-31
JN516x clock module sign error fix

mguc:
This small fix has a huge impact on the etimer events. The current check is wrong for sure since `clock_time_t` is unsigned. For further details check my comment on #1265.


simonduq:
Thanks, I also noticed some problems since the switch to tickless, very possibly caused by this bug :+1:

Issue #1571, opened 2016-03-30
Inefficient CSMA with ContikiMAC which is "off"


bkozak-scanimetrics:
I've been experimenting with disabling duty-cycling on ContikiMAC by calling ```NETSTACK_MAC.off(1)``` at initialization for the purpose of allowing some routers to carry more traffic. 

Unfortunately, this does not work well if there are many nodes transmitting at once since the CSMA algorithm will severely limit the effective bandwidth of the router when a collision is detected. The CSMA backoff period is calculated from the duty cycle period of ContikiMAC which means that we will wait hundreds of milliseconds whenever a collision occurs.  This is a major problem for routers since  a router which has entered CSMA backoff will run out of available transmit buffers after only a handful of packets are received and then will start to drop packets while still transmitting MAC level ACKs. Since the cost of a collision is so severe, it is all but necessary to ensure that any nodes which are generating data always insert a delay between packets to try and prevent routers from detecting a collision (otherwise a collision always occurs when the router attempts to retransmit the first packet in a burst while the second is in the air). 

I think that this can be solved by improving the CSMA code but I wanted to get some feedback on this before I attempted any sort of solution.

My initial idea is to do the following:
- Fix #1142 
- Have ContikiMAC return a different value for ```channel_check_interval()``` when it isn't performing a channel check.
- Truncate the backoff period based on the absolute value of the maximum back-off interval, instead of the value of the exponent.

I think that by doing this, we could allow non duty-cycled routers to function reasonably well without having any major impact on the functioning of nodes which perform the regular ContikiMAC duty-cycling.

PR #1579, opened 2016-04-01
Use collisions for csma backoff and retry

### LINKED ISSUE #1142: CSMA backoff logic and re-TX count ignore MAC_TX_COLLISION

bkozak-scanimetrics:
This PR fixes #1142 

The choice of default value for ```CSMA_MAX_MAX_COLLISIONS``` is working for me but is otherwise arbitrary. Please speak up if you know a better value for this constant.

I have some other concerns about the effectiveness of the exponential back-off constants now that collisions are accounted for (and see also my issue at #1571) but I think that this should be good enough for now.


simonduq:
That's a nice addition, but how much more work would it be to do this as per the IEEE 802.15.4 standard? (figure 11 of IEEE 802.15.4-2011, non-slotted case)

bkozak-scanimetrics:
I haven't had the time to read through the standard but from looking at Figure 11 and reading your comment on #1142 it doesn't sound like this would require any major difficulty.

The major changes that would need to be made are (I think):
- Zero ```backoff_exponent``` and ``` n->collisions``` on a ```NOACK``` failure
- Increment ```n->transmissions``` for every __macMaxCSMABackoffs__ consecutive collisions
- Only consider ```n->transmissions``` when deciding when to abort the transmit attempt

Does this sound right to you?

simonduq:
Sounds right to me, except one thing missing: `backoff_exponent` should be the number of collisions, not the number of transmissions. The exponent increases at every collision until a Tx happens, and is reset before the next attempt (re-enters the state machine in fig 11 at every attempt, max `macMaxCSMABackoffs` times).

To really match the standard, `default_timebase` should not default to `CLOCK_SECOND / NETSTACK_RDC_CHANNEL_CHECK_RATE` but rather aUnitBackoffPeriod (20 symbol times, i.e. 320 usec)

bkozak-scanimetrics:
The current ```default_timebase``` does seem to me to be too high to me but If it is too small we could end up with too much wasted energy on CCA checks since the ContikiMAC strobing and CCA mechanism implies a lower probability that the packet is finished sending until a significant portion of a channel check period has passed. If the energy cost is too high we still have to make sure that the exponent truncates at a much larger value than 3 so as to give a reasonable chance that we don't fail out before the packet (or burst of packets) passes.

I think the ```default_timebase``` probably needs to be higher than 320 usec for ContikiMAC unless perhaps it is in the non-duty-cycled mode since the CCA doesn't use much extra energy for a radio which is already on.

bkozak-scanimetrics:
Otherwise, it makes sense to me to try and match closer to the standard.

simonduq:
Absolutely, I only meant the default value of `default_timebase`, which is for RDCs that do not set a `channel_check_interval`, i.e. nullrdc. Will keep the current timebase for ContikiMAC.

There should also be no truncation of the backoff anymore, as the standard keeps incrmeenting until `macMaxCSMABackoffs`, then transmits and resets the backoff.

bkozak-scanimetrics:
I've attempted to make the csma code match the 802.15.4 standard more closely. By doing this, we will no longer able to track the total number of collisions and the value of ```n->transmissions``` (which is passed up the the higher layers of the stack) takes on a different meaning.

I haven't changed any of the default values for controlling the backoff and make the backoff reset after 3 consecutive collisions (otherwise we would end up with a higher max exponent value than we have now).

bkozak-scanimetrics:
Also added @simonduq's suggestion of setting the default timebase to ~320us. Since 320us is lower than what we can represent on most (all?) platforms this will usually default to 1 clock tick instead.

simonduq:
Nice! Thanks :) I will take a very close look and try this asap.
Don't we have most platforms running rtimer at 32768 Hz, i.e. 30.51 usec tick? Note you could use `US_TO_RTIMERTICKS` when available for usec <-> ticks conversion with rounding to nearest.

bkozak-scanimetrics:
Regarding the Travis failure, it looks like the failure was caused because Travis was unable to download some software package when the test started. I don't believe that it is has anything to do with the actual code.

bkozak-scanimetrics:
> Don't we have most platforms running rtimer at 32768 Hz, i.e. 30.51 usec tick? Note you could use US_TO_RTIMERTICKS when available for usec <-> ticks conversion with rounding to nearest.

Right, I hadn't though of using that approach...

So, just to make sure I have this right, you just want to do the exponent calculation using higher resolution variables then convert to clock ticks so that they can be used by ctimer? Obviously, using the rtimer to directly schedule retransmit events would be better but we can't do that quite yet.

I've made a graph of what would happen if we tried to perform the calculation with rtimer resolution then truncate to clock ticks.  Unfortunately, I don't think that it works very well.
exponent

The only real solution that I see is to have a higher resolution on the clock ticks (which is something else that I'm working on and, in fact, better timer resolution on the csma modules was one of my motivations for doing this).

simonduq:
sorry, my mistake, I forgot the macro was for rtimer but we need ctimer indeed! `CLOCK_SECOND / 3125` is probably good enough then

simonduq:
What is the status on this?

bkozak-scanimetrics:
I don't have any changes planned.

When I get the chance I'll do a rebase to fix Travis but otherwise I don't intend to change anything unless someone has another suggestion.

simonduq:
OK let's see if Travis is happy now :)

simonduq:
I'll take a close look and test this at some point

bkozak-scanimetrics:
The Travis failure is completely different than the last time.

A quick perusal of the Travis build history for Contiki OS shows I'm not the only one getting a timeout error on the RPL build. The exact output from Travis isn't the same for the failure of this PR as it is in some of the others however.

If we consider also that this exact same test did not fail before the rebase, I think it is reasonable to guess that the problem may not be with this particular PR.

alignan:
I just restarted that particular job, let's wait and see

bkozak-scanimetrics:
I've tried this a few times now on my own Travis account and It certainly seems like restarting the build isn't going to help.

I'll take a look into this.

bkozak-scanimetrics:
After much experimentation I have found that, in the large rpl network test, packets will almost inevitably get lost if we do either of the following:
1. Reset the exponentiation after a ```NOACK``` or on some number of consecutive collisions
2. Increment the transmission count after some number of consecutive collisions (thus allowing us to terminate the transmit attempt after a collision). 

Essentially, the very flaws that this PR attempts to fix are actually necessary to make the large network test pass.

In my research into this problem, I had found that, at times, we will see a suspicious number of consecutive collisions (up to 24 in some cases) and I suspect that this may be a bug. Otherwise, my only suggestion to fix this problem is to raise the value of the various constants used in the CSMA algorithm when using Cooja and nullrdc (which the large RPL test uses). 

As I was studying this code, however, I did also find one minor issue which I will fix soon. In the function ```collision``` I will replace
```C
if(n->collisions >= CSMA_MAX_BACKOFF_EXPONENT)
```
with

```C
if(n->collisions > CSMA_MAX_BACKOFF_EXPONENT)
```

bkozak-scanimetrics:
> In my research into this problem, I had found that, at times, we will see a suspicious number of consecutive collisions (up to 24 in some cases) and I suspect that this may be a bug

To be clear, I suspect that this may be a bug in Cooja.

simonduq:
I did take a look in depth and proposed some improvements. I wanted to PR against your branch @bkozak-scanimetrics but for a reason I don't understand Github doesn't allow me to target your repo with my PR... Please take a look at this, cherry pick if you like it and add your modifs on top!
https://github.com/simonduq/contiki/commits/pr/fix_csma_backoff

g-oikonomou:
> I wanted to PR against your branch @bkozak-scanimetrics but for a reason I don't understand Github doesn't allow me to target your repo with my PR.

Did you try starring the target repo?

simonduq:
Thanks @g-oikonomou, it worked :)

bkozak-scanimetrics:
> I did take a look in depth and proposed some improvements. 

This all seems reasonable to me. The new random interval for the backoff time is a really big change however; it probably deserves some testing beyond just Travis.

Speaking of Travis, it seems that your PR also fails on 04-rpl-large-network. I still haven't come up with any good solution for this but I'm starting to understand the problem better.

simonduq:
Note the randomness in the backoff was already there https://github.com/contiki-os/contiki/blob/master/core/net/mac/csma.c#L286

The exact parameters/numbers changed a bit though.

simonduq:
The Travis problem is unrelated to this PR yes, it strikes every now and then, something needs to be done

bkozak-scanimetrics:
I've re-based on your recent fixes for RPL. 

As you can see, we are now failing on a completely different test (large RPL). On my own Travis account, the large RPL test passed but the ipv6 test failed.

simonduq:
Large RPL is actually dependent on external networks and may fail now and then -- I just restarted the job.
As for ipv6 fragmentation tests, I've also noticed failure lately, but these are more annoying as rerunning normally doesn't help...

simonduq:
Are you considering this as well? https://github.com/bkozak-scanimetrics/contiki/pull/2

bkozak-scanimetrics:
> Are you considering this as well? bkozak-scanimetrics#2

Yes, testing  [here](https://github.com/bkozak-scanimetrics/contiki/tree/tmp). It's having the same issues with IPv6 and Large RPL. I've just restarted both of those tests and we'll see if it passes soon.

bkozak-scanimetrics:
I'm now also seeing a failure of the "slip-radio" test once I include your changes.

bkozak-scanimetrics:
After enough restarts of all the failing builds Travis does eventually pass.

I've add your commit to this branch now.

simonduq:
Great! :+1:

alignan:
@g-oikonomou are you OK to merge and close #1142?

g-oikonomou:
I've not followed this one very closely, but if folks think it fixes the problem then who am I to argue? :)

alignan:
:+1:

PR #1587, opened 2016-04-06
Fix failure of ARM-AAPCS travis job

### LINKED ISSUE #1586: Travis failing due to arm-gcc version no longer being available
Since #1504 we are explicitly specifying what version of the toolchain to use for the travis arm job. Unfortunately, this version is no longer available on the PPA, apt fails to install it and the test falls over.

We can obviously revert the change from #1504 but then we will go back to the previous situation of travis using a different toolchain version than the one we specify in various READMEs. We have already agreed that we don't want that.

I've not 100% made up my mind how we want to fix this. @bthebaudeau @alignan any thoughts?

g-oikonomou:
Up till now we have been installing arm-gcc using apt from the team-gcc-arm-embedded ppa.

As discussed in #1453 and implemented in #1504, we have decided to lock travis to use the specific version of the toolchain that is documented in the READMEs of relevant platforms. However, the PPA no longer hosts this version, apt fails to install arm-gcc and as a result the job fails too.

This pull changes the travis job to wget an installation tarball for the desired version, instead of trying to install via apt.

Fixes #1586

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,13 +40,13 @@ before_script:
       arm-none-eabi-gcc --version ;
     fi
 
-  ## Install mainline ARM toolchain.  gcc-arm-none-eabi is available
-  ## in Ubuntu >= 14.04, but this external PPA is needed for 12.04.
-  ## Install srecord
+  ## Install mainline ARM toolchain and srecord.
   - if [ ${BUILD_ARCH:-0} = arm-aapcs ] ; then
-      sudo add-apt-repository -y ppa:team-gcc-arm-embedded/ppa &&
-      sudo apt-get -qq update &&
-      sudo apt-get -qq install gcc-arm-embedded=5-2015q4-1~precise1 srecord &&
+      sudo apt-get -qq install srecord &&
+      $WGET https://launchpad.net/gcc-arm-embedded/5.0/5-2015-q4-major/+download/gcc-arm-none-eabi-5_2-2015q4-20151219-linux.tar.bz2 &&
+      tar xjf gcc-arm-none-eabi-5_2-2015q4-20151219-linux.tar.bz2 -C /tmp/ &&
+      sudo cp -f -r /tmp/gcc-arm-none-eabi-5_2-2015q4/* /usr/local/ &&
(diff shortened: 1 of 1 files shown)


simonduq:
:+1:

.travis.yml
```diff
@@ -40,13 +40,12 @@ before_script:
...
-  ## Install srecord
+  ## Install mainline ARM toolchain and srecord.
   - if [ ${BUILD_ARCH:-0} = arm-aapcs ] ; then
-      sudo add-apt-repository -y ppa:team-gcc-arm-embedded/ppa &&
-      sudo apt-get -qq update &&
-      sudo apt-get -qq install gcc-arm-embedded=5-2015q4-1~precise1 srecord &&
+      sudo apt-get -qq install srecord &&
+      $WGET https://launchpad.net/gcc-arm-embedded/5.0/5-2015-q4-major/+download/gcc-arm-none-eabi-5_2-2015q4-20151219-linux.tar.bz2 &&
+      tar xjf gcc-arm-none-eabi-5_2-2015q4-20151219-linux.tar.bz2 -C /tmp/ &&
+      sudo cp -f -r /tmp/gcc-arm-none-eabi-5_2-2015q4/* /usr/local/ &&
```
bthebaudeau: You could add: `rm -rf /tmp/gcc-arm-none-eabi-* gcc-arm-none-eabi-*-linux.tar.bz2 &&`

g-oikonomou: Hm that was there and I removed it. Why do we need it?

bthebaudeau: This is not mandatory. It just avoids keeping useless files, and this is more consistent with what `.travis.yml` does for the other toolchains.

g-oikonomou: Yes, I too noticed that for other toolchains we remove things, but I simply cannot get why. Keeping useless files in a VM that will get torn down once we're done is pretty irrelevant I would have thought :)

bthebaudeau: Sure. It's just for extra tidiness, nit picking. It does not matter much.

g-oikonomou:
Pushed the removal of the tarball and stuff in /tmp as request.

bthebaudeau:
:+1:

PR #1589, opened 2016-04-07
Allow creating connections on unspecified local port

ibriano:
If the given local port is 0, we get no replies.

### DIFF
--- a/core/net/ip/simple-udp.c
+++ b/core/net/ip/simple-udp.c
@@ -119,7 +119,7 @@ simple_udp_register(struct simple_udp_connection *c,
 
   PROCESS_CONTEXT_BEGIN(&simple_udp_process);
   c->udp_conn = udp_new(remote_addr, UIP_HTONS(remote_port), c);
-  if(c->udp_conn != NULL) {
+  if(c->udp_conn != NULL && local_port) {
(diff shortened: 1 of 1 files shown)


simonduq:
Could you please explain in a bit more details the intent of this PR?

ibriano:
The idea is that if I want to create a simple-udp connection to act as a client, I don't care about the local port, so I'm passing 0 for it and let it assign a random one. The problem with that is that simple-udp always calls bind with the local port passed, and if that is 0, it ends up with a handler that looks unused.

vsaw:
Please update the documentation in the header file as well.

simonduq:
OK great. Yes please update the doxygen in simple-udp.h!

simonduq:
setting a timeout, will close in a month if nothing happens, merge if the doc is updated accordingly

ibriano:
Oops, sorry, got sidetracked and forgot to do this. I'll get to it now.

ibriano:
And I went to do it and the documentation already says that's the case. I guess I can just reword the commit message to state that?

simonduq:
That is true! I'm fine with merging as is, thanks :+1: 
(let's wait for another :+1: or one month deadline)

alignan:
Here you go :+1:

PR #1597, opened 2016-04-13
Removed some unnecessary statics in do_event() and process_post().

rgrr:
This allows the optimizer to put the corresponding variables into registers.
See also discussion about other static variables:
https://sourceforge.net/p/contiki/mailman/message/35010460/


alignan:
:+1: as soon Travis is green

simonduq:
:+1:

PR #1598, opened 2016-04-14
JN516x NETSTACK_LLSEC.init and ccm-star fix

mguc:
I've been experimenting with the adaptivesec llsec driver and I ran into some issues on the jn516x platform. One problem was, that the first hello message was triggered twice on the jn516x platform (netstack_init + NETSTACK_LLSEC.init), causing the authentication to fail since the sending node updated his hello_challenge on the second init already and the receiving node was still working with the old challenge. Is there a reason, why NETSTACK_LLSEC.init should be called twice here?

The second problem was to replace ccm_star_driver_jn516x by the default ccm_star_driver. Both driver files have the same name. Renaming the platform driver file it to jn516x-ccm-star.c fixed the problem.


neophob:
+1

simonduq:
good fixes, thanks!

simonduq:
:+1:

PR #1601, opened 2016-04-18
cc26xx-cc13xx BLE beacon event issue

philrhinehart:
This pull request fixes an issue in the cc26xx-cc13xx BLE beacon process.

The rf_ble_beacon_process yields the thread using PROCESS_WAIT_EVENT() which allows the process to continue when _any_ event is posted rather than checking that the event timer has expired, which is what we really want. This pull request adds a check for ble_adv_et event timer expiration, which prevents broadcasts from sensor updates, etc. from triggering the BLE beacon.


g-oikonomou:
Discussed here: http://thread.gmane.org/gmane.os.contiki.devel/26826/

👍 Merging this after Travis passes.

PR #1602, opened 2016-04-18
cc26xx-cc13xx oscillator control issue

philrhinehart:
This pull request fixes an issue in the cc26xx-cc13xx oscillator control which causes occasional faults and device lockup.

When running both ContikiMAC and the BLE beacon process, the ContikiMAC interrupt will occasionally preempt the beacon process while it is modifying the oscillators. This causes occasional bus faults and infinite loops if it happens at just the right time.

This pull request prevents the conflicts by checking that the BLE beacon is not active prior to modifying the oscillators.

### DIFF
--- a/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
+++ b/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
@@ -1049,12 +1049,6 @@ static int
 on(void)
 {
   /*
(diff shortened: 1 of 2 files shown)


g-oikonomou:
Discussed here: http://thread.gmane.org/gmane.os.contiki.devel/26826/

👍 Merging this after Travis passes.

PR #1609, opened 2016-04-20
Added net-debug.[ch] to provide debug functions even in the non-IP case

simonduq:
This PR moves non IP related debug functions from `ip-debug.[ch]` to `net-debug.[ch]`, enabling the use macros such as `PRINTF` and `PRINTLLADDR` in code that may run without IP.

### DIFF
--- a/core/net/net-debug.c
+++ b/core/net/net-debug.c
@@ -30,41 +30,36 @@
 
 /**
  * \file
- *         A set of debugging tools
+ *         A set of debugging tools for the IP stack
  * \author
  *         Nicolas Tsiftes <nvt@sics.se>
  *         Niclas Finne <nfi@sics.se>
  *         Joakim Eriksson <joakime@sics.se>
+ *         Simon Duquennoy <simon.duquennoy@inria.fr>
  */
 
-#include "net/ip/uip-debug.h"
-#include "debug.h"
+#include "net/net-debug.h"
+
 /*---------------------------------------------------------------------------*/
 void
-uip_debug_ipaddr_print(const uip_ipaddr_t *addr)
+net_debug_lladdr_print(const uip_lladdr_t *addr)
 {
-#if NETSTACK_CONF_WITH_IPV6
-  uint16_t a;
-  unsigned int i;
-  int f;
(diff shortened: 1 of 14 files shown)


core/net/net-debug.c
```diff
@@ -0,0 +1,64 @@
...
+/*---------------------------------------------------------------------------*/
+void
+net_debug_lladdr_print(const uip_lladdr_t *addr)
+{
+#if NETSTACK_CONF_WITH_RIME
+  /* Rime uses traditionally a %u.%u format */
+  PRINTA("%u.%u", addr->addr[0], addr->addr[1]);
+#else /* NETSTACK_CONF_WITH_RIME */
+  unsigned int i;
+  if(addr == NULL) {
```
nfi: What do you think about keeping the NULL check also for Rime? It is rather convenient to not have to check for NULL for debug output.

nfi:
What do you think about adding PRINT6LLADDR() in uip-debug.h that prints a `uip_lladdr_t` and let PRINTLLADDR() in net-debug.h print a `linkaddr_t` ? Then the cast from `linkaddr_t` to `uip_lladdr_t` could be removed in the debug output in code that run without IP.

core/net/net-debug.c
simonduq: sounds good yes

simonduq:
I'm not even sure why we have two different types to be honest.. what would you think about using `linkaddr_t` everywhere instead?

nfi:
I think it would be really nice if the link layer addresses could to be merged into one type. I do not think there should be an important reason to have multiple types anymore.

simonduq:
OK we're on the same page then :)
I implemented your suggestion of keeping the check for NULL addr, thanks!

joakimeriksson:
@simonduq we have two for historic reason I would say. I am definitively positive to have just the linkaddr_t instead of two that cause many ugly casts...

alignan:
Where are we on this one? ready to merge?

simonduq:
yes!

alignan:
:+1:

Issue #1517, opened 2016-02-15
TSCH on wismote - EB synchronization fail


anabildea:
Dear all,

I am trying  to port TSCH on wismote. I have cloned contiki3.0 from https://github.com/contiki-os/contiki.git.

At present, I have successfully compiled the wismote platform  with Orchestra and TSCH mac.

The problem is that I see that nodes associate at first EB received on broadcast  and then leave the network  due to  synchronization issue.
For example I get the error: ??!dl-miss TxBeforeTx? on the coordinator .
Checking  the file hereafter:

./core/net/mac/tsch/tsch-slot-operation.c:          TSCH_SCHEDULE_AND_YIELD(pt, t, current_slot_start, tsch_timing[tsch_ts_tx_offset] - RADIO_DELAY_BEFORE_TX, "TxBeforeTx");

I suspect that my wismote timers are wrong set:

/* Delay between GO signal and SFD: radio fixed delay + 4Bytes preample + 1B SFD -- 1Byte time is 32us
 * ~327us + 129preample = 456 us */
#define RADIO_DELAY_BEFORE_TX ((unsigned)US_TO_RTIMERTICKS(192)) //456
/* Delay between GO signal and start listening
 * ~50us delay + 129preample + ?? = 183 us */
//#define RADIO_DELAY_BEFORE_RX ((unsigned)US_TO_RTIMERTICKS(183))
/* Delay between the SFD finishes arriving and it is detected in software */
//#define RADIO_DELAY_BEFORE_DETECT 0

Do you know  what?s the procedure to compute the right TX/RX delay  timers?

Second, on the node side, the node associates but doesn't synchronize at all:

00:15.571        ID:2    TSCH: {asn-0.4c9 link-1-17-1-1 ch-26} !dl-miss association -31492 5248
00:15.581        ID:2    TSCH: {asn-0.4ca link-1-17-2-1 ch-20} !dl-miss association 28815 328
00:15.591        ID:2    TSCH: {asn-0.4d8 link-2-31-0-2 ch-26} !dl-miss association 28505 4592
00:15.600        ID:2    TSCH: {asn-0.4da link-1-17-1-1 ch-20} !dl-miss association 23933 656
00:26.443        ID:2    TSCH: send packet to 1 with seqno 1, queue 0 1, len 21 21
00:26.446        ID:2    TSCH: sending KA to 1
[...]

PR #1610, opened 2016-04-20
TSCH with Rime

simonduq:
This PR makes TSCH work with short addresses and Rime. Added new example for Rime/TSCH. Useful for testing TSCH without a full IPv6 stack.

Note that this PR includes https://github.com/contiki-os/contiki/pull/1609

### DIFF
--- a/examples/rime-tsch/node.c
+++ b/examples/rime-tsch/node.c
@@ -45,7 +45,6 @@
 #include "net/mac/tsch/tsch.h"
 
 const linkaddr_t coordinator_addr =    { { 1, 0 } };
-const linkaddr_t sender_addr =         { { 2, 0 } };
 const linkaddr_t destination_addr =    { { 1, 0 } };
 
 /*---------------------------------------------------------------------------*/
@@ -89,7 +88,7 @@ PROCESS_THREAD(unicast_test_process, ev, data)
 
     packetbuf_copyfrom("Hello", 5);
 
-    if(linkaddr_cmp(&sender_addr, &linkaddr_node_addr)) {
+    if(!linkaddr_cmp(&destination_addr, &linkaddr_node_addr)) {
       printf("App: sending unicast message to %u.%u\n", destination_addr.u8[0], destination_addr.u8[1]);
       unicast_send(&uc, &destination_addr);
     }
--- a/examples/rime-tsch/node.c
+++ b/examples/rime-tsch/node.c
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2016, Inria.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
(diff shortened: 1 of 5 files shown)


alignan:
We need to wait for https://github.com/contiki-os/contiki/pull/1609 to be merged, right?

simonduq:
I think that would be cleaner, yes. The idea was to give a chance to non-tsch experts to review the non-tsch related part first (https://github.com/contiki-os/contiki/pull/1609)

pablocorbalan:
I have tried the code of this PR and it works nicely. I tested it also with the reliable unicast (runicast) primitive and Contiki Collect and it seems to work fine with all of them. Nice work! 👍 

I may test this a bit more extensively (perhaps, even in some testbed). I may also try to create a simple scheduler for Contiki Collect + TSCH. I'll let you know my results!

One question: are the changes of commit [0c79d23](https://github.com/contiki-os/contiki/pull/1610/commits/0c79d23ed6cf5e0284bbfa0f1f4383c75da8e772) necessary? The functions that were used depend on `nbr-table.[ch]` but not on IP. I feel that we may make the code a bit more complex just to avoid using the `nbr-table`module. Also, if we would eventually update Contiki Collect to use the `nbr-table`module, perhaps, it would make more sense to keep using that module here too.

examples/rime-tsch/node.c
```diff
@@ -0,0 +1,100 @@
...
+  while(1) {
+    static struct etimer et;
+
+    etimer_set(&et, CLOCK_SECOND);
+
+    PROCESS_WAIT_EVENT_UNTIL(etimer_expired(&et));
+
+    packetbuf_copyfrom("Hello", 5);
+
+    if(linkaddr_cmp(&sender_addr, &linkaddr_node_addr)) {
```
pablocorbalan: You could change this line to:
`if(!linkaddr_cmp(&destination_addr, &linkaddr_node_addr)) {`
and then remove the definition of `sender_addr` on top. This way, if someone tries this example with multiple nodes, all the nodes will send unicast messages to the destination.

simonduq: sure! will do

simonduq: done

simonduq:
@pablocorbalan good point regarding the neighbor table for time source selection. Cannot seem to recall why I did this. I was assuming nbr-table was IP-only at some point but this is indeed not true. Will try to look into this again (or if you happen to do, as you mentioned you might use this, please share your conclusions!)

simonduq:
I did as suggested by @pablocorbalan, that is, revert back to using nbr-table for eb-autoselect. A few fixes were needed to get it to work with 2-byte addresses and with other recent changes. This PR is ready IMO (the diff has become very small!)

pablocorbalan:
@simonduq Sorry for the delay! I haven't really found the time to work more on this, so I have no updates. I might push in the future more examples to the `rime-tsch`folder.

I tried the updated code as well with a `runicast` application and `collect`and it also worked fine. This looks great to me with the last changes and using the `nbr-table` module. Thanks!
👍

simonduq:
setting timeout for merging

simonduq:
merging after timeout

sieben:
It doesn't seem to work with wismote as a target platform: http://pastebin.com/XwxSJcZf

simonduq:
That is right, TSCH is not supported on Wismote. More info about an ongoing porting effort at https://github.com/contiki-os/contiki/issues/1517

PR #1611, opened 2016-04-20
Coffee: improve comments and code formatting

nvt:
Improved mainly the comments, debug statements and source code formatting. Also changed type of sector indices to cfs_page_t in case <code>uint16_t</code> is too small.


alignan:
:+1:

PR #1613, opened 2016-04-21
Remove Deluge and associated test files.

nvt:
This Deluge implementation was created in 2007 for research paper experiments with ELF module distribution. It has not been used much since then, and the prevalent method of updating software in real IoT products today is full firmware updates that are typically transferred with more reliable bulk transfer protocols. Anyone interested in using it can of course still access it from the repository history and previous Contiki releases.


simonduq:
:+1: 
(don't hesitate to remove even more unused features btw :))

alignan:
Should this be documented elsewhere? I would suggest mentioning this on the wiki as done with the lost platforms.

simonduq:
good idea :)

nvt:
OK, I'm merging this and creating a Wiki page about lost modules.

nvt:
The new page on lost modules can be found here: https://github.com/contiki-os/contiki/wiki/Lost-modules

alignan:
Thanks!

PR #1619, opened 2016-04-22
Fix bug in TSCH for platforms with 16-bit rtimer

simonduq:
On platforms with 16-bit rtimer, the current `estimated_drift` is wrong when either (but not both) `expected_rx_time` or `rx_start_time` has wrapped. This sends the node straight off sync.

The fix introduces a new macro `RTIMER_CLOCK_DIFF`, defined exactly where `RTIMER_CLOCK_LT` was already set.

### DIFF
--- a/core/sys/rtimer.h
+++ b/core/sys/rtimer.h
@@ -55,10 +55,12 @@
 
 #include "contiki-conf.h"
 
-#ifndef RTIMER_CLOCK_LT
+#ifndef RTIMER_CLOCK_DIFF
 typedef unsigned short rtimer_clock_t;
-#define RTIMER_CLOCK_LT(a,b)     ((signed short)((a)-(b)) < 0)
-#endif /* RTIMER_CLOCK_LT */
+#define RTIMER_CLOCK_DIFF(a,b)     ((signed short)((a)-(b)))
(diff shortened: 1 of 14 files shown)


greg-king5:
I think that you should go further:
- Make `RTIMER_CLOCK_DIFF()` be the only platform-dependent macro.
- Redefine `RTIMER_CLOCK_LT()` to expand `RTIMER_CLOCK_DIFF()`.
- Move `RTIMER_CLOCK_LT()` to outside the preprocessor guard (it will be written in only one place).

simonduq:
True, will do!

simonduq:
Done as you suggested, thanks

core/sys/rtimer.h
```diff
@@ -57,9 +57,11 @@
 
 #ifndef RTIMER_CLOCK_LT
```
greg-king5: `RTIMER_CLOCK_LT` needs to be changed to `RTIMER_CLOCK_DIFF`, there and in the comments in the other files.

simonduq: Right, done!

greg-king5:
Looks good to me. :+1:

alignan:
:+1:

PR #1622, opened 2016-04-23
Fixing 04-rpl-large-network

simonduq:
This PR:
* Disables DAO-ACK by default, as it seemed to cause a failure on test `04-rpl-large-network.csc` with random seed 1, triggering a new run with seed 5. As this is one of the longest tests, this often triggered a Travis timeout;
* Uses a single seed for all simulation tests (instead of two). Now a failure on a single seed will cause a complete failure of the non-regression test.

### DIFF
--- a/core/net/rpl/rpl-conf.h
+++ b/core/net/rpl/rpl-conf.h
@@ -242,7 +242,7 @@
 #ifdef RPL_CONF_WITH_DAO_ACK
 #define RPL_WITH_DAO_ACK RPL_CONF_WITH_DAO_ACK
 #else
-#define RPL_WITH_DAO_ACK 1
+#define RPL_WITH_DAO_ACK 0
 #endif /* RPL_CONF_WITH_DAO_ACK */
 
 /*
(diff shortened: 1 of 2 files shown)


joakimeriksson:
I have taken a quick look at the test and from what I can see it is too many nodes to hold at a single time in the RPL Root which means that it is impossible for all nodes to have routes downwards at any given time. Having DAO ACK enabled will of course cause problems now and then - not having it on might also cause problems. The test seems to send from a node - mostly upwards (will likely work) and then RPL Root will send downward one hop. 

What about trying to setup some documentation per test? It would be very very valuable when they fail to understan what the idea is and if there is specific things such as testing to scale beyond what the Root can store, etc. In this case I am not sure if that is the test or not. If it is - there is 300 / 400 chance that the node that needs its downward route do not get it. But I can be wrong...

simonduq:
OK, I don't know if the intent of this particular test is to check what happens when the root doesn't have enough space for all nodes. If it is I feel like you, we have no way to guarantee things will work. I might also be missing something.

simonduq:
@joakimeriksson do you agree with disabling DAO ACK by default (merging this PR) until we come up with a better solution? (i.e. fixing/adjusting the test if need be, or DAO ACK if the problem lies there)

joakimeriksson:
Sure. As long as it is possible to enable I am fine with that!

simonduq:
ok!

PR #1624, opened 2016-04-24
Prevent uIP buffer over-read with big UDP packets

cedric-d:
When an UDP packet too big to fit in the uIP packet buffer is to be sent, the part fitting in the uIP buffer is copied to it (so no buffer overflow occurs) but uIP actually sends a packet of the original size therefore a buffer over-read occurs.

This modification makes uIP sends only the data copied to the packet buffer. An alternative would be not to send anything and return an error.

### DIFF
--- a/core/net/ip/uip-udp-packet.c
+++ b/core/net/ip/uip-udp-packet.c
@@ -51,12 +51,10 @@ void
 uip_udp_packet_send(struct uip_udp_conn *c, const void *data, int len)
 {
 #if UIP_UDP
-  if(data != NULL) {
+  if(data != NULL && len <= (UIP_BUFSIZE - (UIP_LLH_LEN + UIP_IPUDPH_LEN))) {
     uip_udp_conn = c;
     uip_slen = len;
-    memmove(&uip_buf[UIP_LLH_LEN + UIP_IPUDPH_LEN], data,
-            len > UIP_BUFSIZE - UIP_LLH_LEN - UIP_IPUDPH_LEN?
-            UIP_BUFSIZE - UIP_LLH_LEN - UIP_IPUDPH_LEN: len);
+    memmove(&uip_buf[UIP_LLH_LEN + UIP_IPUDPH_LEN], data, len);
     uip_process(UIP_UDP_SEND_CONN);
 
 #if UIP_CONF_IPV6_MULTICAST


simonduq:
You should consider rebasing, the PR fails on tests that were flaky at the time of your first submission, but are now much more stable

cedric-d:
Well... As you can see, regression tests still do not pass. I wonder if I correctly understand them.
I thought that 2 tests ([06-exp5438-udp-fragmentation-contikimac-large-sender.csc](https://github.com/contiki-os/contiki/blob/master/regression-tests/11-ipv6/07-exp5438-udp-fragmentation-nullrdc-large-sender.csc) and [07-exp5438-udp-fragmentation-nullrdc-large-sender.csc](https://github.com/contiki-os/contiki/blob/master/regression-tests/11-ipv6/07-exp5438-udp-fragmentation-nullrdc-large-sender.csc)) should succeed to transmit big packets with fragmentation and 2 tests ([08-exp5438-udp-fragmentation-contikimac-too-large-sender.csc](https://github.com/contiki-os/contiki/blob/master/regression-tests/11-ipv6/08-exp5438-udp-fragmentation-contikimac-too-large-sender.csc) and [09-exp5438-udp-fragmentation-nullrdc-too-large-sender.csc](https://github.com/contiki-os/contiki/blob/master/regression-tests/11-ipv6/09-exp5438-udp-fragmentation-nullrdc-too-large-sender.csc)) should fail to transmit too big packets. But in all these 4 tests, the test script is _fragmentation-should-receive-none.js_, so none of these tests should transmit packets!?

alignan:
Travis is happy now

simonduq:
I don't like that you had to change the tests, unless there was a flaw in the test (was there?).
Your first commit, that is, your original fix, looked good to me. Simple and with no unwanted side effect. I see no reason Travis would fail on it (the failure was probably on a flaky test). I would suggest to rebase on latest master and remove all commits but the first one (the history needs cleaning anyway as we don't want the merge commit). Happy to support this PR once this is done and if Travis is green.

cedric-d:
To my mind there are 2 problems with the IPv6 large sender tests:
- The buffer size is increased for the sender but not for the receiver which cannot reassemble the packets.
- The "large-sender" tests check that no packet is received like the "too-larger-sender" ones although, with the previous point corrected, they should check that all packets are received.

My first commit alone fails on the "too-large-sender" tests: If a packet is too large, it is no more dropped as expected by the tests, it is now truncated and successfully transmitted.

simonduq:
OK I would also be fine with the option of skipping too big packets

simonduq:
:+1:

alignan:
:+1:

PR #1626, opened 2016-04-24
core/cfs: Disallow seeking past the end of a file unless it is writable

gebart:
Files opened with `CFS_READ` should not allow seeking past the end of the file.

### DIFF
--- a/core/cfs/cfs-coffee.c
+++ b/core/cfs/cfs-coffee.c
@@ -1055,7 +1055,12 @@ cfs_seek(int fd, cfs_offset_t offset, int whence)
   }
 
   if(fdp->file->end < new_offset) {
-    fdp->file->end = new_offset;
+    if(FD_WRITABLE(fd)) {
+      fdp->file->end = new_offset;
+    } else {
+      /* Disallow seeking past the end of the file for read only FDs */
+      return (cfs_offset_t)-1;
+    }
   }
 
   return fdp->offset = new_offset;


core/cfs/cfs-coffee.c
```diff
@@ -1063,7 +1063,12 @@ cfs_seek(int fd, cfs_offset_t offset, int whence)
   }
 
   if(fdp->file->end < new_offset) {
-    fdp->file->end = new_offset;
+    if (FD_WRITABLE(fd)) {
+      fdp->file->end = new_offset;
+    } else {
+      /* Disallow seeking past the end of the file for read only FDs */
+      new_offset = fdp->file->end;
```
nvt: I prefer to <code>return (cfs_offset_t)-1</code> when the seeking fails so it is clear to the caller.

core/cfs/cfs-coffee.c
```diff
@@ -1063,7 +1063,12 @@ cfs_seek(int fd, cfs_offset_t offset, int whence)
   }
 
   if(fdp->file->end < new_offset) {
-    fdp->file->end = new_offset;
+    if (FD_WRITABLE(fd)) {
```
nvt: In Contiki code style, there should be no space after the <code>if</code>.

nvt:
This looks reasonable, but I had a couple of easy-to-fix source code comments.

gebart:
Addressed comments.
OK to squash?

nvt:
Thanks, it can be squashed.

gebart:
squashed

nvt:
:+1:

PR #1632, opened 2016-04-27
[WIP] Update Contiki Collect to use the NBR module

pablocorbalan:
This PR updates `collect-neighbor` to use the `nbr-table` module. This is basically a code refactoring of Contiki Collect to make it easier to understand its code and maintain it. This should also facilitate to extend Contiki Collect for other applications, e.g., adding a routing table for downwards communications or a link statistics module similar to [PR#1403](https://github.com/contiki-os/contiki/pull/1403). 

This is still work in progress (WIP), but I would prefer to get some feedback from the Contiki maintainers to see if these changes could be accepted in the future (after improving the PR) or if it is preferred to keep the current neighbor list as it is. 

The PR also updates all the examples (AFAIK) that depend on Contiki Collect. Moreover, it shortens the `collect-neighbor` function names. As a consequence, it may be a good idea to also change `collect-neighbor.[ch]`to `collect-nbr.[ch]`. I would also prefer to have the Collect code in a folder within the RIME folder to organise the RIME files a bit better. 

Please, let me know if you agree with the main idea of the PR and let me know your comments to improve it.


simonduq:
Haven't looked at the code and I'm not familiar with Contiki Collect, but using the neighbor table also there sounds like the way forward. I see no reason to oppose to such contribution, just keep us posted when you have something ready. Having a separate sub-folder for Collect also sounds good.

niklas88:
I'm interested in this work too since we will be using collect for a development project. Any pointers as to which parts are most work-in-progressy? Giving the code a quick look most changes seem to be quite simple refactoring. Also @simonduq would you be interested in me giving this a more thorough review?

simonduq:
I'll be ready to review as soon as this converges to a more final version (it is still WIP)

niklas88:
Ok so even though this pull request seems to have been abandoned by the author I now took a closer look at it. We are thinking of building an internal version of Rime Collect optimized for our use case and I wanted to begin this with cleaning up a bit (e.g. collect_send(), send_queued_packet() and retransmit_current_parent() overlap a lot).

So what I found is that this code works in simulation but not in my testbed, Debugging it, this seems to be due to the parent being 0.0 by default and nbr_tables treating linkaddr_null as a special value. So what I think is happening is that it's not just a valid key that gets used if the passed linkaddr pointer is NULL but because nbr_table_get_from_lladdr() doesn't check the used bit and entry for 0.0 will always be found.

That said, would there be interest in finishing this pull request and if yes, would it be reasonable to change nbr_table_get_from_lladdr()? Because other than these issues it looks to me to be almost finished with all references to collects own table mechanism already gone.

pablocorbalan:
Hi,

Sorry for all the delay! I have been busy with many things and disconnected from Contiki. Yes, I'd be interested in finishing the PR, but I am not sure how much time I can really spend on it.

When I submitted the PR, I tested the code in `Cooja` with the `collect`example and everything seemed to work, but I never tried the code on a Testbed, that is partially why it is still `WIP`. I'll try to see if I can do some testing on `Flocklab`and I'll get back to you. I'll check wether I get the same problems you are mentioning and in such case how to fix them.

In any case, @niklas88 you are more than welcome to make me a Pull Request to my branch to update/improve the code. If you want, we can also talk to see how to speed this up. You can also comment on specific parts of the code you may want to change in case there is something wrong.

niklas88:
@pablocorbalan so I've looked into this and to me it looks like the nbr_table somehow always has an entry for lladdr = 0.0 that also gets picked as best but only after another neighbor has been seen. I'm still trying to find out how to fix this though and I'm quite a bit confused how this element ends up int the table and why it only gets picked as best once another neighbor has been seen.

niklas88:
@pablocorbalan Ok digging deeper it turns out this was caused by an uninitialized use of the phase_nbr table when using ContikiMAC-Async which we also use in our current version. With the fix mentioned in this [Pull Request](https://github.com/contiki-os/contiki/pull/1549) I now have this working in hardware and without the workaround mentioned in the previous comment.

core/net/rime/collect-neighbor.c
```diff
@@ -76,241 +71,187 @@ MEMB(collect_neighbors_mem, struct collect_neighbor, MAX_COLLECT_NEIGHBORS);
...
-  /*  PRINTF("%d: ", node_id);*/
-  PRINTF("collect_neighbor_best: ");
-
-  /* Find the neighbor with the lowest rtmetric + linkt estimate. */
-  for(n = list_head(neighbors_list->list); n != NULL; n = list_item_next(n)) {
-    PRINTF("%d.%d %d+%d=%d, ",
-           n->addr.u8[0], n->addr.u8[1],
-           n->rtmetric, collect_neighbor_link_estimate(n),
-           collect_neighbor_rtmetric(n));
-    if(collect_neighbor_rtmetric_link_estimate(n) < rtmetric) {
```
niklas88: In the nbr_table version the best neighbor is only calculated based on the rtmetric not the rtmetric+link_estimate score. The rtmetric however is only updated for neighbors we hear, thus if a parent neighbor goes away or we moved we would still try to use it for its potentially better rtmetric even if we can't reach it anymore.

PR #1633, opened 2016-04-27
Renamed and moved the update neighbor table function in RPL. 

joakimeriksson:
This PR moves the function rpl_icmp6_update_nbr_table from RPL into uip-ds6-nbr so that it is in a more logical module. It is also changed to refresh a neighbor entry only - rather than both refreshing and adding if not already there. This makes the function more easy to understand and also more useful for applications or other modules that know that a neighbor is showing forward progress on the IP layer. This is useful for avoiding neighbor entry to go in to non-reachable state and cause NS to be sent (and packets to be lost).

The new function is uip_ds6_nbr_t *uip_ds6_nbr_refresh(const uip_ipaddr_t *ipaddr) which just refresh a IPv6 neighbor entry.

### DIFF
--- a/core/net/ipv6/uip-ds6-nbr.c
+++ b/core/net/ipv6/uip-ds6-nbr.c
@@ -78,6 +78,21 @@ uip_ds6_neighbors_init(void)
 }
 /*---------------------------------------------------------------------------*/
 uip_ds6_nbr_t *
+uip_ds6_nbr_refresh(const uip_ipaddr_t *ipaddr)
+{
+  uip_ds6_nbr_t *nbr;
+
+  if((nbr = uip_ds6_nbr_lookup(ipaddr)) != NULL) {
+#if UIP_ND6_SEND_NA
+    stimer_set(&nbr->reachable, UIP_ND6_REACHABLE_TIME / 1000);
+    nbr->state = NBR_REACHABLE;
+#endif /* UIP_ND6_SEND_NA */
+    return nbr;
+  }
+  return NULL;
+}
+/*---------------------------------------------------------------------------*/
+uip_ds6_nbr_t *
 uip_ds6_nbr_add(const uip_ipaddr_t *ipaddr, const uip_lladdr_t *lladdr,
                 uint8_t isrouter, uint8_t state, nbr_table_reason_t reason,
                 void *data)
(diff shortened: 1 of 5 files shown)


pablocorbalan:
The main idea of this PR is very similar to [1400](https://github.com/contiki-os/contiki/pull/1400). Should we then merge somehow both PRs? After our discussion back then, I think it could be a good idea to refresh the state of a neighbor after every incoming unicast IP message received as in [1400](https://github.com/contiki-os/contiki/pull/1400). That way in the `dis_input`and `dao_input` functions you would not need to worry about refreshing the state.

What do you think about this? I could update [1400](https://github.com/contiki-os/contiki/pull/1400) if required.

joakimeriksson:
It is similar yes - but #1400 will make all incoming packets always keep a neighbor refreshed which is as we also discussed not RFC compliant. In this PR I add the function and call from RPL which is know to do some non compliant tricks already so we are not better nor worse with this fix. The problem with #1400 is that it would need a configuration too allow this non-RFC-compliant refresh (we should in the future also have that in RPL - especially for DIO/DIS). The main thing with this PR is to clean-up some parts of the RPL code that had a function that was badly named and in the wrong module. But I agree - it is adding more or less the same function - maybe we could switch to your suggested name and add configuration for turning on / off the refresh for all packets?

What I have started to do at Yanzi is to allow applications to detect forward progress - e.g. any information in the packet that signals that we do not have a packet stuck in CSMA retransmission or similar. This can be TCP sequence numbers, new payload, timestamps, encryption sequence number or anything else. But we are doing it only at the application layer that knows that it is actually forward progress. We should att this in TCP - which is easy and have some nice examples.

simonduq:
Joakim is there something in this one that will not be covered by https://github.com/contiki-os/contiki/pull/1400 ?
My opinion is we should forward-progress thing aside for now (as it is complex and the problem it solves is not one that is likely to occur nor would have great consequences)

joakimeriksson:
Sure, forward progress is easy to add whenever we think we need it. I think the refresh parts of all unicast packets are ok, as long as we know it is passing via the nbr policy whenever that is active. So merging 1400 would make it possible to remove parts of this PR. I can try to "rebase" this if you pull 1400 ASAP.

simonduq:
Sounds good.
Le 11 juin 2016 10:34, "Joakim Eriksson" <notifications@github.com> a
écrit :

PR #1634, opened 2016-04-27
DAO ACK fix

simonduq:
This fix makes sure the `dao_sequence` stored in `my_dao_seqno` is the same used in the outgoing DAO


alignan:
Nice catch :+1:

joakimeriksson:
Looks good to me also! 👍

PR #1635, opened 2016-04-27
RPL: do not add K flag to no-path DAO as we discard the ACK anyway

simonduq:
We currently do not retransmit nopath DAOs nor do anything when getting an ACK for them. This PR removes the K flag from nopath DAOs to make sure no ACK is sent.


nvt:
:+1:

PR #1636, opened 2016-04-27
Trailing zeros in file leading to unexpected file length get set in destination buffer

frank-freihube-rmcan-com:
We encountered the problem with trailing zeros in files leading to wrong file size. This fix ensures that all bytes will get set to zero and works fine for us as it ensures that there are no zeros in the buffer on adresses we expected to get set to the content of the file. Also as long as there are enough bytes to read in the file the execution time should't increase at all. In cases that the given size only represents the size of the buffer and there are way less bytes to be read in the file the execution time will slightly increase.

### DIFF
--- a/core/cfs/cfs-coffee.c
+++ b/core/cfs/cfs-coffee.c
@@ -1100,7 +1100,7 @@ cfs_read(int fd, void *buf, unsigned size)
   file = fdp->file;
   
 #if COFFEE_IO_SEMANTICS
-  if(fdp->io_flags & CFS_COFFEE_IO_FIRM_SIZE) {
+  if(fdp->io_flags & CFS_COFFEE_IO_ENSURE_READ_LENGTH) {
     while(fdp->offset + size > file->end) {
       ((char*)buf)[--size] = 0;
     }
--- a/core/cfs/cfs-coffee.c
+++ b/core/cfs/cfs-coffee.c
@@ -1098,8 +1098,8 @@ cfs_read(int fd, void *buf, unsigned size)
 
   fdp = &coffee_fd_set[fd];
   file = fdp->file;
-  if(fdp->offset + size > file->end) {
-    size = file->end - fdp->offset;
+  while(fdp->offset + size > file->end) {
+    ((char*)buf)[--size] = 0;
   }
 
   /* If the file is not modified, read directly from the file extent. */
--- a/core/cfs/cfs-coffee.c
+++ b/core/cfs/cfs-coffee.c
@@ -1098,9 +1098,20 @@ cfs_read(int fd, void *buf, unsigned size)
 
   fdp = &coffee_fd_set[fd];
   file = fdp->file;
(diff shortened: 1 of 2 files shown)


nvt:
This PR introduces a semantic change, since it writes to parts of the buffer that have not been set to the file's contents. Hence, I think it is better if this would be configured with a flag in <code>cfs_coffee_set_io_semantics()</code>.

Background regarding trailing zeroes in Coffee:

In order to reduce code complexity, Coffee is designed not to preserve trailing zeroes in files after restarting a system because some kind of end-of-file marker would be required, and this could be costly to update on flash memory. Coffee does, however, preserve the zeroes if the file is re-opened, since it can keep the file size in the <code>struct file</code> object for the file. If cfs_read returns fewer bytes than expected, the caller should handle this condition by inspecting the return value of cfs_read.

frank-freihube-rmcan-com:
This sounds like a good idea and I changed it so it's configurable.

My workaround does not change anything in the file it only changes the way how an EOF during a read gets handled as it will ensure that some bytes in the destination buffer have an undefined state. Literally it only sets `readSize - fileLength` bytes to zero.
For example: We are using coffee to write binary data to the flash, most of the time the data is included in some sort of protocol with trailing checksums so we never have zeros at the end of the file.
In one case we are writing a struct with possible zeros in the last bytes. When we read the files we ignore the return value of `cfs_read()` nearly at all times as we (think we) know how many bytes are in the file because we know how many bytes we've written. We read our data after a restart and now some field is the destination struct (stack variable) have an undefined state because they do not get set by `cfs_read()`.
In this case we would have to do much work because we would have to check the return value at several places.

frank-freihube-rmcan-com:
Can someone give me a hint why the test fails now?
I made a copy/paste-error in commit 5bd7fc8 but the related test **4038** passed. I fixed the commit with commit 59727ed literally only changing `if(fdp->io_flags & 0x2)` to `if(fdp->io_flags & 0x4)` and now test **4039.9** fails. Can someone tell me what **4039.9** is checking?

I found the following `FAIL's` in the report but I'm not sure what they mean:
Line 430: mv: cannot stat `COOJA.testlog': No such file or directory FAIL
Line 658: 16-sky-ipv6-rpl-collect.1.faillog: FAIL
Line 875: ==> 16-sky-ipv6-rpl-collect.1.faillog <== FAIL

There is also a `SimulationCreationException` at `org.contikios.cooja.Cooja` in line 406 `"avax.swing.plaf.BorderUIResource cannot be cast to java.lang.Boolean"`
Might this be the error?

nvt:
The change to use the IO semantics functionality looks good. The Travis error is not related to the changes in this PR, and a separate PR has been made to fix the error. I'll restart this test to see if it goes through now.

frank-freihube-rmcan-com:
Great. It passed now.
Thank you.

nvt:
:+1:

PR #1637, opened 2016-04-27
RPL default lifetime

simonduq:
I'm pretty sure the intended default lifetime was 30 times 60 seconds, not 48 times 96 seconds.


joakimeriksson:
Good catch and you are absolutely correct! 👍

alignan:
:+1:

PR #1638, opened 2016-04-27
Fixing Cooja message list to avoid any Swing components when headless

joakimeriksson:
This fix make Cooja not use the MessageList class that do have a lot of Swing components that now and then cause sporadic exceptions when running in headless mode. This introduce a text-based version that just prints the messages on standard out - the Text based MessageList is used whenever Cooja is run in headless mode.


simonduq:
Well all I can say is that this is the first PR I see succeeding on the ipv6 test for the first time since we switched to single-seed Travis tests.
Re-started the test it to see if we get the same result again :)

joakimeriksson:
Sounds good! I think Cooja in general needs to be more headless when it is run in headless. It is quite a risk to create lots of Swing / AWT components when not having a GUI.

pablocorbalan:
This PR includes in the first commit your previous PR [1633](https://github.com/contiki-os/contiki/pull/1633). It would be good if you can separate them.

joakimeriksson:
@pablocorbalan good catch - my bad - fixed now!

fros4943:
Look great @joakimeriksson! 

:+1: Merging

PR #1639, opened 2016-04-27
Handle the case when log->link is NULL in tsch_log_process_pending

yatch:
This PR prevents possible NULL pointer access for `log->link` in `tsch_log_process_pending()`.

`log->link` would have NULL if `current_link` is NULL when `tsch_log_prepare_add()` is called. This could happen in `tsch_slot_operation` Protothread. Here is the relevant code block, where the TSCH_LOG_ADD macro expands with `tsch_log_prepare_add()` when TSCH_LOG_LEVEL is greater than or equal to 2.

```C
if(current_link == NULL || tsch_lock_requested) { /* Skip slot operation if there is no link
                                                      or if there is a pending request for getting the lock */
  /* Issue a log whenever skipping a slot */
  TSCH_LOG_ADD(tsch_log_message,
                  snprintf(log->message, sizeof(log->message),
                      "!skipped slot %u %u %u",
                        tsch_locked,
                        tsch_lock_requested,
                        current_link == NULL);
  );
```

### DIFF
--- a/core/net/mac/tsch/tsch-log.c
+++ b/core/net/mac/tsch/tsch-log.c
@@ -84,11 +84,15 @@ tsch_log_process_pending(void)
   }
   while((log_index = ringbufindex_peek_get(&log_ringbuf)) != -1) {
     struct tsch_log_t *log = &log_array[log_index];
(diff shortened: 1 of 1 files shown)


simonduq:
Great catch! :+1:

PR #1644, opened 2016-05-02
Do not use setBaudrate

jdede:
The the author of the library *pyserial* removed the command `setBaudrate()` starting from version 3.0. This PR solves problems caused by that in the bsl by setting the baudrate directly.
*pyserial* 3.0 is just upcoming for example in Debian stretch.

### DIFF
--- a/tools/sky/msp430-bsl-linux
+++ b/tools/sky/msp430-bsl-linux
@@ -1162,8 +1162,9 @@ class BootStrapLoader(LowLevel):
         sys.stderr.flush()
         self.bslTxRx(self.BSL_CHANGEBAUD,   #Command: change baudrate
                     a, l)                   #args are coded in adr and len
+        self.serialport.flush()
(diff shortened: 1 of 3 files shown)


alignan:
Does this works also for versions lower than 3.0?

jdede:
Yes. This is exactly the way how the baudrate is set in the corresponding examples in *pyserial* even in version 2.7 and earlier.

jdede:
Can anyone restart the Travis CI build to get rid of the red icon? ;-)

alignan:
Done! the failure is not-related :-)

jdede:
Thanks, I know. But it is ugly anyway- ;-)

g-oikonomou:
Where are we with this one? This seems a bit of a no-brainer?

alignan:
Forgot about this one, we were waiting for Travis go to apple-happy.

PR #1645, opened 2016-05-02
Fix broken include

jdede:
In case of using IPv4,  `contiki-z1-main.c` uses a wrong include.


alignan:
I just restarted the job, let's wait for Travis... good catch :smile:

Are you actively testing IPv4?

jdede:
Not really. Finding it was just a coincidence caused by a copy an paste error for a quick test from my side... ;-)

simonduq:
:+1:

PR #1646, opened 2016-05-02
Fix putchar include when using IPv4

jdede:
In case of using IPv4, `putchar()` is provided by the SLIP device. The missing `#ifndef` result into an error: `multiple definition of 'putchar'`

### DIFF
--- a/cpu/msp430/dev/uart0-putchar.c
+++ b/cpu/msp430/dev/uart0-putchar.c
@@ -1,9 +1,12 @@
-#include "dev/uart0.h"
 #include <stdio.h>
+#include "dev/uart0.h"
 
+#ifndef NETSTACK_CONF_WITH_IPV4
+/* In case of IPv4: putchar() is defined by the SLIP driver */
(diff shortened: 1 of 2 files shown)


alignan:
:+1:

jdede:
When going through the code once again, I figured out that the `#if` makes more sense than an `#ifdef`. So it works properly even in case of `#define NETSTACK_CONF_WITH_IPV4 0`.

greg-king5:
`<stdio.h>` doesn't need to be included when that function is "hidden" (it will give slightly faster building).

jdede:
Of course, I can do so. ;-)

simonduq:
:+1:

greg-king5:
Removing it isn't quite what I meant.

Including `<stdio.h>` is a double-check that the function definition's prototype is correct.  That check isn't needed _when_ the definition isn't compiled.  Therefore, that `#include` should have been moved down to a line just below the `#if` line.

PR #1651, opened 2016-05-04
oma-lwm2m: Use sv for string value, bv for boolean

gebart:
This fixes JSON parsing errors when reading entire instances from the Leshan example server web GUI.

http://dev_devtoolkit.openmobilealliance.org/IoT/LWM2M10/doc/TS/index.html#!Documents/json.htm

PR #1652, opened 2016-05-04
oma-lwm2m: Use Accept header to determine writer object

gebart:
~~depends on #1651~~ - merged
This simplifies using Copper for debugging LWM2M resources since it lets us look at the variables in plain text.

Also changes the decision to use the provided Accept header instead of Content-Format header to determine if we are sending an APPLICATION_LINK_FORMAT or LWM2M_JSON response for dumping entire service instances.


joakimeriksson:
Looks like a very nice improvement - gets a 👍 from me.

nvt:
:+1:

PR #1654, opened 2016-05-04
oma-lwm2m: Add JSON writer

gebart:
A minimal JSON writer for LWM2M resources.

Depends on #1652, ~~#1651~~

### DIFF
--- a/apps/oma-lwm2m/Makefile.oma-lwm2m
+++ b/apps/oma-lwm2m/Makefile.oma-lwm2m
@@ -1,5 +1,13 @@
-oma-lwm2m_src = lwm2m-object.c lwm2m-engine.c \
-	lwm2m-device.c lwm2m-server.c lwm2m-security.c \
-	oma-tlv.c oma-tlv-reader.c oma-tlv-writer.c \
-	lwm2m-plain-text.c
+oma-lwm2m_src = \
+  lwm2m-object.c \
+  lwm2m-engine.c \
+  lwm2m-device.c \
+  lwm2m-server.c \
+  lwm2m-security.c \
+  oma-tlv.c \
+  oma-tlv-reader.c \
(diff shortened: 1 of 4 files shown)


joakimeriksson:
Did you test with Leshan or Wakaama? I think it looks good - nice to see progress on the LWM2M codebase.

gebart:
@joakimeriksson I used Copper for testing the service response. The Leshan server is still using OMA TLV formatted responses with this PR.

nfi:
http://www.gnu.org/software/libc/manual/html_node/Formatted-Output-Functions.html

gebart:
The reason for the greater than rather than greater than or equal is that the CoAP response payload will not contain the trailing null byte and uses the length of the response to determine which bytes to copy, so no trailing null byte is necessary here.

nfi:
I aggree that the trailing null byte should not be necessary here but the problem is that different implementations of `snprintf()` handles this differently. Modern implementations tend to always add the trailing null byte, discarding the last "real" byte if needed.

Example compiled with gcc:
>   static char buf[5];
>   int ret = snprintf(buf, 4, "abcd");
>   printf("ret=%d s='%s'\n", ret, but);

> => ret=4 s='abc'

gebart:
@nfi Yes, you are correct, the output may have been corrupted if the buffer is filled since the last letter may have been replaced by a null byte.

gebart:
Corrected the output length checks.
OK to squash?

nfi:
Thanks for the update. I think this looks good and is ready to squash, but please remove the extra space after `if` (code style).

gebart:
rebased, squashed. #1652 should be reviewed and merged before merging this.

alignan:
#1652 has been merged already, restarting the travis job.

nvt:
:+1:

PR #1657, opened 2016-05-09
Fix for IPv6 compilation and linkage errors

tscheffl:
Two errors have been spotted, when IPv6 is enabled in the **ravenusb**
Project-Makefile:
`#CONTIKI_NO_NET=1`
`CONTIKI_WITH_IPV6=1`

The compile error results from a variable name mismatch in `cdc_task.c`
The variable `r` is undeclared and should be renamed to `route`

The linker also fails with _undefined references_
This has been mediated by adding `core/net` to Modules in the
Project-Makefile.


nvt:
:+1:

Issue #814, opened 2014-10-10
rtimer: Calling rtimer_set before previous rtimer has expired calls wrong function too early


gebart:
Looking at the code for `rtimer_set` in core/sys/rtimer.c it seems like calling `rtimer_set` with a new timer before the first one has expired will cause the second callback function to be called at the time when the first timer was supposed to be called, and the first callback will simply be forgotten. Nothing will be called at the second timer's expiration, since `rtimer_arch_schedule` is never called again.

I believe this is a bug in the core rtimer library.

I am trying to use an rtimer to trigger etimer and clock updates on the Mulle platform (PR #776), but it makes the clock and etimers to stop working after using an rtimer for anything.

Below is the terminal output from the timer-test in examples/cc2538dk, running on my Mulle board:

    2014-10-10 11:52:57,609 - INFO # -----------------------------------------
    2014-10-10 11:52:57,612 - INFO # Rtimer Test, 1 sec (32768 rtimer ticks):
    2014-10-10 11:52:57,614 - INFO # =======================
    2014-10-10 11:52:57,618 - INFO # Now=75788 (clock = 148) - For=108556
    2014-10-10 11:52:57,624 - INFO # Task called at 76289 (clock = 148)

The clock callback is forgotten once the rtimer_set call is made from the test thread, and the rtimer callback is made too early, at the time of the next clock tick (76289) instead of at the rtimer target time (108556). After the above lines are printed the clock stops ticking, and the test app just hangs around waiting for an etimer that never will expire.

yatch:
I also think this is a bug.... And it looks like #1290, a "nominated" PR, will resolve the issue by replacing the problematic code in rtimer.c.

bkozak-scanimetrics:
> I also think this is a bug.... And it looks like #1290, a "nominated" PR, will resolve the issue by replacing the problematic code in rtimer.c.

As the author of #1290, I'm not sure that this should be considered a bug since the current rtimer implementation really doesn't support the use of multiple rtimers simultaneously at all and it doesn't seem that it is meant to either.

Even with #1290 the "buggy" behaviour is still present if ```RTIMER_MULTIPLE_ACCESS``` is false (and this is also the default). However with false ```RTIMER_MULTIPLE_ACCESS``` you would have the option to ```rtimer_cancel``` the old timer before setting a new one.

simonduq:
[...]

PR #1661, opened 2016-05-12
TSCH on Cooja mote 

yatch:
## Introduction

This PR aims to port TSCH to the Cooja mote platform.

Three major changes are made:
- re-implement `rtimer` as a 1 MHz timer to comply with the resolution requirement of TSCH, 32 KHz or higher
- add the radio features required for TSCH, including the polling mode
- add Cooja mote dependent code in the TSCH logic to manage busy waiting

This PR includes a sample simulation file, `examples/ipv6/rpl-tsch/rpl-tsch-cooja.csc`, which is based on `rpl-tsch.csc`.

To run TSCH on Cooja motes, you need to disable the simulated turnaround feature implemented in `cooja-radio.c` with the `COOJA_CONF_SIMULATE_TURNAROUND` macro that is newly introduced. In `examples/ipv6/rpl-tsch/project-conf.h`, the following lines are appended.

``` C
#if CONTIKI_TARGET_COOJA
#define COOJA_CONF_SIMULATE_TURNAROUND 0
#endif /* CONTIKI_TARGET_COOJA */
```
## Concerns / Disadvantages

There are three concerns:

Firstly, this PR apparently changes the `etimer` and `rtimer` logics of the Cooja mote platform, which are very fundamental parts in the Contiki system. Such changes may cause unexpected system behaviors which are not covered by the regression tests.

One noteworthy change is `rtimer_rach_now()`. The function does not call `cooja_mt_yield()` any more. If there is any code expecting to advance the time by calling `rtimer_arch_now()`, it needs to be resolved.

Speaking of the regression tests, I noticed `regression-tests/12-rpl/04-rpl-large-network` took longer time than before. And success rate of `regression-tests/21-large-rpl/01-cooja-http-socket-50` was low. These things may be caused by the changes.

Secondly, Cooja mote dependent code is introduced into the TSCH module. I tried to keep from modifying the module; but I could not find a good way to realize busy waiting on the platform... I borrowed the idea in `core/net/mac/nullrdc.c` that is to set `1` to `simProcessRunValue` and to call `cooja_mt_yield()` in a busy loop. This works in most cases except for `BUSYWAIT_UNTIL_ABS(cond, t0, offset)` with `0` for `cond`.

This type of `BUSYWAIT_UNTIL_ABS()` is used in the `TSCH_SCHEDULE_AND_YIELD` macro. Essentially, the macro makes a process sleep for `offset - RTIMER_GUARD` with `rtimer_set` and wait actively until its target time with `BUSYWAIT_UNTIL_ABS(0, ref_time, offset)`. Since `BUSYWAIT_UNTIL_ABS(0, ref_time, offset)` does not work well as mentioned before, `RTIMER_GUARD` is set to `0` for the Cooja platform to wake up on the exact time it expects.

Lastly, I decided to disable the simulated turnaround implemented around [cooja-radio.c#L204](https://github.com/yatch/contiki/blob/6b1367364dcfbdabe916eb275575d28c60bc15b1/platform/cooja/dev/cooja-radio.c#L204) for TSCH because it causes desynchronizaion. If the feature is important to TSCH as well, it needs another way.
## Screenshot

tsch-cooja-mote
## Memo

Things I noticed during the implementation.
### Transmission duration calculation in ContikiRadio

The comment at [ContikiRadio.java#L357](https://github.com/yatch/contiki/blob/6b1367364dcfbdabe916eb275575d28c60bc15b1/tools/cooja/java/org/contikios/cooja/contikimote/interfaces/ContikiRadio.java#L357) seems obsolete.

``` C
356       /* Calculate transmission duration (us) */
357       /* XXX Currently floored due to millisecond scheduling! */
358       long duration = (int) (Simulation.MILLISECOND*((8 * size /*bits*/) / RADIO_TRANSMISSION_RATE_kbps));
359       transmissionEndTime = now + Math.max(1, duration);
```

In addition, `(Simulation.MILLISECOND*((8 * size /*bits*/) / RADIO_TRANSMISSION_RATE_kbps))` produces microseconds to send the bits unless I am mistaken... I'm also not sure why casting with `int` is used there instead of `long`.
### Handling subsequent tasks passed to `rtimer_set`

Why `next_rtimer` is updated at [rtimer.c#L83](https://github.com/yatch/contiki/blob/6b1367364dcfbdabe916eb275575d28c60bc15b1/core/sys/rtimer.c#L83) regardless of `first`...? When multiple tasks are set with `rtimer_set` at a certain tick, the last task passed to the function would be executed at the expiration of the first task. I might overlook something.

``` C
66 int
67 rtimer_set(struct rtimer *rtimer, rtimer_clock_t time,
68            rtimer_clock_t duration,
69            rtimer_callback_t func, void *ptr)
70 {
71   int first = 0;
...
75   if(next_rtimer == NULL) {
76     first = 1;
77   }
...
83   next_rtimer = rtimer;
84
85   if(first == 1) {
86     rtimer_arch_schedule(time);
87   }
88   return RTIMER_OK;
89 }
```

This is related to #814.


simonduq:
Wow. That is awesome. I never looked very closely at Cooja motes and wasn't really sure how feasible this would be to be honest ;) I will review this is more details at some point, but what I can say now: (1) the feature is super-great (2) the modifications to core TSCH are minor enough to be acceptable (3) we need to think about running some of the Travis tests with TSCH enabled.

yatch:
@simonduq Thanks. By the way, it seems Orchestra also works well with `MAKE_WITH_ORCHESTRA=1`. :-)

simonduq:
Ok I tried and it works very nicely for me. I'm fine with the changes to core. Did not review changes to Cooja motes. I'm concerned with side-effects on regression tests as you mention above. Is there a way to somehow make all changes conditional, and make sure things are just like before as long as TSCH is not enabled?

yatch:
@simonduq Thank you for your trying!

With regard to the regression test, I ran the Travis test again on this branch last night. It succeeded without any failure. I got all green three times in a row so far. The low success rate I experienced before may be due to my bad luck...?

Your idea, making the changes conditional is definitely one of options. It could be sufficient to target the changes related only to clock and timer for the conditional compilation. But my initial impression is that it would not be easy because the changes are not only in the C part but also in Java part....

In any case, I'll take a closer look at RPL behaviors after rebase.

simonduq:
Sounds good, thanks

yatch:
I confirmed there are a couple of behavioral differences between the `master` branch and this PR branch, `pr-tsch-cooja-mote`. My conclusion is that they are not bugs.

Here is what I did.

First of all, I made some measurements on the following concerns about regression test results.

> Speaking of the regression tests, I noticed `regression-tests/12-rpl/04-rpl-large-network` took
> longer time than before. And success rate of
> `regression-tests/21-large-rpl/01-cooja-http-socket-50` was low.

I ran the test of `regression-tests/12-rpl/04-rpl-large-network` and measured the execution time twenty times on a local machine with `master` and `pr-tsch-cooja-mote` respectively. Here is the result, where each number represents the number of seconds. As I experienced before, `pr-tsch-cooja-mote` took more time than `master`.

                   | min   | max   | mean  |
 ------------------|-------|-------|-------|
 master            |543.297|627.602|583.065|
 pr-tsch-cooja-mote|623.192|663.696|638.712|

Next, I ran `regression-tests/21-large-rpl/01-cooja-http-socket-50` on Travis with my account to see any difference of success rate between `master` and `pr-tsch-cooja-mote`. The result is that both of them succeeded ten out of ten times. Now, I think some external condition were not good for the test last month when I sent this PR.

Then, I focused on `04-rpl-large-network`. I applied the patch of #1502 to both of `master` and `pr-tsch-cooja-mote` and examined packets captured during the test of `04-rpl-large-network`. There are two findings.

The first thing is that I saw several nodes of the `master` case sent DIS packets later than the same nodes of the `pr-tsch-cooja-mote` case. This is because `etimer` of the `master` branch occasionally expires later than its expiration. This happens if a packet is received  before the expiration. This is not the case with `pr-tsch-cooja-mote`. 

The other thing is also related to packet reception. I saw a certain node of the `pr-tsch-cooja-mote` case missed an ACK frame, on the other hand, the same node of the `master` case received the ACK frame sent at the exact same timing. The packet reception affects the ACK waiting logic in `nullrdc` of the `master` branch, and this causes such a difference.

These things happen basically because the Cooja mote of the `master` branch ignores the microsecond part of time. To confirm my theory, I applied the following patch to `pr-tsch-cooja-mote` and ran `04-rpl-large-network` with it.


```C
diff --git a/core/net/mac/nullrdc.c b/core/net/mac/nullrdc.c
index 7ba62a4..eecad42 100644
--- a/core/net/mac/nullrdc.c
+++ b/core/net/mac/nullrdc.c
@@ -155,9 +155,17 @@ send_one_packet(mac_callback_t sent, void *ptr)
           rtimer_clock_t wt;
 
           /* Check for ack */
+#if CONTIKI_TARGET_COOJA
+          wt = clock_time();
+#else
           wt = RTIMER_NOW();
+#endif
           watchdog_periodic();
+#if CONTIKI_TARGET_COOJA
+          while(RTIMER_CLOCK_LT(clock_time(), wt + ACK_WAIT_TIME / 1000)) {
+#else
           while(RTIMER_CLOCK_LT(RTIMER_NOW(), wt + ACK_WAIT_TIME)) {
+#endif
 #if CONTIKI_TARGET_COOJA || CONTIKI_TARGET_COOJA_IP64
             simProcessRunValue = 1;
             cooja_mt_yield();
diff --git a/tools/cooja/java/org/contikios/cooja/contikimote/interfaces/ContikiClock.java b/tools/cooja/java/org/contikios/cooja/contikimote/interfaces/ContikiClock.java
index 97cbc61..6715193 100644
--- a/tools/cooja/java/org/contikios/cooja/contikimote/interfaces/ContikiClock.java
+++ b/tools/cooja/java/org/contikios/cooja/contikimote/interfaces/ContikiClock.java
@@ -157,7 +157,7 @@ public class ContikiClock extends Clock implements ContikiMoteInterface, PolledB
 
     /* Request tick next wakeup time for Etimer */
     long etimerNextExpirationTime = (long)moteMem.getInt32ValueOf("simEtimerNextExpirationTime") * Simulation.MILLISECOND;
-    long etimerTimeToNextExpiration = etimerNextExpirationTime - moteTime;
+    long etimerTimeToNextExpiration = etimerNextExpirationTime - (moteTime / 1000) * 1000;
     if (etimerTimeToNextExpiration <= 0) {
       /* logger.warn(mote.getID() + ": Event timer already expired, but has been delayed: " + etimerTimeToNextExpiration); */
       /* Wake up in one millisecond to handle a missed Etimer task


```

I got a packet capture file which has the exact same packet interval with the `master` case. And, the execution time of the patched `pr-tsch-cooja-mote` is close to `master`.

                   | min   | max   | mean  |
 ------------------|-------|-------|-------|
 master            |543.297|627.602|583.065|
 pr-tsch-cooja-mote|623.192|663.696|638.712|
 patched           |541.892|619.882|597.646|

To compare the number of packets of `04-rpl-large-network`, `master` is 33,828 packets, `pr-tsch-cooja-mote` is 37,281 packets. I believe the behavioral differences in `nullrdc` and in `etimer` cause more packets in the case of `pr-tsch-cooja-mote`. More packets lead more events to handle, then longer execution time. This can explain why `pr-tsch-cooja-mote` takes longer time to run `04-rpl-large-network` than `master`.

simonduq:
Thanks a lot for this in-depth analysis!

simonduq:
:+1: , setting a timeout for merging.
Cooja experts, shout if you don't want this merged. Am happy with the PR because of the added feature and the fact Travis still passes makes me confident this does not break the Cooja mote.
@yatch , rebase needed

yatch:
@simonduq Rebased :-)

simonduq:
Merging after timeout, thanks for the awesome contrib!!

hsaxena-vub:
Hello, I am working on TSCH in cooja mote and i want to check energy consumption on tsch. But the problem is i didn't find collect view option. Can anyone help me how i find energy consumption in this situation. Thanks in advance.

PR #1485, opened 2016-01-28
Add support for the CC2650 LaunchPad (LAUNCHXL-CC2650)

g-oikonomou:
This adds support for one more CC26xx board: the new LaunchPad

We support LEDs and Buttons and also the LP's external flash.

The flash is a Macronix part with an instruction set that is basically a superset of the one of the Winbond part present on Sensortags. Thus, we extend the sensortag flash driver to support this additional part and we use the same driver for both boards. We create a `common` dir under `srf06-cc26xx` and we put the flash and SPI drivers therein. These are shared across the two boards for now, but it's really meant as a pool of drivers for devices present on more than one board.

This pull also extends cc26xx examples to support the LP and adds compile tests for LP builds

PR #1663, opened 2016-05-13
Add upload script to Makefile.launchpad

pablocorbalan:
This PR simply adds the upload command to the Makefile of the CC2650 Launchpad addressing the comment of @joakimeriksson on the already merged [#1485](https://github.com/contiki-os/contiki/pull/1485) pull request. I also confirm that the BSL script worked for me on Mac OS X without any problem after addressing the changes described by @g-oikonomou and also setting the `SET_CCFG_BL_CONFIG_BL_LEVEL` to `0x0` in the `ccfg.c` file.

I have set `115200` as predefined baud rate for the script.

### DIFF
--- a/platform/srf06-cc26xx/launchpad/Makefile.launchpad
+++ b/platform/srf06-cc26xx/launchpad/Makefile.launchpad
@@ -4,3 +4,19 @@ CONTIKI_TARGET_DIRS += launchpad common
 
 BOARD_SOURCEFILES += board.c launchpad-sensors.c leds-arch.c button-sensor.c
 BOARD_SOURCEFILES += ext-flash.c board-spi.c
+
+PYTHON = python
+BSL_FLAGS += -e -w -v
+
+ifdef PORT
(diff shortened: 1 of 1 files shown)


g-oikonomou:
Thank you, Pablo, this is something I've meant to do since Joakim's comment, so it's great that you have.

The script should be intelligent enough to default to baud rate 500000 for this board, so I think the -b harms more than it helps. Can you try without it please?

pablocorbalan:
@g-oikonomou I have removed the `-b 115200`option. The script also works nicely with the default baud rate `500000`.

g-oikonomou:
Many thanks, I'll merge this one.

Moving forward (not in this pull, but in the future), I want to have this entire block in `Makefile.srf06-cc26xx`, it is currently duplicated in too many makefukes, The board-specific Makefiles will simply set a make variable to signal that they support the `.upload` target. Default = no I'd suggest.

g-oikonomou:
Oh yay, it passed!

PR #1665, opened 2016-05-14
Consolidate .upload for all CC26xx/CC13xx boards and add login make target

g-oikonomou:
The block that controls the `.upload` target is unnecessarily replicated in multiple sub-board Makefiles. This was originally done because the SmartRF and the Launchpad can be programmed with the c2538-bsl script, whereas the sensortag cannot.

This commit moves the `cc2538-bsl` / `.upload` target logic to the top level platform Makefile (`platform/srf06-cc26xx/Makefile.srf06-cc26xx`). Board makefiles simply set the make variable `BOARD_SUPPORTS_BSL` to 1 to signal that they can be programmed by the BSL script. If `BOARD_SUPPORTS_BSL` is not equal to 1, trying to use the `.upload` target will return an error message.

For example:

```
$ make BOARD=sensortag/cc2650 cc26xx-demo.upload
using saved target 'srf06-cc26xx'
This board cannot be programmed with the ROM bootloader and therefore does not support the .upload target.
```

### DIFF
--- a/cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
+++ b/cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
@@ -81,6 +81,15 @@ CONTIKI_SOURCEFILES += $(CONTIKI_CPU_SOURCEFILES) $(DEBUG_IO_SOURCEFILES)
 TARGET_START_SOURCEFILES += fault-handlers.c $(TI_XXWARE_STARTUP_SRCS)
 TARGET_STARTFILES = $(addprefix $(OBJECTDIR)/,$(call oname, $(TARGET_START_SOURCEFILES)))
 
+PYTHON = python
+BSL_FLAGS += -e -w -v
+
+ifdef PORT
+  BSL_FLAGS += -p $(PORT)
+endif
+
+BSL = $(CONTIKI)/tools/cc2538-bsl/cc2538-bsl.py
+
 ### Don't treat the .elf as intermediate
 .PRECIOUS: %.elf %.hex %.bin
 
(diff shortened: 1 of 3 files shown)


g-oikonomou:
In fact, I cannot see why this should be in the platform makefile instead of the CPU makefile (since it's a CPU feature, common to all chips of both families). 

I've thus moved the logic to `cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx` (hence the close and re-open). This is now ready for your review.

pablocorbalan:
George, this looks good to me! I was in fact implementing your suggestion in a similar way, see [this](https://github.com/pablocorbalan/contiki/commit/e7f38aee617e74099b76912599c73c20f33149be). But your changes look better! I see no problem to move the logic to the CPU Makefile. You will need to adapt [#1666](https://github.com/contiki-os/contiki/pull/1666) to also include `BOARD_SUPPORTS_BSL = 1`in the CC1310 Makefile.
👍

atiselsts:
Hi George, would you mind also including `login` target? Otherwise :+1:

g-oikonomou:
> You will need to adapt #1666 to also include `BOARD_SUPPORTS_BSL = 1` in the CC1310 Makefile.

@pablocorbalan I am fairly certain this is not required ;)

> Hi George, would you mind also including login target? 

@atiselsts Can you open a pull on my repo on top of contrib/cc26xx-bsl. The .upload target should probably also be in the CPU makefile (?)

pablocorbalan:
Sorry, my mistake! I didn't realize that [#1666](https://github.com/contiki-os/contiki/pull/1666) uses the same launchpad subfolder, that's why it is not required. Cool!

g-oikonomou:
I've added Atis' commit that enables the login make target for all CC26xx/CC13xx boards and I've updated the title of this pull accordingly

alignan:
:+1:

PR #1666, opened 2016-05-14
Add support for the CC1310 LaunchPad (LAUNCHXL-CC1310)

g-oikonomou:
This pull request adds support for TI's CC1310 Launchpad.

We support LEDs and Buttons and also the LP's external flash. Examples under `examples/cc26xx` should work off-the-shelf.

According to the CC1310LP design files, the flash on it is the same as the flash on the CC2650 LP (MX25R8035F 8Mbit). However, the flash on some CC1310 LPs reportedly identifies itself as the 16Mbit one (MX25R1635F) instead. Instruction sets are identical, we simply need to tell the driver to recognize this part's Device ID. Thus, this pull also extends the driver to support this part, in case any user happens to have one of those launchpads.

Build with `BOARD=launchpad/cc1310`.

This pull also adds a travis test for the CC1310 LP build.


alignan:
:+1:

PR #1670, opened 2016-05-19
oma-lwm2m: Handle text/plain floatfix numbers without decimal point

gebart:
Fixes bug: If no decimal point is present then the entire number is treated as the decimal part instead of as the integer part


joakimeriksson:
Nice catch and fix. Just a little style issue with a extra space after the if. Please fix that and it should be good to be pulled.

gebart:
removed space after `if`

nfi:
Nice, I think this looks good now.  I think Travis failed for unrelated reasons.

nvt:
:+1: Travis succeeded now. Merging.

Issue #1671, opened 2016-05-20
Application apps/orchestra possible bug


warner83:
Hello, 
in the function new_time_source within the file orchestra-rule-unicast-per-neighbor.c (line 169) the function remove_uc_link is called. 
However, the argument is new_addr, shouldn't it be the address of the old time source?
@simonduq can you comment on this?

Thanks!

simonduq:
It does look like a bug yes. Note that this implementation of Orchestra is not as much tested as the original one (https://github.com/simonduq/orchestra). Thanks for reporting!

simonduq:
fixed via https://github.com/contiki-os/contiki/pull/1672

PR #1672, opened 2016-05-20
Various Orchestra fixes

simonduq:
This PR includes a number of fixes to Orchestra, including a fix to https://github.com/contiki-os/contiki/issues/1671

### DIFF
--- a/apps/orchestra/orchestra-rule-eb-per-time-source.c
+++ b/apps/orchestra/orchestra-rule-eb-per-time-source.c
@@ -83,14 +83,24 @@ new_time_source(const struct tsch_neighbor *old, const struct tsch_neighbor *new
 
   if(old_ts != 0xffff) {
     /* Stop listening to the old time source's EBs */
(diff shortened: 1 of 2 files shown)


yatch:
Is there any possibility that [new_time_source()](https://github.com/simonduq/contiki/blob/c7694e4dbdcfe831b29d8128f3845d9a9806ceac/apps/orchestra/orchestra-rule-eb-per-time-source.c#L75) in `orchestra-rule-eb-per-time-source.c` has NULL for `old` or `new`? If it is the case, I think it would be better to add NULL checks there. Here is an example:

```C
uint16_t old_ts = old != NULL ? get_node_timeslot(&old->addr) : 0xffff
uint16_t new_ts = new != NULL ? get_node_timeslot(&new->addr) : 0xffff
```

simonduq:
Good point, will fix

simonduq:
(or feel free to PR on my branch, your suggested fix is exactly what we need)

yatch:
@simonduq OK, I'll send a PR later :-)

simonduq:
:+1:

PR #1673, opened 2016-05-23
Changed read-only pointers to const in enc28j60

njh:
I have changed the ptr argument passed to init/send/write to const in the enc28j60 driver.

I made the change because I was passing a const from my own code, which was causing a compiler error.


simonduq:
:+1:

alignan:
:+1:

njh:
Thanks!

PR #1676, opened 2016-05-24
update time before return etimer next expiration time

tidyjiang8:
To get more precise etimer next expiration time, call `update_time()` to update time before return `next_expiration`.

### DIFF
--- a/core/sys/etimer.c
+++ b/core/sys/etimer.c
@@ -236,7 +236,11 @@ etimer_pending(void)
 clock_time_t
 etimer_next_expiration_time(void)
 {
-  return etimer_pending() ? next_expiration : 0;
+  if(etimer_pending()){
+  	update_time();
+	return next_expiration;
+  }
+  return 0;
 }
 /*---------------------------------------------------------------------------*/
 void


bkozak-scanimetrics:
Hello @tidyjiang8,

What is the trouble you were having with ```etimer_next_expiration_time```?

tidyjiang8:
Hi bkozak,
   I think it's better to update time before return `next_expiration`, because there is an interval between the last update_time and now

greg-king5:
`update_time()` adds `now` only because it is subtracted from `tdist` while the function looks for the earliest expiration time.  In other words, the effect of the current time is _removed_ from the expiration time.  `update_time()` is called whenever any etimers are changed; therefore, `next_expiration` always is accurate.  This patch isn't needed.

bkozak-scanimetrics:
Ok, I think I get the issue. The ```next_expiration``` value may refer to a timer that has already expired by the time that you run ```etimer_expiration_time()```. 

If you read the discussion on #1265 you'll see why it might not be possible to call ```update_time()``` from ```etimer_next_expiration_time()``` on some platforms. 

What is the reason that you need a more accurate next expiration value? Perhaps you can run the etimer process from your application before calling ```etimer_expiration_time()``` in order to first deal with expired timers?

greg-king5:
I want to point out that `next_expiration` and `etimer_next_expiration_time()` aren't good choices of names.  They are not what those names imply.  They don't refer to the "next time"; they refer to the next etimer.  And, they don't refer to the next etimer that will expire; they refer to the next etimer that will send an event to a process.  [The function should have been called `etimer_next_event_time()`.]

greg-king5:
After looking at `update_time()`, I have concluded that even the `etimer_pending()` test isn't needed in `etimer_next_expiration_time()`.  If `etimer_process()` initiates `next_expiration` to zero when it initiates `timerlist`, then the variable always will be zero when there are no pending etimers.

greg-king5:
Forget two of my comments -- I had forgotten that `etimer_next_expiration_time()` might be called by ISRs.  They might run between an etimer change and a call to `update_time()`.

tidyjiang8:
I see, Thanks all.

PR #1677, opened 2016-05-24
oma-lwm2m: Add functionality to send object instance list as core link format

gebart:
Adds a response to the top level lwm2m object to list all its instances as corelf.

Useful in conjunction with Copper to get the instance list directly without going through an LWM2M server. (Copper automatically adds the new elements to the resource list when it receives a response with content type set to application/link-format)

### DIFF
--- a/apps/oma-lwm2m/lwm2m-engine.c
+++ b/apps/oma-lwm2m/lwm2m-engine.c
@@ -603,6 +603,9 @@ write_object_instances_link(const lwm2m_object_t *object,
   }
 
   for(i = 0; i < object->count; i++) {
+    if((object->instances[i].flag & LWM2M_INSTANCE_FLAG_USED) == 0) {
+      continue;
+    }
     instance = &object->instances[i];
     PRINTF(",</%d/%d>", object->id, instance->id);
 
--- a/apps/oma-lwm2m/lwm2m-engine.c
+++ b/apps/oma-lwm2m/lwm2m-engine.c
@@ -585,6 +585,37 @@ get_resource(const lwm2m_instance_t *instance, lwm2m_context_t *context)
   return NULL;
 }
 /*---------------------------------------------------------------------------*/
+/**
+ * @brief Write a list of object instances as a CoRE Link-format list
+ */
+static int
+write_object_instances_link(const lwm2m_object_t *object,
+                            char *buffer, size_t size)
+{
+  const lwm2m_instance_t *instance;
+  int len, rdlen, i;
+
+  PRINTF("</%d>", object->id);
+  rdlen = snprintf(buffer, size, "</%d>",
+                   object->id);
+  if(rdlen < 0 || rdlen >= size) {
+    return -1;
+  }
+
+  for(i = 0; i < object->count; i++) {
+    instance = &object->instances[i];
+    PRINTF(",</%d/%d>", object->id, instance->id);
+
+    len = snprintf(&buffer[rdlen], size - rdlen,
+                   ",<%d/%d>", object->id, instance->id);
+    rdlen += len;
+    if(len < 0 || rdlen >= size) {
+      return -1;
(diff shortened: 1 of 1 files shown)


simonduq:
@nfi can you take a look?

gebart:
This PR does not implement the whole Discover operation in the LWM2M TS, it only returns the object instances, no resources, but it is a step in the right direction IMO.

The specification states this regarding the Discover operation (GET with Accept: application/link-format):

># 5.4.2 Discover
>The “Discover” operation is used to discover LWM2M Attributes attached to an Object, Object Instances, and Resources. This operation can be used to discover which Resources are implemented for a given Object Instance. The returned payload is a list of application/link-format CoRE Links [RFC6690] for each targeted Object, Object Instance, or Resource, along with their attached Attributes.
>
>The “Discover” operation has the following parameters:
>
>| Parameter          | Required | Default Value | Notes                          |
>|--------------------|----------|---------------|--------------------------------|
>| Object ID          | Yes      | -             | Indicates the Object.          |
>| Object Instance ID | No       | -             | Indicates the Object Instance. |
>| Resource ID        | No       | -             | Indicates the Resource.        |
>
>If Object ID is only specified, the LWM2M Client MUST respond to the “Discover” operation with the list of Object Instances and the list of their respective Resources implemented by the LWM2M Client for that Object. In addition the list of Attributes which have be assigned to this Object level (see section 5.3.1.1) are also returned.
>For example:
>- when the “Discover” operation targets an Object with Object ID of 3, the response to the operation could be:
>`</3>;pmin=10, </3/0/1>, <3/0/2>, </3/0/3>, </3/0/4>, <3/0/6>,<3/0/7>,<3/0/8>,<3/0/11>,<3/0/16>`
>which means that the LWM2M Client implements the Device Info Object (Instance 0)with Resource IDs of 1,2,3, 4 6,7,8,11,and 16 among the Resources of Devive Info Object, with an R-Attributes assigned to the Object level.
>- when the “Discover”operation targets an Object ID and Object Instance ID only, the list of Attributes assigned to that Object Instance MUST be reported, and the list of implemented Resources and their attached Attribute MUST be returned in the response as well. For example: if Object ID is 3 and Object Instance ID is 0, then
>`</3/0>;pmax=60, </3/0/1>, <3/0/2>, </3/0/3>, </3/0/4>, <3/0/6>;dim=8, <3/0/7>;dim=8; gt=50;lt=42.2,<3/0/8>;dim=8,<3/0/11>,<3/0/16>`
>means that regarding the Device Info Object Instance, an R-Attribute has been assigned to this Instance level. And the LWM2M Client implements the multiple Resources 6,7, and 8 with a dimension of 8 and supports 2 additional Notification parameters for Ressource 7
>- when the “Discover”operation targets an Object ID, Object Instance ID and Resource ID, the attributes of that Resource MUST be returned. In addition, the R-Attributes inherited from upper levels (Object and Object Instance level) are also reported for that Resource (The rules of Section 5.3.1.1 apply) For example: if Object ID is 3, Object Instance ID is 2, and Resource ID is 7, then
>`</3/2/7>;dim=8;pmin=10;pmax=60;gt=50;lt=42.2` with pmin assigned at the Object level, and pmax assigned at the Object Instance level
>
>If a Resource, an Object Instance, or an Object has attributes for multiple LWM2M Servers, then one link is returned for each and the ep= attribute is used to indicate the Short Server ID of the LWM2M Server. For example: if Object ID is 3 and Object Instance ID is 0, and Resource ID is 7 with two Observe operations from two Servers, then 

>`</3/0/7>;ep=1;dim=8;gt=50;lt=42.2,`
>`</3/0/7>;ep=2;dim=8;pmax=300;gt=80;lt=75.5`

nfi:
This looks reasonable to me.

nvt:
:+1:

PR #1679, opened 2016-05-25
fix the bug about chameleon raw hdrsize

tidyjiang8:
Fix the bug #1675.
If the length of some packet attribute is larger than 8, chameleon raw module's `hdrsize()` will return an incorrect header size.
This patch make the `hdrsize()` compatiblie with underlying packet attribute that the length is larger than 8.


simonduq:
@atiselsts do you agree with the fix?

atiselsts:
:+1:

PR #1684, opened 2016-05-29
Add CC13xx/CC26xx AUX controller module, ADC support, ALS driver and example

g-oikonomou:
A number of users have tried to use the CC26xx ADC with little success. There are two stumbling points:

* Our LPM module unconditionally powers down the AUX domain when entering deep sleep.
* The CCxxware OSC module unconditionally powers down AUX after every oscillator switch. This means that any radio on/off cycle will leave AUX powered down.

ADC sits within AUX and the two conditions above will make ADC inaccessible, even if the user has previously initialised it correctly.

This pull request adds a controller for the AUX domain and its module clocks. Code modules that require access to the AUX domain should now register themselves as consumers and specify the clocks within AUX that they need.

The controller provides a function that can be used to request AUX power down, either conditionally or unconditionally. We change our LPM logic to manipulate AUX through this function:

* When we want to shut down the device, we request an unconditional AUX power down.
* When we enter deep sleep, we request a conditional AUX power down. The AUX controller will power down the domain only if there are no registered consumers.

Because of this logic, consumer modules must unregister themselves when they no longer require AUX access, in order to allow lowest possible consumption under deep sleep.

Moving forward, this AUX power management logic will make our life easier when we come to the point of adding support for the sensor controller.

This pull changes our oscillator management module to manipulate AUX clocks through the controller.

With all of this in place, we can now actually use ADC, so this pull also adds two ADC examples:
* A generic `adc_sensor` that can be used to perform a single-shot conversion on any ADC channel
* A driver for the SmartRF06EB Ambient Light Sensor, demonstrated in the cc26xx simple demo.


NickTompkins123:
Would you be willing to show a demo as well for DIO23 on the launchpad? 

Also how fast can this be called? I am looking at sampling at 20kHz....

g-oikonomou:
You are catching me unawares here, Nick. What's on the LP's DIO_23?

NickTompkins123:
Nothing it just happens to be a common pin people setup for an ADC giving the examples that we can find for the 7x7 CC2650 launchpad... 

I guess I am more or less curious about how to use the driver in the normal cc26xx-demo project for the launchpad

tonnenpinguin:
If you're looking for data on maximum sample size, most answers lie within the technical reference manual published on Texas Instruments CC2650 product page (it's 200kS/s, restrictions apply).

Also, I'm pretty sure if you want to achieve this kind of sample rate you'll want to have a look at TIs cc26xxware documentation and implement the code necessary on your own, as you'll want to use continuous triggering and probably the adc fifo buffer as well. 

Unfortunatly Texas Instruments for some reason unknown decided to make your life harder as it should have to be and does not provide a downloadlink on their page anymore and you have to download the whole [TI-RTOS](http://www.ti.com/tool/ti-rtos) in order to get a copy of the cc26xxware documentation. Please note that the current cc26xxware shipped with TI-RTOS is not the same as in the contiki submodule, so there might be some differences (consult the [changelog](http://processors.wiki.ti.com/index.php/CC26xxWare) for more info)

One last thing, for info about how to setup the new ADC sensor implementation you can also consult other platform's examples as as of this pull request the sensor uses contikis standard sensor interface.

NickTompkins123:
If I were to create a timer using Contiki could I get a timer at a rate of 20kHz?

tonnenpinguin:
Realistically, no.
You would have to use a rtimer, which has 65536 tics per second on cc2650 (see RTIMER_ARCH_SECOND), however, the minimum amount of tics for any rtimer is 8, which gets you well below your desired sample rate.
Also I assume that the ADC won't be fast enough in manual trigger mode.

That being said this discussion has nothing to do with the actual PR so if you have any further questions I would like to refer you to the contiki mailing list!

alignan:
@g-oikonomou ready to merge?

g-oikonomou:
Yeah, I'm happy with that one as is. @atiselsts any comments here?

atiselsts:
:+1:

alignan:
:+1:

PR #1687, opened 2016-05-30
rpl-ext-header: Use 8-octet unit for HBHO length

tsparber:
According to [RFC 2460](https://tools.ietf.org/html/rfc2460#section-4.3) the length field of the Hop-by-Hop options header uses a 8-octet unit (multiple of 8 byte). This is also correctly handled in [`uip6.c`](https://github.com/contiki-os/contiki/blob/master/core/net/ipv6/uip6.c#L1168), but not when setting up the header.

In a normal configuration the `RPL_HOP_BY_HOP_LEN` define is 8, so the current implementation works, but it is a bit misleading.

This fix is needed, when `RPL_HOP_BY_HOP_LEN` is more than 8. For example when adding extra elements to the `uip_ext_hdr_opt_rpl` for a custom header extension (eq. tracing packets).

### DIFF
--- a/core/net/rpl/rpl-ext-header.c
+++ b/core/net/rpl/rpl-ext-header.c
@@ -78,7 +78,7 @@ rpl_verify_hbh_header(int uip_ext_opt_offset)
   uip_ds6_route_t *route;
   rpl_parent_t *sender = NULL;
 
-  if(UIP_HBHO_BUF->len != RPL_HOP_BY_HOP_LEN - 8) {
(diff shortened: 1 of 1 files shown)


tsparber:
Rebased as #1508 has been merged
 I'm not sure whether to use `>> 3` or `/ 8`, as the rest of `rpl-ext-header.c` is also sometimes using the shift to express the division and sometimes the division directly.

simonduq:
Now that you mention it I think I prefer `>> 3`, even though I guess/hope most compilers would compile the code to a shift anyway.
Happy with this PR regardless, thanks for the fix :+1:

nvt:
:+1:

PR #1688, opened 2016-05-30
Fixed rssi-scanner example

### LINKED ISSUE #1664: RSSI-Scanner example broken, Offset wrong
After the CC2420 driver itself adds its offset the values coming out of the RSSI-Scanner for the JAVA applets are wrong. Easy Fix: replace the 55 offset with 100 (adding RSSI_OFFSET)

tsparber:
Fixes  #1664 

(Tested using sky platform)


alignan:
:+1: but for me it seems like a hack, I have this change done in my personal branch but somehow doesn't seem "correct".  Nevertheless is an example currently broken, and I haven't had the time to fix properly.

simonduq:
:+1:

PR #1691, opened 2016-05-31
fix ipv6 http socket host handling as in RFC2732

casaroli:
we fixed the adding of the [ and ] characters to the Host: of the http request.


tonnenpinguin:
I think some unintended commits (adding a new platform) may have made their way into this pr,

casaroli:
Oh. This is true.

I gonna fix it today.

Sorry for the confusion.

PR #1692, opened 2016-05-31
RIME: Extend unicast example to include a sent callback.

pablocorbalan:
This PR simply adds a packet sent callback to the unicast connection used in the example to better illustrate the possibilities of the unicast primitive. Every time a packet is sent the callback is called and prints the linkaddr_t destination, the MAC status of the message sent, and the link layer number of transmissions of the packet. The PR also removes some unneeded headers for the example.

### DIFF
--- a/examples/rime/example-unicast.c
+++ b/examples/rime/example-unicast.c
@@ -39,11 +39,6 @@
 
 #include "contiki.h"
 #include "net/rime/rime.h"
-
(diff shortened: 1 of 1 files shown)


alignan:
Thanks! I have been using this also in a separate branch.

:+1:

simonduq:
:+1:

PR #1696, opened 2016-06-01
fix ipv6 http socket host handling as in RFC2732

casaroli:
sorry I had to create a new PR.


vsaw:
This is actually not related to your code but I was wondering if you might fix this as well to make your pull-request more robust:

The problem is that the [`parse_url`](https://github.com/momote-io/contiki/blob/03ca795bd6c163eaed116a129ec4d1a53833700c/core/net/http-socket/http-socket.c#L292) might not `\0` terminate the `host` string. The way I see it: If it is passed an URL, where the host is longer than `MAX_HOSTLEN`, then the host string will not be terminated.

Two things should be done to fix this.

1. [`parse_url`](https://github.com/momote-io/contiki/blob/03ca795bd6c163eaed116a129ec4d1a53833700c/core/net/http-socket/http-socket.c#L292)  should check if the host fits into the destination buffer and return the error value `0` if it did not.
2. Your code should consider using [`memchr`](http://www.cplusplus.com/reference/cstring/memchr/) and first check for the `\0` termination before looking for IPv6 addresses.

Ideally [`parse_url`](https://github.com/momote-io/contiki/blob/03ca795bd6c163eaed116a129ec4d1a53833700c/core/net/http-socket/http-socket.c#L292)   should also get the size of the host buffer to remove the assumption that the host and path buffers have a certain size.

casaroli:
@vsaw thanks for the comments.

I now included a check for the \0 char but I think it was simpler to add _just after_ the parsing.

I also changed the my functions to use memchr instead of srdchr.

What do you think?

BR

vsaw:
I'd still prefer having `parse_url` fixed as this is an IMHO unsafe function if used wrong.
But at least your code is now fixed 👍

simonduq:
Looks good to me except for code style: please remove spaces between `if` and `(`

casaroli:
Fixed code style.

@vsaw The first check fixes exactly the parse_url function. I don't see a problem anymore. Is there anything I could do better for that fix?

BR

vsaw:
The danger I see with `parse_url` is that your code invokes it correctly. However it is invoked in two more place which might not take all the precaution you did.

Therefore I would have loved seeing a more "defensive" behaviour of `parse_url`, so that it becomes easier to use the function correctly.

simonduq:
I feel this is a fair concern @vsaw but could go in a different PR, what do you think?

vsaw:
@simonduq Sure.

alignan:
Could you please create an issue about `parse_url` being unsage? just to keep this up.

vsaw:
@alignan Done. See https://github.com/contiki-os/contiki/issues/1707

alignan:
@vsaw Thanks!

:+1:

simonduq:
Thanks! :+1:

PR #1703, opened 2016-06-03
Adaptive 802.15.4 Security

kkrentz:
Currently, the `noncoresec_driver` has various problems:
- At some point, nodes cannot accept further neighbors because anti-replay data of disappeared neighbors is never deleted.
- If a node reboots, its frame counter starts over. This causes neighbors to consider a rebooted node's frames as replayed and also ensues a nonce reuse in 802.15.4 security.
- Moreover, if a node reboots, its anti-replay data gets lost, thus exposing rebooted nodes to replay attacks.

This pull request adds the `adaptivesec_driver`, which avoids the problems above by establishing session keys. For usage details, see https://github.com/kkrentz/contiki/tree/adaptivesec/core/net/llsec/adaptivesec.


apps/reboot-app/Makefile.reboot-app
```diff
@@ -0,0 +1 @@
+reboot-app_src = reboot-app.c
```
alignan: Newline here

alignan:
Please also include the `cc2538dk` and `zoul`, it should be pretty much the same as the `openmote-cc2538`.

core/lib/csprng.c
```diff
@@ -0,0 +1,62 @@
...
+ * \author
+ *         Konrad Krentz <konrad.krentz@gmail.com>
+ */
+
+#include "lib/csprng.h"
+#include "lib/aes-128.h"
+#include <string.h>
+
+#ifndef MIN
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
```
alignan: This is already defined in `sys/cc.h`, see [here](https://github.com/contiki-os/contiki/blob/master/core/sys/cc.h).

core/lib/iq-seeder.c
```diff
@@ -0,0 +1,246 @@
...
+
+#include "lib/iq-seeder.h"
+#include "net/netstack.h"
+#include "lib/aes-128.h"
+#include <string.h>
+
+#ifdef IQ_SEEDER_CONF_WITH_CHANNEL_HOPPING
+#define WITH_CHANNEL_HOPPING IQ_SEEDER_CONF_WITH_CHANNEL_HOPPING
+#else /* IQ_SEEDER_CONF_WITH_CHANNEL_HOPPING */
+#define WITH_CHANNEL_HOPPING 1
```
alignan: Is this a leftover from other module not implemented in Contiki?

core/lib/iq-seeder.c
```diff
@@ -0,0 +1,246 @@
...
+  uint8_t min;
+
+  min = row < column ? row : column;
+  row -= min;
+  column -= min;
+
+  return toeplitz[row ? COLUMN_COUNT - 1 + row : column];
+}
+/*---------------------------------------------------------------------------*/
+#if 0
```
alignan: I would prefer either to remove the commented block, or add some comments about why is disabled

alignan:
There is a frequency hopping block in `core/lib/iq-seeder.c`, is this part of the commit also? are you implementing frequency hop as well? sorry for asking something obvious,

kkrentz:
Thanks for your comments. I will try to address them.

Concerning `iq-seeder.c` - that thing is responsible for generating a cryptographic seed at start up. For this purpose, I currently use I/Q LSBs as an entropy source. Yet, I/Q LSBs are potentially vulnerable to eavesdropping and interference. While I think it should be hard to launch successful attacks on this, channel hopping is intended to further complicate attacks. However, I am currently working on a seeder that is more secure than this one.

kkrentz:
I applied the changes you suggested. Travis should pass after a restart.

kkrentz:
I think it is also a good idea to remove `noncoresec` entirely, isn't it?

examples/llsec/ipv6-demos/rpl-border-router/Makefile
```diff
@@ -0,0 +1,48 @@
...
+#found in the /apps, /platform/$(TARGET)/apps/, or current directory (in that order).
+# WITH_WEBSERVER=webserver for /apps/webserver
+# WITH_WEBSERVER=raven-webserver for /platform/avr-raven/apps/raven-webserver/
+#make clean before changing webservers!
+
+#Note /apps/webserver contains a 2500 byte style sheet which is a severe test
+#of the slip connection. Large MSS together with low baud rates without flow
+#control will overrun the transmit buffer when the style sheet is requested.
+
+WITH_WEBSERVER=0
```
alignan: Using `WITH_WEBSERVER=webserver` produces the following error:

````
../../../../apps/webserver/httpd-cgi.c:158:11: error: 'uip_conns' undeclared (first use in this function)
   conn = &uip_conns[s->u.count];
           ^
../../../../apps/webserver/httpd-cgi.c:158:11: note: each undeclared identifier is reported only once for each function it appears in
../../../../apps/webserver/httpd-cgi.c: In function 'tcp_stats':
../../../../apps/webserver/httpd-cgi.c:197:9: error: 'uip_conns' undeclared (first use in this function)
````

**Comment by alignan** [comment]
I'm testing using 2 nodes with the broadcast application running, I have a couple of questions (sorry in advance if this is something obvious):
- I only receive packets on one node, not in both as expected
- Using a generic sniffer I see the `Test` payload on the open, is this expected?

**Comment by kkrentz** [comment]
Thanks for testing. Indeed, it is expected that you see `Test` since the security level of the example is set to 2. Just change the security level to 5,6, or 7 to enable encryption. However, you should actually receive packets on both nodes.

File: examples/llsec/ipv6-demos/rpl-border-router/Makefile (hunk 5 above)
Comment (kkrentz) [inline examples/llsec/ipv6-demos/rpl-border-router/Makefile]: This is probably because I disabled TCP in the corresponding `project-conf.h`. Also, note that I did not copy some files that are required for enabling the web server over to this example folder. I could do so if you like.

**Comment by alignan** [comment]
Thanks for the fast response.  I don't have a concrete preference, but to try to follow how an user would test the application based on what is documented.  I think just adding a note about enabling back TCP should do.  I will continue testing, thanks!

**Comment by bthebaudeau** [comment]
Building `examples/llsec/ipv6-demos/rpl-border-router` for Zoul Firefly generates a lot of warnings.

Nothing is received with `examples/llsec/ipv6-demos/simple-udp-rpl` on my Zoul Firefly boards, whereas `examples/ipv6/simple-udp-rpl` works fine.

**Comment by vamsi-chikati** [comment]
Is llsec supposed to be working on all platforms, theoretically?

We had limited success testing earlier version of llsec in econotag. If you
could confirm, we'll try testing on econotag.

14 జూన్ 2016 4:49 AMన, "Benoît Thébaudeau" <notifications@github.com>
వ్రాసినది:

>

> Building examples/llsec/ipv6-demos/rpl-border-router for Zoul Firefly
generates a lot of warnings.
>
> Nothing is received with examples/llsec/ipv6-demos/simple-udp-rpl on my
Zoul Firefly boards, whereas examples/ipv6/simple-udp-rpl works fine.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
<https://github.com/contiki-os/contiki/pull/1703#issuecomment-225736642>, or
mute the thread
<https://github.com/notifications/unsubscribe/AH4a2iO8zLH-1GyxkCwgbWy3Pllk2pH7ks5qLeVngaJpZM4ItY88>
.

**Comment by kkrentz** [comment]
@bthebaudeau Regarding `examples/llsec/ipv6-demos/simple-udp-rpl`, you may just have to wait for a little while until session key establishment is done. To see what is going on, you can enable debugging in `akes.c`. Concerning `examples/llsec/ipv6-demos/rpl-border-router`, I only tried this one on TelosB motes, yet. What is also on my TODO list is making SLIP work. I will see what I can do.

@vamsi-chikati Porting the `adaptivesec_driver` to econtags requires implementing an `aes_128_driver`, as well as a `csprng_seeder`. However, I actually could implement a general `csprng_seeder` that is configured with a static seed at compilation time to ease porting to other platforms.

**Comment by alignan** [comment]
@kkrentz send me an email and we'll fix your lack of `RE-Motes` :wink:

File: examples/llsec/ipv6-demos/simple-udp-rpl/project-conf.h [hunk 6]
```diff
@@ -0,0 +1,78 @@
+#ifndef PROJECT_SIMPLE_UDP_CONF_H_
+#define PROJECT_SIMPLE_UDP_CONF_H_
+
+/* set CCA thresholds */
+#undef CC2420_CONF_CCA_THRESH
+#define CC2420_CONF_CCA_THRESH -25
+#undef CC2538_RF_CONF_CCA_THRES
```
alignan: Does this changes anything? checking the `cc2538-rf` there is:

````
#ifdef CC2538_RF_CONF_CCA_THRES
#define CC2538_RF_CCA_THRES CC2538_RF_CONF_CCA_THRES
#else
#define CC2538_RF_CCA_THRES CCA_THRES_USER_GUIDE /** User guide recommendation */
#endif /* CC2538_RF_CONF_CCA_THRES */
````

But in the actual code is pretty much hard-coded

````
REG(RFCORE_XREG_CCACTRL0) = CC2538_RF_CCA_THRES_USER_GUIDE;
````

**Comment by bthebaudeau** [comment]
@kkrentz With `DEBUG` set to `1` in `akes.c`, `examples/llsec/ipv6-demos/simple-udp-rpl` works fine, so there is probably a timing issue.

**Comment by g-oikonomou** [comment]
Hi folks. The bullets in the OP are things that need solved. However, I am finding it very difficult to understand the reasoning behind some of the commits proposed here. This includes  all the pub-sub business and also the reboot app.

I cannot immediately understand why we need an LLsec-specific border router example. All `project-conf.h` files in those examples are very messy, with those `#if 0` blocks.

I see that you enclose entire .c files inside #if guards. Is this really required? In other words, will the file fail to compile if those macros are not set?

I am finding it difficult to understand how you are deriving pairwise keys. Is this some sort of key agreement? If so, is this something well-established? Is this something that you have designed / published yourself?

For channel-hopping in the respective seeder, best not hard-code channels in [11, 26], since we are getting traction with sub-ghz radios. Perhaps get min/max through the radio API (`RADIO_CONST_CHANNEL_MIN`, `RADIO_CONST_CHANNEL_MAX`).

Generally speaking, security and crypto are things difficult to understand (and to therefore configure properly based on requirements). In this context, some documentation would be nice :) That's a general comment that applies to what we already have on master by the way.

Lastly, perhaps of zero consequence, those "Your Name committed" don't look very nice

**Comment by kkrentz** [comment]
* demos: I could remove the demos since I think that the description on https://github.com/kkrentz/contiki/tree/adaptivesec/core/net/llsec/adaptivesec is sufficient for getting started
* reboot app: It is part of the demos, which is why I could remove it, too
* pub-sub: This is a flexible way to handle incoming 802.15.4 MAC command frames. The way it works is that you register for 802.15.4 MAC command frames. Subsequently, you get called upon receipt of a command frame and report report back if you consumed it.
* pairwise keys: The key derivation process is pretty common. It derives a pairwise session key from two random numbers, which are contributed by both sides. Our whole key establishment scheme is explained in a paper, which is linked at the bottom of  https://github.com/kkrentz/contiki/tree/adaptivesec/core/net/llsec/adaptivesec.
* seeder: I could do this.

**Comment by kkrentz** [comment]
Updates:
* I removed all the demo files.
* I updated the README.
* When adapting the current `rpl-border-router` example according to my README, the `rpl-border-router` now works on Sky-, Re-, and Open-Motes.

**Comment by kkrentz** [comment]
@g-oikonomou I was able to remove a couple of #if guards, but removing the remaining ones would cause compiler errors and warnings.

**Comment by bthebaudeau** [comment]
I've tested this update with `examples/ipv6/simple-udp-rpl/broadcast-example.c` on my Zoul Firefly boards:
 - If this example is unchanged, it works fine.
 - If I follow the instructions from your `README`:
    - I still get a range or timing issue with `#define NETSTACK_CONF_RDC contikimac_driver`.
    - The boards keep resetting with `#define NETSTACK_CONF_RDC nullrdc_driver`.

**Comment by bthebaudeau** [comment]
I confirm that `examples/ipv6/simple-udp-rpl/broadcast-example.c` now works fine with `nullrdc_driver` following your latest changes (f315bd1).

**Comment by kkrentz** [comment]
@bthebaudeau Is there anything left that does not work for you?

**Comment by bthebaudeau** [comment]
@kkrentz Now, my Zoul Firefly boards can successfully send and receive messages with all the supported `adaptivesec` configurations. However, I still have the feeling that using `adaptivesec` with ContikiMAC leads to less reliable transmissions than `adaptivesec` with `nullrdc`, or `nullsec` with ContikiMAC. This is only a feeling, because the reliability of transmissions changes from time to time, and my boards lack antennae. I've selected a channel supposed to be free, but there may be uncontrolled remote emitters on any channel from time to time. So I can't say for sure. If someone has a couple of motes with proper antennae, he or she can compare the `adaptivesec` and `nullsec` transmission ranges in an open field environment with ContikiMAC.

**Comment by kkrentz** [comment]
@bthebaudeau I also encountered reliability issues with ContikiMAC, but they seem to be independent of `adaptivesec`. This is one reason why I began to work on a re-implementation of ContikiMAC, as I told you already. For the time being, it might help to switch on CSMA and/or the ContikiMAC framer (which adds padding bytes to short frames).

**Comment by bthebaudeau** [comment]
@kkrentz Thanks: it works much better with your latest additions to the `Troubleshooting` section. Everything seems to be fine now.

**Comment by alignan** [comment]
I can confirm also it is working nicely for the `RE-Mote` platform and generally the `CC2538` platforms.  The troubleshooting section is a great addition, thanks!

@g-oikonomou any additional comment? have your remarks been addressed?

**Comment by kkrentz** [comment]
- I removed channel hopping from `iq-seeder.c` and added more comments to it.
- I revised `csprng.c` to suit the needs of other seeders that I am currently testing.

**Comment by simonduq** [comment]
Two comments:
* I don't like the name of the module `neighbor`, which says little about what the module does. Something in the naming should convey the security aspect. Else things just get confusing, sounds like a neighbor management module.
* Are you 100% confident about the safety of your use of `nbr-table` indexes outside of the module, via `nbr_table_index_from_item`? Indexes are really intended as an internal thing only. Between two consecutive calls, one has no guarantee that a neighbor hasn't changed index. It might have been discarded and re-added at a different index. The safe way to identify a neighbor is with its MAC address.

**Comment by simonduq** [comment]
Thanks!

**Comment by simonduq** [comment]
OK this looks like colliding with https://github.com/contiki-os/contiki/pull/1739
Which suppresses sequence numbers in broadcast to reduce the seqno wrapping pace

**Comment by kkrentz** [comment]
I see. It should be fine by now. Of course, setting the sequence number at the LLSEC layer is debatable, but, at least, the LLSEC layer is a common place for doing this.

**Comment by simonduq** [comment]
works for me, thanks!

**Comment by kkrentz** [comment]
- Since suppressing the sequence numbers of broadcast frames is TSCH specific, I have added the configuration switches `FRAME802154_CONF_SUPPR_UNICAST_SEQNO` and `FRAME802154_CONF_SUPPR_BROADCAST_SEQNO` to `frame802154.h`. They default to what is required by TSCH. (614c8a4ad1a0ef9e063f74568ed403f6f8133d33)
- For setting sequence numbers, I have added two functions to the `mac-sequence` module. Now I can just call these functions from within `nullsec` and `adaptivesec` instead of implementing this twice. (614c8a4ad1a0ef9e063f74568ed403f6f8133d33)
- Lastly, there was a bug in `frame802154.h`. (5038583e9267a97a8278de6c502875c9899e40e4)

**Comment by kkrentz** [comment]
- I revised the suppression of sequence numbers once more. Now, framer-802154.c has a new function for setting sequence numbers. This function also takes care of not setting sequence numbers if suppression is enabled. [Here](https://github.com/kkrentz/contiki/blob/adaptivesec/core/net/mac/framer-802154.c#L81-L93) is a link to that function. As a result, the suppression of sequence numbers is handled locally within `framer-802154.c`, which is somewhat nicer.
- I turned `adaptivesec` into a `MODULE`, which has to be included from Makefiles if needed. This avoids compiling adaptivesec-related files every time.

**Comment by kkrentz** [comment]
rebased and merged

**Comment by alignan** [comment]
What's the status of this?

**Comment by kkrentz** [comment]
This is ready to get merged from my perspective ;)

**Comment by sonvoque** [comment]
Dear kkrentz!
Do you have any stable branch for this adaptivesec? I still get some warnings and errors when compiling the code?

[sudo] password for user: 
  CC        ../../../cpu/cc2538/./ieee-addr.c
  CC        ../../../core/net/llsec/adaptivesec/akes.c
../../../core/net/llsec/adaptivesec/akes.c:57:0: warning: "DEBUG" redefined
 #define DEBUG 0
 ^
In file included from ../../../platform/cc2538dk/./contiki-conf.h:16:0,
                 from ../../../core/./net/mac/frame802154.h:68,
                 from ../../../core/./net/llsec/adaptivesec/adaptivesec.h:43,
                 from ../../../core/./net/llsec/adaptivesec/akes.h:43,
                 from ../../../core/net/llsec/adaptivesec/akes.c:40:
./project-conf.h:95:0: note: this is the location of the previous definition
 #define DEBUG DEBUG_NONE
 ^
../../../core/net/llsec/adaptivesec/akes.c: In function 'prepare_update_command':
../../../core/net/llsec/adaptivesec/akes.c:90:35: warning: passing argument 1 of 'adaptivesec_add_security_header' from incompatible pointer type [-Wincompatible-pointer-types]
   adaptivesec_add_security_header(&entry->refs[status]->anti_replay_info);
                                   ^
In file included from ../../../core/./net/llsec/adaptivesec/akes.h:43:0,
                 from ../../../core/net/llsec/adaptivesec/akes.c:40:
../../../core/./net/llsec/adaptivesec/adaptivesec.h:133:6: note: expected 'struct anti_replay_info *' but argument is of type 'struct anti_replay_info **'
 void adaptivesec_add_security_header(struct anti_replay_info *receiver_info);
      ^
In file included from ../../../core/./net/llsec/adaptivesec/adaptivesec.h:44:0,
                 from ../../../core/./net/llsec/adaptivesec/akes.h:43,
                 from ../../../core/net/llsec/adaptivesec/akes.c:40:
../../../core/net/llsec/adaptivesec/akes.c:129:84: error: request for member 'my_unicast_counter' in something not a structure or union
 ounter.u32 = LLSEC802154_HTONL(entry->refs[status]->anti_replay_info.my_unicast
                                                                     ^
../../../core/./net/llsec/llsec802154.h:87:31: note: in definition of macro 'LLSEC802154_HTONL'
 #define LLSEC802154_HTONL(n) (n)
                               ^
../../../core/net/llsec/adaptivesec/akes.c:132:47: error: 'anti_replay_my_broadcast_counter' undeclared (first use in this function)
     reordered_counter.u32 = LLSEC802154_HTONL(anti_replay_my_broadcast_counter)
                                               ^
../../../core/./net/llsec/llsec802154.h:87:31: note: in definition of macro 'LLSEC802154_HTONL'
 #define LLSEC802154_HTONL(n) (n)
                               ^
../../../core/net/llsec/adaptivesec/akes.c:132:47: note: each undeclared identifier is reported only once for each function it appears in
     reordered_counter.u32 = LLSEC802154_HTONL(anti_replay_my_broadcast_counter)
                                               ^
../../../core/./net/llsec/llsec802154.h:87:31: note: in definition of macro 'LLSEC802154_HTONL'
 #define LLSEC802154_HTONL(n) (n)
                               ^
../../../core/net/llsec/adaptivesec/akes.c: In function 'on_hello':
../../../core/net/llsec/adaptivesec/akes.c:226:5: warning: implicit declaration of function 'anti_replay_restore_counter' [-Wimplicit-function-declaration]
     anti_replay_restore_counter(&entry->permanent->anti_replay_info);
     ^
../../../core/net/llsec/adaptivesec/akes.c: In function 'on_helloack':
../../../core/net/llsec/adaptivesec/akes.c:347:3: warning: implicit declaration of function 'anti_replay_parse_counter' [-Wimplicit-function-declaration]
   anti_replay_parse_counter(payload + ILOS_WAKE_UP_COUNTER_LEN + AKES_NBR_CHALL
   ^
../../../Makefile.include:240: recipe for target 'obj_cc2538dk/akes.o' failed
make: *** [obj_cc2538dk/akes.o] Error 1


Thanks
Vo Que Son

**Comment by kkrentz** [comment]
Currently, my "master" branch only works for OpenMotes. However, this branch should compile flawlessly if you follow the instructions of the README. If not, please send me your `project-conf.h` and TARGET platform.

**Comment by sonvoque** [comment]
Dear Krentz!
I have cloned your master branch and followed the guideline in README.
My project-conf.h:

#ifndef PROJECT_CONF_H_
#define PROJECT_CONF_H_

#ifndef WITH_NON_STORING
#define WITH_NON_STORING 0 /* Set this to run with non-storing mode */
#endif /* WITH_NON_STORING */

#undef NBR_TABLE_CONF_MAX_NEIGHBORS
#undef UIP_CONF_MAX_ROUTES

#ifdef TEST_MORE_ROUTES
/* configure number of neighbors and routes */
#define NBR_TABLE_CONF_MAX_NEIGHBORS     10
#define UIP_CONF_MAX_ROUTES   30
#else
/* configure number of neighbors and routes */
#define NBR_TABLE_CONF_MAX_NEIGHBORS     10
#define UIP_CONF_MAX_ROUTES   10
#endif /* TEST_MORE_ROUTES */

#undef NETSTACK_CONF_RDC
#define NETSTACK_CONF_RDC     nullrdc_driver
#undef NULLRDC_CONF_802154_AUTOACK
#define NULLRDC_CONF_802154_AUTOACK       1

/* Define as minutes */
#define RPL_CONF_DEFAULT_LIFETIME_UNIT   60

/* 10 minutes lifetime of routes */
#define RPL_CONF_DEFAULT_LIFETIME        10

#define RPL_CONF_DEFAULT_ROUTE_INFINITE_LIFETIME 1

#if WITH_NON_STORING
#undef RPL_NS_CONF_LINK_NUM
#define RPL_NS_CONF_LINK_NUM 		40 /* Number of links maintained at the root. Can be set to 0 at non-root nodes. */
#undef UIP_CONF_MAX_ROUTES
#define UIP_CONF_MAX_ROUTES 		0 /* No need for routes */
#undef RPL_CONF_MOP
#define RPL_CONF_MOP RPL_MOP_NON_STORING /* Mode of operation*/
#endif /* WITH_NON_STORING */

/* son define */
#ifndef UIP_CONF_ROUTER
#define UIP_CONF_ROUTER  			1
#endif

#define UIP_CONF_IPV6_RPL  			1
/* ND and Routing */
#define UIP_CONF_ND6_SEND_RA        0
#define UIP_CONF_IP_FORWARD         0


/* Low Power Mode */
#define LPM_CONF_ENABLE       		0		/**< Set to 0 to disable LPM entirely */
#define LPM_CONF_MAX_PM       		0


/* RF parameters define*/
#define CC2538_RF_CONF_CHANNEL    	26
#define CC2538_RF_CONF_TX_POWER		0xFF	// +7dBm
#define CC2538_RF_CONF_AUTOACK 		0 


#ifndef DEBUG
#define DEBUG DEBUG_NONE
#endif

#ifndef STARTUP_CONF_VERBOSE
#define STARTUP_CONF_VERBOSE        0 /**< Set to 0 to decrease startup verbosity */
#endif


#undef ADAPTIVESEC_CONF_UNICAST_SEC_LVL
#define ADAPTIVESEC_CONF_UNICAST_SEC_LVL 2
#undef ADAPTIVESEC_CONF_BROADCAST_SEC_LVL
#define ADAPTIVESEC_CONF_BROADCAST_SEC_LVL 2

#undef LLSEC802154_CONF_USES_AUX_HEADER
#define LLSEC802154_CONF_USES_AUX_HEADER 0

#undef NBR_TABLE_CONF_MAX_NEIGHBORS
#define NBR_TABLE_CONF_MAX_NEIGHBORS 14

#define SINGLE_CONF_KEY { 0x00 , 0x01 , 0x02 , 0x03 , \
                          0x04 , 0x05 , 0x06 , 0x07 , \
                          0x08 , 0x09 , 0x0A , 0x0B , \
                          0x0C , 0x0D , 0x0E , 0x0F }


#undef PACKETBUF_CONF_WITH_UNENCRYPTED_BYTES
#define PACKETBUF_CONF_WITH_UNENCRYPTED_BYTES 1

#undef CSPRNG_CONF_SEEDER
#define CSPRNG_CONF_SEEDER iq_seeder


#endif


the platform I used is a CC2538-based mote (cloned from CC2538DK from TI). 
I compile with: make TARGET=cc2538dk

My target is to integrate the adaptivesec module with my app.

Thank you much

Kind regards
Vo Que Son

**Comment by kkrentz** [comment]
My master branch is currently tied to OpenMotes. Anyway, it might work if you add `#include "net/llsec/adaptivesec/noncoresec-autoconf.h"` to your `project-conf.h` somewhere near the end.

**Comment by sonvoque** [comment]
Dear Krentz!
thank you much. I will try to modify your code to adapt to CC2538DK because OpenMote also uses CC2538 chip

Kind regards
Vo Que Son

**Comment by sonvoque** [comment]
Dear Krentz!

I have compiled successfully the Border-Router and an UDP-echo-server apps. 
I also enable DEBUG 1 in all files of adaptivesec.
This is the debug messages from Border-Router:
Contiki-2.6-4258-g76848c8
TI SmartRF06 + cc2538EM
CC2538: ID: 0xb964, rev.: PG2.0, Flash: 512 KiB, SRAM: 32 KiB, AES/SHA: 1, ECC/RSA: 1
System clock: 16000000 Hz
I/O clock: 16000000 Hz
Reset cause: External reset
 Net: sicslowpan
 LLSEC: adaptivesec
 MAC: CSMA
 RDC: nullrdc
Rime configured with address 00:12:4b:00:05:af:84:06
akes: broadcasting HELLO
*** Address:aaaa::1 => aaaa:0000:0000:0000
Got configuration message of type P
Setting prefix aaaa::
Server IPv6 addresses:
 aaaa::212:4b00:5af:8406
 fe80::212:4b00:5af:8406
akes: broadcasting HELLO
akes: broadcasting HELLO



and the messages from the udp-echo-server node:
Contiki-2.6-4258-g76848c8
TI SmartRF06 + cc2538EM
CC2538: ID: 0xb964, rev.: PG2.0, Flash: 512 KiB, SRAM: 32 KiB, AES/SHA: 1, ECC/RSA: 1
System clock: 16000000 Hz
I/O clock: 16000000 Hz
Reset cause: External reset
 Net: sicslowpan
 LLSEC: adaptivesec
 MAC: CSMA
 RDC: nullrdc
Rime configured with address 00:12:4b:00:05:af:85:70
akes: broadcasting HELLO
akes: broadcasting HELLO
akes: broadcasting HELLO


I saw that your branch is using Contiki 2.6 when we have Contiki 3.0 now.
It seams that 2 nodes only broadcast HELLO without receiving anything.
I believe that my HW platform is similar to OpenMote

Do you have any hints for me?

Danke schoen.
Vo Que Son

**Comment by kkrentz** [comment]
My guess is that this is because you are using nullrdc, which is not compatible with my modified version of `cc2538-rf.c` at the moment. Try using nullrdc instead. You can find another README with instructions in `net/mac/contikimac`.

**Comment by bschwand** [comment]
I am trying to get this adaptivesec working on the nxp JN5168 platform, so far without success. I tried with nullrdc and also contikimac, without success, no traffic seems to get through.
My setup is a border router talking to a mote with slip-radio, and several motes running the fw with adaptivesec enabled. Is the adaptive sec too tied to a particular architecture (cc2538 maybe) ? If it is, what would be needed to bring it up on a different architecture (jn516x) ?

**Comment by kkrentz** [comment]
This branch is not tied to any particular platform. What you should do, however, is implement a hardware-accelerated AES-128 driver for your platform. If no traffic gets through, it may be simply because you are too impatient. It takes a little while until session keys are established. To see what is going on, enable DEBUG in akes.c.

**Comment by bschwand** [comment]
thanks, I was afraid of that. We initially do a broadcast on each frequency to find which frequency other motes are using, and wait only about 250 ms before changing frequency and doing another broadcast. What do you consider a little while, in the order of 100 of milliseconds, or several seconds ?
I will look at akes.c too...

**Comment by kkrentz** [comment]
It may take up to 30s. This is on purpose to prevent HELLO flood attacks.

2017-02-01 20:32 GMT+01:00 bschwand <notifications@github.com>:

**Comment by bschwand** [comment]
I changed my app so only one channel is used, and we wait as long as desired but still nothing gets through.
I see in one of the comments above that you could make SLIP work. Is that still not working ?

That might explain why I get no results, as my setup is a native border router communicating over SLIP with a mote running slip-radio, then a network of other independent motes.
I enabled DEBUG in all adaptivesec files, I also dump in adaptivesec.c input() the type of frame received. The debug log shows the HELLO messages are sent in akes, but the frames coming in input() are of type 0x00 for the slip-radio and of type 0x02, 0x05 for the other motes which is pretty weird.
I only enabled sec level 1 to debug. 
Not much seems to go on, here is an example session log
adaptivesec,input: got a frame type 0x05
adaptivesec,input: got a frame type 0x02
akes-trickle: Doubling interval size
akes-trickle: I=60s t=40s
akes-trickle: Broadcasting HELLO
akes: broadcasting HELLO
akes-trickle: Doubling interval size
akes-trickle: I=120s t=67s
adaptivesec,input: got a frame type 0x05
akes-trickle: Broadcasting HELLO
akes: broadcasting HELLO
akes-trickle: Doubling interval size
akes-trickle: I=240s t=171s
adaptivesec,input: got a frame type 0x05
akes-trickle: Broadcasting HELLO
akes: broadcasting HELLO

Any suggestions ?

**Comment by bschwand** [comment]
one more thing, you said above:
_**My guess is that this is because you are using nullrdc, which is not compatible with my modified version of cc2538-rf.c at the moment. Try using secrdc instead. You can find another README with instructions in net/mac/contikimac.**_

I do not see any other README in that location, nor any reference to secrdc, so I realized the master branch of your contiki fork has more, newer stuff... I will try with that again !

**Comment by kkrentz** [comment]
My master branch is not suitable for your platform. What RDC layer are you using?

**Comment by bschwand** [comment]
Nullrdc

**Comment by kkrentz** [comment]
Usually you should see `akes: Received HELLO` if something gets through. I do not know where `adaptivesec,input: got a frame type` is printed. I suggest you to place some additional printfs to trace down whether HELLOs arrive and where they are discarded. When using nullrdc, I observed that motes sometimes detect preambles within radio noise, which causes them to receive garbage frames. So incoming frames are not necessarily valid.

**Comment by bschwand** [comment]
the message is part of additional trace I am adding to figure where the packet comes and goes.
it is in the input() function of adaptivesec.c . I assume received frames would go through that function, identified by frame type and acted on. But here the frame type is never right. I used a sniffer and verified with binary dumps that the HELLO message and the packet received do match. However in input() the frame type from the packet attributes, which should have been parsed in the framer if I get it right, are not what the dump and the sniffer shows (i.e. the packet dump shows a CMD frame type, but the corresponding attribute is not set.) So I think there is something going on with the framer. I will continue tracing where packets come and how they are processed until I figure it out... thanks for the suggestions. always welcome.

**Comment by bschwand** [comment]
I got the adaptivesec layer working for my purposes (a native border router, talking over slip serial to a mote running slip-radio, in a ipv6 6lowpan network to other motes)
Overall, this is what I had to do:

- Enable adaptivesec in slip-radio and the motes
- Enable LLSEC802154_CONF_ENABLED in the native border router
- the slip-net.c slipnet_input() must be modified to use the total packet length and also copy all data in the uip buffer like here:

+#if ENABLE_ADAPT_SEC
+  uip_len = packetbuf_totlen();
+#else
   uip_len = packetbuf_datalen();
+#endif

+#if ENABLE_ADAPT_SEC
+  memcpy(uip_buf, packetbuf_hdrptr(), uip_len);
+#else
   packetbuf_copyto(uip_buf);
+#endif

- slip-radio needs to include the no_framer declaration

+#if ENABLE_ADAPT_SEC
+extern const struct framer         no_framer;
+#endif

- slip-radio in the slip_radio_cmd_handler() has to use the adaptivese_framer parse()
+#if ENABLE_ADAPT_SEC
+      adaptivesec_framer.parse();
+#else
       no_framer.parse();
+#endif

Some of those fixes might be due to not setting correctly framers and al, or simply that this setup is a little unusual as the NBR and radio mote are separate.

**Comment by bschwand** [comment]
I think I might have found a bug in this..

akes.c :
122 generate_pairwise_key(uint8_t *result, uint8_t *shared_secret)
 123 {
 124   ADAPTIVESEC_SET_KEY(shared_secret);
 125   AES_128.encrypt(result);
 126 }

ADAPTIVE_SET_KEY() is defined in adaptivesec.h

50 #if AKES_NBR_KEY_LEN == 16
  51 #define ADAPTIVESEC_SET_KEY(key)      CCM_STAR.set_key(key)
  52 #else /* AKES_NBR_KEY_LEN == 16 */
  53 #define ADAPTIVESEC_SET_KEY(key)      aes_128_set_padded_key(key, AKES_NBR_KEY_LEN)
  54 #endif /* AKES_NBR_KEY_LEN == 16 */

so for a key len of 16, we set the key in the CCM_STAR driver, but encrypt with the AES driver. Which explains why I was getting successful session key negotiation even when setting different keys in both motes.

**Comment by kkrentz** [comment]
You are right. Thanks for pointing this out. I added a fix and rebased.

**Comment by bschwand** [comment]
my local fix was to not use the ADAPTIVESEC_SET_KEY anymore as it is only used in two places, and in each of those it made sense to appropriately set the key for AES128 or CCM_STAR engine.

**Comment by kkrentz** [comment]
Okay. I did it your way and fixed the key length.

**Comment by msolters** [comment]
Hi @bschwand, I too have a debian host (ARM v7) tied to a SLIP radio (CC2650).

I will be trying to replicate your success as documented here.  In my case, however, I will be using 6LBR (not the native border router as found in Contiki).  It seems that this does not matter, however, as the `adaptivesec` is apparently something that *only* the SLIP radio knows about?

Why is it necessary to enable `LLSEC` in the border router?  **Is** this still necessary, after the recent commits?

**Comment by msolters** [comment]
I've managed to compile `slip-radio.bin`
*  For the CC2650
*  With the `adaptivesec` module
*  With @bschwand 's suggested changes in `slip-radio.c` and `slip-net.c`
*  The following configuration overriding the `project-conf.h` of the SLIP radio, and the node leafs that I am using to test traffic:

```c
#define ENABLE_ADAPT_SEC 1

#undef ADAPTIVESEC_CONF_UNICAST_SEC_LVL
#define ADAPTIVESEC_CONF_UNICAST_SEC_LVL 2
#undef ADAPTIVESEC_CONF_BROADCAST_SEC_LVL
#define ADAPTIVESEC_CONF_BROADCAST_SEC_LVL 2


#define SINGLE_CONF_KEY { 0x00 , 0x01 , 0x02 , 0x03 , \
                          0x04 , 0x05 , 0x06 , 0x07 , \
                          0x08 , 0x09 , 0x0A , 0x0B , \
                          0x0C , 0x0D , 0x0E , 0x0F }

#undef CSPRNG_CONF_SEEDER
#define CSPRNG_CONF_SEEDER iq_seeder

#undef NETSTACK_CONF_FRAMER
#define NETSTACK_CONF_FRAMER contikimac_framer
#undef CONTIKIMAC_FRAMER_CONF_DECORATED_FRAMER
#define CONTIKIMAC_FRAMER_CONF_DECORATED_FRAMER adaptivesec_framer
#undef ADAPTIVESEC_CONF_DECORATED_FRAMER
#define ADAPTIVESEC_CONF_DECORATED_FRAMER framer_802154

#include "net/llsec/adaptivesec/noncoresec-autoconf.h"
```

The result: SLIP radio does communicate as I expect (I get a MAC address, 6LBR starts up fine).  However, I never see my node appear as a "sensor" under 6LBR, and the SLIP traffic is as follows:

Wireless input, generally a good sign that traffic is at least occuring and on the right channels.
```
2017-04-08 0:13:42.798159: PACKET: PF: wireless_input
2017-04-08 0:13:42.798251: PACKET: PF: wireless_input : broadcast
2017-04-08 0:13:42.798370: PACKET: PF: wireless_input: processing frame
```

Every time one of the nodes sends a CoAP request to a server on the SLIP's native host, I see this:
```
2017-04-08 0:13:45.186270: PACKET: PF: bridge_output: Sending packet to Broadcast
2017-04-08 0:13:45.186471: PACKET: PF: wireless_output: sending packet
2017-04-08 0:13:45.186682: PACKET: BR-RDC: write: 80 (sid: 0, cb: 1)
2017-04-08 0:13:45.186775: DUMP: BR-RDC: 
         7a3b3a1a 9b01c1b4 1efb0100 90f00000 
         aaaa0000 00000000 02124b00 07904a01 
         040e0008 0c0a0700 01000001 001e0100 
         081e4040 00000000 00000000 00000000 
         aaaa0000 00000000 00000000 00000000 
         
2017-04-08 0:13:45.189374: PACKET: SLIP: write: 108
2017-04-08 0:13:45.189443: DUMP: SLIP: 
         21530003 0a00580b 00010e00 0141d858 
         cdabffff 014a9007 004b1200 7a3b3a1a 
         9b01c1b4 1efb0100 90f00000 aaaa0000 
         00000000 02124b00 07904a01 040e0008 
         0c0a0700 01000001 001e0100 081e4040 
         00000000 00000000 00000000 aaaa0000 
         00000000 00000000 00000000 
2017-04-08 0:13:45.340795: PACKET: SLIP: read: 5
2017-04-08 0:13:45.341021: DUMP: SLIP: 
         21520000 01
2017-04-08 0:13:45.341409: PACKET: SCMD: Packet data report for sid:0 st:0 tx:1
2017-04-08 0:13:45.341517: PACKET: BR-RDC: sid ack: 0 (0, 1)
2017-04-08 0:13:53.964694: PACKET: PF: bridge_output: Sending packet to Broadcast
2017-04-08 0:13:53.964931: PACKET: PF: wireless_output: sending packet
2017-04-08 0:13:53.965099: PACKET: BR-RDC: write: 80 (sid: 1, cb: 1)
```

I think this is a fantastic alternative to LLSEC, but I must be missing something to make it work through a SLIP radio still.

kkrentz:
I once had SLIP running, but unlike suggested here I ran adaptivesec natively. Anyway, I was not satisfied with this solution because it was incompatible with my reimplementation of ContikiMAC. BTW, I am currently trying to accelerate session key establishment (while preserving AKES' HELLO flood defense).

msolters:
Thanks!

Will you please clarify what you mean by running adaptivesec "natively"?  You were running adaptivesec on the host machine, or on the SLIP radio attached to the host machine?

As a separate request, do you think you can provide an opinion as to the best choices for the following defines, which are the default for the Contiki `slip-radio` `project-conf.h`:

```c
/* configuration for the slipradio/network driver */
#undef NETSTACK_CONF_MAC
#define NETSTACK_CONF_MAC     nullmac_driver

#undef NETSTACK_CONF_RDC
#define NETSTACK_CONF_RDC     contikimac_driver

#undef NETSTACK_CONF_NETWORK
#define NETSTACK_CONF_NETWORK slipnet_driver

#undef NETSTACK_CONF_FRAMER
#define NETSTACK_CONF_FRAMER no_framer
```

In your opinion (or best guess), what *should* those be set to for adaptivesec to execute successfully on the SLIP radio?  At this time I'm just assuming RDC is acceptable across the mesh, to keep things simple.

kkrentz:
Here is a preliminary solution: https://github.com/kkrentz/contiki/commits/akes-tests. I changed slip-radio to implement everything from RDC to LLSEC - `no_framer` is no more. This simplifies matters a lot.

bschwand:
@msolters 
"Why is it necessary to enable LLSEC in the border router? Is this still necessary, after the recent commits?"

if I remember correctly, this had to do with enabling LLSEC AUX headers. We need certain llsec related headers in the packets and having it at the border router level already made it so the packets can get parsed correctly. Otherwise you are correct, adaptive sec is only enabled in the nodes and the slip-radio with this setup.

I also had to change the max payload size, so that the fragmenter (6cslowpan) makes fragment small enough that the additional security headers do not cause the fragments to go over the maximum 802.15.4 frame size.

Also, for debugging I recommend enabling DEBUG in akes.c and akes-nbr.c and also akes-trickle.c 
Most of the fixes I made, were because of log generated that way !

msolters:
Thank you so much to both of you!  I have now a working setup with 6LBR+SLIP communicating to nodes, with adaptivesec on the SLIP radio and the nodes (https://github.com/cetic/6lbr/issues/115#issuecomment-292767833).

It seemed the magic ingredient was using the `csma_driver` on both sides.

However, I will definitely be taking a look at the akes-tests branch!  It seems a much more robust approach.  @bschwand I will definitely consider those payload sizes moving forward.

msolters:
Hi @kkrentz, unfortunately the SLIP radio in the `akes-tests` branch does not work with 6LBR; I just get the following error over and over:

```
2017-04-09 14:27:18.618695: ERROR: BR-RDC: br-rdc: send failed, slip ack timeout (7)
```

This seems a like a serial-specific problem, rather than a networking one.

kkrentz:
In my branch, the PRINTF `br-rdc: failed to parse` no longer exists. Maybe you need recompile something.

msolters:
Yes, I believe these error messages are germane to 6LBR, and not printed by the SLIP radio itself.  Nevertheless, one result here is that the SLIP radio in `akes-tests` is incompatible with 6LBR as it stands.

Nevertheless, `adaptivesec` *does* work fine with 6LBR when using the "vanilla" Contiki SLIP radio, after being modified as per @bschwand.  I'm going to revert back to that setup to continue testing, as it is the only functional configuration I can manage with this hardware.

I understand, of course, that this PR is a Contiki-specific offering, and no support or compatibility with 6LBR was offered or implied.  But, it is one of the most popular BRs in use, and I think it would certainly help to have `adaptivesec` play nicely with 6LBR as a lot of developers use it.

bschwand:
@msolters : enable debug in framer-80215 and maybe hexdump the packets to find out where the packets are dumped instead of forwarded. When I added several changes for HW mac and ACK support on our platform, the packets were not parsed correctly because of additional fields. Debugging what happened in the framer allowed to find the issue.

kkrentz:
I appended some minor fixes. The acceleration of AKES is not part of these fixes, but will be released in the near future ...

kkrentz:
AKES works quicker now. Security is preserved.

kkrentz:
I have added more documentation on crypto as suggested by @g-oikonomou.

wdimmit:
I am attempting to bring this patch set up on a pair of cc1350 Launchpads.  For testing, I'm using the rpl-udp demo, with one server and one client node.   I've modified the project-conf.h file to use contikimac and csma along with the other suggested settings.  I've tested with both security level 2 & 5. I've also enabled DEBUG in all files in the adaptivesec directory.

If I include `noncoresec-autoconf.h` to use a group-level key, everything works as expected and sessions generally are fully established within a few seconds.  

If I include `coresec-autoconf.h` I see the pairwise and group keys logged with in a second or so of booting the second board.  After that, the two boards continue to acknowledge HELLOs from each other, but always report `akes: No need to start a new session` after HELLOACKs.  The one error I do see pretty frequently is `coresec-strategy: Inauthentic broadcast`.  No application-level traffic ever flows between the nodes.

Is there any basic reason pairwise encrypted communication would not work on a cc1350?

kkrentz:
My guess is that this is because of the unreliability of the current ContikiMAC implementation. This problem aggravates when using coresec since each broadcast frame is then preceded with an ANNOUNCE frame, which must also arrive. For comparison, you may try nullrdc.

wdimmit:
@kkrentz On your suggestion I have just tested a number of configurations using nullrdc.  Sadly, all have failed in similar fashion. The biggest difference compared with ContikiMAC seems to be that these two lines always appear together in the debugging output with nullrdc: 

`coresec-strategy: Received ANNOUNCE`
`coresec-strategy: Inauthentic broadcast`

Enabling or disabling CSMA did not seem to have any impact.  The relevant portion of my project-conf.h file for a recent attempt is here: https://gist.github.com/wdimmit/bac8b6af2be62c7177c9a2323b747e80

kkrentz:
I tried your configuration using OpenMotes and Cooja, but did not encounter any problems. Maybe this issue is really specific to cc1350, which is strange since noncoresec works on your platform. If you use a hardware-implemented AES-128 driver, you may try a software-implemented AES-128 instead.
```c
#undef AES_128_CONF
#define AES_128_CONF aes_128_driver
```

Johan-Henry:
Hi @kkrentz 
I am experimenting with a slip radio - native border router setup on cc2538 chips. I have tried some of the modifications as mentioned by bschwand and msolters here, but have not quite gotten to a solid solution with adaptive sec implemented (my normal border router on chip solution works much better).  

I am curious about your "preliminary solution: https://github.com/kkrentz/contiki/commits/akes-tests", mentioned here, however that link does not seem to work. Do you perhaps have some other link or reference that I can have a look at?

kkrentz:
@Johan-Henry I moved my "preliminary solution" to https://github.com/kkrentz/contiki/commits/secrdc-demos. My preliminary solution seems to work, but not as reliable as compared to using "examples/ipv6/rpl-border-router". Maybe this is true for the current implementation, as well.

amitbhanja:
Hello, I have a setup where there are some leaf nodes which directly communicate with the border router. The border router then sends the data on mqtt through http format. 
I have used adaptivesec and have enabled the security level to 1 for both broadcast and unicast packets. However, is it possible to dynamically decide security level for different kinds of unicast packets?
One unicast packet could use security level 1 and the other 4( authentication and encryption). Is it possible to add this feature? I saw that if the security levels mismatch, the packet is not discarded in adaptivesec as was done in noncoresec. 

Could some one please give a way through in this?

kkrentz:
Hi there, in principle that is possible, but I did not implement that.

amitbhanja:
Thanks for your quick reply. Will it be correct if change the adaptivesec_get_sec_lvl to give the security level according to the one set by the user for unicast packets? I have not tested yet. Do you see any other implications?

kkrentz:
In effect, that function is just used to check whether encryption is necessary or not. It should be feasible to adapt this function to encrypt only a subset of unicast frames. If you also want to adjust the length of message integrity codes (MICs) at runtime, things get more complicated since I use the statically configured MIC length everywhere.

amitbhanja:
Thanks for the answer. Will secrdc be working with potr_enabled on cc26xx-cc13xx cpu?

kkrentz:
This will not work out of the box because secrdc uses my own custom radio interface (core/dev/radio-async.h), which is only implemented for cc2538, yet (cpu/cc2538/dev/cc2538-rf-async.c). I am not planning to implement further radio_async_drivers, but I do plan to implement another MAC protocol on top of this interface (coordinated sampled listening).

amitbhanja:
Hello,

I am trying to establish communication between two nodes. One of them acting as a border router to ip connection. I am establishing adaptivesec and giving the following settings in the project_conf for both the nodes. 

LEAF NODE
#define NETSTACK_CONF_MAC 						csma_driver

//#define NETSTACK_CONF_RDC 					nullrdc_driver
#define NETSTACK_CONF_RDC 						contikimac_driver
#define NETSTACK_CONF_RDC_CHANNEL_CHECK_RATE 	8

#define UIP_CONF_TCP							0

//#define RPL_CONF_LEAF_ONLY               	    1

#define RPL_CONF_PROBING_INTERVAL           	(600 * CLOCK_SECOND)

#define UIP_CONF_ROUTER 						0

#define NETSTACK_CONF_WITH_IPV6					1

#define UIP_CONF_IP_FORWARD                 	0

#define CC26XX_UART_CONF_ENABLE             	0                               //Enable/Disable UART I/O

#define RF_CHANNEL	           					26

// configurations for security feature

// Enable the security level
// Currently enabling authentication only for unicast and
// broadcast communication
#undef ADAPTIVESEC_CONF_UNICAST_SEC_LVL
#define ADAPTIVESEC_CONF_UNICAST_SEC_LVL 5
#undef ADAPTIVESEC_CONF_BROADCAST_SEC_LVL
#define ADAPTIVESEC_CONF_BROADCAST_SEC_LVL 5

// Enable pairwise keys
#include "net/llsec/adaptivesec/coresec-autoconf.h"

#undef NETSTACK_CONF_FRAMER
#define NETSTACK_CONF_FRAMER contikimac_framer
#undef CONTIKIMAC_FRAMER_CONF_ENABLED
#define CONTIKIMAC_FRAMER_CONF_ENABLED 1
#undef CONTIKIMAC_FRAMER_CONF_DECORATED_FRAMER
#define CONTIKIMAC_FRAMER_CONF_DECORATED_FRAMER adaptivesec_framer
#undef ADAPTIVESEC_CONF_DECORATED_FRAMER
#define ADAPTIVESEC_CONF_DECORATED_FRAMER framer_802154


#undef SINGLE_CONF_KEY
#define SINGLE_CONF_KEY { 0x10 , 0x21 , 0x32 , 0x43 , \
                     0x54 , 0x65 , 0x76 , 0x87 , \
                     0x98 , 0xA9 , 0xBA , 0xCB , \
                     0xCC , 0xDD , 0xED , 0xFC }

BORDER_ROUTER
/* Comment this out to use Radio Duty Cycle (RDC) and save energy */
//#define NETSTACK_CONF_RDC          			nullrdc_driver
#define NETSTACK_CONF_RDC 						contikimac_driver
#define NETSTACK_CONF_RDC_CHANNEL_CHECK_RATE 	16

#define QUEUEBUF_CONF_NUM          				4

#define UIP_CONF_BUFFER_SIZE       				256

#define UIP_CONF_RECEIVE_WINDOW    				60

#define RF_CHANNEL	           					26

// configurations for security feature

// Enable the security level
// Currently enabling authentication only for unicast and
// broadcast communication
#undef ADAPTIVESEC_CONF_UNICAST_SEC_LVL
#define ADAPTIVESEC_CONF_UNICAST_SEC_LVL 5
#undef ADAPTIVESEC_CONF_BROADCAST_SEC_LVL
#define ADAPTIVESEC_CONF_BROADCAST_SEC_LVL 5

// Enable pairwise keys
#include "net/llsec/adaptivesec/coresec-autoconf.h"

#undef NETSTACK_CONF_FRAMER
#define NETSTACK_CONF_FRAMER contikimac_framer
#undef CONTIKIMAC_FRAMER_CONF_ENABLED
#define CONTIKIMAC_FRAMER_CONF_ENABLED 1
#undef CONTIKIMAC_FRAMER_CONF_DECORATED_FRAMER
#define CONTIKIMAC_FRAMER_CONF_DECORATED_FRAMER adaptivesec_framer
#undef ADAPTIVESEC_CONF_DECORATED_FRAMER
#define ADAPTIVESEC_CONF_DECORATED_FRAMER framer_802154


#undef SINGLE_CONF_KEY
#define SINGLE_CONF_KEY { 0x10 , 0x21 , 0x32 , 0x43 , \
                     0x54 , 0x65 , 0x76 , 0x87 , \
                     0x98 , 0xA9 , 0xBA , 0xCB , \
                     0xCC , 0xDD , 0xED , 0xFC }


But I am seeing on the wireshark that there are lot of exchanges of RIT Data Request and DSME information request, but no acknowledgements or any data exchange

Can you suggest what can be wrong here

kkrentz:
The reason why you see RIT Data Requests and DSME information requests maybe is that adaptivesec uses some command IDs that are in use. Another problem probably is that the ContikiMAC framer adds a non-compliant byte for handling its padding bytes. These things could be fixed by switching to reserved command IDs and moving ContikiMAC's padding data to an information element (IE).

If you want to run adaptivesec on cc26xx-cc13xx, you will have to implement an appropriate `csprng_seeder`. However, I had just uploaded an improved `null_seeder`, which suffices if you configure adaptivesec to exclusively use pairwise session keys. The `csprng_seeder` is mainly required to generate secret group session keys and improves the opaqueness of pairwise session keys.

firmwareguru:
AdaptiveSec seems to be already available on cetic/6lbr, why is it taking so long to get integrated into Contiki mainline?

kkrentz:
I, meanwhile, ported adaptivesec, along with my implementations of ContikiMAC and CSL to Contiki-NG https://github.com/kkrentz/contiki-ng. I will create a new pull request there.

firmwareguru:
Excellent kkrentz. Your efforts to bring Contiki towards IIoT-level robustness are invaluable.  I have a couple of comment-questions:

1. Where do we sit with bringing adaptive key exchange and compromise resliency to TSCH?

2. I notice no references between your papers and this one: 
Cryptographic Key Exchange in IPv6-Based Low Power,
Lossy Networks
Panagiotis Ilia, George Oikonomou, Theo Tryfonas
>>>
 I'm sure you're aware of it, however.
How does your system compare?  
I would like to see a system based on ECC/PKI on the TI 13x2 platforms.  However if your system is just as compromise resilient without any central key distribution system, then fine by me.

amitbhanja:
@kkrentz , thanks for the update. Please also inform us if you could, about the new pull request in contiki-ng.

amitbhanja:
@kkrentz will the CSL work for cc26xx-cc13xx cpu drivers?

kkrentz:
@firmwareguru I have little hope that we can add compromise resilience to TSCH anytime soon. Even if TSCH was changed to use pairwise keys, its time synchronization could still be attacked by compromised nodes. Because of this, I rather added channel hopping (and other nice features) to CSL. Regarding the paper, I think I saw this paper a long time ago. They are using unauthenticated ECDH and hence do not prevent unauthorized nodes from joining the network. Another limitation is that ECDH is quite energy- and time-consuming. By contrast, our system ensures authorization and avoids public-key cryptography.

@amitbhanja For the time being, I only implemented a driver for the CC2538: https://github.com/kkrentz/contiki-ng/blob/master/arch/cpu/cc2538/dev/cc2538-rf-async.c.

amitbhanja:
@kkrentz , Thanks for the info. I was actually trying the adaptivesec (https://github.com/kkrentz/contiki) with rpl-border-router. The border router is on cc1310 and sensortag on cc1350. I enabled the prints of the security module. 

#include "net/llsec/adaptivesec/coresec-autoconf.h"

#define AKES_SINGLE_CONF_KEY { 0x12 , 0x37 , 0x59 , 0x65 , \
                               0xC4 , 0xD5 , 0x8A , 0x76 , \
                               0x19 , 0xE7 , 0xA9 , 0x98 , \
                               0x1D , 0x51 , 0xCB , 0xFE }

#undef CSPRNG_CONF_SEEDER
#define CSPRNG_CONF_SEEDER null_seeder

#undef CONTIKIMAC_CONF_WITH_PHASE_OPTIMIZATION
#define CONTIKIMAC_CONF_WITH_PHASE_OPTIMIZATION 1

#undef NETSTACK_CONF_FRAMER
#define NETSTACK_CONF_FRAMER contikimac_framer
#undef CONTIKIMAC_FRAMER_CONF_ENABLED
#define CONTIKIMAC_FRAMER_CONF_ENABLED 1
#undef CONTIKIMAC_FRAMER_CONF_DECORATED_FRAMER
#define CONTIKIMAC_FRAMER_CONF_DECORATED_FRAMER adaptivesec_framer
#undef ADAPTIVESEC_CONF_DECORATED_FRAMER
#define ADAPTIVESEC_CONF_DECORATED_FRAMER framer_802154

I tried the above with both phase optimization and without that. However, I always get the prints of the ACK being received. But then a series of Invalid ACK prints, which I believe is because the entry becomes permanent and then it is not any more tentative. There has been occasions of ACK was not acknowledged prints. This i believe is because of no mac layer acknowledgments being sent. Would this be an issue?

but after that, I dont see the sicslowpan connection getting through. Dont see any sicslowpan input prints on either side. 

Could you give some idea about what could be the issue?

kkrentz:
This sounds like the ACK is not being acknowledged, although it was received successfully ...

firmwareguru:
@amitbhanja I am looking at using adaptivesec + contikimac + cc13xx too; please update here any advances or resolutions, as will I.

firmwareguru:
@kkrentz CSL-MAC: A low power channel hopping MAC with adaptive security and compromise resiliency?  Sounds great.  Where is it, and how can we get it rolled into here, or Contiki-NG, which seems to have only TSCH as the MAC layer??

amitbhanja:
Hello @kkrentz , I have been using the adaptivesec protocol from ur github repository https://github.com/kkrentz/contiki. Thanks a lot for providing this. It is working fine. I have a setup with border router and 6 nodes. The nodes send data every 1 minute. There is a mechanism I implemented which deletes the parent from the neighbor list if UPDATE ACK is not received after two checks of update interval. After about 2 days, it has been observed that even though node is sending HELLO message with a different session key than what was present due to deletion of the border router from neighbor table, the border router is identifying it as new session with permanent neighbor but the HELLO ACK being sent is not received at the node.  This is creating issue with starting a new session with the border router.
Could you give any tip about why it is happening?

kkrentz:
Hello,

thanks for using adaptivesec. It may happen that a HELLOACK is silently
discarded because there is no need to start a new session. Maybe this
mechanism interferes with your deletion of parents. I do not know what you
are trying to achieve, but, normally, it is unnecessary to delete neighbors
by hand since akes-delete.c already takes care of deleting uncommunicative
neighbors. To trace down the problem, it would be useful to have the DEBUG
output of akes.c. Also, I would need to know what MAC protocol, platform,
and seeder you are using.

That said, I meanwhile ported adaptivesec to Contiki-NG and
actually stopped maintaining my Contiki repository. You can find the
latest version there: https://github.com/kkrentz/contiki-ng,
https://github.com/kkrentz/contiki-ng/wiki/CSL. As Contiki-NG also contains
a lot of other fixes to Contiki, it probably makes sense for you to move to
Contiki-NG, too.

Kind Regards

Konrad

2018-07-23 7:31 GMT+02:00 Amit Bhanja <notifications@github.com>:

amitbhanja:
Hi @kkrentz ,

Thanks for your quick response. 
Mysetup: a node connecting to the border router. Trying to keep the radio off most of the time. Only on during the initial setup and then when data is being sent.
MAC: Contikimac along with CSMA. Channel check rate is 8.
However, I might have confused in the previous post. I am not deleting any entry on my own from the application layer. The only thing i did is added a callback to determine which neighbor has been deleted. If the neighbor is the preferred parent from RPL, I just reset the akes trickle timer , change the adaptivesec group key and reset the rpl timers. I do this because my node keeps the RADIO off most of the time, but when it knows that preferred parent has been deleted, it keeps the radio on. I checked that the restarts of the border router work properly. But after some time i see this issue. 
The HELLO message from the node is received at the border router and it determines that a new session has started and sends the HELLO ACK. But nothing is received at the node. In fact, nothing is received at the node (although the radio is on) after the deletion of the border router happens due to the akes-delete thread. Please see the logs below of akes.

From the node:

Line 1: akes-delete: #permanent = 1
	Line 4: akes-delete: Sent UPDATE
	Line 44: AKES: deleted preferred parent
	Line 45: adaptivesec_group_key = 6ffe9215bdf060a9e8f998e053619a11akes-trickle: Resetting Trickle
	Line 53: akes-delete Setting update check timer
	Line 54: akes-trickle: I=30s t=27s
	Line 136: akes-trickle: Broadcasting HELLO
	Line 139: akes: broadcasting HELLO
	Line 148: akes-trickle: I=60s t=55s
	Line 299: akes-trickle: Broadcasting HELLO
	Line 302: akes: broadcasting HELLO
	Line 316: akes-trickle: I=120s t=75s
	Line 513: akes-trickle: Broadcasting HELLO
	Line 516: akes: broadcasting HELLO
	Line 641: akes-trickle: I=240s t=238s
	Line 878: akes-delete: #permanent = 0
	Line 879: akes-delete Setting update check timer
	Line 1279: akes-trickle: Broadcasting HELLO
	Line 1282: akes: broadcasting HELLO
	Line 1286: akes-trickle: I=480s t=287s
	Line 1682: akes-delete: #permanent = 0
	Line 1683: akes-delete Setting update check timer
	Line 2049: akes-trickle: Broadcasting HELLO
	Line 2052: akes: broadcasting HELLO
	Line 2482: akes-delete: #permanent = 0
	Line 2483: akes-delete Setting update check timer
	Line 2569: akes-trickle: I=960s t=504s
	Line 3294: akes-delete: #permanent = 0
	Line 3295: akes-delete Setting update check timer
	Line 3924: akes-trickle: Broadcasting HELLO
	Line 3927: akes: broadcasting HELLO
	Line 4097: akes-delete: #permanent = 0
	Line 4098: akes-delete Setting update check timer
	Line 4904: akes-delete: #permanent = 0
	Line 4905: akes-delete Setting update check timer
	Line 5151: akes-trickle: I=1920s t=1390s
	Line 5712: akes-delete: #permanent = 0
	Line 5713: akes-delete Setting update check timer
	Line 6516: akes-delete: #permanent = 0
	Line 6517: akes-delete Setting update check timer
	Line 7322: akes-delete: #permanent = 0
	Line 7323: akes-delete Setting update check timer

From the border router:
Jul 27 08:54:19 raspberrypi bert[341]: akes: Received UPDATE from 00:12:4b:00:18:00:82:8f
Jul 27 08:54:20 raspberrypi bert[341]: His unicast counter 271
Jul 27 08:54:20 raspberrypi bert[341]: Received Counter 272
Jul 27 08:54:20 raspberrypi bert[341]: His unicast counter 272
Jul 27 08:54:20 raspberrypi bert[341]: Received Counter 272
Jul 27 08:54:20 raspberrypi bert[341]: akes-nbr: prolonging
Jul 27 08:54:20 raspberrypi bert[341]: akes-nbr: prolonged
Jul 27 08:54:20 raspberrypi bert[341]: Sending UPDATE ACK to 00:12:4b:00:18:00:82:8f
Jul 27 08:54:20 raspberrypi bert[341]: Setting unicast counter 503
Jul 27 08:55:02 raspberrypi bert[341]: akes: Received HELLO from 00:12:4b:00:18:00:82:8f
Jul 27 08:55:02 raspberrypi bert[341]: Deleting neighbor Status 1 00:12:4b:00:18:00:85:0d
Jul 27 08:55:02 raspberrypi bert[341]: noncoresec-strategy: Inauthentic frame
Jul 27 08:55:02 raspberrypi bert[341]: akes: Starting new session with permanent neighbor
Jul 27 08:55:02 raspberrypi bert[341]: akes: Will send HELLOACK in 2s
Jul 27 08:55:05 raspberrypi bert[341]: akes: Sending HELLOACK to 00:12:4b:00:18:00:82:8f
Jul 27 08:55:05 raspberrypi bert[341]: Setting unicast counter 504
Jul 27 08:56:00 raspberrypi bert[341]: akes: Received HELLO from 00:12:4b:00:18:00:82:8f
Jul 27 08:56:00 raspberrypi bert[341]: Deleting neighbor Status 1 00:12:4b:00:18:00:82:8f
Jul 27 08:56:00 raspberrypi bert[341]: noncoresec-strategy: Inauthentic frame
Jul 27 08:56:00 raspberrypi bert[341]: akes: Starting new session with permanent neighbor
Jul 27 08:56:00 raspberrypi bert[341]: akes: Will send HELLOACK in 1s
Jul 27 08:56:02 raspberrypi bert[341]: akes: Sending HELLOACK to 00:12:4b:00:18:00:82:8f
Jul 27 08:56:02 raspberrypi bert[341]: Setting unicast counter 505
Jul 27 08:57:20 raspberrypi bert[341]: akes: Received HELLO from 00:12:4b:00:18:00:82:8f
Jul 27 08:57:20 raspberrypi bert[341]: Deleting neighbor Status 1 00:12:4b:00:18:00:82:8f
Jul 27 08:57:20 raspberrypi bert[341]: noncoresec-strategy: Inauthentic frame
Jul 27 08:57:20 raspberrypi bert[341]: akes: Starting new session with permanent neighbor
Jul 27 08:57:20 raspberrypi bert[341]: akes: Will send HELLOACK in 1s
Jul 27 08:57:21 raspberrypi bert[341]: akes: Sending HELLOACK to 00:12:4b:00:18:00:82:8f
Jul 27 08:57:21 raspberrypi bert[341]: Setting unicast counter 506
Jul 27 08:59:09 raspberrypi bert[341]: akes: Received HELLO from 00:12:4b:00:16:7c:46:51
Jul 27 08:59:09 raspberrypi bert[341]: Deleting neighbor Status 1 00:12:4b:00:18:00:82:8f
Jul 27 08:59:09 raspberrypi bert[341]: noncoresec-strategy: Inauthentic frame
Jul 27 08:59:09 raspberrypi bert[341]: akes: Starting new session with permanent neighbor
Jul 27 08:59:09 raspberrypi bert[341]: akes: Will send HELLOACK in 2s
Jul 27 09:02:03 raspberrypi bert[341]: akes: Received HELLO from 00:12:4b:00:18:00:82:8f
Jul 27 09:02:03 raspberrypi bert[341]: Deleting neighbor Status 1 00:12:4b:00:16:7c:46:32
Jul 27 09:02:03 raspberrypi bert[341]: noncoresec-strategy: Inauthentic frame
Jul 27 09:02:03 raspberrypi bert[341]: akes: Starting new session with permanent neighbor
Jul 27 09:02:03 raspberrypi bert[341]: akes: Will send HELLOACK in 0s
Jul 27 09:02:03 raspberrypi bert[341]: akes: Sending HELLOACK to 00:12:4b:00:18:00:82:8f
Jul 27 09:02:03 raspberrypi bert[341]: Setting unicast counter 529
Jul 27 09:02:12 raspberrypi bert[341]: Setting unicast counter 530
Jul 27 09:02:12 raspberrypi bert[341]: Sending UPDATE to 00:12:4b:00:18:00:82:8f
Jul 27 09:02:13 raspberrypi bert[341]: akes-delete: Sent UPDATE
Jul 27 09:02:28 raspberrypi bert[341]: Deleting neighbor Status 0 00:12:4b:00:18:00:82:8f
Jul 27 09:18:29 raspberrypi bert[341]: akes: Received HELLO from 00:12:4b:00:18:00:82:8f
Jul 27 09:18:29 raspberrypi bert[341]: Deleting neighbor Status 1 00:12:4b:00:16:7c:46:32
Jul 27 09:18:29 raspberrypi bert[341]: akes: Will send HELLOACK in 2s
Jul 27 09:18:32 raspberrypi bert[341]: akes: Sending HELLOACK to 00:12:4b:00:18:00:82:8f
Jul 27 09:18:32 raspberrypi bert[341]: Setting unicast counter 544

I have some prints at the input of adaptivesec to check if any packet has been received.
I will try enabling prints at the possible error in contikimac or framer802154 parser.

Kindly please provide your inputs.

I will try to move to contiki-ng meanwhile.

kkrentz:
The problem seems to be outside of adaptivesec. It could be due to ContikiMAC, which is known to suffer from some timing issues. Besides, I have also encountered a lot of trouble with CSMA in the past. If you are using a CC2538-based platform, you may try my CSL implementation https://github.com/kkrentz/contiki-ng/wiki/CSL instead.

amitbhanja:
@kkrentz , thanks a lot for your advice. I am actually using srf06-cc26xx platform. What settings would you suggest to enable in contiki-ng for proper working? I am observing in the original contiki version that after using security, there have been issues of sometimes no packet being received at the gateway. Upon restarting those nodes, packets are being received. It seems to be issues with contikimac along with radio driver. 

Well , I will test with contiki-ng. Please recommend what settings to do for cc26xx platforms.
Using null-seeder now.

PR #1704, opened 2016-06-03
One fix to CC2538 hardware security

kkrentz:
First, I adapted the `cc2538_aes_128_driver` to reboot if an error occurs. I also considered handling errors, but I think that this would result in a lot of rarely tested code. After all, I have never actually encountered errors when calling `cc2538_aes_128_driver`. Nevertheless, I am afraid that an attacker may provoke errors and then bad things may happen.

### DIFF
--- a/cpu/cc2538/dev/cc2538-aes-128.c
+++ b/cpu/cc2538/dev/cc2538-aes-128.c
@@ -40,6 +40,7 @@
 #include "contiki.h"
 #include "dev/ecb.h"
 #include "dev/cc2538-aes-128.h"
+#include "dev/sys-ctrl.h"
 
 #include <stdint.h>
 #include <stdio.h>
@@ -82,6 +83,7 @@ set_key(const uint8_t *key)
                       CC2538_AES_128_KEY_AREA);
   if(ret != CRYPTO_SUCCESS) {
     PRINTF("%s: aes_load_keys() error %u\n", MODULE_NAME, ret);
+    sys_ctrl_reset();
   }
 
   restore_crypto(crypto_enabled);
@@ -99,13 +101,13 @@ encrypt(uint8_t *plaintext_and_result)
                         plaintext_and_result, AES_128_BLOCK_SIZE, NULL);
   if(ret != CRYPTO_SUCCESS) {
     PRINTF("%s: ecb_crypt_start() error %u\n", MODULE_NAME, ret);
-    restore_crypto(crypto_enabled);
-    return;
+    sys_ctrl_reset();
   }
 
   while((res = ecb_crypt_check_status()) == CRYPTO_PENDING);
   if(res != CRYPTO_SUCCESS) {
     PRINTF("%s: ecb_crypt_check_status() error %d\n", MODULE_NAME, res);
+    sys_ctrl_reset();
   }
 
   restore_crypto(crypto_enabled);


cpu/cc2538/dev/cc2538-aes-128.c
```diff
@@ -99,13 +101,14 @@ encrypt(uint8_t *plaintext_and_result)
                         plaintext_and_result, AES_128_BLOCK_SIZE, NULL);
   if(ret != CRYPTO_SUCCESS) {
     PRINTF("%s: ecb_crypt_start() error %u\n", MODULE_NAME, ret);
-    restore_crypto(crypto_enabled);
+    watchdog_reboot();
     return;
```
bthebaudeau: This `return` becomes useless.

cpu/cc2538/dev/cc2538-ccm-star.c
```diff
@@ -75,49 +75,18 @@ restore_crypto(uint8_t enabled)
 static void
 set_key(const uint8_t *key)
 {
-  cc2538_aes_128_driver.set_key(key);
+  ccm_star_driver.set_key(key);
```
bthebaudeau: Do you have test vectors or something that would help to duplicate and debug the issue?

kkrentz: I do not have a concise example, but [this](https://github.com/kkrentz/contiki/tree/potr/examples/secmac/tests/potr) one. It crashes after sending 100-10000 frames when using the old implementation.

cpu/cc2538/dev/cc2538-aes-128.c
```diff
@@ -82,6 +83,7 @@ set_key(const uint8_t *key)
                       CC2538_AES_128_KEY_AREA);
   if(ret != CRYPTO_SUCCESS) {
     PRINTF("%s: aes_load_keys() error %u\n", MODULE_NAME, ret);
+    watchdog_reboot();
```
bthebaudeau: Rethinking about this, it's actually better to use `sys_ctrl_reset()` from `cpu/cc2538/dev/sys-ctrl.h` for CC2538-specific code because:
 - It's immediate (vs. Twdt x 64).
 - `SYS_CTRL_CLOCK_STA.RST` then indicates `CLD/SW` instead of `WDT`, so it's possible to know that the reset was intentionally triggered by software, rather than by the watchdog timer because of the software accidentally getting out of control.

Out of curiosity, I've looked at what `rom_util_reset_device()` from `cpu/cc2538/dev/rom-util.h` does by disassembling the ROM, and it actually uses an undocumented hardware feature (reserved bit):
```
SYS_CTRL_CLOCK_CTRL = 0x8;
while (1)
  ;
```
According to http://www.ti.com/lit/pdf/swru333, this `performs a system reset equivalent to an external pin reset`, so `sys_ctrl_reset()` is still more informative here.

Note that there is yet another means of resetting the CC2538: `NVIC_APINT.SYSRESETREQ = 1`.

cpu/cc2538/dev/cc2538-ccm-star.c
bthebaudeau: Thanks. I'll look into it. In the meantime, if there is an urgent need to fix this, your commit can be used, then reverted once a proper fix using the hardware acceleration has been found.

I have a few questions:
 - With which board have you tested this?
 - Do both the sender and the receiver crash, or only one of them?
 - Why do I get the following errors with a clean checkout of your branch, if building for Zoul Firefly?

Sender:
```
obj_zoul/netstack.o: In function `netstack_init':
netstack.c:(.text.netstack_init+0x30): undefined reference to `adaptivesec_driver'
obj_zoul/secmac.o: In function `on_fifop':
secmac.c:(.text.on_fifop+0x66): undefined reference to `potr_parse_and_validate'
obj_zoul/secmac.o: In function `on_rxpktdone':
secmac.c:(.text.on_rxpktdone+0x16): undefined reference to `potr_shall_send_ack'
obj_zoul/secmac.o: In function `process_thread_post_processing':
secmac.c:(.text.process_thread_post_processing+0x160): undefined reference to `adaptivesec_framer'
secmac.c:(.text.process_thread_post_processing+0x168): undefined reference to `adaptivesec_driver'
sender.co: In function `process_thread_sender_process':
sender.c:(.text.process_thread_sender_process+0x2c): undefined reference to `neighbor_new'
sender.c:(.text.process_thread_sender_process+0x98): undefined reference to `adaptivesec_group_key'
sender.c:(.text.process_thread_sender_process+0xa4): undefined reference to `adaptivesec_driver'
```

Receiver:
```
obj_zoul/netstack.o: In function `netstack_init':
netstack.c:(.text.netstack_init+0x30): undefined reference to `adaptivesec_driver'
obj_zoul/secmac.o: In function `on_fifop':
secmac.c:(.text.on_fifop+0x66): undefined reference to `potr_parse_and_validate'
obj_zoul/secmac.o: In function `on_rxpktdone':
secmac.c:(.text.on_rxpktdone+0x16): undefined reference to `potr_shall_send_ack'
obj_zoul/secmac.o: In function `process_thread_post_processing':
secmac.c:(.text.process_thread_post_processing+0x160): undefined reference to `adaptivesec_framer'
secmac.c:(.text.process_thread_post_processing+0x168): undefined reference to `adaptivesec_driver'
receiver.co: In function `process_thread_receiver_process':
receiver.c:(.text.process_thread_receiver_process+0x20): undefined reference to `neighbor_new'
receiver.c:(.text.process_thread_receiver_process+0x48): undefined reference to `adaptivesec_group_key'
```

bthebaudeau: Another question: Why don't you just `#define CCM_STAR_CONF ccm_star_driver`, either in `project-conf.h` or in the `contiki-conf.h` of the platform, instead of altering this driver? The only difference is the crypto enable/restore, but the `cc2538_aes_128_driver` driver already takes care of that, so this should not be needed twice.

kkrentz: * I am using OpenMotes
* Only the receiver crashes
* The compilation errors should now be gone - I added `adaptivesec` to the MODULES list of Zoul
* If I used `ccm_star_driver` directly, every call to AES would enable and disable the crypto engine.

cpu/cc2538/dev/cc2538-aes-128.c
kkrentz: I replaced `watchdog_reboot()` with `sys_ctrl_reset()`

cpu/cc2538/dev/cc2538-ccm-star.c
bthebaudeau: - Thanks: it indeed now builds correctly. I'll look into it.
 - With your patch, it's `cc2538_ccm_star_driver` that enables and disables the crypto engine instead of `cc2538_aes_128_driver`. If you want to keep the crypto engine permanently enabled, you have to call either `crypto_init()` or `crypto_enable()` in some initialization code without calling `crypto_disable()` afterwards as currently done in `contiki-main.c`. But disabling the crypto engine when unused is not an issue (the key store seems to be retained). This optimizes the power consumption, and enabling or disabling it is extremely quick as this is only a matter of 3 modifications of registers changing the active clock domains. Or am I missing something?

cpu/cc2538/dev/cc2538-aes-128.c
bthebaudeau: Good, thanks.

cpu/cc2538/dev/cc2538-ccm-star.c
bthebaudeau: @kkrentz I've just tried your `potr` tests. I can't get a single packet through. I only get NAKs on the sender, and the receiver gets nothing:
```
Contiki-3.x-2589-g125569b
Zolertia Firefly platform
Rime configured with address 00:12:4b:00:06:16:0f:9c
 Net: nullnet
 MAC: Sec-MAC
 RDC: nordc
OK 0 / COLLISION 0 / NOACK 1 / DEFERRED 0 / ERR 0 / ERR_FATAL 0
OK 0 / COLLISION 0 / NOACK 2 / DEFERRED 0 / ERR 0 / ERR_FATAL 0
OK 0 / COLLISION 0 / NOACK 3 / DEFERRED 0 / ERR 0 / ERR_FATAL 0
```

I've changed the addresses in both the sender and the receiver to the Rime addresses. I've also defined `RTIMER_CONF_GUARD_TIME` to `7` in `contiki-conf.h` as you did it for `openmote-cc2538`. Is there anything else to do?

kkrentz: The first thing you could check is if #1703 works on Firefly motes, e.g., by running examples/llsec/ipv6-demos/simple-udp-rpl. Since I only have OpenMotes, I can not try this on my own. Furthermore, if you enabled the debug output in potr.c and secmac.c, I might be able to identify the problem.

bthebaudeau: Actually, I can get a few `OK`s if I wait a very long time, but I get `NOACK` most of the time.

> The first thing you could check is if #1703 works on Firefly motes, e.g., by running examples/llsec/ipv6-demos/simple-udp-rpl. Since I only have OpenMotes, I can not try this on my own.

It does not work. See https://github.com/contiki-os/contiki/pull/1703#issuecomment-225736642.

> Furthermore, if you enabled the debug output in potr.c and secmac.c, I might be able to identify the problem.

See `sender.log` and `receiver.log` there:
https://gist.github.com/bthebaudeau/1d8d9669c3b8ca40aaf0133a849d0fe6

The frequency of the `OK`s seems to be a little bit higher with `DEBUG`, but it's still very far from working smoothly. The nodes are only 1 cm apart.

kkrentz: Something goes completely wrong with your platform. With OpenMotes, I do not see any `NOACK`s at such a distance. In particular, what should not be shown is `FIFOP unexpected`.

alignan: @bthebaudeau If you are using `fireflies` (the prototype A) keep in mind you need a pigtail to connect an external 2.4 GHz antenna.  This perhaps explains it.  Send me an email and we can check this

bthebaudeau: @kkrentz I had a range issue because of the lack of antenna as suggested by @alignan, but I do not have antennae at hand right now. Yet, the basic examples like `examples/ipv6/simple-udp-rpl` work fine without these antennae, so there had to be something else.

I've found that `secmac_driver.init()` calls `configure_radio()`, which calls
`cc2538_rf_driver.set_value(RADIO_PARAM_CCA_THRESHOLD, CCA_THRESHOLD)`,
which does this:
```
REG(RFCORE_XREG_CCACTRL0) = ((int8_t) value) + RSSI_OFFSET;
```

This is wrong. It should be:
```
REG(RFCORE_XREG_CCACTRL0) = (uint8_t)(value + RSSI_OFFSET);
```

`RSSI_OFFSET` is defined as `73`, as per the CC2538 UG. `CCA_THRESHOLD` is defined as `(-50)`. Hence, the `potr` examples set the CCA threshold to -50 dBm (0x17), instead of the -81 dBm (0xF8) recommended by the UG. I've tested with `CCA_THRESHOLD` set to `(-81)`, and the RF range is back to normal. However, there are still `NOACK`s for about 1/3 of packets. These `NOACK`s correspond to `no SHR detected` on the receiver side.

I've also tried to duplicate the reboot issue on the receiver with the original `cc2538-ccm-star.c`, but to no avail. It never reboots, even after a few tens of thousands of `OK`s.

bthebaudeau: I've been able to duplicate the reboot issue with `examples/ipv6/simple-udp-rpl` and `noncoresec` in the `master` branch. The watchdog timer is triggered by a deadlock caused by `aes_auth_crypt_check_status()` never indicating an end of operation. I think that the root cause is that the status flags are cleared after the final register write access starting the operation, so they can be cleared after having been set, depending on the timing. I will investigate this by identifying the precise register write access starting each operation, thanks to the UG and some tests.

bthebaudeau: @kkrentz #1737 Fixes this reboot issue for me. Can you please confirm that it works for your use cases too? If so, you can drop the first commit of this PR (cf9fece27a9475f064252cefb740248ce21a6488), and then I can merge it.

bthebaudeau: The description of this PR would have to be updated too in this case.

kkrentz:
@bthebaudeau #1737 fixes my reboot issue, too. Also thanks for pointing out the casting bug in my RF driver.

bthebaudeau:
:+1:

PR #1708, opened 2016-06-06
TSCH enhancements

atiselsts:
This patch fixes some minor bugs and prepares TSCH code for a port to the CC2650 platform.

### DIFF
--- a/core/net/mac/tsch/tsch-adaptive-timesync.c
+++ b/core/net/mac/tsch/tsch-adaptive-timesync.c
@@ -92,6 +92,10 @@ timesync_learn_drift_ticks(uint32_t time_delta_asn, int32_t drift_ticks)
   int32_t last_drift_ppm = (int32_t)((int64_t)real_drift_ticks * TSCH_DRIFT_UNIT / time_delta_ticks);
 
   drift_ppm = timesync_entry_add(last_drift_ppm, time_delta_ticks);
+
+  TSCH_LOG_ADD(tsch_log_message,
+      snprintf(log->message, sizeof(log->message),
+          "drift %ld", drift_ppm / 256));
 }
 /*---------------------------------------------------------------------------*/
 /* Either reset or update the neighbor's drift */
--- a/core/net/mac/tsch/tsch-adaptive-timesync.c
+++ b/core/net/mac/tsch/tsch-adaptive-timesync.c
@@ -86,8 +86,8 @@ timesync_entry_add(int32_t val, uint32_t time_delta)
 static void
 timesync_learn_drift_ticks(uint32_t time_delta_asn, int32_t drift_ticks)
 {
-  /* should fit in 32-bit unsigned integer */
-  uint32_t time_delta_ticks = time_delta_asn * tsch_timing[tsch_ts_timeslot_length];
+  /* should fit in a 32-bit integer */
+  int32_t time_delta_ticks = time_delta_asn * tsch_timing[tsch_ts_timeslot_length];
   int32_t real_drift_ticks = drift_ticks + compensated_ticks;
   int32_t last_drift_ppm = (int32_t)((int64_t)real_drift_ticks * TSCH_DRIFT_UNIT / time_delta_ticks);
 
--- a/core/net/mac/tsch/tsch-conf.h
+++ b/core/net/mac/tsch/tsch-conf.h
@@ -191,4 +191,11 @@
 #define TSCH_RADIO_ON_DURING_TIMESLOT 0
 #endif
 
+/* How long to scan each channel in the scanning phase */
+#ifdef TSCH_CONF_CHANNEL_SCAN_DURATION
+#define TSCH_CHANNEL_SCAN_DURATION TSCH_CONF_CHANNEL_SCAN_DURATION
+#else
+#define TSCH_CHANNEL_SCAN_DURATION CLOCK_SECOND
+#endif
+
 #endif /* __TSCH_CONF_H__ */
--- a/core/net/mac/tsch/tsch-conf.h
+++ b/core/net/mac/tsch/tsch-conf.h
@@ -184,4 +184,11 @@
 #define TSCH_HW_FRAME_FILTERING 1
 #endif /* TSCH_CONF_HW_FRAME_FILTERING */
 
+/* Keep radio always on within TSCH timeslot (1) or turn it off between packet and ACK? (0) */
+#ifdef TSCH_CONF_RADIO_ON_DURING_TIMESLOT
+#define TSCH_RADIO_ON_DURING_TIMESLOT TSCH_CONF_RADIO_ON_DURING_TIMESLOT
+#else
+#define TSCH_RADIO_ON_DURING_TIMESLOT 0
+#endif
+
 #endif /* __TSCH_CONF_H__ */
--- a/core/net/mac/tsch/tsch-conf.h
+++ b/core/net/mac/tsch/tsch-conf.h
@@ -53,6 +53,8 @@
 #define TSCH_HOPPING_SEQUENCE_4_16 (uint8_t[]){ 20, 26, 25, 26, 15, 15, 25, 20, 26, 15, 26, 25, 20, 15, 20, 25 }
 /* 4 channels, sequence length 4 */
 #define TSCH_HOPPING_SEQUENCE_4_4 (uint8_t[]){ 15, 25, 26, 20 }
+/* 2 channels, sequence length 2 */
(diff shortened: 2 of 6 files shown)


alignan:
Travis unhappy :disappointed:

atiselsts:
I think the failure was related to bad RNG quality on msp430. Tried to fix that by changing `random_rand()` to `random_rand() >> 2` - not using the lowest-order bits usually gives better results.

core/net/mac/tsch/tsch-adaptive-timesync.c
```diff
@@ -86,12 +86,16 @@ timesync_entry_add(int32_t val, uint32_t time_delta)
...
+  /* should fit in a 32-bit integer */
+  int32_t time_delta_ticks = time_delta_asn * tsch_timing[tsch_ts_timeslot_length];
   int32_t real_drift_ticks = drift_ticks + compensated_ticks;
   int32_t last_drift_ppm = (int32_t)((int64_t)real_drift_ticks * TSCH_DRIFT_UNIT / time_delta_ticks);
 
   drift_ppm = timesync_entry_add(last_drift_ppm, time_delta_ticks);
+
+  TSCH_LOG_ADD(tsch_log_message,
+      snprintf(log->message, sizeof(log->message),
+          "drift %ld", drift_ppm / 256));
```
simonduq: Could we instead add this info to the tx/rx messages that already contain some drift info ("dr"/"edr")?

core/net/mac/tsch/tsch-slot-operation.c
```diff
@@ -370,6 +371,24 @@ update_neighbor_state(struct tsch_neighbor *n, struct tsch_packet *p,
...
+}
+/*---------------------------------------------------------------------------*/
+static void
+tsch_radio_off(int inside_slot, int after_slot)
+{
+  if((!TSCH_RADIO_ON_DURING_TIMESLOT && inside_slot)
+      || (TSCH_RADIO_ON_DURING_TIMESLOT && after_slot)) {
+    NETSTACK_RADIO.off();
+  }
+}
```
simonduq: I find the two functions a bit hard to get, after all, if you're inside a slot how could you also be after?
How abut two parameters `inside_slot`, `force`?

Also, have you considered using the global `tsch_in_slot_operation`?

core/net/mac/tsch/tsch-queue.h
```diff
@@ -39,6 +39,7 @@
 #include "lib/ringbufindex.h"
 #include "net/linkaddr.h"
 #include "net/mac/tsch/tsch-schedule.h"
+#include "net/mac/mac.h"
```
simonduq: Why was this needed in this `.h` file? (I find this odd given there are no other changes to the file)
Shouldn't this be rather included in some particular `.c` files?

core/net/mac/tsch/tsch-private.h
```diff
@@ -112,4 +112,8 @@ void tsch_disassociate(void);
 #define BUSYWAIT_UNTIL_ABS(cond, t0, offset) \
   while(!(cond) && RTIMER_CLOCK_LT(RTIMER_NOW(), (t0) + (offset))) ;
 
+#ifndef CLOCK_LT
+#define CLOCK_LT(a, b) ((int32_t)((a)-(b)) < 0)
+#endif
```
simonduq: I'm afraid this breaks things. `CLOCK_LT` is not standard in Contiki, platforms don't define it (yet?). Therefore, what happens with platforms that don't have `int32_t` as clock type?

core/net/mac/tsch/tsch-conf.h
```diff
@@ -191,4 +191,11 @@
 #define TSCH_RADIO_ON_DURING_TIMESLOT 0
 #endif
 
+/* How long to scan each channel in the scanning phase */
+#ifdef TSCH_CONF_ASSOCIATION_CHANNEL_SCAN_DURATION
+#define TSCH_ASSOCIATION_CHANNEL_SCAN_DURATION TSCH_CONF_ASSOCIATION_CHANNEL_SCAN_DURATION
+#else
+#define TSCH_ASSOCIATION_CHANNEL_SCAN_DURATION CLOCK_SECOND
+#endif
```
simonduq: how about renaming to `TSCH_CHANNEL_SCAN_DURATION`?

core/net/mac/tsch/tsch.c
```diff
@@ -609,33 +609,37 @@ PT_THREAD(tsch_scan(struct pt *pt))
...
+          (random_rand() >> 2) % sizeof(TSCH_JOIN_HOPPING_SEQUENCE)];
       if(current_channel != scan_channel) {
         NETSTACK_RADIO.set_value(RADIO_PARAM_CHANNEL, scan_channel);
         current_channel = scan_channel;
         PRINTF("TSCH: scanning on channel %u\n", scan_channel);
       }
-      current_channel_since = now_seconds;
+      /* select a random interval of time to spend scanning the channel */
+      next_channel_time = now_time + TSCH_ASSOCIATION_CHANNEL_SCAN_DURATION / 8
+          + (random_rand() >> 2) % (TSCH_ASSOCIATION_CHANNEL_SCAN_DURATION);
```
simonduq: I'm no fan of `random_rand() >> 2`, we already had a similar thing in `tsch-queue.c` and I was not happy about it, now making this more general doesn't feel good. Shouldn't we rather fix `random_rand` for platforms that have the problem?
As for getting rid of `CLOCK_LT`, would it work to add a variable `next_scan_duration`, and do `now_time - last_time > next_scan_duration`? I think this works regardless of the type because we now that `last_time` is in the past, that is, semantically lower than `now_time` (may be higher if there is wrap but means an earlier clock).

core/net/mac/tsch/tsch-queue.h
atiselsts: Hi Simon,
the code in `tsch-queue.h` file depends on the declaration of `mac_callback_t`, which may lead to compilation error for applications that directly include `tsch-queue.h` before `net/mac/mac.h` header. If the file is private and should not be included directly by applications maybe mark it as such.

core/net/mac/tsch/tsch-adaptive-timesync.c
atiselsts: There are some big differences here:
- `dr` prints drift correction in ticks, this prints the actual drift in ppm
- the drift value is only printed for packets that received from the time source node and used for time correction (and even then not always - not if the delta time is too small). There is no function to estimate the drift ppm of arbitrary neighbors in the network.
So for most of packets the ppm drift will just be "undetermined".
Also, the buffer for log message lines is quite but (48 bytes) and would need to be increased further if we want print more stuff for each tx/rx message.
I just find the printing this drift quite useful, especially for people starting with TSCH.
Maybe can somehow disable this by default, if you think it's too much?

core/net/mac/tsch/tsch-queue.h
simonduq: right I missed that, thanks!

core/net/mac/tsch/tsch-adaptive-timesync.c
simonduq: But wait, this is also only done when receiving from the time source isn't it?
A good thing with the Rx/Tx logs is that they don't suffer from the 48 byte limitation as they report the numbers via a dedicated struct, so I feel it would be convenient so simply add a field for ppms next to estimated_drift.

But I'm also fine if you don't want to change this.

atiselsts:
I addressed your comments. Changed parameter names and commented `tsch_radio_on` and `tsch_radio_off` functions, hopefully clearer now.
Randomization in scanning stage is reverted back to the one-second-per channel behavior, only now the scan duration is configurable. Looked like randomization helps to join the network faster, but this needs more investigation.

simonduq:
OK, thanks! I wish we add a general definition of `CLOCK_LT` at some point, will help for features such as what you had with random scanning times.

simonduq:
Could you change the commit log for this?
https://github.com/contiki-os/contiki/pull/1708/commits/bcab9d62d1b06fbf329433639d0e58c1d5922cca

atiselsts:
Now I had an idea that having an `enum` argument for `tsch_radio_on` and `tsch_radio_off` makes the calling code easier to read - updated the code, what do you think?

simonduq:
Right, I forgot about this part. Yes, I like this much more!

simonduq:
:+1:

g-oikonomou:
@simonduq are you happy for this to get merged?

simonduq:
Yep sure!

PR #1709, opened 2016-06-06
TSCH port for CC26xx  

### LINKED ISSUE #1615: A number of problems with CC26xx CCA
We currently use:

```
  /* Configure CCA settings */
  cmd->ccaOpt.ccaEnEnergy = 1;
  cmd->ccaOpt.ccaEnCorr = 1;
  cmd->ccaOpt.ccaEnSync = 0;
  cmd->ccaOpt.ccaCorrOp = 1;
  cmd->ccaOpt.ccaSyncOp = 1;
```

So, we are basically relying on observed RSSI and the number of correlation peaks observed during the last 8-symbol period. For the channel to be busy, we require both of those signals to be busy. SFD reception is ignored altogether. This has a number of problems:

* We are not standards-compliant. The standard (and common sense) demands the channel to be always reported as busy when an SFD is found.
* It is possible to receive a frame correctly and in its entirety, but `receiving_packet()` will be signaling 'false' throughout the entire packet reception. Thanks to @atiselsts for reporting this in the first place, which is what led to discovering this bug

In order to be standards-compliant, we need to start by setting:
```
  cmd->ccaOpt.ccaEnSync = 1;
  cmd->ccaOpt.ccaSyncOp = 0;
```

We then need to modify `receiving_packet()` to ignore RSSI and Correlator and to simply inspect the value of `ccaInfo.ccaSync` returned by `CMD_IEEE_CCA_REQ`.

Lastly, we need to choose a CCA mode among the 3 specified in .15.4. If a sync has _not_ been found, then (edit: fixed list formatting):

1. Based on RSSI value only
2. Based on correlator only
3. Based on both of the above
  * Using an OR logic: Busy if either the RSSI or Correlator test signal is busy, clear otherwise.
  * Using an AND logic: Busy if both the RSSI and Correlator test signals are busy, clear otherwise.

Unless I've missed something, `channel_clear()` will not need changed.

atiselsts:
This pull request adds a full-featured (HW timetamps, polling mode, etc.) TSCH port on CC26xx. This port has been tested for several months in a small testbed with custom CC26xx SoC based devices at the University of Bristol.

Depends on #1708 and #1684 (#1708 is used as the base branch, and some cherry-picked commits included from #1684).

Fixes #1615.

### DIFF
--- a/platform/srf06-cc26xx/contiki-main.c
+++ b/platform/srf06-cc26xx/contiki-main.c
@@ -59,6 +59,7 @@
 #include "uart.h"
 #include "sys/clock.h"
 #include "sys/rtimer.h"
+#include "sys/node-id.h"
 #include "lib/sensors.h"
 #include "button-sensor.h"
 #include "dev/serial-line.h"
@@ -67,6 +68,8 @@
 #include "driverlib/driverlib_release.h"
 
 #include <stdio.h>
+/*---------------------------------------------------------------------------*/
+unsigned short node_id = 0;
 /*---------------------------------------------------------------------------*/
 /** \brief Board specific iniatialisation */
 void board_init(void);
@@ -123,6 +126,10 @@ set_rf_params(void)
     printf("%02x\n", linkaddr_node_addr.u8[i]);
   }
 #endif
+
+  /* also set the global node id */
+  node_id = short_addr;
+  printf(" Node ID: %d\n", node_id);
 }
 /*---------------------------------------------------------------------------*/
 /**
--- a/platform/srf06-cc26xx/contiki-main.c
+++ b/platform/srf06-cc26xx/contiki-main.c
@@ -149,6 +149,11 @@ main(void)
   /* Set the LF XOSC as the LF system clock source */
   oscillators_select_lf_xosc();
 
+#if CC2650_FAST_RADIO_STARTUP
+  /* Also request HF XOSC to start up */
+  oscillators_request_hf_xosc();
+#endif
+
   lpm_init();
 
   board_init();
--- a/core/net/mac/tsch/README.md
+++ b/core/net/mac/tsch/README.md
@@ -38,6 +38,7 @@ It has been tested on the following platforms:
   * Zolertia Z1 (`z1`, tested in cooja only)
   * CC2538DK (`cc2538dk`, tested on hardware)
   * Zolertia Zoul (`zoul`, tested on hardware)
+  * CC2650 (`srf06-cc26xx`, tested on hardware)
 
 This implementation was present at the ETSI Plugtest
 event in Prague in July 2015, and did successfully inter-operate with all
@@ -78,7 +79,7 @@ Orchestra is implemented in:
 
 A simple TSCH+RPL example is included under `examples/ipv6/rpl-tsch`.
 To use TSCH, first make sure your platform supports it.
-Currently, `jn516x`, `sky`, `z1`, `cc2538dk` and `zoul` are the supported platforms.
+Currently, `jn516x`, `sky`, `z1`, `cc2538dk`, `zoul` and `srf06-cc26xx` are the supported platforms.
 To add your own, we refer the reader to the next section.
 
 To add TSCH to your application, first include the TSCH module from your makefile with:
@@ -164,7 +165,7 @@ Finally, one can also implement his own scheduler, centralized or distributed, b
 ## Porting TSCH to a new platform
 
 Porting TSCH to a new platform requires a few new features in the radio driver, a number of timing-related configuration paramters.
-The easiest is probably to start from one of the existing port: `jn516x`, `sky`, `z1`, `cc2538dk`, `zoul`.
+The easiest is probably to start from one of the existing port: `jn516x`, `sky`, `z1`, `cc2538dk`, `zoul`, `srf06-cc26xx`.
 
 ### Radio features required for TSCH
 
--- a/cpu/cc26xx-cc13xx/clock.c
+++ b/cpu/cc26xx-cc13xx/clock.c
@@ -118,6 +118,9 @@ clock_init(void)
   /* GPT0 / Timer B: One shot, PWM interrupt enable */
   HWREG(GPT0_BASE + GPT_O_TBMR) =
         ((TIMER_CFG_B_ONE_SHOT >> 8) & 0xFF) | GPT_TBMR_TBPWMIE;
+
+  /* enable sync with radio timer */
+  HWREGBITW(AON_RTC_BASE + AON_RTC_O_CTL, AON_RTC_CTL_RTC_UPD_EN_BITN) = 1;
 }
 /*---------------------------------------------------------------------------*/
 static void
--- a/cpu/cc26xx-cc13xx/dev/soc-rtc.c
+++ b/cpu/cc26xx-cc13xx/dev/soc-rtc.c
@@ -81,11 +81,13 @@ soc_rtc_init(void)
 
   ti_lib_aon_rtc_event_clear(AON_RTC_CH0);
   ti_lib_aon_rtc_event_clear(AON_RTC_CH1);
+  ti_lib_aon_rtc_event_clear(AON_RTC_CH2);
 
   /* Setup the wakeup event */
   ti_lib_aon_event_mcu_wake_up_set(AON_EVENT_MCU_WU0, AON_EVENT_RTC_CH0);
   ti_lib_aon_event_mcu_wake_up_set(AON_EVENT_MCU_WU1, AON_EVENT_RTC_CH1);
-  ti_lib_aon_rtc_combined_event_config(AON_RTC_CH0 | AON_RTC_CH1);
+  ti_lib_aon_event_mcu_wake_up_set(AON_EVENT_MCU_WU2, AON_EVENT_RTC_CH2);
+  ti_lib_aon_rtc_combined_event_config(AON_RTC_CH0 | AON_RTC_CH1 | AON_RTC_CH2);
 
   HWREG(AON_RTC_BASE + AON_RTC_O_SEC) = SOC_RTC_START_TICK_COUNT;
 
@@ -123,7 +125,7 @@ soc_rtc_get_next_trigger()
 void
 soc_rtc_schedule_one_shot(uint32_t channel, uint32_t ticks)
 {
-  if((channel != AON_RTC_CH0) && (channel != AON_RTC_CH1)) {
+  if((channel != AON_RTC_CH0) && (channel != AON_RTC_CH1) && (channel != AON_RTC_CH2)) {
     return;
   }
 
@@ -168,6 +170,12 @@ soc_rtc_isr(void)
     ti_lib_aon_rtc_channel_disable(AON_RTC_CH0);
     HWREG(AON_RTC_BASE + AON_RTC_O_EVFLAGS) = AON_RTC_EVFLAGS_CH0;
     rtimer_run_next();
+  }
+
+  if(ti_lib_aon_rtc_event_get(AON_RTC_CH2)) {
+    /* after sleep; since a rtimer is already scheduled, do nothing */
+    ti_lib_aon_rtc_channel_disable(AON_RTC_CH2);
+    HWREG(AON_RTC_BASE + AON_RTC_O_EVFLAGS) = AON_RTC_EVFLAGS_CH2;
   }
 
   ENERGEST_OFF(ENERGEST_TYPE_IRQ);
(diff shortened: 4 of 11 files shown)


atiselsts:
Now that #1708 and #1684 have been merged, how about reviewing this? I rebased to the latest Contiki `master` just now. Thanks!

cpu/cc26xx-cc13xx/putchar.c
```diff
@@ -37,6 +37,7 @@ int
 putchar(int c)
 {
   cc26xx_uart_write_byte(c);
+  while(cc26xx_uart_busy() == UART_BUSY);
```
g-oikonomou: This is definitely not required.

platform/srf06-cc26xx/contiki-main.c
```diff
@@ -123,6 +126,10 @@ set_rf_params(void)
     printf("%02x\n", linkaddr_node_addr.u8[i]);
   }
 #endif
+
+  /* also set the global node id */
+  node_id = short_addr;
```
g-oikonomou: I cannot immediately see why we need this. Is this required by TSCH?

The CPU makefile does this:
```
ifdef NODEID
  CFLAGS += -DIEEE_ADDR_NODE_ID=$(NODEID)
endif
```

Is this not suitable?

g-oikonomou:
Atis, thanks for your efforts with this! 

I went through most of it and it looks great. I know I still owe you taking a look at fast RF startup and I'd also like to see what logic you decided to use for CCA (OR vs AND).

Other than that, superb.

Closes #1615

platform/srf06-cc26xx/contiki-main.c
simonduq: The node ID isn't required anymore in the `rpl-tsch` example

simonduq:
Great contribution. Now that https://github.com/contiki-os/contiki/pull/1708 is in I'm fine with this PR getting merged as it doesn't touch core.

platform/srf06-cc26xx/contiki-main.c
atiselsts: We discussed this with George off-line and he's ok with this change. I personally think it would be nice to have `node_id` for all Contiki platforms. There are other examples besides `rpl-tsch` (at least one even in Contiki mainline) that rely on `node_id`. A nice way select address-specific functionality in platform and even network stack independent way (works with Rime too).

simonduq: well, I don't really mind!

g-oikonomou:
My objections about the changes to the UART driver have been addressed and I'm happy with the `node_id` thing as Atis says. I'll merge this when travis goes green

alignan:
I like the idea of `node_id` being standard, happy to merge a PR on that

simonduq:
Now we have TSCH supported on all major platforms! Let's make it the default MAC ;)

niklas88:
I'm seeing a regression in power consumption with these changes. I'm using a TI SensorTag with a simple RIME Collect app for collecting sensor data. In the first figure power consumption is plotted with a version before the TSCH changes, the second is with them and the third figure is a zoom in of the higher power consumption area in the second. I've tried with all sensor reading disabled so it's not an interference with that.

contiki_async_first
contiki_async_second
contiki_async_second_zoom

My project-conf.h looks like this:  http://pastebin.com/0JBthp66

I've tried bisecting but got compile errors at several of the commits so that only narrowed it down to the following.

The first bad commit could be any of:
97c1cfc3b4fe4ea430f5ed6413a5d9867ed6728d
e19fbc996e240c0222409ac2b51a44645f7caf89
25c5f0b744970343841edd3d7d26215d60fceefc
3a99639294bef07938ed9ccb54603060ef9226c1
80aa30c5da2aeaf2f6688930c54d6a39cb47944b
a47fb723e468368808f44e48712552c52519a31a
54e4b5f3510ecb9f249eaf8e465aef3344c65b99
95b66657aab656e3ca9b2a44d3c2e8cf093ff4d2
ac6f8008fd8e4cf9c3802353de1c1ddebd6b7d63
d85667d5356b3a08773adadce69eab552b43d805
e233c3f27f9adc475e9d6d6d0b4cac560644eb0c

atiselsts:
Dear Niklas, the patchset introduces an "early wakeup" logic for
CC26xx. The lpm.c code exits deep sleep mode 1 ms before the actual
wakeup is requested in order to initialize HF crystal oscillator so
that it can be used soon enough after the start of a TSCH timeslot.
That means spending 1ms extra time before each wakeup in what Contiki
calls "sleep mode" (not deep sleep) with >1 mA current consumption.

The new logic is enabled by default. Add this to project-conf.h to disable it:
#define CC2650_FAST_RADIO_STARTUP     0
Would be great if you could verify that this fixes the problem for you
- there are also some other changes and bugfixes in lpm.c file and the
IEEE mode radio driver.

I agree that this is unfortunate that this change affects completely
unrelated applications. The problem is that there is no easy way to
produce different default CC26xx platform's configs for TSCH
applications and for non-TSCH applications - there is no compile-time
constant that tells whether TSCH is enabled. There is also no way to
tell before a wakeup whether the radio is actually going to be used
immediately after it, so the code in lpm.c just assumes the worst case
and always starts the HF crystal oscillator.

bkozak-scanimetrics:
I've just had a quick look at the changes that have been made to lpm.c and I think I know what @niklas88's problem is.

The updated lpm code appears to have removed the default sleep period that I had added to reduce power consumption when no etimers are running.

You can see the relevant changes on [this diff](https://github.com/contiki-os/contiki/commit/3a99639294bef07938ed9ccb54603060ef9226c1#diff-fa46fc35ebe51649ba292cbb938277cd) (scroll down to the removed line 452 to see the code which had been removed).

This will only tend to affect RIME code since we'll usually have at least one etimer scheduled at all times when running IPv6. @niklas88 __is__ running RIME so this is probably his issue. If you add back the else clause that was removed with this PR then you should see the problem go away.

Although the code under ```CC2650_FAST_RADIO_STARTUP``` is likely to increase power consumption also, I don't believe that it is the cause of your issue.

niklas88:
Thanks all for looking into this. Unfortunately both suggestions didn't fix the problem.
I used the following diff for @bkozak-scanimetrics hint:
http://pastebin.com/RhtxX4TL

Unfortunately I'm not sure I can  share my code publicly even though I'd consider it a pretty trivial derivative mix of example-collect.c and cc26xx-demo.c. As for no etimers running. In my design there are two processes. One has an explicit etimer with a period of 5 seconds which triggers sensor readouts. After all sensor data is retrieved (the sensors_events have fired and the data been read) this process posts to another process which sends all sensor data as a single message via collect_send().

simonduq:
@atiselsts don't be afraid of adding some platform-specific config for tsch in the tsch example's project-conf for now. This is how we do it currently. Ofc in the long run we need something better, e.g. a platform specific conf file included after project-conf, or a make flag.

atiselsts:
I will look into energy consumption and try to prepare a patch in the next
week when I have access to hardware again. One problem in your setup is
CC2650_FAST_RADIO_STARTUP for sure; if there is something else we need to
find out.

Simon, now that I think about it, I still would like to avoid the need to
have application-specific config to get TSCH running. Not particularly
new-user friendly, and does not scale with the number of platforms.
I feel that a better option is either to reduce the "early wakeup" time as
much as possible (it's definitely possible to go below 1ms, but we need
longer-term experiments to confirms that there's no bad effect on PDR), or
to have slightly nonstandard timing. Using 15 ms TSCH slots was one
solution, but reducing the TSCH guard time size just a little (say, to 2ms
instead of 2.2ms) for CC26xx would also allow help. That way CC26xx could
get rid of CC2650_FAST_RADIO_STARTUP as the default config option. (It
would still keep this setting as an option, as some applications may need
it.)

Getting rid of the FAST_STARTUP trick also means better energy efficiency
for TSCH applications. Interestingly, 6tisch RFCs actually talk about 1ms
guard time instead of 2.2 ms. Given that TSCH syncs on EBs, and EB period
in Contiki is just 4 seconds, 2.2ms long guard time is not really needed
for reliability by default; the only reason to have it is "because the IEEE
802.15.4 standard says so" in my opinion. If having it makes stuff much
worse on a specific platform, it makes sense to ignore this recommendation
on that platform.

bkozak-scanimetrics:
> One has an explicit etimer with a period of 5 seconds which triggers sensor readouts. 

In that case, I think I may see the issue. It's still sort of in the same vein as the last issue I pointed out. The issue will occur whenever you have no etimer which is set to expire at less than 1 second (i.e. ```MAX_SLEEP_TIME```) into the future.

If you trace the logic in ```setup_sleep_mode()```, you'll find that with a 5 second etimer, you won't end up setting a wakeup interrupt at all. This will cause you to wake up on the next clock tick.

See the code snippet from ```setup_sleep_mode()``` below with my comments added:
```C
next_etimer = soc_rtc_last_isr_time() + (until_next_etimer * (RTIMER_SECOND / CLOCK_SECOND));
    if(RTIMER_CLOCK_LT(next_etimer, now + STANDBY_MIN_DURATION)) {
      soc_rtc_schedule_one_shot(AON_RTC_CH1, now + MINIMAL_SAFE_SCHEDULE);
      return LPM_MODE_SLEEP;
    }
    /* With a 5 second etimer, you'll end up in this conditional here. The variable
        max_sleep is set one second into the future. */
    if(RTIMER_CLOCK_LT(max_sleep, next_etimer)) {
      
      /* *next_timer holds the time of the next clock tick, this will be
          less than max sleep so you'll skip this */
      if(RTIMER_CLOCK_LT(max_sleep, *next_timer)) {
        soc_rtc_schedule_one_shot(AON_RTC_CH1, max_sleep);
      }
    } else {
      soc_rtc_schedule_one_shot(AON_RTC_CH1, next_etimer);
    }

    /* then we end up here without having scheduled AON_RTC_CH1.
       the CH1 interrupt will remain set to the next clock tick in 512 RTC ticks. 
       There are no more opportunities to set CH1 after this. */

    if(RTIMER_CLOCK_LT(next_etimer, *next_timer)) {
      *next_timer = next_etimer;
    }
```

Once again, this is something that is unlikely to occur with IPv6 since we will most often have some etimer scheduled at a point less than 1 second into the future.

bkozak-scanimetrics:
> Simon, now that I think about it, I still would like to avoid the need to have application-specific config to get TSCH running.

I'm not at all familiar with the TSCH code so, forgive me if I say something stupid, but would it not be better to avoid doing this in lpm.c altogether? Having to pay the cost of the early wakeup on every single sleep cycle seems wasteful.

Surely this fast wakeup can be accomplished by setting an rtimer to occur 1ms before your planned radio turn on then using a reigstered lpm module to stay awake until the actual time slot. If you're already using an rtimer then #1290 will allow you to set more than one. 

Again, forgive me if I've misunderstood something about the problem. Perhaps the issue is that you don't know for sure when you will next need to turn on the radio?

niklas88:
@bkozak-scanimetrics your reasoning makes sense so does this mean the DEFAULT_SLEEP_TIME else needs to be somewhere else to actually work as compared with the change I tried?

bkozak-scanimetrics:
> DEFAULT_SLEEP_TIME else needs to be somewhere else to actually work as compared with the change I tried?

That'll work, but you could also leave the other ```DEFAULT_SLEEP_TIME``` you added before since that one covers the seperate case where no etimers are set.

It should fix your issue if you add your else clause after
```C
      if(RTIMER_CLOCK_LT(max_sleep, *next_timer)) {
        soc_rtc_schedule_one_shot(AON_RTC_CH1, max_sleep);
      }
```
which is at line 307 of the current lpm.c

atiselsts:
Billy, I believe you have now found the problem - let's wait for the
results. Somehow missed the fact the CH1 interrupts are generated by
default on each tick.
The code can in `setup_sleep_mode` can be simplified as a result - it
should always reschedule the CH1 interrupt before sleeping.

With regards to the design decisions behind all this, I think Simon is the
best person to answer and explain that. I believe the main constraints are
to keep the core TSCH code platform-independent and to not schedule any
timers related to a specific timeslot before the start of that timeslot
(i.e. do not occupy the previous timeslot, as it might be busy).
​

bkozak-scanimetrics:
> I believe the main constraints are to keep the core TSCH code platform-independent 

Right, completely slipped my mind that the lpm modules aren't platform independent.

niklas88:
Good news everyone! I've just done a power consumption test with the suggested changes and it seems to indeed solve the problem. I used both the older change as well as the new one. I guess however that since this makes the setup_sleep function pretty convoluted it would be best to get it properly organized? I'd of course be happy to test such changes.

PR #1711, opened 2016-06-08
One fix to sensorTag board_i2c.c

xukai871105:
According to “CC1310, CC26XX Simplelink Wireless MCU Technical Reference Manual” In Figure 21-10 In Master RECEIVE With Repeated Start Condition. After reading len-1 bytes from slave device(such as mpu9250), CC13XX/CC26XX should write 101 to I2C_MCTRL firstly(ti_lib_i2c_master_control(I2C0_BASE, I2C_MASTER_CMD_BURST_RECEIVE_FINISH)),then read the data from I2C_MDR(ti_lib_i2c_master_data_get(I2C0_BASE)).There is something wrong in board_i2c.c board_i2c_read() and board_i2c_write_read()
On sensorTag platform mpu9250 works fine.But when we monitor the I2Cbus, we find that mpu9250 gives one more byte and CC2650 sends I2C_STOP at a worng time.
We have done the test on PN7120(NFC controller from NXP), new board_i2c_read() and board_i2c_write_read work fine.


g-oikonomou:
This looks reasonable. We're looking into it.

PR #1715, opened 2016-06-09
TSCH port for Zolertia ReMote/Zoul.

sumanpanchal:


simonduq:
Cool :)
Two things:
* please add the new supported platforms to the TSCH readme (several locations within the file)
* regarding the node-id problem in rpl-tsch, I'm not particularly keen on keeping it as it is. We could for instance replace it with a hardcoded coordinator address, set by default to the MAC address of node 1 in Cooja for z1 nodes (because this is what is used for the regression tests)

alignan:
Great contrib! I thought this was already included in the previous CC2538/TSCH commit, so great to have :smile:

sumanpanchal:
@simonduq @alignan 
I have done all the necessary changes as per your suggestion.

alignan:
:+1: from me, @simonduq?

simonduq:
:+1:

sumanpanchal:
Thanks @simonduq and @alignan :)

PR #1719, opened 2016-06-12
RPL: Remove useless `else' block as def_route will be removed before if ex…

jcook:
RPL: Remove useless `else' block as def_route will be removed before if existed.
Keep this logic will bring confusion.

### DIFF
--- a/core/net/rpl/rpl-dag.c
+++ b/core/net/rpl/rpl-dag.c
@@ -554,13 +554,6 @@ rpl_set_default_route(rpl_instance_t *instance, uip_ipaddr_t *from)
     if(instance->def_route == NULL) {
       return 0;
     }
-  } else {
-    PRINTF("RPL: Removing default route\n");
-    if(instance->def_route != NULL) {
-      uip_ds6_defrt_rm(instance->def_route);
-    } else {
-      PRINTF("RPL: Not actually removing default route, since instance had no default route\n");
-    }
   }
   return 1;
 }


simonduq:
The intent of this block was to enable deleting the default route by simply passing a null argument. Closing. Please re-open if I missed something!

greg-king5:
`rpl_set_default_route()` replaces an old default-route.  It does that by deleting the old route -- unconditionally; then, adding a new route -- conditionally.  There is no point in _trying_ to delete the old route _twice_.

simonduq:
Right, so I was indeed missing something! Makes sense.
Sorry.

simonduq:
:+1: if Travis is green

MohamedSeliem:
i do not think removing the else statement will do the fix, 
you may try this:
int
 rpl_set_default_route(rpl_instance_t *instance, uip_ipaddr_t *from)
 {

   if(from != NULL) {
     PRINTF("RPL: Adding default route through ");
     PRINT6ADDR(from);
     PRINTF("\n");
     uip_ds6_defrt_rm(instance->def_route); //first remove the old route
     instance->def_route = uip_ds6_defrt_add(from,
         RPL_DEFAULT_ROUTE_INFINITE_LIFETIME ? 0 : RPL_LIFETIME(instance, instance->default_lifetime)); // add the new route assuming it has value that is not equal to null
     if(instance->def_route == NULL) {
       return 0; // however the new def_route is equal null return 0 to indicate that the default route is not added
     }
  } else {
   // in case from is equal to null, then we need to remove the def_route
    PRINTF("RPL: Removing default route\n");
    if(instance->def_route != NULL) {
      uip_ds6_defrt_rm(instance->def_route);
    } else {
      PRINTF("RPL: Not actually removing default route, since instance had no default route\n");
   }
   }
   return 1;
 }

greg-king5:
Why would it not work?  After simply removing that redundant else-block, the old code will continue to do the same thing that your code does; but, it is smaller:

1. If there is an old route, then remove it.
2. If a new route is requested, then attempt to add it.
3. Return "false" if that attempt failed.
4. Return "true" if the attempt succeeded, or it was not requested.

MohamedSeliem:
:+1:

jcook:
@simonduq so you will merge this change?

simonduq:
waiting for another pair of eyes from the merge team to validate, see the merging policy:
https://github.com/contiki-os/contiki/blob/master/CONTRIBUTING.md#pull-request-merging-policy

nvt:
:+1:

PR #1720, opened 2016-06-12
Fix for compilation with clang - issue reported by Olaf Bergmann

joakimeriksson:
Clang seems not to discover that the statement is always false (while GCC does) and therefore complains about lack of the function called within the always false block e.g. rpl_ns_periodic();. Olaf Bergmann reported it to me and I got the same problem compiling with clang on my OS-X. This fix the issue.

### DIFF
--- a/core/net/rpl/rpl-timers.c
+++ b/core/net/rpl/rpl-timers.c
@@ -84,11 +84,13 @@ handle_periodic_timer(void *ptr)
   rpl_dag_t *dag = rpl_get_any_dag();
 
   rpl_purge_dags();
-  if(dag != NULL && RPL_IS_STORING(dag->instance)) {
-    rpl_purge_routes();
-  }
-  if(dag != NULL && RPL_IS_NON_STORING(dag->instance)) {
-    rpl_ns_periodic();
+  if(dag != NULL) {
+    if(RPL_IS_STORING(dag->instance)) {
+      rpl_purge_routes();
+    }
+    if(RPL_IS_NON_STORING(dag->instance)) {
+      rpl_ns_periodic();
+    }
   }
   rpl_recalculate_ranks();
 


alignan:
:+1:

simonduq:
Why doing this only for this particular instance of `RPL_IS_NON_STORING`? Why not do it for the `RPL_IS_STORING` above?

The purpose of `RPL_IS_NON_STORING` is to yield simple code that does the compile-time and run-time check. If the other code locations with `RPL_IS_NON_STORING` don't cause any problem with Clang I'm confident there is a cleaner solution. 

How adding a variable `instance` assigned at the beginning of the function instead? Let it be `NULL` if `dag` is `NULL`.

joakimeriksson:
Assigning a runtime variable will not help that function will still be problematic. I can do it like this:

     if(RPL_WITH_STORING && dag != NULL && RPL_IS_STORING(dag->instance)) {
       rpl_purge_routes();
     }
     if(RPL_WITH_NON_STORING && dag != NULL && RPL_IS_NON_STORING(dag->instance)) {
       rpl_ns_periodic();
     }

To make it more "fair" between storing and non storing but it needs to be at compile time - clang seems not be the best at figuring out the truth value of the DEFINE RPL_IS_NON_STORING()... It is obviously always false but clang seems to fail seeing that here.

simonduq:
but `RPL_WITH_NON_STORING` is checked in `RPL_IS_NON_STORING` so I don't really understand

joakimeriksson:
Neither do I. Ask the clang team ;-)

simonduq:
Have you tried:
```
struct rpl_instance *instance = dag != NULL dag->instance : NULL;
if(RPL_IS_STORING(instance)) {
  rpl_purge_routes();
}
if(RPL_IS_NON_STORING(instance)) {
  rpl_ns_periodic();
}
```

joakimeriksson:
Clang is intestesting... Changed the PR a bit now. Since it just needs not to evaluate the dag in the same if... That seems to solve the problem - thanks for the suggestion.

simonduq:
Yep there are many other instances of `if(RPL_IS_NON_STORING())` so I found it odd you'd only need to fix this one.

joakimeriksson:
Likely due to combining it with something more. I guess we should have a OS-X machine under control of Travis somewhere ;-)

simonduq:
Can't we do it in Linux?

joakimeriksson:
Yes, I guess the same clang as I am running on OS-X should be possible to install in linux also.

joakimeriksson:
Also fails with this:
Ubuntu clang version 3.5.0-4ubuntu2~trusty2 (tags/RELEASE_350/final) (based on LLVM 3.5.0)
Target: i386-pc-linux-gnu
Thread model: posix

So - yes will also work fine "failing" in linux.

joakimeriksson:
Another message list failure in Cooja - I do think we should add a class method in Simulator that know the visibility - otherwise we will have to throw in that argument in all methods... (Simulator.hasGUI() or similar).

simonduq:
annoying yes... restarted the job

joakimeriksson:
I will try to do another Cooja PR during the weekend (if I get the time).

simonduq:
:+1:

PR #1723, opened 2016-06-13
core/net/mac/frame802154: Fix typo in PAN ID compression condition

gebart:
I stumbled upon this warning when trying out the new GCC 6.1 release:

```
../../../contiki/core/net/mac/frame802154.c: In function ‘frame802154_has_panid’:
../../../contiki/core/net/mac/frame802154.c:151:30: warning: self-comparison always evaluates to true [-Wtautological-compare]
       if(fcf->dest_addr_mode == fcf->dest_addr_mode) {
                              ^~
```

I believe this is a typo and that the correct fix is the supplied PR, but please validate this independently.

### DIFF
--- a/core/net/mac/frame802154.c
+++ b/core/net/mac/frame802154.c
@@ -146,30 +146,18 @@ frame802154_has_panid(frame802154_fcf_t *fcf, int *has_src_pan_id, int *has_dest
   }
 
   if(fcf->frame_version == FRAME802154_IEEE802154E_2012) {
+    /* IEEE 802.15.4e-2012, Table 2a, PAN ID Compression */
     if(!fcf->panid_compression) {
-      /* Compressed PAN ID == no PAN ID at all */
-      if(fcf->dest_addr_mode == fcf->dest_addr_mode) {
-        /* No address or both addresses: include destination PAN ID */
-        dest_pan_id = 1;
-      } else if(fcf->dest_addr_mode) {
-        /* Only dest address, include dest PAN ID */
+      if(fcf->dest_addr_mode) {
+        /* Use destination PAN ID if destination address is present */
         dest_pan_id = 1;
       } else if(fcf->src_addr_mode) {
         /* Only src address, include src PAN ID */
         src_pan_id = 1;
       }
-    }
-    if(fcf->dest_addr_mode == 0 && fcf->dest_addr_mode == 1) {
-      /* No address included, include dest PAN ID conditionally */
-      if(!fcf->panid_compression) {
-        dest_pan_id = 1;
-      }
-    }
-    /* Remove the following rule the day rows 2 and 3 from table 2a are fixed: */
-    if(fcf->dest_addr_mode == 0 && fcf->dest_addr_mode == 0) {
-      /* Not meaningful, we include a PAN ID iff the compress flag is set, but
-       * this is what the standard currently stipulates */
-      dest_pan_id = fcf->panid_compression;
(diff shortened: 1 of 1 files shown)


simonduq:
I confirm this is a bug, and not the only one. It is something I had seen and fixed before but forgot to upstream it seems.

I think the test should be:
`if((fcf->dest_addr_mode == 0) == (fcf->src_addr_mode == 0)) {`
(because we don't want equality, we only want both 0 or both non-0)

Then, the next if `if(fcf->dest_addr_mode == 0 && fcf->dest_addr_mode == 1) {` should basically be removed together with its content.

The third one should also become:
`if(fcf->dest_addr_mode == 0 && fcf->src_addr_mode == 0) {`

gebart:
I rewrote the conditions based on table 2a in [IEEE 802.15.4e-2012](https://standards.ieee.org/getieee802/download/802.15.4e-2012.pdf)

core/net/mac/frame802154.c
```diff
@@ -146,30 +146,18 @@ frame802154_has_panid(frame802154_fcf_t *fcf, int *has_src_pan_id, int *has_dest
...
-        /* Only dest address, include dest PAN ID */
+      if(fcf->dest_addr_mode) {
+        /* Use destination PAN ID if destination address is present */
         dest_pan_id = 1;
       } else if(fcf->src_addr_mode) {
         /* Only src address, include src PAN ID */
         src_pan_id = 1;
       }
-    }
-    if(fcf->dest_addr_mode == 0 && fcf->dest_addr_mode == 1) {
```
gebart: This condition could never be true

core/net/mac/frame802154.c
```diff
@@ -146,30 +146,18 @@ frame802154_has_panid(frame802154_fcf_t *fcf, int *has_src_pan_id, int *has_dest
...
-      /* No address included, include dest PAN ID conditionally */
-      if(!fcf->panid_compression) {
-        dest_pan_id = 1;
-      }
-    }
-    /* Remove the following rule the day rows 2 and 3 from table 2a are fixed: */
-    if(fcf->dest_addr_mode == 0 && fcf->dest_addr_mode == 0) {
```
gebart: I guess this is another typo, likely the second part of the condition would have been src instead of dest

simonduq: Yes, as mentioned in my previous comment. But the block is needed to comply with 802.15.4-e-2012

core/net/mac/frame802154.c
simonduq: Right this needs removing as mentioned in my previous comment

core/net/mac/frame802154.c
simonduq: OK I see now that you have reworked the whole thing, I like it this way. Only thing I would add is a comment on the case with no address included, which is a bug in 802.15.4e-2012 (probably fixed now in 802.15.4-2015, but this needs double-checking)

gebart: Could you post a link to the 2015 standard? I only have the 2011 document for reference together with the 2012e addendum

simonduq: Unfortunately not, it has not yet become free

simonduq:
:+1:

gebart:
OK to squash?

alignan:
Yes, please do

gebart:
squashed

simonduq:
Thanks!

PR #1724, opened 2016-06-13
Coffee: minor enhancements

nvt:
* Simplify Coffee by changing certain CPP conditionals to C conditionals, which should be possible for the compiler to optimize based on their constant values.
* Remove the I/O semantics conditional because it is good to have consistent API across Coffee ports.
* Change the struct cfs_dir internal variable name to something more sensible.
* Minor clean up of the source code.


alignan:
:+1:

PR #1725, opened 2016-06-13
Refactor tcpip

bkozak-scanimetrics:
This PR performs some re-factoring of ```tcpip.c```, mainly focused on ```tcpip_ipv6_output()```.

The main purpose of this PR is to prepare for the changes proposed in #1186. As per the discussion on #1186, these commits were originally part of #1186 but were split apart.

These commits accomplish the following:
1. Improve the readability of ```tcpip_ipv6_output()``` so as to make future modifications to this function easier.
2. Reduce the size of generated code, esp. on the 6502 platforms where it is easy to run out of memory.

### DIFF
--- a/core/net/ip/tcpip.c
+++ b/core/net/ip/tcpip.c
@@ -116,7 +116,14 @@ enum {
 #else
 #define NEXTHOP_NON_STORING(addr) 0
 #endif
-
+/*---------------------------------------------------------------------------*/
+static void
+setup_appstate(uip_tcp_appstate_t* as, void* state)
+{
+  as->p = PROCESS_CURRENT();
+  as->state = state;
+}
+/*---------------------------------------------------------------------------*/
 /* Called on IP packet output. */
 #if NETSTACK_CONF_WITH_IPV6
 
@@ -236,8 +243,7 @@ tcp_connect(const uip_ipaddr_t *ripaddr, uint16_t port, void *appstate)
     return NULL;
   }
 
-  c->appstate.p = PROCESS_CURRENT();
-  c->appstate.state = appstate;
+  setup_appstate(&c->appstate, appstate);
 
   tcpip_poll_tcp(c);
 
@@ -288,8 +294,8 @@ tcp_attach(struct uip_conn *conn,
   uip_tcp_appstate_t *s;
 
   s = &conn->appstate;
-  s->p = PROCESS_CURRENT();
-  s->state = appstate;
+
+  setup_appstate(s, appstate);
 }
 
 #endif /* UIP_TCP */
@@ -302,8 +308,8 @@ udp_attach(struct uip_udp_conn *conn,
   uip_udp_appstate_t *s;
 
   s = &conn->appstate;
-  s->p = PROCESS_CURRENT();
-  s->state = appstate;
+
+  setup_appstate(s, appstate);
 }
 /*---------------------------------------------------------------------------*/
 struct uip_udp_conn *
@@ -318,8 +324,7 @@ udp_new(const uip_ipaddr_t *ripaddr, uint16_t port, void *appstate)
   }
 
   s = &c->appstate;
-  s->p = PROCESS_CURRENT();
-  s->state = appstate;
+  setup_appstate(s, appstate);
 
   return c;
 }
@@ -347,8 +352,7 @@ udp_broadcast_new(uint16_t port, void *appstate)
 uint8_t
 icmp6_new(void *appstate) {
   if(uip_icmp6_conns.appstate.p == PROCESS_NONE) {
-    uip_icmp6_conns.appstate.p = PROCESS_CURRENT();
-    uip_icmp6_conns.appstate.state = appstate;
+    setup_appstate(&uip_icmp6_conns.appstate, appstate);
     return 0;
   }
   return 1;
@@ -863,14 +867,8 @@ PROCESS_THREAD(tcpip_process, ev, data)
   PROCESS_BEGIN();
 
 #if UIP_TCP
-  {
-    unsigned char i;
-
-    for(i = 0; i < UIP_LISTENPORTS; ++i) {
-      s.listenports[i].port = 0;
-    }
-    s.p = PROCESS_CURRENT();
-  }
+  memset(s.listenports, 0, UIP_LISTENPORTS*sizeof(*(s.listenports)));
+  s.p = PROCESS_CURRENT();
 #endif
 
   tcpip_event = process_alloc_event();
--- a/core/net/ip/tcpip.c
+++ b/core/net/ip/tcpip.c
@@ -111,6 +111,12 @@ enum {
   PACKET_INPUT
 };
 
+#if UIP_CONF_IPV6_RPL && RPL_WITH_NON_STORING
+#define NEXTHOP_NON_STORING(addr) rpl_srh_get_next_hop(addr)
+#else
+#define NEXTHOP_NON_STORING(addr) 0
+#endif
+
 /* Called on IP packet output. */
 #if NETSTACK_CONF_WITH_IPV6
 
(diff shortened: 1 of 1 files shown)


simonduq:
Looks great! This will need thorough review and testing but I believe this is well worth it. Feels good to see such a streamlined `tcpip_ipv6_output()`!

core/net/ip/tcpip.c
```diff
@@ -282,8 +294,8 @@ tcp_attach(struct uip_conn *conn,
   uip_tcp_appstate_t *s;
 
   s = &conn->appstate;
-  s->p = PROCESS_CURRENT();
-  s->state = appstate;
+
+  setup_appstate(s, appstate);
```
greg-king5: In this function and the ones below it, the local variable `s` no longer is needed.
```c
  setup_appstate(&conn->appstate, appstate);
```

core/net/ip/tcpip.c
```diff
@@ -526,222 +536,268 @@ tcpip_input(void)
...
+#if UIP_CONF_IPV6_RPL
+  rpl_dag_t *dag;
+  rpl_instance_t *instance;
 
-  if(uip_len > UIP_LINK_MTU) {
-    UIP_LOG("tcpip_ipv6_output: Packet to big");
-    uip_clear_buf();
-    return;
+  /* If we are running RPL, and if we are the root of the
+     network, we'll trigger a global repair berfore we remove
```
greg-king5: Typo: berfore -> before.

core/net/ip/tcpip.c
```diff
@@ -526,222 +536,268 @@ tcpip_input(void)
...
-      }
-#endif /* TCPIP_CONF_ANNOTATE_TRANSMISSIONS */
-    }
+  if(uip_len == 0) {
+    return;
+  }
 
-    /* End of next hop determination */
+  if(uip_len > UIP_LINK_MTU) {
+    UIP_LOG("tcpip_ipv6_output: Packet to big");
```
greg-king5: Typo: to -> too.

bkozak-scanimetrics:
Fixed some typos and removed some unnecessary temporary vars thanks to @greg-king5

core/net/ip/tcpip.c
```diff
@@ -276,44 +288,30 @@ tcp_listen(uint16_t port)
...
+  setup_appstate(&conn->appstate, appstate);
 }
 /*---------------------------------------------------------------------------*/
 struct uip_udp_conn *
 udp_new(const uip_ipaddr_t *ripaddr, uint16_t port, void *appstate)
 {
   struct uip_udp_conn *c;
-  uip_udp_appstate_t *s;
 
   c = uip_udp_new(ripaddr, port);
```
greg-king5: You can change that expression into an initializer of the variable's definition.

tonnenpinguin: I'm not sure, but from what I've read in other issues older compilers seem to have problems with initializing variables with non-const values.

greg-king5: The cc65 compiler has that problem, but only with structs and unions.  It can handle non-constant expressions in the  initiations of "scalar" variables.

tonnenpinguin: I see I see. Learn something new every day...

bkozak-scanimetrics:
Made change suggested by @greg-king5 in ```udp_new()```

> You can change that expression into an initializer of the variable's definition

alignan:
I like it as well, I haven't tested but the code looks OK and Travis is happy :+1:

core/net/ip/tcpip.c
```diff
@@ -526,222 +522,268 @@ tcpip_input(void)
...
-        uip_packetqueue_free(&nbr->packethandle);
-        tcpip_output(uip_ds6_nbr_get_ll(nbr));
-      }
-#endif /*UIP_CONF_IPV6_QUEUE_PKT*/
+  tcpip_output(linkaddr);
 
-      uip_clear_buf();
-      return;
-    }
+  if(nbr) {
```
simonduq: I would prefer having this block conditioned by `UIP_CONF_IPV6_QUEUE_PKT` (rather than the body of `send_queued`).
Would be clearer IMO

core/net/ip/tcpip.c
```diff
@@ -230,8 +243,7 @@ tcp_connect(const uip_ipaddr_t *ripaddr, uint16_t port, void *appstate)
     return NULL;
   }
 
-  c->appstate.p = PROCESS_CURRENT();
-  c->appstate.state = appstate;
+  setup_appstate(&c->appstate, appstate);
```
simonduq: naming suggestion: `init_appstate`?

core/net/ip/tcpip.c
```diff
@@ -526,222 +522,268 @@ tcpip_input(void)
...
+  tcpip_output(linkaddr);
 
-      uip_clear_buf();
-      return;
-    }
+  if(nbr) {
+    send_queued(nbr);
   }
-  /* Multicast IP destination address. */
-  tcpip_output(NULL);
```
simonduq: why is this not needed anymore?

simonduq: ok, got it, it's in `send_packet`

core/net/ip/tcpip.c
```diff
@@ -526,222 +522,268 @@ tcpip_input(void)
...
-         DELAY, or PROBE). See RFC 4861, section 7.3.3 on node behavior. */
-      if(nbr->state == NBR_STALE) {
-        nbr->state = NBR_DELAY;
-        stimer_set(&nbr->reachable, UIP_ND6_DELAY_FIRST_PROBE_TIME);
-        nbr->nscount = 0;
-        PRINTF("tcpip_ipv6_output: nbr cache entry stale moving to delay\n");
-      }
+  if(nbr->state == NBR_INCOMPLETE) {
+    PRINTF("tcpip_ipv6_output: nbr cache entry incomplete\n");
+    queue_packet(nbr);
```
simonduq: same here, I'd make this conditional (`UIP_CONF_IPV6_QUEUE_PKT`)

core/net/ip/tcpip.c
```diff
@@ -526,222 +522,268 @@ tcpip_input(void)
...
-      uip_len = 0;
-      return;  
-#endif /* UIP_ND6_SEND_NA */
+  nbr = uip_ds6_nbr_lookup(nexthop);
+  if(nbr == NULL) {
+    if(send_nd6_ns(nexthop)) {
+      PRINTF("tcpip_ipv6_output: failed to add neighbor to cache\n");
+      goto exit;
     } else {
+      goto send_packet;
```
simonduq: Can we add a comment above clarifying that what we send here is NS (not the original data packet)

core/net/ip/tcpip.c
```diff
@@ -526,222 +522,268 @@ tcpip_input(void)
...
-        /* Send the first NS try from here (multicast destination IP address). */
-      }
-#else /* UIP_ND6_SEND_NA */
-      PRINTF("tcpip_ipv6_output: neighbor not in cache\n");
-      uip_len = 0;
-      return;  
-#endif /* UIP_ND6_SEND_NA */
+  nbr = uip_ds6_nbr_lookup(nexthop);
+  if(nbr == NULL) {
+    if(send_nd6_ns(nexthop)) {
```
simonduq: To clarify, I would call this only if `UIP_ND6_SEND_NA` is set (rather than make the body of the function completely conditional)

simonduq: This way you might even be able to refactor this with the `#if UIP_ND6_SEND_NA` at line 755

simonduq:
Very very nice. I reviewed this and left a few comments. Happy to :+1: after they are answered/addressed

bkozak-scanimetrics:
Hello @simonduq,

Regarding your comments of the form:
> I would prefer having this block conditioned by [SOME MACRO] rather than the entire the body of [SOME STATIC FUNCTION] 

For the most part it isn't possible to have the condition in [SOME MACRO] _instead_ of [SOME STATIC FUNCTION], you would actually need to have the conditional both at the function call site and around the entire function (i.e. not just the body but also the signature and enclosing block) .

This is because:
1. The compiler will (probably) emit unused static function warnings when the preprocessor makes the function call(s) disappear.
2. Some of the functions, variables and/or struct members used in [SOME STATIC FUNCTION] may not be available when [SOME MACRO] is false (or undefined).

There are several other ways to work around number one (like adding compiler specific pragmas for instance), but with number two we still need some preprocessor conditionals inside of [SOME STATIC FUNCTION].

Personally I think it is kind of ugly to add redundant preprocessor conditionals (and also to wrap entire functions in preprocessor conditionals), but if you are ok with this approach then I'll go ahead and do it.

Regarding the rest of your comments, I think that they make good sense and I'll implement your suggestions.

simonduq:
Great case for your current approach, I agree. True that what I was suggesting isn't ideal for static functions.

bkozak-scanimetrics:
I've now added in the remainder of @simonduq's suggestions.

core/net/ip/tcpip.c
```diff
@@ -230,8 +243,7 @@ tcp_connect(const uip_ipaddr_t *ripaddr, uint16_t port, void *appstate)
     return NULL;
   }
 
-  c->appstate.p = PROCESS_CURRENT();
-  c->appstate.state = appstate;
+  setup_appstate(&c->appstate, init_appstate);
```
simonduq: oops sorry for being unclear, the naming suggestion `init_appstate` was for the function name (replaces `setup_appstate`), not for the parameter.

simonduq:
Sorry for the unclarify of my comment on `init_appstate`. Can you please take a look at my latest comment?

bkozak-scanimetrics:
> oops sorry for being unclear, the naming suggestion init_appstate was for the function name (replaces setup_appstate), not for the parameter.

That does make more sense than what I did.

The function name is now changed.

simonduq:
:+1:

arurke:
@bkozak-scanimetrics If you want to resolve the conflicts I can test this on my nodes. Kind of interested in the functionality of #1186, which depends on this one - so will contribute what I can to get this merged

PR #1726, opened 2016-06-14
zoul: Add support for flashing to multiple devices

arthurfabre:
This adds support for flashing to multiple devices, by implementing the
$MOTES option used by the Z1.
This replaces the $PORT interface, as it only allowed flashing to one
device at a time, and the $MOTES interface was still used for other
tasks (such as getting a list of connected devices).

A foreach loop is used to dynamically create an upload rule per device
that needs to be flashed.
The "main" upload rule depends on the device specific ones.
This allows for parallel flashing.

### DIFF
--- a/platform/zoul/Makefile.zoul
+++ b/platform/zoul/Makefile.zoul
@@ -12,10 +12,6 @@ endif
 PYTHON = python
 BSL_FLAGS += -e -w -v
 
-ifdef PORT
-  BSL_FLAGS += -p $(PORT)
-endif
-
 ### Configure the build for the board and pull in board-specific sources
 CONTIKI_TARGET_DIRS += . dev
 CONTIKI_TARGET_DIRS += . $(BOARD)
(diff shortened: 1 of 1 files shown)


arthurfabre:
Just found a bug with the `BSL_ADDRESS` generation, I'll push a fix when I've got it sorted.

arthurfabre:
Should be fixed now - there was some weird behavior with the `eval` used to create the `BSL_ADDRESS_ARG`, I've turned it into a shell variable now.

platform/zoul/Makefile.zoul
```diff
@@ -69,33 +65,40 @@ else
...
+	@# Dummy recipe to prevent "No rule to make *.upload errors"
 endif
 
+### Variable that exands into a pattern rule to upload to a given MOTE.
+### Requires $(MOTE) to be defined
+### $$$$ Double escapes $s that need to be passed to the shell - once for when make parses UPLOAD_RULE, and once for when the expanded rule is parsed by make.
+define UPLOAD_RULE
+%.$(MOTE): %.bin %.elf
+	@echo "Flashing $(MOTE)"
+	BSL_ADDRESS=`$(OBJDUMP) -h $$*.elf | grep -B1 LOAD | \
```
alignan: Add an `@` to avoid printing the command

platform/zoul/Makefile.zoul
```diff
@@ -69,33 +65,40 @@ else
...
-	$(eval BSL_ADDRESS_ARG := -a $(shell $(OBJDUMP) -h $*.elf | grep -B1 LOAD | \
-	                               grep -Ev 'LOAD|\-\-' | awk '{print "0x" $$5}' | \
-	                               sort -g | head -1))
-	$(PYTHON) $(BSL) $(BSL_FLAGS) $(BSL_ADDRESS_ARG) $<
+### Upload to every MOTE
+%.upload: $(foreach MOTE,$(MOTES),%.$(MOTE))
+	@# Dummy recipe to prevent "No rule to make *.upload errors"
 endif
 
+### Variable that exands into a pattern rule to upload to a given MOTE.
```
alignan: `expands`

platform/zoul/Makefile.zoul
```diff
@@ -69,33 +65,40 @@ else
             perl -ne 'print $$1 . " " if(m-(/dev/\w+)-);')
 endif
```
alignan:
````
ifdef PORT		
  MOTES = $(PORT)
endif
````

alignan:
I've tested and works great! :smile:

Works better if updating the `cc2538-bsl`, as the extra sorting operations adds a small delay and sometimes it overran the on-board PIC timing to put the CC2538 ready to be flashed.

I have mixed feelings as I prefer to keep the `PORT` argument, so when specified it just flashes one mote, and when `PORT` is not defined then use `MOTES`.  I know defining `MOTES=/dev/ttyUSBx` when compiling does the same effect, but this is a concern mostly based on not breaking the existing documentation and scripts relying on `PORT`, also commonly used for the cc13xx/cc26xx platforms.

Please add:
````
### If PORT is defined, override to keep backward compatibility
ifdef PORT		
  MOTES = $(PORT)
endif
````

Before this line:
````
### Check the BSL script exists
````

Thanks,, great job!

arthurfabre:
Updated with all the changes you mentioned :)

Also changed the `MOTES`, `MOTELIST`, and `SERIALDUMP` variables to not be recursively expanded (so make only runs the `$(shell)` command once).

alignan:
Sweet, thanks!

alignan:
:+1:

PR #1727, opened 2016-06-14
Add CC13xx/CC26xx TRNG Support

### LINKED ISSUE #1238: CC26xx static initial DIS interval, synchronization in larger networks
I might be stating the obvious, but I'll post it anyways as reference/backlog - also, people might experience some issues related to it. 
I noticed I was always getting the same initial "random" DIS interval on my CC26xx sensortag each boot, this is caused by the seed being hardcoded in main(). This can cause synchronization of the DIS messages and problems in larger networks, ref 2a2175a2149e0a11af0fb5dfb514a3e5d182f46d. 

I see there is some true RNG implementation in the driverlib, but I am not confident enough on it to create a pull request atm.

Regards,
Andreas

g-oikonomou:
This pull request adds support for the CC13xx/CC26xx hardware RNG.

The hardware has 24 free-running oscillators (FROs) and generates random numbers by sampling them. The longer we let the hardware sample the FROs, the more entropy we will get in our random number. Thus, the hardware can generate low-entropy random numbers super quickly, or it can generate high-entropy random numbers, but this takes longer.

In this context, the driver can work in two modes:

* Synchronous mode: We issue a request to the hardware to generate a new, minimum-entropy random number and we block till it's ready. This function is thread-safe and can be called from inside an interrupt context. It is recommended to use this mode when a low-entropy number is required and therefore generation can be performed quickly. 
* Asynchronous: We issue a request to the RNG to generate a number with configurable degree of entropy and we return immediately. Once the number is ready, the caller will be notified through a callback. This mode is recommended for random numbers of higher entropy. This call should not be made from within an interrupt context. While generation is in progress, the driver will allow CM3 to sleep, but it will prohibit Deep Sleep.

Only one asynchronous generation can be in progress at any given point in time. If a request is made while a previous one is still in progress, the new request will be rejected.

Before an asynchronous generation starts, the driver will cache a small number of randoms. If a synchronous request is made while an asynchronous generation is in progress, the synchronous call will return one of those cached numbers. The number of randoms to be cached is configurable through `SOC_TRNG_CONF_CACHE_LEN`.

Asynchronous mode accepts the `samples` argument, which basically controls the amount of time the RNG will spend sampling the FROs. Minimum entropy can be achieved when `samples` is `SOC_TRNG_REFILL_CYCLES_MIN` and this will take about 1.5usecs to generate a number. Increasing the value of `samples` will increase entropy, but it will also increase latency. Maximum entropy is achieved when the value of `samples` is 0 (`SOC_TRNG_REFILL_CYCLES_MAX`), in which case generation takes approximately 350msec.

The intended usage is for synchronous mode to be used when we need a low-entropy number very quickly. These numbers are suitable for retransmission back-offs etc, but not suitable for crypto operations. When a high-entropy number is required, it is recommended to use asynchronous mode.

With all of this in place, this pull also implements the random API. `random_rand()` will use synchronous mode and will return the 2 LS bytes of the generated number. 

Closes #1238

### DIFF
--- a/cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
+++ b/cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
@@ -73,6 +73,7 @@ CONTIKI_CPU_SOURCEFILES += putchar.c ieee-addr.c batmon-sensor.c adc-sensor.c
 CONTIKI_CPU_SOURCEFILES += slip-arch.c slip.c cc26xx-uart.c lpm.c
 CONTIKI_CPU_SOURCEFILES += gpio-interrupt.c oscillators.c
 CONTIKI_CPU_SOURCEFILES += rf-core.c rf-ble.c ieee-mode.c
+CONTIKI_CPU_SOURCEFILES += random.c soc-trng.c
 
 DEBUG_IO_SOURCEFILES += dbg-printf.c dbg-snprintf.c dbg-sprintf.c strformat.c
 
--- a/cpu/cc26xx-cc13xx/dev/soc-trng.c
+++ b/cpu/cc26xx-cc13xx/dev/soc-trng.c
@@ -1,0 +1,315 @@
+/*
+ * Copyright (c) 2016, University of Bristol - http://www.bristol.ac.uk
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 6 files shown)


atiselsts:
:+1: 

I like that after reboot `random_rand()` does not return the same sequence of numbers anymore.

PR #1728, opened 2016-06-16
CoAP OTA

msolters:
Rewrote the OTA mechanism with a few major changes:

*  OTA uses CoAP instead of HTTP
*  OTA download buffer only has a 1Kb SRAM footprint (instead of 4Kb)
*  OTA feature is repackaged as a Contiki app
*  Can now write firmware to external flash in units smaller than a full flash page


msolters:
Sorry meant to merge to my fork.

PR #1729, opened 2016-06-17
CC2538: Fix external flash support

arthurfabre:
External flash is supported through the `COFFEE_CONF_CUSTOM_PORT` macro,
which when used disables the program flash based Coffee.
The linker script directly uses `COFFEE_START` and `COFFEE_SIZE` to reserve
space for the program flash based Coffee.

Consequently, space is reserved in program flash even for external flash
chips, and things break when the linker script attempts to reserve
several MB of space.

This change adds a new macro, `CC2538_FLASH_OFFSET`, that is used by the
linker script.
If the program flash based Coffee is used, it is set to `COFFEE_SIZE`,
else set to 0.


bthebaudeau:
Thanks. That was on my TODO list, but for different reasons (OTA FW upgrade). There may be other things to take into consideration. I'll think about that and I'll come back to you, probably this weekend.

bthebaudeau:
I see the following issues with these changes:
 - They break the current behavior if `COFFEE_CONF_START` is used to place Coffee after the beginning of the flash memory.
 - The introduced definitions should not be in `cfs-coffee-arch.h`, because they are unrelated, but rather in `flash.h`.
 - The naming must reflect the fact that it's not a general flash definition, but one related to firmware placement.
 - We need to be able to easily choose alternative firmware locations in flash memory, for OTA or whatever.

So how about the following commit?
https://github.com/bthebaudeau/contiki/commit/8a273a248e33cb57a16650aeb8ea4b911a3e5809

arthurfabre:
That works for our use-case and seems a bit neater :)

alignan:
Just to keep this on the list :smile_cat:

PR #1732, opened 2016-06-20
cc2538: Add configuration for firmware location

### LINKED ISSUE #1731: [CC2538] External flash support broken
> External flash is supported through the COFFEE_CONF_CUSTOM_PORT macro,
> which when used disables the program flash based Coffee.
> The linker script directly uses COFFEE_START and COFFEE_SIZE to reserve
> space for the program flash based Coffee.

> Consequently, space is reserved in program flash even for external flash
> chips, and things break when the linker script attempts to reserve
> several MB of space.

Comes from #1729

bthebaudeau:
Introduce `FLASH_CONF_FW_ADDR` and `FLASH_CONF_FW_SIZE` in order to make it possible to place the firmware anywhere, regardless of Coffee, and without having to write a custom linker script. Also, handle the default values properly in order to fix the link breakage reported by @arthurfabre with `COFFEE_CONF_CUSTOM_PORT`.

Closes #1731.


alignan:
Looks sweet, I will test locally and update.  Please @arthurfabre could you test as well?

arthurfabre:
I merged this in to our branch this morning, and the generated linker script / binaries are as expected.

alignan:
:+1:

PR #1737, opened 2016-06-23
cc2538: aes: Fix possibly missing result available status

bthebaudeau:
Depending on the use case and on the timings, `auth_crypt_check_status()` sometimes never reported an available result, leading to a deadlock of any protothread waiting for this event, and to a `WDT` reset if a protothread was polling it.

This was caused by `aes_auth_crypt_start()` clearing the result available interrupt after operations that may rightfully trigger it, leading to a missed interrupt.

This PR also fixes build errors discovered in `noncoresec` with `DEBUG` set to `1` while investigating this issue. Even if `noncoresec` is removed soon, it's better to fix this in the meantime.


kkrentz:
:+1:

alignan:
:+1:

PR #1738, opened 2016-06-24
mac-seqno: add a lifetime to sequence numbers in history

simonduq:
Without a lifetime, there are cases where node B receives from A with seqno N, then does not hear anything for an arbitrary period of time, and then hears a new, different packet with seqno N again (seqnos wrap quick given they are a single byte).

This PR adds a configurable lifetime for received seqnos in history.


nvt:
:+1:

PR #1739, opened 2016-06-24
Improved TSCH sequence number handling

simonduq:
This PR does two things, in TSCH:
* move from ad-hoc seqno handling to the generic and standard-compliant mac-sequence module
* use sequence numbers only for unicast packets. We suppress it in broadcast. This has two benefits, (1) slows down the pace at which seqnos wrap and (2) save one byte in broadcast frames.

Together with https://github.com/contiki-os/contiki/pull/1738 this avoids spurious duplicate drops.


nvt:
:+1:

PR #1743, opened 2016-06-24
Updated cc2538-bsl to version 2.1 (c6100a7)

alignan:
Bring in the latest changes in `v2.1`


alignan:
@JelmerT @g-oikonomou are you OK with updating?

PR #1745, opened 2016-06-24
Minor changes to z1/contiki-conf.h

alignan:
Minor changes:

* Moved `project-conf.h` inclusion to the top of `contiki-conf.h`, as configurable parameters are defined beforehand, thus requiring to `#undef` first.
* Made common parameters overridable
* Added `BOARD_STRING` platform definition

### DIFF
--- a/platform/z1/contiki-conf.h
+++ b/platform/z1/contiki-conf.h
@@ -105,7 +105,9 @@
 #define CC2420_CONF_CCA_THRESH      -45
 #endif /* CC2420_CONF_CCA_THRESH */
 
+#ifndef IEEE802154_CONF_PANID
 #define IEEE802154_CONF_PANID       0xABCD
+#endif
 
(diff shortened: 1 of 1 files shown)


alignan:
:+1:

simonduq:
I think this might cause problems with all `project-conf.h` files that assume they are included after this `contiki-conf.h`, as the latter sometimes sets some flags without `ifndef`. I feel we need to agree on a config architecture and stick to it in all examples and platforms, else this soon getting messy.

alignan:
Agree, I changed this to follow how the `RE-Mote` and newer platforms like the `srf06-cc26xx`, as I think it is better implemented (it is ugly to #undef a `CONF_XXX` value from a `project-conf.h`).  

At the moment at least it should be easy to spot as it will yield a lot of warnings :smile:

simonduq:
Yes true, at least easy to spot :)

PR #1748, opened 2016-06-27
CC26xx: Make sure UART buffers are flushed before sleeping

atiselsts:
This fixes a problem with corrupted characters occasionally appearing in UART output.

### DIFF
--- a/cpu/cc26xx-cc13xx/lpm.c
+++ b/cpu/cc26xx-cc13xx/lpm.c
@@ -51,6 +51,7 @@
 #include "dev/watchdog.h"
 #include "dev/soc-rtc.h"
 #include "dev/oscillators.h"
+#include "cc26xx-uart.h"
 
 #include <stdint.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


g-oikonomou:
And now I can see why you were getting garbage chars...

How about clocking the UART during sleep by changing the driver instead of adding peripheral-specific awareness to the LPM module? I'm just thinking this may open a can of worms, now we do this for the UART, next we may do it for SPI or GPTs and the story goes.

What do you think?

atiselsts:
I see your point. Adding this line in `cc26xx_uart_init()` does seem to be an alternative way to solve this problem:
`ti_lib_prcm_peripheral_deep_sleep_enable(PRCM_PERIPH_UART0);`

I measured energy consumption with a multimeter after this change, did not detect any difference in either deep sleep or sleep mode.

There are some more calls to `ti_lib_prcm_peripheral_sleep_enable() / ti_lib_prcm_peripheral_sleep_disable()` in the driver which now can be removed. But not sure about this line, should keep it or remove it?:
https://github.com/contiki-os/contiki/blob/master/cpu/cc26xx-cc13xx/dev/cc26xx-uart.c#L234

g-oikonomou:
After some discussion with Atis, this needs a little more work. Closing for now as this is a "don't merge" at the moment.

PR #1749, opened 2016-06-27
Add checksum to SLIP packets when SLIP_CRC_ON is defined

atiselsts:
Helps to avoid invalid packets when the underlying serial connection is not reliable. Unlike the IP checksum, this covers not only the header, but also the body of a packet.

This feature is disabled by default and the patch add Contiki side code only. Modifications in Contiki's `tunslip` code are necessary to enable and use this feature. We have tested it only using a custom Python script instead of `tunslip` for the host side.


g-oikonomou:
This sounds like a good idea. 👍

alignan:
I like the idea, but it would be even better if fixing `tunslip` as well, or providing a new script that makes actual use of the checksum check, else it would be just unused.  Any chance to update this PR?

simonduq:
Out of curiosity, is there a standard way of doing this? (maybe this is?)
Have you tried H/W flow control or S/W flow control? Very reliable in my experience (and supported by existing tools, even already in tunslip6)

atiselsts:
Updated the patch after #920 was merged, and added the change for `tunslip6` as well.
Not sure if flow control can replace this.

neophob:
any plans to merge this?

atiselsts:
For our project I have updated this to a 16 bit CRC, let me know if you're interested.

PR #1750, opened 2016-06-27
Improve RPL extension headers handling

yatch:
This PR addresses mainly two bugs in `rpl-ext-header.c`.

 1. The system crashes or falls into an infinite loop when it encounters an unsupported protocol header
(roughly speaking).
 2. A RPL header to be processed is not handled properly unless it comes just after an IPv6 header.

You can confirm the first one with a packet unsupported by Contiki. For example, on sending a simple UDP-Lite packet to a mote running on Cooja, Cooja crashes by SIGSEGV. When the mote runs with Non-Storing mode enabled, Cooja stalls because the mote does not release CPU.

For the second bug, I used a packet which had a Destination Options header followed by a RPL Routing header. The Destination Options header had only one PadN option. Contiki was not able to find the RPL Routing header.


simonduq:
:+1:

laurentderu:
Haven't tested it yet, but 👍 from me too

Issue #1736, opened 2016-06-22
CC26xx/CC13xx lpm wakeup scheduling issues


bkozak-scanimetrics:
As first reported in the discussion on #1549, then later on #1709 by @niklas88 there is an issue in lpm.c with scheduling the RTC interrupt to wake in time for the next etimer.

The issue has three aspects:
1. No interrupt is scheduled if no etimer is pending thus causing the MCU to wake on the next clock tick when it could have slept much longer.
2. No interrupt is scheduled if the next-to-expire etimer is set at a point more than one second into the future from the time of the lpm switch.
3. The MCU is limited to sleep for at most one second even when sleeping in deep sleep mode. Although we have to limit sleep time for ```LPM_MODE_SLEEP``` to prevent watchdog expiry, no such restriction exists for ```LPM_MODE_DEEP_SLEEP``` where the watchdog clock is off.

g-oikonomou:
#1751

PR #1751, opened 2016-06-28
CC26xx: fix a regression in and refactor LPM code

atiselsts:
This addresses all three points of issue #1736


bkozak-scanimetrics:
Looks to me like it'll work. I don't have a setup to reproduce #1736 so I haven't tested it.

I'm sure @niklas88 will be able to confirm that this fixes the problem though.

niklas88:
Indeed, it seems like the power regression is gone and the code works well in my setup. :+1:

PR #1752, opened 2016-06-28
Add CCxxware macros for cpu.h

g-oikonomou:
Allows us to do things such as enquire the current state of the global interrupt enable/disable flag


bthebaudeau:
How about switching all the Cortex-M3 platforms to the following?
https://github.com/contiki-os/contiki/blob/master/cpu/arm/common/CMSIS/core_cm3.h
https://github.com/contiki-os/contiki/blob/master/cpu/arm/common/CMSIS/core_cmFunc.h
https://github.com/contiki-os/contiki/blob/master/cpu/arm/common/CMSIS/core_cmInstr.h

g-oikonomou:
Apparently we had those already...

PR #1746, opened 2016-06-26
Disable CC2650_FAST_RADIO_STARTUP by default

g-oikonomou:
`CC2650_FAST_RADIO_STARTUP` is required for TSCH, but it currently breaks ContikiMAC. We set the default value to the one that will work with the default choice of RDC.

I'm also moving the define to a place in `contiki-conf.h` closer to where the user will look in order to configure the RDC driver.

Fixes #1744

PR #1754, opened 2016-06-28
CC26xx: change default TSCH guard time to 1.8ms

atiselsts:
This change enables to run TSCH on CC26xx without application-specific configuration settings (i.e. without setting `CC2650_FAST_RADIO_STARTUP` to 1).

It fixes two real problems:
1) The early wakeup due to `CC2650_FAST_RADIO_STARTUP` requires additional energy compared to the "normal" wakeup at the start of an active TSCH slot.
2) Since `CC2650_FAST_RADIO_STARTUP` is not the default setting, it means that TSCH applications have to be explicitly ported to CC26xx by adding an application-specific setting. This is potentially confusing to new users. It is also not a good design since this approach would not scale well (in terms of more platforms).

The change introduces slightly "nonstandard" TSCH guard times for this platform. Numerous experiments have shown that slightly reducing the TSCH guard time while keeping the other Contiki TSCH configuration settings unchanged (or even changed, but not extensively) does not have bad effects on interoperability and packet delivery rate, but instead helps to save energy. For example, the 6tisch RFC 7554 talks about even smaller 1ms guard time: https://datatracker.ietf.org/doc/rfc7554/?include_text=1

### DIFF
--- a/platform/srf06-cc26xx/contiki-conf.h
+++ b/platform/srf06-cc26xx/contiki-conf.h
@@ -66,8 +66,8 @@
 #endif
 
 /*
- * Disable turning off HF oscillator when the radio is off:
- * You need to set this in order to use TSCH, disable to save more energy.
+ * If set, the systems keeps the HF crystal oscillator on even when the radio is off.
+ * You need to set this to 1 to use TSCH with its default 2.2ms or larger guard time.
  */
 #ifndef CC2650_FAST_RADIO_STARTUP
 #define CC2650_FAST_RADIO_STARTUP               0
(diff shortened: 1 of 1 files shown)


simonduq:
I agree in general that 2.2 ms is long, and I've worked with e.g 0.6 ms guards with no problems, but why introduce a change for this specific platform?
BTW in RFC 7554 my understanding is +/- 1ms, that is, a 2 ms guard (that's what I get from the calculations that follow the statement -- I can be wrong)

atiselsts:
@simonduq the other platforms* do not have such a long system / radio wakeup time, so a long guard time is not a problem for them.
The fact is that since #1746 has been merged, the rpl-tsch example is broken for CC26xx platform. We need to fix it somehow, and I'm not happy to have CC26XX specific config in the example for the reasons I explained.
Good point about the RFC, I'm really not sure.

[*] - Except JN516x, which also does these "early wakeups" for the same reason, and where guard time reduction could be an alternative solution - but would require some tinkering with the platform's code.

simonduq:
Right, I see!
:+1:

g-oikonomou:
@simonduq @atiselsts where are we on this one guys? Happy to merge?

simonduq:
fine with me!

Issue #1744, opened 2016-06-24
Example cc26xx-demo Continually Reboots with Latest Contiki


davidjnix:
The cc26xx-demo on the SmartRF06 Eval Board reboots every 12 seconds or so. I am using the latest pull (f0b19a1). There was a similar report (#1465) from January 2016 that was resolved at that time. I verified that the fix for #1465 is in my repo. The serial output below demonstrates the reboot.

Starting Contiki-3.x-2647-gf0b19a1
With DriverLib v0.44336
TI SmartRF06EB + CC26xx EM
 Net: sicslowpan
 MAC: CSMA
 RDC: ContikiMAC, Channel Check Interval: 16 ticks
 RF: Channel 25
 Node ID: 34057
CC26XX demo
-----------------------------------------
Bat: Temp=25 C
Bat: Volt=3292 mV
ALS: 687 raw
Starting Contiki-3.x-2647-gf0b19a1
With DriverLib v0.44336
TI SmartRF06EB + CC26xx EM
 Net: sicslowpan
 MAC: CSMA
 RDC: ContikiMAC, Channel Check Interval: 16 ticks
 RF: Channel 25
 Node ID: 34057
CC26XX demo
-----------------------------------------
Bat: Temp=25 C
Bat: Volt=3292 mV
ALS: 685 raw

g-oikonomou:
What toolchain version are you using?

g-oikonomou:
I can confirm this. I suspect it's something to do with #1709

g-oikonomou:
Can you try with `#define CC2650_FAST_RADIO_STARTUP 0`?

You can change the respective line in `contiki-conf.h` or set it in `project-conf.h`

davidjnix:
George,


Will do later today. Thanks for the quick response.

David


________________________________

davidjnix:
That fixed the issue. Verified using (f0b19a1) with SmartRF06 Eval Board and toolchain gcc-arm-none-eabi-5_2-2015q4.

PR #1755, opened 2016-06-28
CC26xx: always request HF crystal oscillator at radio on()

atiselsts:
This fixes a problem that appears when CC2650_FAST_RADIO_STARTUP is set to 1: radio `on()` could occasionally be called without requesting HF crystal oscillator before trying to use it, leading to reboots.

This change unconditionally requested the HF crystal oscillator in `on()`, which is very fast in case the oscillator already has been requested before and is ready.

### DIFF
--- a/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
+++ b/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
@@ -1222,13 +1222,11 @@ on(void)
     return RF_CORE_CMD_OK;
   }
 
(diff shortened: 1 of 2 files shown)


g-oikonomou:
This is basically a proper fix for #1744 in a sense that `CC2650_FAST_RADIO_STARTUP` set to 1 will not break ContikiMAC. We still leave it as 0 for ContikiMAC builds, since it's not required there, but this pull does improve the logic of things.

Extensively discussed with Atis earlier, so I'll merge this when Travis goes green.

Thanks 👍

PR #1757, opened 2016-06-29
RPL border HTTP server: support for non-storing

simonduq:
With this PR, the BR will include the list of (RPL non-storing) links in its HTTP response.


nvt:
:+1:

PR #1759, opened 2016-06-30
Implementation of optimized neighbor discovery based on RFC 6775

MohamedSeliem:
IPv6 Neighbor Discovery (ND) based on RFC4861 was developed primarily for wired traffic on the shared medium. Its heavily use the multicast transmission in periodic router-advertisement multicast addresses with router solicitation, neighbor solicitation, address resolution, and duplicate-address detection. While periodic multicast signaling and solicited-node multicast signaling are useful for network stability in the standard Ethernet-based shared network, the limited-lifetime, battery-operated devices in the IEEE 802.15.4 network conserve energy with less signaling and by sending broadcast messages only once in a while. that makes makes it inefficient and sometimes impractical for IPv6 over Low power Wireless Personal Area Networks (6LoWPAN). 
you could find all details in the following paper http://eece.cu.edu.eg/~akhattab/files/ND.pdf.


alignan:
You had some unwanted commits in your PR, please fix and create a new PR anytime, thanks!

PR #1763, opened 2016-07-02
Bug fix for uninitialized variable

MohamedSeliem:
change is made to busy_wait_time to be static.
this will solve the problem of
../../../core/net/mac/tsch/tsch-slot-operation.c: In function ‘tsch_get_lock’:
../../../core/net/mac/tsch/tsch-slot-operation.c:192:20: error: ‘busy_wait_time’ may be used uninitialized in this function
make[1]: *** [obj_z1/tsch-slot-operation.o] Error 1

### DIFF
--- a/core/net/mac/tsch/tsch-slot-operation.c
+++ b/core/net/mac/tsch/tsch-slot-operation.c
@@ -189,7 +189,7 @@ int
 tsch_get_lock(void)
 {
   if(!tsch_locked) {
-    static rtimer_clock_t busy_wait_time;
+    rtimer_clock_t busy_wait_time=0;
     
     int busy_wait = 0; /* Flag used for logging purposes */
     /* Make sure no new slot operation will start */
--- a/core/net/mac/tsch/tsch-slot-operation.c
+++ b/core/net/mac/tsch/tsch-slot-operation.c
@@ -189,7 +189,7 @@ int
 tsch_get_lock(void)
 {
   if(!tsch_locked) {
-    rtimer_clock_t busy_wait_time;
+    static rtimer_clock_t busy_wait_time;
     int busy_wait = 0; /* Flag used for logging purposes */
     /* Make sure no new slot operation will start */
     tsch_lock_requested = 1;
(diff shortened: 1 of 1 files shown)


alignan:
What test did you run that caused this error? I'm wondering about why it was not shown in the regression tests.

MohamedSeliem:
this error appears when i run (01-compile-base) regression test locally. specifically, the error appears in (ipv6/rpl-tsch/z1 , ipv6/rpl-tsch/z1:MAKE_WITH_ORCHESTRA=1) fail logs.

I think  it was not detected by travis check as it appears as warning. however in local regression tests:
cc1: warnings being treated as errors.

In general, Uninitialized variables make a program non-deterministic. Each time the program runs, it may behave differently. Unrelated changes to operating environment, time of day, phase of the moon and permutations of such affect how and when these daemons manifest. The program may run a million times before the defect presents, them may do it every time, or run another million. 

So Initializing a variable can prevent bugs. especially, If the variable is a reference type, initializing it can prevent null reference errors down the line.

simonduq:
Hello,

* I would expect a modern compiler to find out that the variable is never used unless initialized. It is used only if `busy_wait != 0`, which happens only `busy_wait_time` is also set. What compiler version are you using? mspgcc 4.7.2 recommended.
* Travis compiles with -Werror enabled so it would notice such a warning.
* I wouldn't be happy with the fix anyway, as the variable has no reason to be static. If initialization were needed, I'd simply set it to 0.

Thanks

MohamedSeliem:
I use gcc-4.7 (Ubuntu/Linaro 4.7.2-22ubuntu4) 4.7.2.
i do not know about travis, but i think initialization were needed. primarily, i though about set it to 0. then, i thought if the variable is defined to be static, it will by automatically set to 0 at initialization, besides it will  keeps its value between invocations.

i agree with you, it'd be simply set to 0.

Thanks

g-oikonomou:
@simonduq any advances here?

simonduq:
I don't see the point of the PR as there is no warning on an up-to-date msp430-gcc.

PR #1767, opened 2016-07-05
RPL: added missing return, to not join instances with unsupported MOP

simonduq:


alignan:
:+1: once Travis is happy

nvt:
:+1:

PR #1768, opened 2016-07-05
JN516x: Exception handler causing infinite loop

mguc:
This issue was reported some time ago ([jn516x platform: Software Reset raising exception](https://sourceforge.net/p/contiki/mailman/message/34612894/)) and I ran into it. By removing the exception handler for SYSCALLs the JN516x restarts as expected.

### DIFF
--- a/platform/jn516x/dev/exceptions.c
+++ b/platform/jn516x/dev/exceptions.c
@@ -182,7 +182,6 @@ vEXC_Register(void)
   BUS_ERROR = (uint32)exception_handler;
   UNALIGNED_ACCESS = (uint32)exception_handler;
   ILLEGAL_INSTRUCTION = (uint32)exception_handler;
-  SYSCALL = (uint32)exception_handler;
   TRAP = (uint32)exception_handler;
   GENERIC = (uint32)exception_handler;
   STACK_OVERFLOW = (uint32)exception_handler;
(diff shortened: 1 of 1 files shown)


simonduq:
Nice catch.
As an alternative fix, hat would you think of keeping the handler, but in `exception_handler`, in case of `E_EXC_SYSCALL`, return at the end of the function? (right before `#if EXCEPTION_STALLS_SYSTEM`)

simonduq:
I don't know more than what you found :)

platform/jn516x/dev/exceptions.c
```diff
@@ -354,6 +354,9 @@ exception_handler(uint32 *pu32Stack, eExceptionType eType)
   }
 #endif
 
+  if(eType == E_EXC_SYSCALL)
```
simonduq: code style: please add `{` and `}`

g-oikonomou:
This needs a code style fix @mguc but otherwise looks like a no-brainer (@simonduq ?)

simonduq:
Right @mguc , the `if` block is still not compliant with Contiki code style. Space needed between `(` and `{`

simonduq:
@mguc ready to merge as soon as you fix the code style!

simonduq:
Thanks! :+1:

PR #1769, opened 2016-07-06
rest-engine: Allow adding periodic resources after starting REST engine

gebart:
Without this fix, the timers for periodic resources added after the REST engine was already running were never started. As a result, any observe requests over CoAP were never updated after the initial GET.

### DIFF
--- a/apps/rest-engine/rest-engine.c
+++ b/apps/rest-engine/rest-engine.c
@@ -114,6 +114,7 @@ rest_activate_resource(resource_t *resource, char *path)
            resource->periodic->resource->url);
     list_add(restful_periodic_services, resource->periodic);
   }
+  process_poll(&rest_engine_process);
 }
 /*---------------------------------------------------------------------------*/
 /*- Internal API ------------------------------------------------------------*/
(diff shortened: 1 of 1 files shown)


apps/rest-engine/rest-engine.c
```diff
@@ -188,26 +189,39 @@ PROCESS_THREAD(rest_engine_process, ev, data)
...
     PROCESS_WAIT_EVENT();
 
-    if(ev == PROCESS_EVENT_TIMER) {
+    if(ev == PROCESS_EVENT_POLL) {
+      PRINTF("Periodic: poll\n");
+      for(periodic_resource =
+            (periodic_resource_t *)list_head(restful_periodic_services);
+          periodic_resource; periodic_resource = periodic_resource->next) {
+        if(periodic_resource->periodic_handler && periodic_resource->period) {
+          if(!etimer_expired(&periodic_resource->periodic_timer)) {
```
greg-king5: That test causes "race" problems:  If an existing, running timer expires immediately after a new periodic resource is added, then this loop will think that it is a new timer; it will be reset.  Then, when its event is handled, it won't be "expired" anymore -- its resource handler won't be called!
That `if` statement should check for new, never-set timers, not expired timers.

But, I wonder:  wouldn't it be better if `res_activate_resource()` starts the timer of the periodic resource that it just has added to the list?

gebart: Adding the resource from (and thus setting the timer from) a different protothread than the `rest_engine_process` will bind the timer to that thread instead of the `rest_engine_process`.
I don't know of a way to differentiate between new, never used, etimers and old, expired, etimers. Suggestions?

greg-king5: There are two macroes that change the process context, temporarily.  Read starting at https://github.com/contiki-os/contiki/blob/master/core/sys/process.h#L406

gebart: Thanks for the hint! I didn't know that was possible. I will update the PR accordingly.

gebart:
@greg-king5 I updated the PR to set the timer from `rest_activate_resource` instead of polling the process.

greg-king5:
:+1:

alignan:
:+1: 

Merging in a week if there is no objection, thanks for the patience!

alignan:
And a week has passed, happy to merge

PR #1771, opened 2016-07-07
Modify obsolete  macro related to neighbor cache table size

MohamedSeliem:
UIP_CONF_DS6_NBR_NBU, the number of elements in neighbor cache requested by the user in contiki configuration, is obsolete.
NBR_TABLE_CONF_MAX_NEIGHBOR is the latest Macro which is related to neighbor cache table


MohamedSeliem:
This simple fix, set the maximum number of neighbor entries  in the neighbor table as defined in the project configuration file for (ipv6/ multicast example)
and without this fix, the defined value of the neighbor entries  in the neighbor table will be set to the default value as defined in (ipv6/ nbr-table.h)

MohamedSeliem:
@g-oikonomou  can i got your feed back on this.

g-oikonomou:
You are right, those are the old defines. You probably want to be using
`NBR_TABLE_CONF_MAX_NEIGHBORS` and `UIP_CONF_MAX_ROUTES`

MohamedSeliem:
great.
I do not think we have a problem in case of (UIP_CONF_DS6_ROUTE_NBU), as routing table defined it

/* Routing table */
#ifndef UIP_CONF_MAX_ROUTES
#ifdef UIP_CONF_DS6_ROUTE_NBU
#define UIP_DS6_ROUTE_NB UIP_CONF_DS6_ROUTE_NBU
#else /* UIP_CONF_DS6_ROUTE_NBU */
#define UIP_DS6_ROUTE_NB 4
#endif /* UIP_CONF_DS6_ROUTE_NBU */
#else /* UIP_CONF_MAX_ROUTES */
#define UIP_DS6_ROUTE_NB UIP_CONF_MAX_ROUTES
#endif /* UIP_CONF_MAX_ROUTES */

even though, updating to recent macros is a good way to move forward.

P.s. I redefined the Macros of nbr table, to avoid such a problem (a hidden, or unwilling used (UIP_CONF_DS6_NBR_NBU) ).

MohamedSeliem:
@g-oikonomou any progress here?

simonduq:
I would rather fully retire the two old macros, by:
1) reverting the change to `nbr-table.h`
2) removing all definitions of `UIP_CONF_DS6_NBR_NBU` and `UIP_CONF_DS6_ROUTE_NBU` (not only in the multicast example)

simonduq:
now replaced with https://github.com/contiki-os/contiki/pull/1849

PR #1779, opened 2016-07-16
CC26xx: enable MCU sleeping while transmitting if and only if not in poll mode

atiselsts:
This pull request fixes the current, invalid logic in CC26xx IEEE mode radio driver, which messed up the global interrupt state.

### DIFF
--- a/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
+++ b/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
@@ -763,9 +763,9 @@ check_rat_overflow(bool first_time)
 {
   static uint32_t last_value;
   uint32_t current_value;
-  uint8_t interrupts_enabled;
+  uint8_t interrupts_disabled;
 
-  interrupts_enabled = ti_lib_int_master_disable();
+  interrupts_disabled = ti_lib_int_master_disable();
   if(first_time) {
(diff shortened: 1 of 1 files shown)


cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
```diff
@@ -765,7 +765,7 @@ check_rat_overflow(bool first_time)
   uint32_t current_value;
   uint8_t interrupts_enabled;
```
g-oikonomou: Variable needs renamed here.

cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
```diff
@@ -908,16 +908,8 @@ transmit(unsigned short transmit_len)
   cmd.startTime = 0;
   cmd.startTrigger.triggerType = TRIG_NOW;
 
-  /* XXX: there seems to be no function that gets interrupt state w/o changing it */
-  interrupts_enabled = ti_lib_int_master_disable();
```
g-oikonomou: `interrupts_enabled` is now an unused variable and generates a warning. Travis will fail. The definition needs removed.

g-oikonomou:
👍 I'll merge this when travis goes green.

PR #1781, opened 2016-07-16
TSCH: add RADIO_DELAY_BEFORE_DETECT to rx guard time when waiting for a packet to be received

atiselsts:
This essentially fixes the guard time for ACK packets in TSCH. The default guard time led to a large number of missed ACKs because the first call to `receiving_packet()` returns with 352 usec delay on CC26xx, and the unadjusted ACK guard time is just 400 usec.


simonduq:
:+1:

PR #1784, opened 2016-07-17
Add CC1350 support

g-oikonomou:
This pull adds support for the CC1350.

We only focus on CPU capability here.

Board support to follow in a separate pull request. We've got a little bit of work to do before we can properly support CC1350 boards, eg the brand new CC1350 LP (LAUNCHXL-CC1350). The main reason for this is that we need to make RF front-end / bias more configurable and we also need a clever way to support antenna switches, which is what the CC1350LP uses.

### DIFF
--- a/cpu/cc26xx-cc13xx/rf-core/rf-core.c
+++ b/cpu/cc26xx-cc13xx/rf-core/rf-core.c
@@ -97,6 +97,8 @@
 #define cc26xx_rf_cpe0_isr RFCCPE0IntHandler
 #define cc26xx_rf_cpe1_isr RFCCPE1IntHandler
 /*---------------------------------------------------------------------------*/
+typedef ChipType_t chip_type_t;
+/*---------------------------------------------------------------------------*/
 /* Remember the last Radio Op issued to the radio */
 static rfc_radioOp_t *last_radio_op = NULL;
 /*---------------------------------------------------------------------------*/
@@ -371,7 +373,7 @@ uint8_t
 rf_core_set_modesel()
 {
   uint8_t rv = RF_CORE_CMD_ERROR;
-  ChipType_t chip_type = ti_lib_chipinfo_get_chip_type();
+  chip_type_t chip_type = ti_lib_chipinfo_get_chip_type();
 
   if(chip_type == CHIP_TYPE_CC2650) {
     HWREG(PRCM_BASE + PRCM_O_RFCMODESEL) = PRCM_RFCMODESEL_CURR_MODE5;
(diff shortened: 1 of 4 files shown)


g-oikonomou:
Sits on top of #1756

alignan:
:+1:  from me, if no objections I'm merging this by the end of the week

tussock-cam:
Has there been any work on adding support for CC1350 launchpad? I need to get this going and don't want to duplicate effort! Do you have a mental plan of what needs to be done?

g-oikonomou:
I just rebased this one. We have done some (a lot of) work on the CC1350LP, yes. The antenna switch is one thing that's still not finalised.

tussock-cam:
I've started to have a look at this, too. Have you had a lot of luck with the prop-mode driver? I've been getting `ERROR_SYNTH_PROG` when trying to use 863.125 MHz. Works fine at 868 MHz. Haven't been able to figure it out yet, do you have any ideas?

Also a little confused about the modesel for CC1310/CC1350. Going off the values listed here  https://e2e.ti.com/support/wireless_connectivity/proprietary_2-4_ghz/f/964/t/51971 it seems that the CC1310 is in IEEE 802.15.4/BLE mode, and the CC1350 is Proprietary/BLE. Seems a little weird, but I can't find much official documentation for the configuration values. 

I've also noticed that `smartrf_settings_cmd_prop_radio_div_setup.formatConf.whitenMode` is set to 0x07 for the CC13xx chips, which SmartRF lists as unknown (it's the 802.15.4 whitener). 

I'm pretty new to the rf-core stuff, so sorry if I'm missing something..

g-oikonomou:
> I've been getting `ERROR_SYNTH_PROG` when trying to use 863.125 MHz

It's been a while since I played around with this, but I don't remember having a problem like this. Are you using a US launchpad? Those synth errors can be related to overrides.

> Also a little confused about the modesel 

Join the club :) the values we're using were provided by TI. The info you pasted from that thread makes it look as though modesel may be a bitmask.

We control the whitener from within the prop mode driver and it can be configured through `PROP_MODE_CONF_DW` (https://github.com/contiki-os/contiki/blob/master/cpu/cc26xx-cc13xx/rf-core/prop-mode.c#L98). So basically the value in smartrf-settings is not ultimately used.

alignan:
I'm ready to merge, would you prefer to keep the conversation here and wait until findings may be added, or do you prefer to deal with this later?

tussock-cam:
> Are you using a US launchpad? Those synth errors can be related to overrides.

Yeah, I am. I guess that makes sense, it seems to work from SmartRF studio, though. 

> Join the club :) the values we're using were provided by TI. The info you pasted from that thread makes it look as though modesel may be a bitmask.

Well that's confusing! I'll just treat them as magic and hope for the best :)

> We control the whitener from within the prop mode driver

Ah, I should've noticed that. Thanks!

> I'm ready to merge, would you prefer to keep the conversation here and wait until findings may be added, or do you prefer to deal with this later?

I think George has cleared most things up, so happy to merge. Thanks!

g-oikonomou:
@tussock-cam see early CC1350LP work in this branch https://github.com/g-oikonomou/contiki/tree/contrib/cc1350/lp-and-st

Feel free to report back!

PR #1786, opened 2016-07-17
Discuss: Provide a reason with calls to channel_clear()

g-oikonomou:
On a number of occasions I've found it desirable to know (within the RF driver) the reason why `channel_clear()` was called by the RDC layer.

This function is called in two contexts:
* For collision avoidance before TX
* For incoming frame detection

Knowing the reason could be used to change the CCA logic depending on what we are trying to achieve. For example, if using energy level as a criterion, we can use a different threshold for each scenario. Or we can switch between OR and AND logic when we are using both Correlator-based and Energy-based CCA.

I thought I'd put it on the table, this is what it might look like. I'll provide examples for Cc13xx/CC26xx IEEE and PROP modes once I've had some feedback.

Looking forward to reading your thoughts.


kkrentz:
I agree that multiple CCA thresholds may be needed. However, configuring these thresholds at the RADIO layer seems difficult because every RDC layer has its particularities. Moreover, RADIO layers probably do not have sufficient context information for "self-configuring" CCA thresholds. Instead, I suggest that RDC layers (self-)configure CCA thresholds by themselves. This way, RADIO layers remain simple. We might even replace `channel_clear` with `get_rssi` to free RADIO layers from managing CCA thresholds entirely.

PR #1787, opened 2016-07-17
Use `ccaInfo.ccaState` to decide whether CCA is complete

g-oikonomou:
This commit changes the logic of `get_cca_info()` in the CC26xx IEEE mode driver. We now use the command's return status bits to determine whether the radio's CCA monitoring function has completed, instead of drawing conclusions based on RSSI readings.

The logic that dictates whether CCA will signal an IDLE or BUSY channel is unchanged.


g-oikonomou:
I'll merge this in a month unless I receive an objection.

PR #1793, opened 2016-07-22
Minor data type correction.

sumanpanchal:


bthebaudeau:
Technically speaking, this change is useless, but it's cleaner, so :+1:. Thanks.

sumanpanchal:
@bthebaudeau Thanks!!

PR #1795, opened 2016-07-24
Add generic sensniff example and remove legacy sniffers

g-oikonomou:
This pull adds a generic sensniff example.

This implements the latest version of sensniff's serial line protocol and allows us to send commands from the host to the device to e.g. retrieve min/max/current channel or to change channel at runtime.

To test you will need to update your copy of [sensniff](https://github.com/g-oikonomou/sensniff) to its latest version.

This example will work with any radio that implements the extended NETSTACK radio API. More specifically, we need to be able to turn off frame filtering and automatic h/w ACKs by calling `NETSTACK_RADIO.set_value(RADIO_PARAM_RX_MODE, 0)`. For full functionality, this example needs the radio to also support:

```
NETSTACK_RADIO.get_value(RADIO_PARAM_CHANNEL, ...)
NETSTACK_RADIO.set_value(RADIO_PARAM_CHANNEL, ...)
NETSTACK_RADIO.get_value(RADIO_CONST_CHANNEL_MIN, ...)
NETSTACK_RADIO.get_value(RADIO_CONST_CHANNEL_MAX, ...)
```

With the approach proposed here, we no longer need to bother with changing the RF driver at all (unless required to implement the above functionality). Basically the example is meant to work with _zero_ changes to platform/CPU code.

I have also added support for quite a few platforms. I aimed for some of the popular ones, but also to cover a wide range of radios and CPUs in order to make it easier to add new ones:
* zoul (tested on remote over UART as well as USB)
* srf06-cc26xx (tested CC2650EM, CC1310EM, CC1310LP, CC2650LP)
* CC2538DK (tested UART as well as USB)
* CC2530DK (Srf05+CC2530EM, CC2531 USB dongle)
* Z1 (untested)
* jn516x (untested)
* openmote-cc2538 (untested)
* ev-aducrf101mkxz (untested)

The example should work with platforms equipped with the following radios:
* CC1200 (tested with zoul/remote)
* CC13xx/CC26xx IEEE and PROP modes
* CC2538
* CC2420 (untested)
* jn516x radio (untested)
* aducrf101 radio (untested)

The CC2520 RF driver doesn't support the extended API and cannot be supported as things stand. Perhaps this could be part of the future ToDo list.

I should add, I did not even look into any of the avr-based devices.

Assuming the radio is supported, adding support for new platforms is now trivial and requires minimal code replication. Literally all we need is a new directory with one or two header files and a small number of defines.

This example obsoletes platform- , cpu-, radio-specific implementations and examples, so I am removing them all.


arurke:
A much more elegant solution indeed. Tested on my cc1350 in PROP mode, works fine when combined with #1794.

examples/sensniff/zoul/target-conf.h
```diff
@@ -0,0 +1,71 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+/*---------------------------------------------------------------------------*/
+#ifndef TARGET_CONF_H_
+#define TARGET_CONF_H_
+/*---------------------------------------------------------------------------*/
+/* Set to 1 for a sub-ghz sniffer with the CC1200 */
+#ifndef ZOUL_CONF_SUB_GHZ_SNIFFER
+#define ZOUL_CONF_SUB_GHZ_SNIFFER     1
```
alignan: Please set to zero as default, as the `RE-Mote` and alike will default to 2.4GHz

alignan:
Great job! thanks for the work put into making sensniff more generic.  I can confirm works on the `RE-Mote` for both 2.4GHz and 868MHz.   I would suggest having sensniff as submodule, despise the reasons against if any, I think it would be appreciated generally, and easier to work with.

:+1:  after the minor change requested in my inline comment, and after travis is happy.

g-oikonomou:
Ooops not sure how that one escaped me. Update now. Regarding adding as submodule, perhaps one day :)

joakimeriksson:
We have just opened up Yanzi's codebase for the Sparrow application layer - there is a significantly updated serial radio there including support for sniffing, RSSI scanning, etc. There is also a 3D RSSI scanner demo as an example there. 

Documentation is not fully done yet but there is some there.
https://github.com/sics-iot/sparrow

Maybe we should try to make a "grand" combined super-duper sniffer/serial-radio/RSSI scanner thing to avoid fragmentation?

alignan:
Normally I prefer to merge working applications now rather than wait for a newer one that would supersedes it.  That said, I have been testing sparrow and it is working great! so I hope to see more of it in Contiki :smile: 

I keep my :+1:  if @g-oikonomou addresses my suggestion

PR #1805, opened 2016-08-01
galileo: Build link-stats.c when IPv6 is enabled

mdlemay:
This is required by the IPv6 implementation in Contiki OS.


nvt:
:+1:

PR #1810, opened 2016-08-03
x86: Support configurable UART baud rate and fix initialization

mdlemay:
The first patch extends the Intel Quark X1000 SoC UART initialization API to accept a numeric baud rate specification.  This is useful for projects that require configuring a serial port with a particular baud rate.

The second patch fixes UART system call authorization initialization for applications that initialize both UARTs.


nvt:
:+1:

PR #1813, opened 2016-08-08
Cc26xx/cc13xx I2C fix

arurke:
Fixes I2C multiple-write bug, as reported by Ron Segal: https://sourceforge.net/p/contiki/mailman/message/34633862/
plus minor opt.

### DIFF
--- a/platform/srf06-cc26xx/sensortag/board-i2c.c
+++ b/platform/srf06-cc26xx/sensortag/board-i2c.c
@@ -174,8 +174,6 @@ board_i2c_write(uint8_t *data, uint8_t len)
 bool
 board_i2c_write_single(uint8_t data)
 {
-  bool success;
-
   /* Write slave address */
   ti_lib_i2c_master_slave_addr_set(I2C0_BASE, slave_addr, false);
 
@@ -188,9 +186,8 @@ board_i2c_write_single(uint8_t data)
   /* Assert RUN + START + STOP */
   ti_lib_i2c_master_control(I2C0_BASE, I2C_MASTER_CMD_SINGLE_SEND);
   while(ti_lib_i2c_master_busy(I2C0_BASE));
-  success = i2c_status();
 
-  return success;
+  return i2c_status();
 }
 /*---------------------------------------------------------------------------*/
 bool
--- a/platform/srf06-cc26xx/sensortag/board-i2c.c
+++ b/platform/srf06-cc26xx/sensortag/board-i2c.c
@@ -256,12 +256,11 @@ board_i2c_write_read(uint8_t *wdata, uint8_t wlen, uint8_t *rdata, uint8_t rlen)
   for(i = 1; i < wlen && success; i++) {
     /* Write next byte */
     ti_lib_i2c_master_data_put(I2C0_BASE, wdata[i]);
-    if(i < wlen - 1) {
-      /* Clear START */
-      ti_lib_i2c_master_control(I2C0_BASE, I2C_MASTER_CMD_BURST_SEND_CONT);
-      while(ti_lib_i2c_master_busy(I2C0_BASE));
-      success = i2c_status();
-    }
+
+    /* Clear START */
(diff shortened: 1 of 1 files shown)


arurke:
I would also like to suggest moving the board-i2c files into the common-folder.

Also adding a write_single_read() function as this is the most common usage (at least in my experience). Alternatively move the sensor-common.* files as well, but I am having trouble understanding the need to do all the copying in sensor_common_write_reg()...

g-oikonomou:
I'm not following you here, Andreas

arurke:
The PR or the comments?

Regarding the PR. I had trouble finding doc. for exactly what the different
master control commands do, but I (and Ron I guess) believe data_put puts
data into master reg. and I2C_MASTER_CMD_BURST_SEND_CONT does the actual
transmission.
Before this PR the I2C_MASTER_CMD_BURST_SEND_CONT command would not be
executed after the last byte is put into the reg. if more than 1 byte was
being transmitted.


Den tirsdag 9. august 2016 skrev George Oikonomou <notifications@github.com>
følgende:

arurke:
Digging deeper I think I am following why you are not following.

In board_i2c_write() it seems I2C_MASTER_CMD_BURST_RECEIVE_FINISH will indeed transmit whatever is in the i2c master register. So me extrapolating Ron's report onto board_i2c_write() seems incorrect.

However in board_i2c_write_read() there are no such finish command. The command after the writes are setting slave address and I2C_MASTER_CMD_BURST_RECEIVE_START. If these does not cause the byte in the register to be transmitted (which my crude testing indicates they do not), I believe it is a bug.

g-oikonomou:
Andreas, I have only found the time to look at this very superficially, so it's probably me :) However, just double-checking, could this be in any way related with #1711?

g-oikonomou:
OK I think I'm with you now.

Those `I2C_MASTER_CMD_xxx` manipulate the `ACK`, `START`, `STOP`, `RUN` bits of the `MCTRL` register. A transmission or reception of a byte is executed by setting the `RUN` bit here.

For the first byte of a burst TX transaction we want `START` and `RUN`, but not `STOP`. For intermediate bytes we want simply `RUN`, for the last byte we want `RUN` and `STOP`. This will basically send the I2C start condition before the first byte and the stop condition after the last one.

It appears as though we are missing a trigger of the `RUN` for the last TX byte. This should be a `_BURST_SEND_CONT` and not a `_FINISH`, since we do _not_ want the stop condition sent between direction changes. Thus, your recommended change seems valid to me.

Have a look at 21.3.5 of revision F of the TRM (SWCU117F), those flow charts make it nicely clear.

Regarding moving the file to somewhere more central, the reason it is located where it currently is, is because the MPU's I2C pins on the sensortag are connected to different DIOs than the I2C pins of all other sensors. Hence this interface selection in `board_i2c_select()`. If we were to move it, we should also come up with a more generic fashion that would allow boards to specify number of I2C interfaces and corresponding pins. If we were to move it as-is, it would fail to compile due to those sensortag-specific defines.

arurke:
Regarding i2c (and spi):
Could we make the libraries be almost agnostic to the pins (except storing the current selected), and let the application be responsible for using them correctly? e.g. having functions such as:
```
board_i2c_select(uint8_t sda_pin, uint8_t scl_pin, uint8_t address)
    selects and stores current pins
board_i2c_write/read/etc()
    acts on the current pins
board_i2c_deselect() and/or close()
   deselects/closes the current pins
```
And identically
```
board_spi_select(uint32_t bit_rate, uint8_t clk_pin, uint8_t miso_pin, uint8_t mosi_pin)
board_spi_write/read/flush()
board_spi_deselect() and/or close()
```

This would be completely generic and we don't have to worry about storing away lists of ports pr board - at the expense of slightly uglier code (having to specify pins instead of I2C_INTERFACE_0 etc.) and slightly more responsibility to the application (not messing up pins)

arurke:
As I made the comment on a closed thread I am not sure if a notif. is sent out or not, so please allow me to tag your name @g-oikonomou 
You are of course perfectly welcome not to respond :)

PR #1823, opened 2016-08-14
Fix duplicate register definition warnings for atmega256rfr2 + rf230bb

vikovo:
I am working on a platform using the said combination and the build produces a bunch of
warnings about duplicate register definitions. Judging by the comment above the definitions it seems to me that someone forgot to exclude them from the radio registermap.

I am not sure if the attached simple fix is valid for all the toolchain versions but it works for me
using the recent avr-libc. Update: it is not.


nvt:
I'm closing since this did not pass the compilation test, and has not been updated. We can re-open it if you at some point would like to provide another fix for the warnings.

PR #1832, opened 2016-08-22
TI ADS7828 test application with Zolertia Remote.

sumanpanchal:
Test application to expand ADC channels using TI ADS7828 on Zolertia Remote and read sensor data using Remote I2C interface.
ADS7828 Datasheet: http://www.ti.com/lit/ds/symlink/ads7828.pdf


alignan:
Sorry for taking so long to answer, my intention was to make a PR on top of your contribution to show you my comments, but I got tangled with lots of work.

I like the contribution and the module itself, but I'm normally against having the sensor/actuator implementation in the same example application, as this difficult its reuse.  I would suggest separating the `ADS7828` driver in `platforms/zoul/dev/`, and provide a test example using the `ADS7828` with Contiki's sensor's API.

See the examples in `platforms/zoul/dev` for guidance, I will be happy to chip in anytime.

sumanpanchal:
@alignan 
Thanks for the feedback. I have updated the code as per your suggestion. Please look into it.

sumanpanchal:
@alignan 
Please provide your feedback so it can be merge.

platform/zoul/dev/ads7828.c
```diff
@@ -0,0 +1,79 @@
...
+/*---------------------------------------------------------------------------*/
+#define DEBUG 0
+#if DEBUG
+#define PRINTF(...) printf(__VA_ARGS__)
+#else
+#define PRINTF(...)
+#endif
+/*---------------------------------------------------------------------------*/
+void
+ads7828_init()
```
msloth: this should be `ads7828_init(void)`; leaving out the `void` means something else than "no argument". Same thing in the header file.

examples/zolertia/zoul/test-ads7828.c
```diff
@@ -0,0 +1,108 @@
...
+
+  uint16_t data;
+
+  printf("TI ADS7828 test application\n");
+
+  ads7828_init();
+  etimer_set(&et, LOOP_INTERVAL);
+
+  while(1) {
+    PROCESS_YIELD();
```
msloth: looks like little unusual use of the abstractions, how about simply

```
while(1) {
  static struct etimer et;
  etimer_set(&et, LOOP_INTERVAL);
  PROCESS_WAIT_EVENT_UNTIL(etimer_expired(&et));

  if(ads7828_read_channel(ADS7828_CMD_0, &data) == ADS7828_SUCCESS) {
    printf("ADC=%d\n", data);
  }
}
```

this would greatly increase readability imo.

msloth:
I don't have the hardware to test it, but I added a couple of suggestions on how to improve look and form of the code.

PR #1833, opened 2016-08-23
CC2520: Added offset to calculate correct RSSI value

sumanpanchal:
In CC2520 datasheet (http://www.ti.com/lit/ds/symlink/cc2520.pdf), according to section 5.7 
Real RSSI = Register value - offset
where offset = 76 dBm


nvt:
:+1:

PR #1835, opened 2016-08-23
galileo: Add general protection exception handler

mdlemay:
This patch adds a general protection exception handler that prints a
message indicating the faulting instruction and the error code. This is
useful when debugging general protection exceptions.


nvt:
:+1:

PR #1836, opened 2016-08-23
galileo: Add process to perform DHCP configuration

mdlemay:
This patch adds a process that is started automatically to request DHCP
configuration.  It also moves the IP configuration ahead of autostart processes
in case some autostart process depends on the IP configuration.


nvt:
:+1:

Issue #1505, opened 2016-02-07
cc2538: pwm: Improve API


bthebaudeau:
This issue is a discussion about the CC2538 PWM driver API improvements suggested in #1419 to @alignan following #1399.

I have regrouped my suggestions below.

As further improvements to the PWM driver, I think that it would be good to simplify the API to use only the following functions:
 - `pwm_start(timer, a_b, port, pin, freq, polarity)`, which would be the equivalent of `pwm_enable(0%)` + `pwm_set_direction(polarity)` + `pwm_stop()` with the existing API,
 - `pwm_set_duty_cycle(timer, a_b, duty_cycle)`, which would dynamically adjust the duty cycle without stopping the timer, except for 0% and 100%, which would be the equivalent of `pwm_stop()` with the existing API,
 - `pwm_stop(timer, a_b)`, which would be the equivalent of `pwm_disable()` with the existing API, the used port and pin having been saved by `pwm_start()` in order not to have to pass them twice, or they could also be derived from the MUX + GPIO registers.

The rationale for this is that there should be a PWM session between `pwm_start()` and `pwm_stop()` during which the chosen pin is always generating a PWM signal between 0% and 100%. Apart from the duty cycle, all the PWM settings should be constant during a session. During this PWM session, the regular usage of a PWM requires to be able to change the duty cycle without stopping the timer, so that this operation is simple, quick and glitch-free. `pwm_set_duty_cycle()` could automatically start and stop the PWM timer when switching between 0%/100% and the other duty cycle values, in order to optimize power consumption. After `pwm_stop()`, the end of the PWM session should revert the timer and the pin to their reset states.

It would probably be more efficient to use a fractional data type rather than a percentage to express the duty cycle passed to this API, because:
 - % are very human while the duty cycle values passed to this API will probably be software-generated rather than human-generated, especially if the duty cycle has to change in real time,
 - % are inefficient for the CPU, requiring a division, while binary-oriented representations would only require a bit shift instead,
 - the accuracy would be better and could better benefit from the timer resolution if the frequency is well chosen.

It would also be great to have the following options:
 - hardware synchronization of timers,
 - rounding the start counter value to a power of 2, minus 1, preferably to get full 8- or 16-bit resolution for the duty cycle (using the 8-bit prescaler can help here, and the accuracy of the frequency is often unimportant for PWM signals), in order to ease the use of duty cycle values resulting from computations,
 - updating the duty cycle through µDMA.

PR #1837, opened 2016-08-24
Zoul: generic servo driver

Zolertia:
This PR adds a basic servo driver using the PWM library

[](https://www.youtube.com/watch?v=LMnSEBkATvA)

There are still suggestions open in https://github.com/contiki-os/contiki/issues/1505 to improve the PWM driver, but this PR doesn't significantly prevents so.


alignan:
:+1:

PR #1844, opened 2016-08-31
Zoul: added driver to control an AC light dimmer with zero-crossing

alignan:
This PR adds support for an AC light dimmer using a zero-crossing trigger.  The specific AC dimming module is the [Krida Electronic's AC dimmer module](http://www.inmojo.com/store/krida-electronics/item/ac-led-light-dimmer-module-with-heatsink/).

[](https://youtu.be/cvkhv_EkM6Y)

The timings were verified using a logic analyser, for a 50Hz AC frequency.


alignan:
:+1:

PR #1846, opened 2016-09-01
MSP430/F2xxx: fix ubr msb byte being hardcoded to zero

alignan:
This PR request fixes UART0 initialization with a hard-coded UBR's MSB register to zero, preventing baudrates like 9600 from properly being configured. Tested for 115200 and 9600 baud rates, using any of the Z1 test examples with putty configured for the tested baudrates.


alignan:
:+1:

PR #1848, opened 2016-09-05
fix misplaced bracket causing compilation error when probing is disabled

afonsomota:
When configuring a project to not use RPL parent probing (ex: we set RPL_CONF_WITH_PROBING to 0 in our project configuration), a misplaced bracket causes a compilation error in function rpl_select_parent of the file core/net/rpl/rpl-dag.c.


simonduq:
that seems correct to me but please also fix the new indentation of the block

afonsomota:
Sorry, should have done it from the start. The new commit does the fix.

simonduq:
:+1:

PR #1852, opened 2016-09-06
Port efm32

niwotahao:


simonduq:
Please check the rules for new platforms:
https://github.com/contiki-os/contiki/wiki/Code-Contributions#new-platforms

Please add some text in the PR to motivate the addition of this new platform. Please avoid changes to apps and core files (acceptable only if strictly needed). Am also unsure about the whole set of emf-related header files, which seem like standard lib rather than contiki-specific, it might be a better choice to keep these outside of the contiki tree. Please rebase on master (not 2.7 nor 3.0). I invite you to submit a new PR once all the requirements are met.

PR #1854, opened 2016-09-08
x86: Fix section alignment for EFI binaries

mdlemay:
This patch fixes the section alignment specification in the linker script for
the Intel Quark X1000 SoC so that the UEFI GenFw program generates correct EFI
binaries.  See the added comment in quarkX1000.ld for details.


nvt:
:+1:

PR #1856, opened 2016-09-12
use the arm-none-eabi toolchain by default

yongli3:
Signed-off-by: Yong Li <sdliyong@gmail.com>


yongli3:
Since the InstantContiki does not include the arm-elf toolchain, we need to change it

PR #1859, opened 2016-09-13
Zoul: RTCC with trigger selection and second-based time increase

alignan:
This PR adds two RTCC functionalities:

- Allows to select which pin to trigger upon an interrupt (INT1 or INT2)
- Allows to increment a date using seconds, useful to schedule an alarm interrupt

These features are used in the shutdown mode of the RE-Mote revision B (see https://github.com/contiki-os/contiki/pull/1857), which will be implemented in a next PR


alignan:
Platform specific, merging :+1:

PR #1864, opened 2016-09-17
Wismote: Added support for Battery voltage measurement

sumanpanchal:
Have added Wismote battery voltage measurement support. Tested with `test-battery.c` example code and it works fine.


nvt:
:+1:

PR #1866, opened 2016-09-20
Improved serialdump.c

yschroeder:
This PR fixes some issues we found when working with serialdump on our own Contiki fork.

First, it adresses the issue that serialdump sometimes injects newlines while running. This results from not configured `c_iflag` in terminos. Depending on your system it might be configured to replace carriage returns with newlines or vice versa. This results in excessive newlines in the terminal output and wrong output when transferring binary data. While at it I made sure to initialize all `c_iflag` values to make it work reproducible.

Second, it makes serialdump terminate when the serial device gets disconnected, e.g. when the USB cable is unplugged. The original behaviour was that serialdump runs indefinitely without recovering even when replugging the USB device.

Third, serialdump exits cleanly (returns 0) when terminated via `SIGINT` (`ctrl-c`). This is especially handy when used via `make login` as make will not assume anymore that the build failed.

### DIFF
--- a/tools/sky/serialdump.c
+++ b/tools/sky/serialdump.c
@@ -208,6 +208,8 @@ main(int argc, char **argv)
   options.c_cflag |= CS8;
 
   /* Raw input */
+  options.c_iflag &= ~(IGNBRK | BRKINT | PARMRK | ISTRIP
+                       | INLCR | IGNCR | ICRNL | IXON);
   options.c_lflag &= ~(ICANON | ECHO | ECHOE | ISIG);
   /* Raw output */
   options.c_oflag &= ~OPOST;
--- a/tools/sky/serialdump.c
+++ b/tools/sky/serialdump.c
@@ -281,6 +281,11 @@ main(int argc, char **argv)
         perror("could not read");
         exit(-1);
       }
+      if(n == 0) {
+        errno = EBADF;
+        perror("serial device disconnected");
+        exit(-1);
+      }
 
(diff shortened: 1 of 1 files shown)


alignan:
Cool! I can relate to the problems you found :+1:

yschroeder:
Have you reviewed the code? Do you think this can be merged?

alignan:
Tested in Linux and works as intended :+1: 

Could someone test for OSX?

alignan:
Probably it would make sense to rebuild the OS binaries before merging

yschroeder:
Yes, but the binaries should be built by a member of the project right? Also I cannot build or test for OS X.

alignan:
I can build the linux one, but no OSX on my side either... perhaps @g-oikonomou?

g-oikonomou:
It builds fine on El Capitan

g-oikonomou:
So I tested this with a Zolertia Firefly. I am getting identical behaviour with master and with this PR:

```
$ make login
using saved target 'zoul'
../../../tools/sky/serialdump-macos -b115200 /dev/tty.SLAB_USBtoUART
connecting to /dev/tty.SLAB_USBtoUART (115200) [OK]
??P???P???P?could not read: Device not configured
make: *** [login] Error 255
````

So as far as I'm concerned this doesn't break OS X, but it fixes other stuff so looks good.

alignan:
I will rebuild the linux binary on a separate PR, so :+1:  from me

PR #1869, opened 2016-09-22
Added support in Cooja to include Wismote and Z1 motes when exporting simulations

nfi:
Cooja lacked support for including Wismote and Z1 motes when exporting simulations as executable JAR files. Earlier Cooja only supported Tmote Sky motes and this PR adds support also for Wismote and Z1 motes.


alignan:
:+1:

PR #1870, opened 2016-09-22
CC1200 added config for 1.2 kbps

simonduq:


alignan:
Cool! I will test and add a few things to the PR :smile:

alignan:
In the [CC1200](https://github.com/simonduq/contiki/blob/7c58c0ece99bddbc70be59e87b81d422f65cc32c/dev/cc1200/cc1200.c#L801) I had to increase the waiting time to swap back to RX up to i.e `RTIMER_SECOND` in earlier tests, I added something like:

````
#if CC1200_RF_CFG == cc1200_868_fsk_1_2kbps
    BUSYWAIT_UNTIL_STATE(STATE_RX, RTIMER_SECOND);
#else
    BUSYWAIT_UNTIL_STATE(STATE_RX, RTIMER_SECOND/100);
#endif
````

simonduq:
`#if CC1200_RF_CFG == cc1200_868_fsk_1_2kbps` won't work unfortunately so we need to add one more field to `cc1200_rf_cfg_t` instead. What is this particular busy wait exactly waiting for? At first I thought we wanted to use `tx_pkt_lifetime` but no, as the packet transmission is already over when reaching there..

alignan:
I found out that it takes a bit longer to actually swap back to RX after transmitting a packet, enabling the interruptions earlier gave me some problems at that time, as we were returning the `rf_flags` ahead of time

simonduq:
OK, I basically added a new field `tx_rx_turnaround`

alignan:
:+1: and merge as soon as you approve

simonduq:
thanks!

PR #1873, opened 2016-09-22
Doxygen Fix

MohamedSeliem:
this minor fixation to avoid Doxygen bug, which appears when doing this in the documented code:
/**
/* \exception \c text
*/
void foo();


alignan:
:+1:

PR #1874, opened 2016-09-23
Zoul: added new Firefly revision A

alignan:
This PR adds support for the new revision of the Firefly platform (namely revision A), it comprises more or less the same features as the Firefly, with a different pin-out, and the following improvements:
* A USB type A connector instead of Micro-USB to directly connect to devices i.e Raspberry Pi, BeagleBone Black, USB hubs, etc
* On-board 2.4GHz ceramic chip antenna
* Compatible with 2xAA/AAA battery holders
* Improvements in the RF impedance matching


alignan:
Platform specific, merging :+1:

PR #1876, opened 2016-09-26
stf06-cc26xx: Enable input chars in serial input process for srf06-cc…

naguirre:
…26xx platform


naguirre:
Ok ! Make sense. Thanks for the info !

PR #1877, opened 2016-09-28
Zoul: updated pin-out images

alignan:
Zoul's back pin-out image was rotated and incorrect, fixing and adjusting both images to same pan-out


alignan:
Merging :+1:

PR #1879, opened 2016-09-30
Change on DIO trickle timer condition for DIO sending

pernin:
The two changes are into the condition for the transmission of DIO. The third is commented in the code.
1) The first is to be compliant to the RFC6550 that says: "In RPL, when k (i.e. DIORedundancyConstant) has the value of 0x00, this is to be treated as a redundancy constant of infinity in RPL, i.e., Trickle never suppresses messages). So, when the  instance->dio_redundancy is equal to 0, we can't suppress DIO transmission. This is the way to allow user to disable DIO suppression.
2) If we don't want to set the default route lifetime to infinite, in the case of a very large network (nodes > 40), there will be a high probability that instance->dio_counter is greater than instance->dio_redundancy. In this case the node will not send DIO anymore in the future (more so if this node is the root). So, other nodes will not receive anymore DIO from this node, and this node will be removed from the NB list. So, when we reach the maximum doubling of DIO interval, it is recommended to send DIO anyway to keep other nodes updated about this node.
3) Another possible change alternative to this is to multiply instance->dio_redundancy to the difference between actual interval and minimum interval (the longer the interval, the more numerous will be the received DIO... we can't keep the redundancy constant to a "constant" value!!!!).

In any case it is recommended to keep the RPL_CONF_DIO_REDUNDANCY high (>= nodes) when the network contains a large number of nodes.

### DIFF
--- a/core/net/rpl/rpl-timers.c
+++ b/core/net/rpl/rpl-timers.c
@@ -174,7 +174,7 @@ handle_dio_timer(void *ptr)
 
   if(instance->dio_send) {
     /* send DIO if counter is less than desired redundancy */
-    if(instance->dio_redundancy != 0 && instance->dio_counter < instance->dio_redundancy) {
+    if(instance->dio_redundancy == 0 || instance->dio_counter < instance->dio_redundancy) {
 #if RPL_CONF_STATS
       instance->dio_totsend++;
 #endif /* RPL_CONF_STATS */


simonduq:
I didn't get the point with `instance->dio_intcurrent >= instance->dio_intmin + instance->dio_intdoubl`: when is this supposed to happen and why do we want not to suppress in such case?

pernin:
Probably this condition can be seen as a forcing, but RFCs aren't fully clear.
I introduced this condition in my work (and so I have proposed it) after a lot of tests and observations on the network with a fairly large number of nodes (> 40).
If you don't want to set the default route lifetime to infinite (in this way the network can be more dynamic), but at the same time you want to use the DIO sending suppression (that in any case with the original condition can't be disabled), after some time the root stops to send its DIO (cause it receives DIOs from nodes) and when the default route lifetime expires the nodes remove their default route. In this case, they continue to send DIOs, and the root continues to receive DIOs: so the root will not send DIO anymore, and nodes will never know who is the root! Until the network contains a few nodes, this is not a problem, because the DIO sending suppression happens rarely, but if you have lot of nodes the suppression interveins very often!
The condition I've proposed plans to send a DIO in anycase when the DIO timer expires and the root has already reached the maximum of DIO timer doubling. So in this case we don't suppress the sending, but this happens very rarely (we are at the maximum of doubling!).
The other possible change is the increase of the threshold for the suppression, by multiplying it to the number of doubling (the longer the interval, the greater the number of received DIO). But this probably goes against RFC that talks about a "constant" for the threshold of the suppression.

simonduq:
Sounds like a hack to me, and not rfc-compliant.
But I like the new condition `instance->dio_redundancy == 0`, this was definitely a bug.

pernin:
Probably it is so, but RFC aren't clear at all (RFC only talk about a "k" that is configured starting from `DIORedundancyConstant`.
In my opinion, remains the case that in any case a mechanism to allow nodes to know the root after ending their default route lifetime is needed.
Thankyou for your feedback.

simonduq:
To sum up, I will support this PR if it get simplified significantly:
* single commit that fixes the case of redundancy == 0
* no extra feature on `dio_intcurrent` nor the commented code

simonduq:
where are we on this?

pernin:
I'm using the changes I proposed on my network for over a month and I noticed good improvement in stability. I have not made any further changes to those proposed.

simonduq:
OK but this is not just a stability issue, but rather code style and standard compliance.
Are you intending to make the changes or not?

pernin:
Yes, today I will change the code following your comment of 12 October. Sorry for the late

> To sum up, I will support this PR if it get simplified significantly:
> 
> single commit that fixes the case of redundancy == 0
> no extra feature on dio_intcurrent nor the commented code

simonduq:
Great, thanks so much!
Would be awesome if you could squash all commits into a single one :) (interactive rebase followed by push -f)

pernin:
It's embarrassing but I could not run the interactive rebase! I used the git reset feature. Please check if it is all ok.

simonduq:
:+1: , setting timeout for merging

nvt:
:+1:

PR #1885, opened 2016-10-08
Cc26xx/cc13xx generic I2C library

arurke:
This PR modify the existing sensortag-specific board-i2c library to become generic and move it to /common/i2c.[c|h]. Usage is similar as before, i.e. user must call i2c_select() before i2c operations and should call i2c_deselect() afterwards. Added possibility to specify bus speed: i2c_select() now requires SCL & SDA pin IOID, bus speed (normal or fast), slave address and pin pull (internal pull to set at shutdown/deselect).
Usage of underlying I2C functions has been mostly carried over from the board-specific version. As I don't feel 100 % confident on the TI I2C library any reviews are welcome - in particular i2c_deselect() which has a somewhat different approach than the previous version.

Updated sensortag sensors to use the new library. Also added i2c_write_single_read_multi() for reading registers, which is a common task - this could replace sensor_common_read_reg() in sensor-common.c.

Tested on cc2650 sensortag.

### DIFF
--- a/platform/srf06-cc26xx/common/i2c.c
+++ b/platform/srf06-cc26xx/common/i2c.c
@@ -297,6 +297,12 @@ i2c_write_read(uint8_t *wdata, uint8_t wlen, uint8_t *rdata, uint8_t rlen)
   return success;
 }
 /*---------------------------------------------------------------------------*/
+bool
+i2c_write_single_read_multi(uint8_t wdata, uint8_t *buf, uint8_t len)
+{
+  return i2c_write_read(&wdata, 1, buf, len);
+}
+/*---------------------------------------------------------------------------*/
 void
 i2c_select(uint32_t new_pin_sda,
                  uint32_t new_pin_scl,
--- a/platform/srf06-cc26xx/common/i2c.c
+++ b/platform/srf06-cc26xx/common/i2c.c
@@ -29,25 +29,43 @@
  */
 /*---------------------------------------------------------------------------*/
 /**
- * \addtogroup sensortag-cc26xx-i2c
+ * \author Andreas Urke <arurke@gmail.com>
+ *
+ * \addtogroup cc26xx-cc13xx-i2c
  * @{
  *
+ *
  * \file
- * Board-specific I2C driver for the Sensortag-CC26xx
+ * Generic I2C driver for cc26xx/13xx
  */
 /*---------------------------------------------------------------------------*/
 #include "contiki-conf.h"
 #include "ti-lib.h"
-#include "board-i2c.h"
+#include "i2c.h"
 #include "lpm.h"
 
 #include <string.h>
 #include <stdbool.h>
 /*---------------------------------------------------------------------------*/
-#define NO_INTERFACE 0xFF
+#define DEBUG 0
+#if DEBUG
+#include <stdio.h>
+#define PRINTF(...) printf(__VA_ARGS__)
+#else
+#define PRINTF(...)
+#endif
 /*---------------------------------------------------------------------------*/
-static uint8_t slave_addr = 0x00;
-static uint8_t interface = NO_INTERFACE;
+#define I2C_PIN_NOT_SET           IOID_UNUSED
+#define I2C_SLAVE_ADDRESS_NOT_SET 0x00
+#define I2C_PIN_PULL_NOT_SET      0x00
+#define I2C_DEFAULT_SPEED         I2C_SPEED_FAST
+/*---------------------------------------------------------------------------*/
+static uint8_t selected_slave_addr = I2C_SLAVE_ADDRESS_NOT_SET;
+static uint32_t selected_pin_sda = I2C_PIN_NOT_SET;
+static uint32_t selected_pin_scl = I2C_PIN_NOT_SET;
+static bool selected_speed = I2C_DEFAULT_SPEED;
+static uint32_t selected_pin_pull = I2C_PIN_PULL_NOT_SET;
+
 /*---------------------------------------------------------------------------*/
 static bool
 accessible(void)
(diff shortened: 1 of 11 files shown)


platform/srf06-cc26xx/common/i2c.c
```diff
@@ -293,38 +300,82 @@ board_i2c_write_read(uint8_t *wdata, uint8_t wlen, uint8_t *rdata, uint8_t rlen)
...
-  if(new_interface != interface) {
-    interface = new_interface;
+  /* Configure for new values only */
+  if(new_pin_sda != selected_pin_sda ||
+      new_pin_scl != selected_pin_scl ||
+      new_speed != selected_speed) {
+
+    /* If already selected, deselect first */
+    if(selected_pin_sda != I2C_PIN_NOT_SET ||
+      selected_pin_sda != I2C_PIN_NOT_SET) {
```
tonnenpinguin: Are you sure this shouldn't be selected_pin_scl? ;)

platform/srf06-cc26xx/common/i2c.c
```diff
@@ -29,25 +29,43 @@
...
+#if DEBUG
+#include <stdio.h>
+#define PRINTF(...) printf(__VA_ARGS__)
+#else
+#define PRINTF(...)
+#endif
+/*---------------------------------------------------------------------------*/
+#define I2C_PIN_NOT_SET           0x00
+#define I2C_SLAVE_ADDRESS_NOT_SET 0x00
+#define I2C_PIN_PULL_NOT_SET      0x00
```
tonnenpinguin: Please note that 0x00 is actually a valid IOID (IOID_0)!

platform/srf06-cc26xx/common/i2c.c
```diff
@@ -29,25 +29,43 @@
...
-#define NO_INTERFACE 0xFF
+#define DEBUG 0
+#if DEBUG
+#include <stdio.h>
+#define PRINTF(...) printf(__VA_ARGS__)
+#else
+#define PRINTF(...)
+#endif
+/*---------------------------------------------------------------------------*/
+#define I2C_PIN_NOT_SET           0x00
```
tonnenpinguin: Please note that 0x00 is actually a valid IOID (IOID_0)!

platform/srf06-cc26xx/common/i2c.c
```diff
@@ -293,38 +300,82 @@ board_i2c_write_read(uint8_t *wdata, uint8_t wlen, uint8_t *rdata, uint8_t rlen)
...
+/*---------------------------------------------------------------------------*/
+void
+i2c_deselect()
+{
+  PRINTF("I2C: Deselecting sda: %lu, scl: %lu, addr: %x\n",
+         selected_pin_sda, selected_pin_scl, selected_slave_addr);
+
+  /* Deselect only if selected */
+  if(selected_pin_sda != I2C_PIN_NOT_SET &&
+      selected_pin_scl != I2C_PIN_NOT_SET) {
```
tonnenpinguin: Most of the time you are checking whether at least one of these pins is not set. Why did you choose to change logic here?

platform/srf06-cc26xx/common/i2c.c
arurke: Good find! Fixed.

arurke:
Thanks a lot @tonnenpinguin ! I made some changes to deselect() to make it consistent, simpler and a tiny bit more efficient.

platform/srf06-cc26xx/common/i2c.c
```diff
@@ -29,25 +29,43 @@
...
+#define I2C_PIN_PULL_NOT_SET      0x00
+#define I2C_DEFAULT_SPEED         I2C_SPEED_FAST
 /*---------------------------------------------------------------------------*/
-static uint8_t slave_addr = 0x00;
-static uint8_t interface = NO_INTERFACE;
+static uint8_t selected_slave_addr = I2C_SLAVE_ADDRESS_NOT_SET;
+static uint32_t selected_pin_sda = I2C_PIN_NOT_SET;
+static uint32_t selected_pin_scl = I2C_PIN_NOT_SET;
+static bool selected_speed = I2C_DEFAULT_SPEED;
+static uint32_t selected_pin_pull = I2C_PIN_PULL_NOT_SET;
```
tonnenpinguin: ti libs define IOC_NO_IOPULL for no pull, so why not set it to a valid value? :)

arurke: I wanted to be able to identify if a value was set and not (which will not be possible if using IOC_NO_IOPULL since it is a valid value). At the moment this has no implications but it seems reasonable just for the sake of consistency. If you have any arguments otherwise I am all yours :)

tonnenpinguin: Fair enough!

platform/srf06-cc26xx/common/i2c.c
```diff
@@ -98,9 +117,11 @@ i2c_status()
 }
 /*---------------------------------------------------------------------------*/
 void
-board_i2c_shutdown()
+i2c_shutdown()
```
arurke: 👍

platform/srf06-cc26xx/common/i2c.h
```diff
@@ -99,33 +105,41 @@ bool board_i2c_write_single(uint8_t data);
...
  */
-bool board_i2c_write_read(uint8_t *wdata, uint8_t wlen, uint8_t *rdata,
+bool i2c_write_read(uint8_t *wdata, uint8_t wlen, uint8_t *rdata,
                           uint8_t rlen);
 
 /**
+ * \brief Write one byte and read len bytes in one operation
+ * \param wdata Pointer to the buffer to be written
+ * \param buf Pointer to the buffer to be written
+ * \param len Number of bytes to write
```
greg-king5: Those "param" descriptions are wrong: `wdata` isn't a pointer; `buf` and `len` are for the "read" half of the operation.

arurke: 👍 boiler-plate symptom

PR #1896, opened 2016-10-19
Renamed driver and reworked to support both TSL2561/TSL2563

alignan:
This PR extends the existing light sensor driver to support both TSL2563 and TSL2561, the later one used in Seeedstudio's grove digital light sensor and currently tested.


alignan:
Merging now :+1:

PR #1897, opened 2016-10-19
galileo: Shrink pinmux array

mdlemay:
The pins array in galileo-gen2-pinmux.c:galileo_brd_to_cpu_gpio_pin is
unnecessarily large.  This patch reduces its size.


alignan:
:+1:

PR #1898, opened 2016-10-21
Contiki-6top : Added 6top (6TiSCH Operation Sublayer) Implementation 

shalurajendran:
Handles 2-step 6top transaction on a basic RPL + TSCH network. The current implementation supports dynamic cell negotiation (addition/deletion) with one-hop neighbor. Tested working with Z1 platform on Cooja simulator using the example code node-sixtop.c (examples/ipv6/rpl-tsch-sixtop/).

### DIFF
--- a/core/net/mac/frame802154e-ie.c
+++ b/core/net/mac/frame802154e-ie.c
@@ -60,6 +60,7 @@ enum ieee802154e_header_ie_id {
 enum ieee802154e_payload_ie_id {
   PAYLOAD_IE_ESDU = 0,
   PAYLOAD_IE_MLME,
+  PAYLOAD_IE_SIXTOP,
   PAYLOAD_IE_LIST_TERMINATION = 0xf,
 };
 
@@ -186,6 +187,20 @@ frame80215e_create_ie_payload_list_termination(uint8_t *buf, int len,
   int ie_len = 0;
   if(len >= 2 + ie_len && ies != NULL) {
     create_payload_ie_descriptor(buf, PAYLOAD_IE_LIST_TERMINATION, 0);
+    return 2 + ie_len;
+  } else {
+    return -1;
+  }
+}
+
+/* Payload IE. 6top. Used to nest sub-IEs */
+int
+frame80215e_create_ie_sixtop(uint8_t *buf, int len,
+    struct ieee802154_ies *ies)
+{
+  int ie_len = 0;
+  if(len >= 2 + ie_len && ies != NULL) {
+    create_payload_ie_descriptor(buf, PAYLOAD_IE_SIXTOP, ies->ie_mlme_len);
     return 2 + ie_len;
   } else {
     return -1;
@@ -529,6 +544,11 @@ frame802154e_parse_information_elements(const uint8_t *buf, uint8_t buf_size,
             len = 0; /* Reset len as we want to read subIEs and not jump over them */
             PRINTF("frame802154e: entering MLME ie with len %u\n", nested_mlme_len);
             break;
+          case PAYLOAD_IE_SIXTOP:
+				    /* Now expect 'len' bytes of Sixtop sub-IEs */
+				    nested_mlme_len = len;
+				    PRINTF("frame802154e: entering Sixtop ie with len %u\n", nested_mlme_len);
+				    return 0;     /* Sixtop IE detected, return success(0) */
           case PAYLOAD_IE_LIST_TERMINATION:
             PRINTF("frame802154e: payload ie list termination %u\n", len);
             return (len == 0) ? buf + len - start : -1;
--- a/core/net/mac/frame802154e-ie.h
+++ b/core/net/mac/frame802154e-ie.h
@@ -45,6 +45,7 @@
 #include "net/mac/tsch/tsch-private.h"
 
 #define FRAME802154E_IE_MAX_LINKS       4
+#define SIXTOP_IE_MAX_LINKS           	3
 
 /* Structures used for the Slotframe and Links information element */
 struct tsch_slotframe_and_links_link {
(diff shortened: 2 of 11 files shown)


simonduq:
Sounds amazing! Will review whenever I find the time. One comment now: please check code style vs Contiki guidelines. Thanks :)

shalurajendran:
@simonduq  Updated the files and solved the build issues. Please let me know if any more modifications required.

core/net/mac/frame802154e-ie.c
```diff
@@ -529,6 +544,11 @@ frame802154e_parse_information_elements(const uint8_t *buf, uint8_t buf_size,
             len = 0; /* Reset len as we want to read subIEs and not jump over them */
             PRINTF("frame802154e: entering MLME ie with len %u\n", nested_mlme_len);
             break;
+          case PAYLOAD_IE_SIXTOP:
+				    /* Now expect 'len' bytes of Sixtop sub-IEs */
```
yatch: From here to line-551, tabs needs to be replaced with spaces.

core/net/mac/frame802154e-ie.h
```diff
@@ -45,6 +45,7 @@
 #include "net/mac/tsch/tsch-private.h"
 
 #define FRAME802154E_IE_MAX_LINKS       4
+#define SIXTOP_IE_MAX_LINKS           	3
```
yatch: It seems there is another tab in my editor...

yatch:
That's cool. I was about to look into 6P and SF0 when I saw your PR. :-)

I just noticed the mode bits (file permissions) of tsch.h have been changed somehow...
> core/net/mac/tsch/tsch.h 100644 → 100755

shalurajendran:
@yatch Thanks a lot for reviewing the code. Made the changes as suggested.

yatch:
Although I'm in the middle of reading your code, honestly, I think there is some room for improvement. I'll send a PR to your branch when it's ready....

shalurajendran:
@yatch Thanks for the reply. Kindly point out the areas for updation.

yatch:
@shalurajendran From the point of view of structure, it seems sixtop does too many things. It has a Scheduling Function in it as well as the 6top processing. It would be better for sixtop to do only the 6top processing and to have more primitive APIs rather than `sixtop_add_links()` and `sixtop_remove_link()`. The Scheduling Function part could be defined separately as a SF implementation using sixtop APIs.

In addition, from the coding aspect, I saw some redundant code which, I guess, is result of copy-and-past from the existing code. For example, an assignment for `nested_mlme_len` is not necessary in the following code block which is a part of `frame802154e_parse_information_elements()` in frame802154e-ie.c. If you want to print the length of a 6top IE, you can refer to `len` directly instead. Another specific comment on this code block is that this case block could do more useful things than just debug printing, by the way.
```C
          case PAYLOAD_IE_SIXTOP:
            /* Now expect 'len' bytes of Sixtop sub-IEs */
            nested_mlme_len = len;
            PRINTF("frame802154e: entering Sixtop ie with len %u\n", nested_mlme_len);
            return 0;     /* Sixtop IE detected, return success(0) */	
``` 

That's my two cents....

shalurajendran:
@yatch Thank you. Now I got it.

The current version includes only sixtop implementation. The scheduling function decides when to add/delete cells based on its algorithm, this is not implemented right now.

Second one, as you suggested we could use the variable len itself. A new variable is not required since further parsing is done at sixtop layer.

yatch:
@shalurajendran I've sent a PR to the branch, https://github.com/shalurajendran/contiki/pull/1. It's eventually become a huge change... The PR aims to untangle 6P and a (sample) schedule function which reside in sixtop.c.

I can see the schedule function there. One of obvious parts is the code starting at [L733 of sixtop.c](https://github.com/contiki-os/contiki/pull/1898/files#diff-b6b6548fa12e02e644ed8b0ee8354b8dR733). This block reads the CellList field values out of the buffer. However, the CellList field is defined to be opaque to 6P. In other words, 6P has no idea how to handle with the field. Here is an excerpt from draft-ietf-6tisch-6top-protocol-02.

```
CellList:  A list of 0, 1 or multiple 6P Cells.  The CellList is an
           opaque set of bytes, sent unmodified to the SF.  The
           RECOMMENDED format of each 6P Cell is defined in Section 4.2.5.
           The SF MAY redefine the format of the CellList field.
```

In my PR, your schedule function is implemented in sixtop-sf-simple.c. And, 6P message manipulation is done in sixtop-6p.c. 

Regarding of redundant code, you will find in the PR what I thought was redundant, but I can give you another example here. From [L133](https://github.com/contiki-os/contiki/pull/1898/files#diff-b6b6548fa12e02e644ed8b0ee8354b8dR133), the sixtop module builds an IEEE 802.15.4 frame to be sent. However, it doesn't necessarily have to do by itself. It can use APIs provided by the lower layer. In this case, it can use `send_packet()` of TSCH, instead.

By the way, you code helped me understand 6P deeply! Thanks a lot :-)

shalurajendran:
@yatch  Good work!!! Seems more change...

Need to go through in detail and will get back...

simonduq:
I left some comments at https://github.com/shalurajendran/contiki/pull/1
Please ping me when this is merged so I can do a full review of this PR!

yatch:
@simonduq Thanks! I'll check your comments soon!!

It's still at the "work-in-progress" stage. For example, it doesn't have the schedule generation management as you noticed. I'll push other changes made in the local repo if @shalurajendran  is happy for the structural changes. :-)

shalurajendran:
@simonduq Thanks for the inputs...

Surely will let you know once it is done.

shalurajendran:
@yatch Ofcourse, happy to merge your PR.;)

A few suggestions :
1.	For more readability, it would be better to use either “cell” or “link” in function and variable names. May be we could use “link”, since it is already been utilized for TSCH implementation. Similarly for the case of “delete” / ”remove”.

2.	The keywords listed below may be replaced as referred in the draft 6top-protocol-01.
	PAYLOAD_IE_IANA_IETF  	 ->  	IANA_IETF_IE_GROUP_ID 
	SIXTOP_6P_SUBIE_ID		->	IANA_6TOP_SUBIE_ID  
	SIXTOP_6P_VERSION		->	IANA_6TOP_6P_VERSION

4.	The add and delete functions in the structure of sixtop_sf_t may be referred as given in sixtop-sf-simple.c, to avoid confusion

5.	A small typo at line 506 of sixtop.c.

simonduq:
For context: link is used in IEEE 802.15.4-2015, while cell is used in IETF 6TiSCH. Not sure what is the best option in our case, maybe stick to link in TSCH and cell in 6TiSCH?

yatch:
@shalurajendran Thanks  for the comments.

With regard to link/cell, I'm in favor of using "cell" there, in 6TiSCH, and using "link" in TSCH as the latter one in @simonduq comment.

Another thing is IANA_ prefix. I guess this prefix will be gone when values for the constants are assigned by IANA. So, I'd keep them without "IANA_" prefix.

I'll make some updates.... hopefully soon...

shalurajendran:
Ok, we could fix it to be "cell" for sixtop files.

shalurajendran:
@yatch Please find the modified Copyright statement and update the files.

yatch:
@shalurajendran OK, thanks!

PR #1905, opened 2016-10-25
Fixed compiler warnings for CC2538-based platforms

nfi:


bthebaudeau:
Regarding the first commit, there is indeed an issue, but putting `#if`s everywhere is the wrong way of handling this. The root cause of these warnings is the way the LPM functions are disabled in `lpm.h`. See #1906.

Regarding the second commit, did you actually get warnings or anything wrong? This change does not seem to be required, maybe because `<string.h>` is already `#include`d somewhere in the other `#include`s, but I agree that the code is cleaner with it because of the calls to `memcpy()` and the like.

So if you drop your first commit, I'd be glad to merge this PR with only the second commit, and we'll use #1906.

nfi:
Nice way of solving the LPM issue in #1906. I have removed the first commit.

For the second remaining commit, I got warnings in another CC2538 based platform I am working on about memcpy. But it is not a big issue and it can be avoided by including ```string.h``` globally in the platform contiki-conf.h. So whatever you would prefer. It was the LPM issue I really wanted solved.

bthebaudeau:
:+1: Thanks. Waiting for Travis to pass, and you'll be all set.

PR #1906, opened 2016-10-25
cc2538: lpm: Fix build warnings with LPM_CONF_ENABLE == 0

bthebaudeau:
Some arguments passed to the functions disabled with LPM_CONF_ENABLE == 0 could trigger build warnings because they became unused with this configuration option. Disable these functions by using empty static inline functions instead of empty macros, so that the function arguments are always considered by the compiler as used, without having to #if-out code in many places.


bthebaudeau:
:+1:

PR #1918, opened 2016-11-02
Contiki websocket client

adamdunkels:
This PR adds support for the websocket protocol (RFC 6455).

The websocket protocol is a lightweight framing protocol that runs over HTTP or HTTPS that allows maintaining an active connection over an HTTP/HTTPS transaction. The protocol is primarily intended as a way to communicate between a web browser and a web server, but because it runs over the standard HTTP/HTTPS protocol, it can pass through most firewalls and even many HTTP/HTTPS proxies. Websocket server software exists for most server-side frameworks such as node.js.

The intended use case for this code is for a device behind a Contiki `ip64` NAT gateway to connect to a server on the Internet. It only works with IPv6 and if a client asks to connect to an IPv4 address, the code converts it to an IPv6 address to make it pass through the `ip64` gateway. The Thingsquare system uses this code to connect to mesh devices to a backend server in the cloud.

This PR adds:

* Websocket client code, with support for sending and receiving data, custom HTTP/HTTPS headers (e.g. for sending cookies), user-supplied websocket subprotocols, per-request configurable port numbers, multiple len bytes, `ping`/`pong` opcodes, `binary` and `text` message formats, connection to URLs with either a direct IP address or a hostname. The code works well even if the websocket frames are split across multiple TCP segments.
* Support for going through `CONNECT`-proxies.
* Pluggable HTTP/HTTPS clients, with the HTTP client included in this PR.
* Contiki client example code.
* Really tiny node.js example websocket server code example.
* Small bug fix to the `tcp-socket.c` code.

### DIFF
--- a/core/net/ipv6/websocket-http-client.c
+++ b/core/net/ipv6/websocket-http-client.c
@@ -1,0 +1,337 @@
+/*
+ * Copyright (c) 2014, Thingsquare, http://www.thingsquare.com/.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#include "websocket-http-client.h"
+#include "net/ip/uiplib.h"
+#include "net/ip/resolv.h"
+
+#include "ip64-addr.h"
+
+#include <stdio.h>
+#include <string.h>
+
+#define DEBUG DEBUG_NONE
+#include "net/ip/uip-debug.h"
+
+enum {
+  STATE_WAITING_FOR_HEADER,
+  STATE_WAITING_FOR_CONNECTED,
+  STATE_STEADY_STATE,
+};
+/*---------------------------------------------------------------------------*/
+static void
+send_get(struct websocket_http_client_state *s)
+{
+  struct tcp_socket *tcps;
+
+  tcps = &s->s;
+  tcp_socket_send_str(tcps, "GET ");
+  tcp_socket_send_str(tcps, s->file);
+  tcp_socket_send_str(tcps, " HTTP/1.1\r\n");
+  tcp_socket_send_str(tcps, "Host: ");
+  tcp_socket_send_str(tcps, s->host);
+  tcp_socket_send_str(tcps, "\r\n");
+  if(strlen(s->header) > 0) {
+    tcp_socket_send_str(tcps, s->header);
+  }
+  tcp_socket_send_str(tcps,
+                      "Connection: Upgrade\r\n"
+                      "Upgrade: websocket\r\n"
+                      "Sec-WebSocket-Key: x3JJHMbDL1EzLkh9GBhXDw==\r\n"
+                      "Sec-WebSocket-Version: 13\r\n"
+                      "Sec-WebSocket-Protocol:");
+  tcp_socket_send_str(tcps, s->subprotocol);
+  tcp_socket_send_str(tcps, "\r\n");
+  tcp_socket_send_str(tcps, "\r\n");
+  PRINTF("websocket-http-client: send_get(): output buffer left %d\n", tcp_socket_max_sendlen(tcps));
+}
+/*---------------------------------------------------------------------------*/
+static void
+send_connect(struct websocket_http_client_state *s)
+{
+  struct tcp_socket *tcps;
+  char buf[20];
+
+  tcps = &s->s;
+  tcp_socket_send_str(tcps, "CONNECT ");
+  tcp_socket_send_str(tcps, s->host);
+  tcp_socket_send_str(tcps, ":");
+  sprintf(buf, "%d", s->port);
+  tcp_socket_send_str(tcps, buf);
+  tcp_socket_send_str(tcps, " HTTP/1.1\r\n");
+  tcp_socket_send_str(tcps, "Host: ");
+  tcp_socket_send_str(tcps, s->host);
+  tcp_socket_send_str(tcps, "\r\n");
+  tcp_socket_send_str(tcps, "Proxy-Connection: Keep-Alive\r\n\r\n");
+}
+/*---------------------------------------------------------------------------*/
+static void
+event(struct tcp_socket *tcps, void *ptr,
+      tcp_socket_event_t e)
+{
+  struct websocket_http_client_state *s = ptr;
+
+  if(e == TCP_SOCKET_CONNECTED) {
+    if(s->proxy_port != 0) {
+      send_connect(s);
+    } else {
+      send_get(s);
+    }
+  } else if(e == TCP_SOCKET_CLOSED) {
+    websocket_http_client_closed(s);
+  } else if(e == TCP_SOCKET_TIMEDOUT) {
+    websocket_http_client_timedout(s);
+  } else if(e == TCP_SOCKET_ABORTED) {
+    websocket_http_client_aborted(s);
+  } else if(e == TCP_SOCKET_DATA_SENT) {
+  }
+}
+/*---------------------------------------------------------------------------*/
+static int
+parse_header_byte(struct websocket_http_client_state *s,
+                  uint8_t b)
+{
+  static const char *endmarker = "\r\n\r\n";
+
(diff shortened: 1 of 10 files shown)


core/net/ipv6/websocket-http-client.c
```diff
@@ -0,0 +1,337 @@
...
+  tcp_socket_send_str(tcps, "Host: ");
+  tcp_socket_send_str(tcps, s->host);
+  tcp_socket_send_str(tcps, "\r\n");
+  if(strlen(s->header) > 0) {
+    tcp_socket_send_str(tcps, s->header);
+  }
+  tcp_socket_send_str(tcps,
+                      "Connection: Upgrade\r\n"
+                      "Upgrade: websocket\r\n"
+                      "Sec-WebSocket-Key: x3JJHMbDL1EzLkh9GBhXDw==\r\n"
```
alignan: Where does this key comes from? if this is something configurable perhaps it would be better to have defined as a string at `project-conf.h`

core/net/ipv6/websocket-http-client.c
```diff
@@ -0,0 +1,337 @@
...
+  tcp_socket_send_str(tcps, "\r\n");
+  tcp_socket_send_str(tcps, "\r\n");
+  PRINTF("websocket-http-client: send_get(): output buffer left %d\n", tcp_socket_max_sendlen(tcps));
+}
+/*---------------------------------------------------------------------------*/
+static void
+send_connect(struct websocket_http_client_state *s)
+{
+  struct tcp_socket *tcps;
+  char buf[20];
```
alignan: The `s->port` takes up to 20 bytes? perhaps we could reduce the `buf` length

core/net/ipv6/websocket-http-client.c
```diff
@@ -0,0 +1,337 @@
...
+    } else {
+      send_get(s);
+    }
+  } else if(e == TCP_SOCKET_CLOSED) {
+    websocket_http_client_closed(s);
+  } else if(e == TCP_SOCKET_TIMEDOUT) {
+    websocket_http_client_timedout(s);
+  } else if(e == TCP_SOCKET_ABORTED) {
+    websocket_http_client_aborted(s);
+  } else if(e == TCP_SOCKET_DATA_SENT) {
```
alignan: This check is unnecessary, for code readiness perhaps  adding just as a comment would be just fine

core/net/ipv6/websocket-http-client.c
```diff
@@ -0,0 +1,337 @@
...
+
+  /* Skip the first part of the HTTP response */
+  while(b != ' ') {
+    PT_YIELD(&s->parse_header_pt);
+  }
+
+  /* Skip the space that follow the first part */
+  PT_YIELD(&s->parse_header_pt);
+
+  /* Read the first three bytes that consistute the HTTP status code. */
```
alignan: `constitute`

core/net/ipv6/websocket-http-client.c
```diff
@@ -0,0 +1,337 @@
...
+  s->http_status = (b - '0');
+  PT_YIELD(&s->parse_header_pt);
+  s->http_status = s->http_status * 10 + (b - '0');
```
alignan: What does this exactly does? adding a comment here would be better for readiness

core/net/ipv6/websocket-http-client.c
```diff
@@ -0,0 +1,337 @@
...
+    /* This is a websocket request, so the server should have answered
+       with a 101 Switching protocols response. */
+    PRINTF("Websocket HTTP client didn't get the 101 status code (got %d), closing connection\n",
+           s->http_status);
+    websocket_http_client_close(s);
+    while(1) {
+      PT_YIELD(&s->parse_header_pt);
+    }
+  }
+
```
alignan: Extra space

core/net/ipv6/websocket-http-client.c
```diff
@@ -0,0 +1,337 @@
...
+websocket_http_client_get(struct websocket_http_client_state *s)
+{
+  uip_ip4addr_t ip4addr;
+  uip_ip6addr_t ip6addr;
+  uip_ip6addr_t *addr;
+  uint16_t port;
+
+  PRINTF("websocket_http_client_get: connecting to %s with file %s subprotocol %s header %s\n",
+         s->host, s->file, s->subprotocol, s->header);
+
```
alignan: Extra space

core/net/ipv6/websocket.c
```diff
@@ -0,0 +1,719 @@
...
+    memcpy(&buf[6], data, datalen);
+    return websocket_http_client_send(&s->s, buf, 6 + datalen);
+  }
+  return -1;
+}
+/*---------------------------------------------------------------------------*/
+int
+websocket_send_str(struct websocket *s, const char *str)
+{
+  //  PRINTF("websocket_send_str %s\n", str);
```
alignan: Replace with `/* */`

core/net/ipv6/websocket.c
```diff
@@ -0,0 +1,719 @@
...
+{
+  websocket_http_client_close(&s->s);
+  s->state = WEBSOCKET_STATE_CLOSED;
+}
+/*---------------------------------------------------------------------------*/
+static int
+send_data(struct websocket *s, const void *data,
+          uint16_t datalen, uint8_t data_type_opcode)
+{
+  uint8_t buf[WEBSOCKET_MAX_MSGLEN + 4 + 4];
```
alignan: Where does the `buf` lenght comes from? perhaps adding a comment here explaining the `4 +4`

core/net/ipv6/websocket.c
```diff
@@ -0,0 +1,719 @@
...
+/*---------------------------------------------------------------------------*/
+int
+websocket_ping(struct websocket *s)
+{
+  uint8_t buf[sizeof(struct websocket_frame_hdr) +
+              sizeof(struct websocket_frame_mask)];
+  struct websocket_frame_hdr *hdr;
+  struct websocket_frame_mask *mask;
+
+  if(2 + 4 > websocket_http_client_sendbuflen(&s->s)) {
```
alignan: Here as well, add a comment about the `2 + 4`

core/net/ipv6/websocket.h
```diff
@@ -0,0 +1,115 @@
...
+  WEBSOCKET_DATA = 5,
+  WEBSOCKET_RESET = 6,
+  WEBSOCKET_TIMEDOUT = 7,
+  WEBSOCKET_CLOSED = 8,
+  WEBSOCKET_PINGED = 9,
+  WEBSOCKET_DATA_RECEIVED = 10,
+  WEBSOCKET_PONG_RECEIVED = 11,
+} websocket_result_t;
+
+
```
alignan: Extra space here

core/net/ipv6/websocket.h
```diff
@@ -0,0 +1,115 @@
...
+enum {
+  WEBSOCKET_STATE_CLOSED = 0,
+  WEBSOCKET_STATE_DNS_REQUEST_SENT = 1,
+  WEBSOCKET_STATE_HTTP_REQUEST_SENT = 2,
+  WEBSOCKET_STATE_WAITING_FOR_HEADER = 3,
+  WEBSOCKET_STATE_RECEIVING_HEADER = 4,
+  WEBSOCKET_STATE_HEADER_RECEIVED = 5,
+  WEBSOCKET_STATE_RECEIVING_DATA = 6,
+};
+
```
alignan: Extra space

examples/websockets/websocket-example.c
```diff
@@ -0,0 +1,63 @@
+#include "contiki.h"
```
alignan: Missing the license header here :smile:

core/net/ipv6/websocket-http-client.c
adamdunkels: The maximum size of a 32-bit port number is 10, and with one null byte at the end it makes 11 characters. So 20 is just to be on the safe side. No need to optimize 9 bytes on the stack imo, particularly since stack arrays are always problematic. Better to be on the safe side.

adamdunkels:
@alignan : pushed a bunch of fixes to your comments - didn't squash the commits, to make it easier to see them as separate commits, but will rebase once reviewed and ready for merge.

joakimeriksson:
quick question - is there a reason not to put the websocket code as an app - similar to er-coap and other higher layer protocols?

adamdunkels:
Not really, except that the websocket protocol isn't really meant to be used on its own, unlike coap and friends. It always needs something running on top of it. So in that sense it is almost a transport protocol, so makes sense to put it next to the other transport protocols. But there is no really strong reason either way.

joakimeriksson:
Yes, true. But even CoAP needs something on top - like LWM2M or some kind of JSON formats - but I agree - web socket is very similar to plain TCP in a sense - just better att sneaking through firewalls.

alignan:
:+1:

core/net/ipv6/websocket-http-client.c
```diff
@@ -0,0 +1,347 @@
...
+     hardcoded value here. */
+  tcp_socket_send_str(tcps,
+                      "Connection: Upgrade\r\n"
+                      "Upgrade: websocket\r\n"
+                      "Sec-WebSocket-Key: x3JJHMbDL1EzLkh9GBhXDw==\r\n"
+                      "Sec-WebSocket-Version: 13\r\n"
+                      "Sec-WebSocket-Protocol:");
+  tcp_socket_send_str(tcps, s->subprotocol);
+  tcp_socket_send_str(tcps, "\r\n");
+  tcp_socket_send_str(tcps, "\r\n");
```
greg-king5: For efficiency and consistency with other calls to `tcp_socket_send_str()`, those two lines should be a single function-call (with a four-character string).

core/net/ipv6/websocket-http-client.c
```diff
@@ -0,0 +1,347 @@
...
+    /* We could feed this information up to the websocket.c layer, but
+       we currently do not do that. */
+  }
+}
+/*---------------------------------------------------------------------------*/
+static int
+parse_header_byte(struct websocket_http_client_state *s,
+                  uint8_t b)
+{
+  static const char *endmarker = "\r\n\r\n";
```
greg-king5: `endmarker` is a constant pattern; therefore, it should be defined as
```c
static const uint8_t endmarker[4] = "\r\n\r\n";
```
And, the `while` loop should test against `sizeof(endmarker)`.

core/net/ipv6/websocket-http-client.c
```diff
@@ -0,0 +1,347 @@
...
+  if(subprotocol == NULL) {
+    return -1;
+  }
+  strncpy(s->subprotocol, subprotocol, sizeof(s->subprotocol));
+
+  if(header == NULL) {
+    strncpy(s->header, "", sizeof(s->header));
+  } else {
+    strncpy(s->header, header, sizeof(s->header));
+  }
```
greg-king5: I would like to see that test refactored into:
```C
strncpy(s->header, header == NULL ? "" : header, sizeof(s->header));
```

alignan:
ping?

joakimeriksson:
I would suggest using the 30-days single merger rule here - to get the code in even if these fairly minor issues are not fixed. Otherwise it might be stuck for a while which to me would be sad. I would like to make use of the code myself ;-). From what I can see it will not break any already existing functionality.

alignan:
My requests were addressed so I'm happy as it is, but I was waiting for comments on https://github.com/contiki-os/contiki/pull/1918#pullrequestreview-7123986, but yeah let's set a timeout and move this further.

g-oikonomou:
We want to be merging this one folks, I have not looked at it closely but it strikes me as though most comments have been addressed @alignan ?

alignan:
My comments were addressed, yes... but curious about https://github.com/contiki-os/contiki/pull/1918#pullrequestreview-7123986

g-oikonomou:
I think it may have merit, but I think this is too major a feature to keep on hold till it gets addressed, plus it can be addressed by others. I say 👍

alignan:
Ok, agree... This could be addressed in a future PR :+1:

PR #1920, opened 2016-11-03
zoul/remote-revb: initialize unused pins to minimize current draw

alignan:
This is a **work in progress** to reduce the current draw from pins not used, please feel free to drop any comment or suggestion!

Testing with the `examples/zolertia/zoul/zoul-demo` with the following changes:

- `NETSTACK_MAC.off(1);` in [L128](https://github.com/contiki-os/contiki/blob/master/examples/zolertia/zoul/zoul-demo.c#L128) commented out
- `PROCESS_YIELD();` in [L145](https://github.com/contiki-os/contiki/blob/master/examples/zolertia/zoul/zoul-demo.c#L145) changed to `PROCESS_WAIT_EVENT_UNTIL(0)`
- `NullRDC` commented out in [project-conf.h](https://github.com/contiki-os/contiki/blob/master/examples/zolertia/zoul/project-conf.h#L44)

So far in PM2 the current drops from 2.73mA to 0.672mA, but some work is still required.  There is around 500uA consumed by the XTAL oscillator (32KHz), as gatting the pads reduces the current consumption to 50uA.  Dropping to PM3 yields around 32uA.

The values are measured on the battery input of the RE-Mote.

### DIFF
--- a/platform/zoul/remote-revb/board.c
+++ b/platform/zoul/remote-revb/board.c
@@ -39,13 +39,111 @@
 /*---------------------------------------------------------------------------*/
 #include "contiki-conf.h"
 #include "antenna-sw.h"
+#include "dev/ioc.h"
 #include <stdint.h>
 #include <string.h>
 /*---------------------------------------------------------------------------*/
+#define CLEAR(x, y, z) GPIO_SOFTWARE_CONTROL(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y)); \
+                       ioc_set_over(x, y, z);     \
+                       GPIO_SET_OUTPUT(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y));  \
+                       GPIO_CLR_PIN(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y));
+
+#define SET(x, y, z)   GPIO_SOFTWARE_CONTROL(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y)); \
+                       ioc_set_over(x, y, z);     \
+                       GPIO_SET_OUTPUT(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y));  \
+                       GPIO_SET_PIN(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y));
+
+#define INPUT(x, y, z) GPIO_SOFTWARE_CONTROL(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y)); \
+                       GPIO_SET_INPUT(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y));
+/*---------------------------------------------------------------------------*/
+/* The following functions configure the pins to minimize current draw, modules
+ * and libraries using these pins are in charge of configuring as corresponding
+ */
 static void
 configure_unused_pins(void)
 {
-  /* FIXME */
+  /* Port A pins */
+#if UART_CONF_ENABLE == 0
+  INPUT(GPIO_A_NUM, 0x00, IOC_OVERRIDE_DIS);
+  INPUT(GPIO_A_NUM, 0x01, IOC_OVERRIDE_DIS);
+#endif
+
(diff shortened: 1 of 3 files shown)


alignan:
note to self: Scheduling a `rtimer` spikes the current consumption to almost 8mA

bthebaudeau:
Yes, a tickless implementation would help (as previously discussed).

platform/zoul/remote-revb/board.c
```diff
@@ -39,13 +39,111 @@
...
 #include "contiki-conf.h"
 #include "antenna-sw.h"
+#include "dev/ioc.h"
 #include <stdint.h>
 #include <string.h>
 /*---------------------------------------------------------------------------*/
+#define CLEAR(x, y, z) GPIO_SOFTWARE_CONTROL(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y)); \
+                       ioc_set_over(x, y, z);     \
+                       GPIO_SET_OUTPUT(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y));  \
+                       GPIO_CLR_PIN(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y));
```
bthebaudeau: `do`/`while(0)` or a GCC compound statement `({})` would be better.

platform/zoul/remote-revb/board.c
```diff
@@ -39,13 +39,111 @@
...
+
+#define SET(x, y, z)   GPIO_SOFTWARE_CONTROL(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y)); \
+                       ioc_set_over(x, y, z);     \
+                       GPIO_SET_OUTPUT(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y));  \
+                       GPIO_SET_PIN(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y));
+
+#define INPUT(x, y, z) GPIO_SOFTWARE_CONTROL(GPIO_PORT_TO_BASE(x), GPIO_PIN_MASK(y)); \
```
bthebaudeau: `z` is unused.

bthebaudeau: I mean that `z` should be used with `ioc_set_over()`, not that it should be removed.

platform/zoul/remote-revb/board.c
```diff
@@ -39,13 +39,111 @@
...
+  CLEAR(GPIO_D_NUM, 0x00, IOC_OVERRIDE_DIS);
+  CLEAR(GPIO_D_NUM, 0x01, IOC_OVERRIDE_DIS);
+
+#if ANTENNA_SW_SELECT_DEFAULT == ANTENNA_SW_SELECT_2_4GHZ
+  CLEAR(GPIO_D_NUM, 0x02, IOC_OVERRIDE_DIS);
+#endif
+
+  INPUT(GPIO_D_NUM, 0x03, IOC_OVERRIDE_DIS); /* RTCC interrupt */
+  CLEAR(GPIO_D_NUM, 0x04, IOC_OVERRIDE_DIS); /* LED */
+  CLEAR(GPIO_D_NUM, 0x05, IOC_OVERRIDE_DIS);
```
bthebaudeau: `IOC_OVERRIDE_DIS` is fine for the outputs, but it's not for the inputs (besides the issue with the last parameter of `INPUT()`), unless the pin is always driven externally, whether it is by the output of another chip or by an external pull resistor, etc.

Except for PC0-3, all the pins are set as GPIO inputs with internal pull-up resistors by the CC2538 after reset, which is already optimal for power consumption, unless there are external components forcing a low level on the pin or enabled by a high level on an input of theirs.

Also, all the components of the board could be switched off by default, and only enabled when needed (dynamically), but this sometimes lengthens the access times because of an additional wake-up time.

The sleep modes could be better used.

The µDMA and interrupts should be used instead of polling where possible.

The system frequencies could be tuned according to the application requirements, even dynamically.

alignan: Thanks for the pointers!, I will fix the `INPUT` macro left-over

PR #1924, opened 2016-11-06
Update to latest CC13xxware/CC26xxware

g-oikonomou:
This pull updates the CC13xx/CC26xx ports to use the latest versions of TI's CC13xxware and CC26xxware respectively.

Recent changes in both xxwares are exclusively internal and there are no API changes. Therefore it appears we don't need to make any modifications to Contiki's sources.

Can I ask the community to please help test this pull?

Cheers


arurke:
Tested it for ~24h on one CC1310 EM node. Transmitting a frame directly to sink every sec - my metrics are basically lost frames and unknown restarts. Both are comparable to our other nodes running earlier cc13xxware.
(RSSI on received from sink seems to be the same as well (ref. the 10dB problem))

g-oikonomou:
Tested with:

* CC2650 STK
* SmarfRF + CC2650EM
* CC1310LP
* Something else that I'm not telling you about :)

From my PoV this is ready for consideration and merging.

alignan:
:+1:

PR #1934, opened 2016-11-14
RPL: Basic security features implementation

arenantonio92:
Implementation of basic security features over RPL. The current implementation supports only secure preinstalled mode. Example code (with a README) is provided examples/ivp6/rpl-sec


simonduq:
Thanks for the contribution!
Before we can take a closer look, please make sure to pass all Travis tests, to cleanup the diff (e.g. the changes to the `rpl-udp` don't seem intended). Also, as this is an new feature of an important core module (RPL), I will ask for an extra Travis test that ensured non-regression on the new feature. Please ping us when this is done so we can provide detailed feedback.

arenantonio92:
Hi! I have done most of what you ask (cleanup diffs, pass all regression tests). I have one question about the extra Travis test. 
Does it have to be a regression test on the specific implementation of the security features?

simonduq:
Thanks! Great that you got Travis to pass. Yes, I meant a test of the new security feature. But before you spend time on this, it may be a good to get a feel what the PR exactly adds:
* what is implemented exactly?
* what does this bring compared to using link-layer security?

core/lib/ccm-star.c
```diff
@@ -172,6 +173,7 @@ aead(const uint8_t* nonce,
 /*---------------------------------------------------------------------------*/
 const struct ccm_star_driver ccm_star_driver = {
   set_key,
-  aead
+  aead,
+  mic
```
simonduq: cannot get the mic using aead?
besides, if you update this interfaces, all other implementations of it must also be updated (I'm surprised Travis did not catch this)

core/net/rpl/rpl-dag.c
```diff
@@ -140,6 +140,8 @@ rpl_dag_init(void)
 {
   nbr_table_register(rpl_parents, (nbr_table_callback *)nbr_callback);
 }
+
+
```
simonduq: no needed changes in this file

simonduq:
please also note the code style is a bit off, at least in `rpl-icmp.c`

arenantonio92:
This PR implements part of RFC 6550 Section 10:

- Authenticity, Integrity and (optionally) confidentiality are implemented over RPL messages: RPL messages are enlarged with a MAC at the end of the packet and optionally are encrypted using CCM* library. 
- The MAC covers the entire IPv6 packet, i.e. IP header, ICMP header, Security Section of RPL messages and the RPL payload.
- The RPL secure message use different ICMP6 type codes
- The messages are protected by using a 128 bit key shared between all the network nodes, and different security levels are implemented

- While link-layer security protects every type of packets, this implementation is compliant to RFC 6550 and only RPL messages are protected.

I changed the CCM-star interface because RPL optionally encrypts, it can only add a MAC for integrity and authentication. The mic function in the interface was already implemented (aead itself uses it), but it was private; with this change the function is now public and I can use it in this implementation avoiding useless encryption when not needed. That's why tests are still passed, I think. 

About code style, I actually forgot to use the beautifier on `rpl-icmp.c`. I will update soon to fix this.

simonduq:
Right, interesting. If you're familiar with IPsec, I'm curious, what are the differences in terms of protection with IPsec AH?

As for adea, 2 things:
(1) right, but other implementations of ccm-star (in cc2538 and jn516x) also need to implement your interface now
(2) if I'm not mistaken aead can also compute mic without encrypting any payload (look at the a and b parameters)

arenantonio92:
Speaking about the RPL Security vs Link layer security there's another important difference:
Link layer security mechanisms are link layer specific, RPL security is not.
Example scenario:

- In a DODAG two different link layer are used (like Bluetooth and 802.15.4). 

- RPL security mechanism protects its messages despite the link layer. Otherwise, i have to set two different link layer security mechanisms.

simonduq:
Yes, that is a good point. How about IPsec AH?

arenantonio92:
The suggestion about CCM* star library was really good, I didn't think on that possibility so I make that  function public, I will revert the changes on that library.

Speaking about IpSec, I'm not very familiar with it, but RPLSec protects only RPL messages, in order to protect the routing protocol state. 
IpSec protects every IP packet. If we don't need it, in order to protect RPL state we suffer of a greater computational cost.

simonduq:
Right. I need to take a closer look at RPLsec and will then review this PR at some point.
Thanks!

mcr:
IPsec AH is largely dead.  IPsec AH made a fundemental mistake: if the SPI is unknown then the entire packet is discarded, rather than continuing to process as if there was no security. That makes it hard to implement listening for DIOs, but not authenticating them until you have the key.  Plus RPL really needs asymmetric methods to do security well. (and 6550 security team didn't know IPsec... le sigh).
As for differences against l2 security: it crosses link types, which might matter if two nodes have two paths, across two link types.  Also while symmetric methods can't work with A=1 to protect node identity (you need assymetric methods), the symmetric methods do keep non-router (host) nodes from impersonating routers. (this is not important at this point in contiki and RPL evolution, but it will matter).

core/net/rpl/rpl-icmp6.c
```diff
@@ -996,23 +1444,97 @@ dao_input_nonstoring(void)
...
+  lvl = (buffer[pos++] & RPL_LVL_MASK);
+  ++pos;  /* Flags */
+
+/*  Discard packets with time stamp instead of simple counters,
+ *  Only accept packets with pre installed key
+ *  Discard unassigned security levels
+ */
+
+  if((timestamp == 1) || (kim != 0) || (lvl > 3)) {
+    goto discard;
```
mcr: would be nice to have some counters here, perhaps one for each reason to discard.

core/net/rpl/rpl-icmp6.c
```diff
@@ -1021,12 +1543,20 @@ dao_input(void)
   }
 
   if(RPL_IS_STORING(instance)) {
-    dao_input_storing();
```
mcr: Stupid question, if RPL_SECURITY is defined (compiled in), is there a mode where I don't set a key and operate insecurely?  Can I do it differently on different interfaces?

examples/ipv6/rpl-sec/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+	#undef RPL_CONF_SECURITY
+		
+	#define RPL_CONF_SECURITY 1
+
+You can also set the preinstalled key (128 bit, CCM* uses AES-128) in your project-conf.h with
+
+	#undef RPL_CONF_SECURITY_K
+
+	#define RPL_CONF_SECURITY_K { 0xF7, 0x8E, 0xBA, 0xC9, 0xED, 0xE3, 0xE6, 0x68, 0xDF, 0x15, 0x72, 0xE0, 0x78, 0xB7, 0x80, 0x9E }
```
mcr: Could this be moved to a single (additional) file so that we can innovate properly: the key should be stored in some platform specific eeprom outside of the code?  This kind of macro will result in another IoT disaster of constant keys.

simonduq:
Thanks for the input @mcr !

arenantonio92:
Hi! Thanks for you answer. The basic features about security section in rfc 6550 are quite clear, yes. 
By the way, I'm working on an improvement of this implementation, where a standard-compliant replay protection mechanism is deployed. In this work I had more doubts, that are not clarified in the rfc. 
One of this is about this sentence:

>  If the received message counter
> value is non-zero and less than the maintained incoming counter
> watermark, a potential packet replay is indicated and the node MUST
> discard the incoming packet. 

Why a received counter equal to the watermark is accepted?

core/net/rpl/rpl-icmp6.c
arenantonio92: Yes, you're right. 
I think the best choice is to use the rpl_stat structure, already implemented in Contiki, and increment the malformed_msgs counter. 
I'll update code as soon as I can.

examples/ipv6/rpl-sec/README.md
arenantonio92: Yes, it is possible to do this.
However in this first implementation the scheme of TiSCH is used.

core/net/rpl/rpl-icmp6.c
arenantonio92: If I understood your first question, you're asking if I can enable security in RPL with_RPL_SECURITY_ macro without setting the preinstalled key through_RPL_CONF_SECURITY_K_ macro. The answer is yes. If someone forgets to set the key, a default key is set in `rpl-conf.h`, to prevent possible uncorrect behaviours.

Speaking about the second question, I think Contiki assumes that nodes have only one network interface, but I might be wrong. Anyway, in this implementation if security is enabled, nodes drop non-recurring packets.

mcr:
arenantonio92 <notifications@github.com> wrote:
    > If I understood your first question, you're asking if I can enable
    > security in RPL with_RPL_SECURITY_ macro without setting the
    > preinstalled key through_RPL_CONF_SECURITY_K_ macro. The answer is
    > yes. If someone forgets to set the key, a default key is set in
    > rpl-conf.h, to prevent possible uncorrect behaviours.

No, I mean: Can I compile the security in, but not enable it until the
key is set *at runtime*.

    > Speaking about the second question, I think Contiki assumes that nodes
    > have only one network interface, but I might be wrong. Anyway, in this
    > implementation if security is enabled, nodes drop non-recurring
    > packets.

Contiki nodes regularly have two interfaces: the second one being a SLIP
interface to a host, or an ethernet interface.   I have run RPL with
two interfaces as well.

If you don't have two interfaces, then you can't run RPL on different media,
so one of the points of RPL security goes away!

--

arenantonio92:
> No, I mean: Can I compile the security in, but not enable it until the
> key is set *at runtime*.

Not in this version.
We are planning to develop a version which allows for authenticated mode (A=1). In this mode, setting and enabling a key at runtime is needed, so we will probably develop this feature.

arenantonio92:
A light replay protection mechanism has been added to the project. The mechanism is very similar to the 802.15.4 replay protection one. In particular:

1- The first message of a new neighbor is always processed, and a local copy of the current value of the incoming counter is stored. 
2- After this first message, the next ones are checked on counter value, i.e. packets are processed iff the incoming counter is greater then the stored local copy. 
3- Packets with incoming counter equal to 0xffff are always discarded

Also, CC messages are implemented. 
- In case of reboot, a node restarts with its secure counter equal to zero. If neighbor nodes have a local copy for the counter of rebooted node, they send a CC response with the value of their local copy. In this way the rebooted node can resync its counter the latest value used.
- In case of interOS interactions, if a CC request is received, nodes send a CC response as defined in RFC6550.

PR #1935, opened 2016-11-14
ContikiMac: Schedule next cycle after last CCA check

arthurfabre:
After the last CCA check, ContikiMac schedules another wake up and
yields.
When it wakes up again, it just schedules the next cycle check.

This wastes power and time, so once we are done with the last CCA check,
we immediately schedule the next cycle check and yield.
This saves us from performing an extra yield / sleep / wakeup cycle.

On a power meter, the 3rd spike seen after the first 2 CCA checks
disappears. On a muntjac (remote-like platform), this reduces our average power consumption by
~0.4mA.

### DIFF
--- a/core/net/mac/contikimac/contikimac.c
+++ b/core/net/mac/contikimac/contikimac.c
@@ -429,6 +429,13 @@ powercycle(struct rtimer *t, void *ptr)
         }
         powercycle_turn_radio_off();
       }
+
+      // If this is the last CCA check, avoid yielding and go straight to scheduling
+      // our next cycle
+      if (count == CCA_COUNT_MAX - 1) {
+          break;
+      }
+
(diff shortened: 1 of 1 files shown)


arthurfabre:
Any update on this? The Travis checks failed due to a `org.contikios.cooja.Cooja$SimulationCreationException: Unknown error: javax.swing.plaf.BorderUIResource cannot be cast to java.awt.Color` which I doubt was caused by this change.

niklas88:
Seeing as ContikiMAC-Async seems to have been abandoned I'm thinking of switching back to mainline ContikiMAC and would be interested in this change as well. It seems to me there is a bit of interest in very low power operations and consequently reviews for ContikiMAC currently so I think such small step changes would be especially welcome. Let me know if additional testing would help bring things along.

PR #1469, opened 2016-01-22
Add support for nRF52 DK platform

wbober:
This PR adds support for nRF52 SOC and nRF52 Development Kit from Nordic Semiconductor.

The nRF52 is a Bluetooth SoC which supports IPv6 over Bluetooth Low Energy. Therefore this port demonstrates how Contiki can be used for IPv6 over BLE communication.

The PR adds support for nRF52832 SoC and nRF52 DK. Examples and docummentation that demonstrate how to use the DK are also provided.

The following features have been implemented:
* Support for IPv6 over BLE using Contiki 6LoWPAN implementation
* Contiki system clock and rtimers (using 32kHz and 1MHz timers)
* UART driver
* Watchdog driver
* Hardware RNG
* Temperature sensor driver
* DK LED driver
* DK Buttons driver
* Segger Real Time Transfer (RTT) I/O support

This port is officially supported by Nordic Semiconductor. 

This PR depends on two eariler PR, which were already reviewed:
https://github.com/contiki-os/contiki/pull/1428
https://github.com/contiki-os/contiki/pull/1429

PR #1936, opened 2016-11-15
Add IPv6 over BLE support for SensorTag CC26xx platform

spoerk:
This pull request adds IPv6 over BLE support according to the [RFC 7886](https://tools.ietf.org/html/rfc7668) standard for the Texas Instruments SensorTag CC2650.

This PR adds a BLE implementation of the radio layer for the SensorTag CC2650.
A hardware-independent MAC layer implementation for IPv6 over BLE that implements packet fragmentation/reassembly is also included.

### DIFF
--- a/core/net/ipv6/uip-nd6.c
+++ b/core/net/ipv6/uip-nd6.c
@@ -517,10 +517,16 @@ na_input(void)
       if(nd6_opt_llao == NULL || !extract_lladdr_from_llao_aligned(&lladdr_aligned)) {
         goto discard;
       }
+
+#if NETSTACK_RADIO != ble_mode
+      /**
+       * The neighbor table is not needed in ble mode
+       */
       if(nbr_table_update_lladdr((const linkaddr_t *)lladdr, (const linkaddr_t *)&lladdr_aligned, 1) == 0) {
         /* failed to update the lladdr */
         goto discard;
       }
+#endif
 
       if(is_solicited) {
         nbr->state = NBR_REACHABLE;
--- a/cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
+++ b/cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
@@ -61,7 +61,7 @@ endif
 CLEAN += symbols.c symbols.h *.d *.elf *.hex
 
 ### CPU-dependent directories
-CONTIKI_CPU_DIRS = . dev rf-core rf-core/api $(TI_XXWARE_STARTUP_DIR)
+CONTIKI_CPU_DIRS = . dev rf-core rf-core/api rf-core/ble-hal $(TI_XXWARE_STARTUP_DIR) net
 
 ### Use the existing debug I/O in cpu/arm/common
 CONTIKI_CPU_DIRS += ../arm/common/dbg-io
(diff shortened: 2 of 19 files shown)


adamdunkels:
This looks cool! Can this be used to make a smartphone act as a bridge into an IPv6 / 802.15.4 network, or is it intended only as a point-to-point connection to a single cc26xx device? The example seems to indicate that it needs a special hardware solution on the other side (a Raspberry Pi with an nRF dongle?).

core/net/ipv6/uip-nd6.c
```diff
@@ -517,10 +517,16 @@ na_input(void)
       if(nd6_opt_llao == NULL || !extract_lladdr_from_llao_aligned(&lladdr_aligned)) {
         goto discard;
       }
+
+#if NETSTACK_RADIO != ble_mode
```
adamdunkels: This change should not be in a core file like this. If the neighbor table is not needed, it would be better to add a generic API for turning off the neighbor table functionality, which would be called from one of the BLE driver files.

Also, `NETSTACK_RADIO` probably is defined to be `ble_mode_driver`, and not `ble_mode`, so this particular `#if` would seem to never trigger anyway. So maybe it isn't needed?

spoerk:
This IPv6 over BLE stack works with every IPv6 over BLE border router that is compliant to the RFC 7668 standard. 

I only included the Raspberry Pi BR with BLE/USB-Dongle because the same border router was used for https://github.com/contiki-os/contiki/pull/1469, and because I used it to test my IPv6 over BLE stack.

Therefore, it is absolutely possible to use a smartphone as a bridge into another IPv6 network.

core/net/ipv6/uip-nd6.c
```diff
@@ -518,7 +518,7 @@ na_input(void)
         goto discard;
       }
 
-#if NETSTACK_RADIO != ble_mode
+#if NETSTACK_RADIO != ble_mode_driver
```
spoerk: I fixed the typo in the #if check.

I agree that this #if should not be in the official Contiki repo. Does anyone have any suggestions or recommendations on how to create such a generic API to disable the neighbor table functionality?

atiselsts: That's an interesting pull request!

I don't think C preprocessor can see symbolic names, so I suspect this check cannot be done this way. For an example, try compiling this code - it will print "equals":
```
int abc, def;
#define ABC abc
#if ABC == def
#warning equals
#endif
```

g-oikonomou:
Oh wow, I'll definitely look at the platform side of this one!

core/net/ipv6/uip-nd6.c
greg-king5: > I don't think the C preprocessor can see symbolic names; so, I suspect that check cannot be done that way. For an example, try compiling this code -- it will print "equals":
```C
int abc, def;
#define ABC abc
#if ABC == def
#warning equals
#endif
```
`#if` sees macro names.  And, arithmetic operators see only numbers.  `abc` and `def` aren't #defined; therefore, they're "expanded" into zeroes.
```C
#if 0 == 0
```

LanshunNie:
This looks great. Can this be used to support point-to-point communication between CC26xx devices? If it can, how to do collision avoidance such as LBT? Thanks spoerk.

PR #1938, opened 2016-11-17
Fixed checking of the cc1200 GPIO0 pin status.

makrog:
Driver of cc1200 was mistakenly comparing gpio0 pin high status with 1. When the pin is high you should compare it with the pin number.

### DIFF
--- a/platform/zoul/dev/cc1200-zoul-arch.c
+++ b/platform/zoul/dev/cc1200-zoul-arch.c
@@ -241,7 +241,7 @@ cc1200_arch_gpio2_disable_irq(void)
 int
 cc1200_arch_gpio0_read_pin(void)
 {
-  return GPIO_READ_PIN(CC1200_GDO0_PORT_BASE, CC1200_GDO0_PIN_MASK);
+  return (GPIO_READ_PIN(CC1200_GDO0_PORT_BASE, CC1200_GDO0_PIN_MASK) ? 1 : 0);
 }
 /*---------------------------------------------------------------------------*/
 int


msloth:
Semantically, the original seems to be right, since return value should rather indicate pin being high or low, not return a pin bitfield. Looking at the zoul arch file for example, it returns

`GPIO_READ_PIN(CC1200_GDO0_PORT_BASE, CC1200_GDO0_PIN_MASK);`

So I'd say instead of your proposed change, do either change the cc1200-arch's to return strictly 0/1 for low/high, or simply check for 0/non-zero in `cc1200.c`.

alignan:
Yup, I think that would be preferable to avoid having cpu-specific in the driver.

platform/zoul/dev/cc1200-zoul-arch.c
```diff
@@ -241,7 +241,7 @@ cc1200_arch_gpio2_disable_irq(void)
 int
 cc1200_arch_gpio0_read_pin(void)
 {
-  return GPIO_READ_PIN(CC1200_GDO0_PORT_BASE, CC1200_GDO0_PIN_MASK);
+  return (GPIO_READ_PIN(CC1200_GDO0_PORT_BASE, CC1200_GDO0_PIN_MASK)? 1 : 0);
```
alignan: 'return (GPIO_READ_PIN(CC1200_GDO0_PORT_BASE, CC1200_GDO0_PIN_MASK) ? 1 : 0);'

alignan:
Thanks for making the change, just fix the space missing between the operator and ready to merge :smile:

alignan:
The travis error is not related to the PR, I just restarted the job and once it is green I will merge

alignan:
:+1:

PR #1939, opened 2016-11-17
Add global SD/MMC and FAT support, with RE-Mote as an example

bthebaudeau:
FatFs is used.

Tested on RE-Mote rev. B.


cpu/cc2538/dev/spi.c
```diff
@@ -314,6 +320,37 @@ spix_set_mode(uint8_t spi,
 }
 /*---------------------------------------------------------------------------*/
 void
```
alignan: Great addition!

dev/disk/mmc/mmc.c
```diff
@@ -0,0 +1,592 @@
...
+#include <stddef.h>
+#include <stdint.h>
+#include "contiki.h"
+#include "sys/clock.h"
+#include "sys/rtimer.h"
+#include "dev/watchdog.h"
+#include "mmc-arch.h"
+#include "mmc.h"
+
+#define SECTOR_SIZE     512
```
alignan: Is this the de-facto size? just asking out of curiosity

dev/disk/mmc/mmc.c
```diff
@@ -0,0 +1,592 @@
...
+#define BLOCK_LEN       SECTOR_SIZE
+
+#define CLOCK_FREQ_CARD_ID_MODE         400000
```
alignan: Where are this values coming from?

dev/disk/mmc/mmc.c
```diff
@@ -0,0 +1,592 @@
...
+
+#define TOK_DATA_RESP_MASK              0x1f
+#define TOK_DATA_RESP_ACCEPTED          0x05
+#define TOK_DATA_RESP_CRC_ERROR         0x0b
+#define TOK_DATA_RESP_WR_ERROR          0x0d
+#define TOK_RD_SINGLE_WR_START_BLOCK    0xfe
+#define TOK_MULTI_WR_START_BLOCK        0xfc
+#define TOK_MULTI_WR_STOP_TRAN          0xfd
+
+#define SD_STATUS_SIZE                  (512 / 8)
```
alignan: Could you please add a comment explaining this value?

examples/zolertia/zoul/rev-b/test-power-mgmt.c
```diff
@@ -162,36 +157,11 @@ PROCESS_THREAD(test_remote_pm, ev, data)
   /* Configure the RTCC to schedule a "hard" restart of the shutdown mode,
    * waking up from a RTCC interrupt to the low-power PIC
    */
-  printf("PM: System date: %s\n", DATE);
```
alignan: Thanks for this re-work :smile:

regression-tests/18-compile-arm-ports/Makefile
```diff
@@ -54,6 +54,8 @@ sensniff/ev-aducrf101mkxz \
 cfs-coffee/cc2538dk \
 cfs-coffee/openmote-cc2538 \
 cfs-coffee/zoul \
+fat/zoul:BOARD=remote-reva \
```
alignan: The `remote-reva` doesn't support the micro-SD enable/disable, but the `remote-revb` should be fully working, officially I would prefer to support the `remote-revb` only

examples/fat/example-fat.c
```diff
@@ -0,0 +1,96 @@
...
+ */
+/*---------------------------------------------------------------------------*/
+#include <stdio.h>
+#include "contiki.h"
+#include "ff.h"
+/*---------------------------------------------------------------------------*/
+PROCESS(example_fat_process, "FAT example");
+AUTOSTART_PROCESSES(&example_fat_process);
+/*---------------------------------------------------------------------------*/
+#define TEST_FILENAME   "test"
```
bthebaudeau: Will do.

examples/fat/README.md
```diff
@@ -0,0 +1,6 @@
+FAT File System Example
+=======================
+
+Supported Hardware (tested or known to work)
+--------------------------------------------
+* zoul remote
```
bthebaudeau: Will do.

dev/disk/mmc/mmc.c
bthebaudeau: I will check in the standard, but I think that other sizes are possible and that this is a limitation of this driver. I can add a comment about this.

dev/disk/mmc/mmc.c
bthebaudeau: From the SD/MMC specifications. I can add a comment.

dev/disk/mmc/mmc.c
bthebaudeau: Will do.

regression-tests/18-compile-arm-ports/Makefile
bthebaudeau: I have not checked the schematic of the rev. A. Does it support SD, but not its power switch, or is SD not supported at all for this revision? In the former case, we can simply have a different code handling the power for rev. A. In the latter case, we should remove all the SD stuff for rev. A, not only in the READMEs and the tests, but also in the MAKEFILEs and in the code (but maybe not in `board.h`).

alignan: It has a bug in the design, so even if some features of the Micro-SD are working, the SD card detection is not working.  We are not manufacturing `remote-reva` anymore, but I'm OK to leave the code as it is for the `remote-reva`, in case anyone is willing to try it out

bthebaudeau: On rev. A, Q4 is a MOSFET of a wrong type if the schematic symbol corresponds to the actual BoM (can you confirm?), so the SD card cannot even be powered properly. However, the card could be detected with the internal pull-down resistor of the MCU pad enabled.

alignan: Yup, bypassing the MOSFET would be required to properly power the MicroSD, I didn't remembered that.

bthebaudeau:
Rework done.

alignan:
Great!

:+1: 

Setting the `timeout` to merge

PR #1949, opened 2016-11-25
RPL non-storing: more conservative DTSN update

simonduq:
This is as per RFC 6550, section 9.6: in non-storing mode, only the root triggers DTSN updates.


nvt:
:+1:

PR #1827, opened 2016-08-16
Harmonize RF_CHANNEL definition

alignan:
This PR brings back the `RF_CHANNEL` configuration parameter to newly platforms missing this, towards having a common channel definition to avoid platform-specific overrides.

@simonduq  `MICROMAC_CONF_CHANNEL` is the jn516x channel definition?

PR #1952, opened 2016-11-27
Move RF_CHANNEL check in cc13xx/cc26xx such that it applies for cc13xx prop mode 

arurke:
RF_CHANNEL added by #1827 will not be applied to CC13xx prop mode since it is checked after the RF_CORE_CONF_CHANNEL definition. This PR moves it above the prop mode section.


alignan:
:+1:

PR #1954, opened 2016-11-28
Disable DIO suppression by default

simonduq:
This might be controversial, so discussions are highly welcome here :)

The rationale is that in Contiki, nodes have a single DAO parent, and can only notice DTSN increment from that one particular parent. This means that DTSN increments triggered by the route struggle to propagate. The problem gets more pronounced in denser network.

In large IoT-LAB deployments, such as Grenoble (352 nodes, 6-7 hops diameter), the network even fails to bootstrap in the first place. Disabling redundancy enables quick convergence and swift maintenance. When the signaling traffic gets too heavy, decreasing the Trickle frequency also helps.

Note that this is only the default value; Nodes will adopt whichever value is advertised by the BR, i.e. they will still be able to adapt to networks with specific Trickle-redundancy needs.

### DIFF
--- a/core/net/rpl/rpl-conf.h
+++ b/core/net/rpl/rpl-conf.h
@@ -197,14 +197,19 @@
 /*
  * DIO redundancy. To learn more about this, see RFC 6206.
  *
- * RFC 6550 suggests a default value of 10. It is unclear what the basis
- * of this suggestion is. Network operators might attain more efficient
- * operation by tuning this parameter for specific deployments.
+ * RFC 6550 suggests a default value of 10. It is unclear what the basis of
+ * this suggestion is. Instead, we disable the suppression mechanism by default.
+ * In Contiki, nodes maintain a single DAO parent. They need to hear a DIO from
+ * this parent to notice DTSN updates. DIO suppression can in result in large
+ * delays (many Trickle periods, in dense networks) in DTSN update. Network
+ * operators might attain more efficient operation by tuning this parameter
+ * for specific deployments.
+ *
  */
 #ifdef RPL_CONF_DIO_REDUNDANCY
 #define RPL_DIO_REDUNDANCY          RPL_CONF_DIO_REDUNDANCY
 #else
-#define RPL_DIO_REDUNDANCY          10
+#define RPL_DIO_REDUNDANCY          0
 #endif
 
 /*


yatch:
It's an interesting topic, which is similar to what I discussed somewhere else :-)

With regard to the change, I'm in favor of keeping the current value, 10, for the default RPL_DIO_REDUNDANCY because it's the default value defined in RFC 6550, even though "it is unclear what the basis of this suggestion is."

The comment that you added is useful. That would be a nice guideline for RPL_CONF_DIO_REDUNDANCY configuration.

simonduq:
Is the discussion you already had on the topic available online by any chance?

yatch:
Yes, here is the mail related to the discussion:

  https://mailarchive.ietf.org/arch/msg/roll/-CtpgAUTiU0OXJdPtXFEuYghC5I

While your observation and the discussion are in different contexts, I see the common between them, that is both disable the suppression mechanism in order to propagate information quickly.

By the way, in my opinion, the DTSN case is a bug in RFC 6550... From the point of view of the DAO Triggering mechanism, it makes no sense for DAO parents of a child node to suppress their DIOs because of DIOs from other nodes. In the other view, if the Non-Storing mode is used, I cannot see any reason that RFC 6550 forces child nodes to take DTSN changes only from their DAO parents. In the Non-Storing mode, (typically) only the root increments DTSN, which is a global value in the network.

simonduq:
Thanks for the link.

I share your concern. In general, I'm doubtful about the usefulness of suppression on DIOs, because each DIO contains different information, related to the sender, which is unlike what Trickle was originally designed for (flooding / multicast). With DIOs, suppressing means losing content (children don't get rank update, nor DTSN increments etc). This is why I would disable it by default in Contiki, regardless of the default in RFC 6550 (RFC 5706 points out that defaults are getting less and less needed anyway).

Good point on non-storing with DTSN being global and no need to wait for an update from the preferred parent, indeed.

yatch:
I agree with you on the gap between what RFC 6550 does and the original idea of Trickle timer.

Just out of curiosity, what happened in the network bootstrap? A good amount of nodes were not able to receive DIOs from their expected preferred parents because of DIO suppression?

simonduq:
What happened is the BR would not see DAOs from all nodes before a *very* long time. First DAOs would come in quick and then the convergence would virtually almost never end (maybe 90% of nodes discovered after 1h). The problem occurred both in storing and non-storing if I remember right.

What I believe was exactly going is:
(1) all nodes would receive some DIO and join;
(2) they would then send a DAO;
(3) many DAOs get lost because at bootstrap there are too many of them (was without DAO-ACK);
(4) subsequent DTSN increment were unfruitful because of DIO suppression. At a network density of 70-80, the k=10 would basically suppress most DIOs, and some nodes have to wait many iterations until finally hearing from their parent (and the iterations take longer and longer due to interval doubling...).

yatch:
Thanks for the information. In that case, I would enable DAO-ACK if possible (I like DAO-ACK). If DAO-ACK is disabled, it may be better to set DAO sending interval short. But, short DAO interval could cause congestion...

Anyway, DAO is independent of the issue found at (4). Still, there is surely something needed to make the DTSN thing succeed in such an environment.

simonduq:
Yes, DAO-ACK would have helped!

nvt:
I think that we can disable the suppression by default if the given advantages are clear, and no counter-examples are shown where keeping the RFC-recommended value is better.

PR #1958, opened 2016-11-30
Fixed buffer-overflow in asking for command for ping6 app

Fathardie:
Buffer command is only 20 characters big. Malicious input could be much larger and overflow the buffer, corrupting memory.


nvt:
This application could probably be removed, but restricting the input to stay within bounds is good while it is kept.

Issue #1955, opened 2016-11-29
#define RPL_CONF_OF_OCP RPL_OCP_OF0 crash cooja simulator


meetbshah:
Hi, 
I am trying to use OF0 routing. 

I have added "#define RPL_CONF_OF_OCP RPL_OCP_OF0" in rpl conf.h and tried to run ipv6/rpl-udp example in cooja. Cooja crashes every time. I can run same example with mrhop config.

pablocorbalan:
To use OF0, you can directly add the following to your application `project-conf.h` file:
```
#undef RPL_CONF_OF
#define RPL_CONF_OF rpl_of0
```

This should work. In fact, some examples such as `er-rest-example` are configured to use OF0 that way.

simonduq:
Hi,

Not anymore @pablocorbalan :)
The new way to configure OF0 is:
(1) embed support for OF0, by making sure rpl_of0 is in the array RPL_CONF_SUPPORTED_OFS
(2) ask the root to advertise OF0 as the OF in use in its network , setting RPL_CONF_OF_OCP to RPL_OCP_OF0

Now, the fact you got a crash smells like a bug anyway. Nodes should have simply complained that they don't support the OF, at runtime, when receiving a DIO with the OF0 OCP. Worth investigating.

pablocorbalan:
It looks like I am getting very rusty with Contiki. I need to catch up with the latest changes :). Sorry for my outdated comment. Then, I guess you just need to add the following to the `project-conf.h`:
```
#define RPL_CONF_SUPPORTED_OFS {&rpl_of0}
#define RPL_CONF_OF_OCP RPL_OCP_OF0
```

@simonduq shall we then remove all the references to `RPL_CONF_OF`? See:
https://github.com/contiki-os/contiki/search?utf8=✓&q=RPL_CONF_OF
I could make a PR if required.

Btw, I confirm that only setting `#define RPL_CONF_OF_OCP RPL_OCP_OF0` makes Cooja crash with the following errors:

### Contiki Error:
```
Writing to read-only P1IN

Stack Trace: number of calls: 5 PC: $0eb50
  memset (memset.c) called from PC: $0982e (elapsed: 43)
  rpl_set_prefix (memset.c) called from PC: $04458 (elapsed: 261)
  process_thread_udp_server_process (local in udp-server.c) called from PC: $061b6 (elapsed: 16033)
  call_process (local in process.c) called from PC: $06346 (elapsed: 16068)
[...]

PR #1960, opened 2016-12-01
Remove definitions of outdated RPL_CONF_OF

pablocorbalan:
The way to configure the RPL Objective Function to be used has changed and now uses the definitions of  `RPL_CONF_SUPPORTED_OFS` and `RPL_CONF_OF_OCP`. Previously, to change the objective function `RPL_CONF_OF` was used. This definition is not used anymore.

This PR removes the outdated definitions of `RPL_CONF_OF` from several platforms and the `er-rest-example`. Note that this should not affect the platforms because MRHOF is by default the objective function of RPL.

For more information see the appearance of `RPL_CONF_OF`in the code:
https://github.com/contiki-os/contiki/search?utf8=✓&q=RPL_CONF_OF

And also the discussion in issue [#1955](https://github.com/contiki-os/contiki/issues/1955).

### DIFF
--- a/examples/er-rest-example/project-conf.h
+++ b/examples/er-rest-example/project-conf.h
@@ -96,12 +96,9 @@
 #undef COAP_PROXY_OPTION_PROCESSING
 #define COAP_PROXY_OPTION_PROCESSING   0
 
-/* Turn of DAO ACK to make code smaller */
+/* Turn off DAO ACK to make code smaller */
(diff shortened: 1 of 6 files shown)


simonduq:
:+1:

simonduq:
only minor comment is you might want to set of0 on the rest example, as initially intended. But doesn't matter much IMO

alignan:
:+1:

PR #1961, opened 2016-12-01
Do not initialize and use the wrong UART (z1)

jdede:
In contrast to the sky mote, z1 uses uart0. This leads to a link error when compiling the `rpl-udp/udp-client.c` example.
As the uart is initialized in the corresponding `contiki-<platform>-main.c` file, the corresponding include and initialization in `udp-client.c` are unnecessary and can be removed.

### DIFF
--- a/examples/ipv6/rpl-udp/udp-client.c
+++ b/examples/ipv6/rpl-udp/udp-client.c
@@ -40,9 +40,7 @@
 #include <stdio.h>
 #include <string.h>
 
-/* Only for TMOTE Sky? */
 #include "dev/serial-line.h"
-#include "dev/uart1.h"
 #include "net/ipv6/uip-ds6-route.h"
 
 #define UDP_CLIENT_PORT 8765
@@ -203,11 +201,6 @@ PROCESS_THREAD(udp_client_process, ev, data)
   PRINT6ADDR(&client_conn->ripaddr);
   PRINTF(" local/remote port %u/%u\n",
 	UIP_HTONS(client_conn->lport), UIP_HTONS(client_conn->rport));
-
-  /* initialize serial line */
-  uart1_set_input(serial_line_input_byte);
-  serial_line_init();
-
 
 #if WITH_COMPOWER
   powertrace_sniff(POWERTRACE_ON);


examples/ipv6/rpl-udp/udp-client.c
```diff
@@ -40,9 +40,7 @@
 #include <stdio.h>
 #include <string.h>
 
-/* Only for TMOTE Sky? */
 #include "dev/serial-line.h"
-#include "dev/uart1.h"
```
MohamedSeliem: #if CONTIKI_TARGET_Z1
#include "dev/uart0.h"
#else
#include "dev/uart1.h"
#endif

examples/ipv6/rpl-udp/udp-client.c
```diff
@@ -204,11 +202,6 @@ PROCESS_THREAD(udp_client_process, ev, data)
   PRINTF(" local/remote port %u/%u\n",
 	UIP_HTONS(client_conn->lport), UIP_HTONS(client_conn->rport));
 
-  /* initialize serial line */
```
MohamedSeliem: #if CONTIKI_TARGET_Z1
  uart0_set_input(serial_line_input_byte);
#else
  uart1_set_input(serial_line_input_byte);
#endif
 serial_line_init();

examples/ipv6/rpl-udp/udp-client.c
MohamedSeliem: what do you think about such modification.

examples/ipv6/rpl-udp/udp-client.c
MohamedSeliem: same here.

examples/ipv6/rpl-udp/udp-client.c
jdede: I though about that solution as it is done in this was in `rpl-collect` example. But I though that the examples in this directory should be independent of the platforms. And as the initialization of the UART is platform-specific, I think the correct place to do so is in the platform directory.
In fact, the include of `dev/uart?.h` is done in the corresponding `contiki-*-main.c`files. So I think that this part is unnecessary here. 
Or am I missing an important point? Just tried the given example with two TelosB nodes (real hardware): works perfectly!

alignan: Yes, this example should be platform-independent, thus I don't like this approach... normally the `serial-line-init` is done at platform's initialisation:

````
$ grep -lr "serial_line_init(" .
./srf06-cc26xx/contiki-main.c
./openmote-cc2538/contiki-main.c
./nrf52dk/contiki-main.c
./zoul/contiki-main.c~
./zoul/dev/mp3-grove.c~
./zoul/contiki-main.c
./eval-adf7xxxmb4z/contiki-main.c
./cc2538dk/contiki-main.c
./cooja/contiki-cooja-main.c
./jn516x/contiki-jn516x-main.c
./cooja-ip64/contiki-cooja-ip64-main.c
./ev-aducrf101mkxz/contiki-main.c
./mbxxx/contiki-main.c
./wismote/contiki-wismote-main.c
./native/contiki-main.c
./avr-rss2/contiki-main.c
./z1/contiki-z1-main.c~
./z1/contiki-z1-main.c
./sky/contiki-sky-main.c
./exp5438/contiki-exp5438-main.c
./cc2530dk/contiki-main.c
````

Issue #1929, opened 2016-11-09
SPI lines mixed in MSP430 based radio platforms (exp5438)


RameO:
There are lines in spix.c and /cpu/msp430/f5xxx/spi.c

  P3SEL |= BV(SCK) | BV(MOSI) | BV(MISO); // Select Peripheral functionality
  P3DIR |= BV(SCK) | BV(MISO);  // Configure as outputs(SIMO,CLK).

I guess the lower line should be 
  P3DIR |= BV(SCK) | BV(MOSI); 

Because the controller acts as a spi master.

PR #1971, opened 2016-12-04
Fix wrong GPIO port direction for SPI master

MohamedSeliem:
This is  a bug, although apparently a harmless one in typical conditions (as the SPI bus still works).
- the MOSI and SCK pins are actually set as outputs, even though the code appears to say something else.


MohamedSeliem:
@g-oikonomou can you look at this please.

MohamedSeliem:
@g-oikonomou please give me your feed back on this,

atiselsts:
@MohamedSeliem it's not as simple.

First of all, to me it looks like the code on Sky is already correct in terms of functionality, it's just the names of the pins that are wrong (a double bug cancelling out?). See this discussion: https://sourceforge.net/p/contiki/mailman/message/33197414/

Second, this code has been in Contiki for so long and so widely used, I would be very careful when applying this change - even though it's correct, it may still break someones system in an obscure way.

On the other hand, this issue crops up periodically (see also #1929) and I think it's embarrassing to have such a problem in a very core part of Contiki.

MohamedSeliem:
@atiselsts thanks for this insightful response. 👍

PR #1973, opened 2016-12-04
Add sensniff as a submodule

g-oikonomou:
This pull adds sensniff as a submodule.

This allows us to invoke the script with the correct command line arguments (e.g. baudrate) by running `make sniff` from inside the example dir.

This pull also:

* Specifies correct baudrate for jn516x
* Specifies default baudrate for CC2538DK when not using native USB
* Extends documentation

I'm not feeling too strongly about this, but it has been asked of me so I thought I'd knock something together. What do you think?


alignan:
I'm happy to finally have this as a sub-module :+1: 

Setting a timeout to merge, if anyone wants to merge sooner or provide additional feedback it would be most welcomed

nvt:
Couldn't it be added directly to the tools directory since the firmware part is already in the source tree? Having it as a module would be fine too, however.

g-oikonomou:
Nicolas, I do strongly prefer using a submodule here so that users can easily (locally) checkout newer/older versions from the origin repo (e.g. for debugging) without having to manually copy files between repos.

PR #1984, opened 2016-12-10
Bugfix in Address Resolution / Neighbor Unreachability Detetion

yatch:
# Summary
As a result of the address resolution with NS/NA, a Contiki node has a neighbor cache of the REACHABLE state for the peer, the link-layer address of which is the broadcast address or `linkaddr_null`. This PR resolves this bug.

# Test
`regression-tests/11-ipv6/22-cooja-ping6.csc` provides a test for this bugfix. You can run the test as follows:
```shell
$ cd regression-tests/11-ipv6/
$ make 22-cooja-ping6.testlog
Running test 22-cooja-ping6 with random Seed 1: ............. OK
```
In the simulation, four pairs of IPv6 nodes are generated: "CSMA", "CSMA + IEEE 802.15.4e-2012", "TSCH", and "RPL". A node in each pair, called requester, sends a ICMPv6 Echo Request to the other node one time. NS/NA exchange happens if necessary.

The test criteria are
- Each requester receives an ICMPv6 Echo Reply from its peer.
- Each requester has a neighbor cache for its peer in which a non-broadcast address is registered as a link-layer address.

Here is the summary of node configurations:

|Node Type|MAC|Frame Version|RPL|
|---|---|---|---|
|CSMA|CSMA|FRAME802154_IEEE802154_2006|Disabled|
|CSMA + IEEE 802.15.4e-2012|CSMA|FRAME802154_IEEE802154E_2012|Disabled|
|TSCH|TSCH|FRAME802154_IEEE802154E_2012|Disabled|
|RPL|CSMA|FRAME802154_IEEE802154_2006|Enabled|

Only "RPL" passes the test, others fails on the tip of the master branch. The following is a part of the test log.
```shell
14980456 8 ping6 OK - 20 bytes from fe80::207:7:7:7 ttl=64
14980456 8 post-ping: uip_ds6_nbr_get_ll() returns 00:07:00:07:00:07:00:07 for fe80::207:7:7:7
14980456 8 0.8: Contiki>
14988616 2 ping6 OK - 20 bytes from fe80::201:1:1:1 ttl=64
14988616 2 post-ping: uip_ds6_nbr_get_ll() returns 00:00:00:00:00:00:00:00 for fe80::201:1:1:1
14988616 2 0.2: Contiki>
14988680 6 ping6 OK - 20 bytes from fe80::205:5:5:5 ttl=64
14988680 6 post-ping: uip_ds6_nbr_get_ll() returns 00:00:00:00:00:00:00:00 for fe80::205:5:5:5
14988680 6 0.6: Contiki>
15007392 4 ping6 OK - 20 bytes from fe80::203:3:3:3 ttl=64
15007392 4 post-ping: uip_ds6_nbr_get_ll() returns 00:00:00:00:00:00:00:00 for fe80::203:3:3:3
CSMA:                       NG
CSMA + IEEE 802.15.4e-2012: NG
TSCH:                       NG
RPL:                        OK
TEST FAILED
```

# Fix
There are two essential changes:
- make `uip_ds6_nbr_refresh_reachable_state()` not to change the state of a specified neighbor cache to REACHABLE if its (current) state is INCOMPLETE
- use the newly introduced function called `uip_ds6_nbr_update_lladdr()` to change the link-layer address of a specified neighbor cache instead of `nbr_table_update_lladdr()`

It seems that `uip_ds6_nbr_refresh_reachable_state()` is supposed to be used by a protocol other than Neighbor Unreachability Detection (NUD) so that its packet can refresh the neighbor cache state for a peer with REACHABLE. The problem here is, While transition from INCOMPLETE to REACHABLE should be done only by NUD, the function changes the neighbor cache state regardless of its current state. This is one of things to fix. Then, it'll be compliant with the following paragraph in [Section 7.3.3, RFC 4861](https://tools.ietf.org/html/rfc4861#section-7.3.3)

```
   When a reachability confirmation is received (either through upper-
   layer advice or a solicited Neighbor Advertisement), an entry's state
   changes to REACHABLE.  The one exception is that upper-layer advice
   has no effect on entries in the INCOMPLETE state (e.g., for which no
   link-layer address is cached).
```

The other thing is `nbr_table_update_lladdr()`. This function is called to update the link-layer address of a neighbor cache with a specified address when a solicited NA is received.

Typically, a neighbor cache has the broadcast address, `linkaddr_null`, as its link-layer address when a NS is being sent. On receiving a solicited NA, the broadcast address is going to be replaced (updated) with a link-layer address stored in the target link-layer address field of the NA. `nbr_table_update_lladdr()` is called for this purpose, however, it fails because there is a *nbr_table* entry of the new address for some reason. I suspect `sicslowpan_driver.input()` installs an entry for the new address before `nbr_table_update_lladdr()` is called since I saw  a `nbr_table` entry with the reason of `NBR_TABLE_REASON_LINK_STATS` in a failed test.

The point is, a neighbor cache should be able to have a new link-layer address even if it's already registered in `nbr_table` as long as there is no neighbor cache associated with the new address. `uip_ds6_nbr_update_lladdr()`, which is newly introduced, takes this into account and update the link-layer address of a neighbor cache as expected. This PR replaces `nbr_table_update_lladdr()` with `uip_ds6_nbr_update_lladdr` in `uip-nd6.c`.

# Discussion
This is a topic related to this PR; but no change has been made for this.

I don't think `uip_process()`  is an appropriate place where `uip_ds6_nbr_refresh_reachable_state()` is called. The IPv6 header of an incoming packet alone does not provide "forward progress" on the connection. IPv6 is not an upper-layer protocol of NUD/ICMPv6 in the first place. 

I'd say that we should put `uip_ds6_nbr_refresh_reachable_state()` in TCP, RPL, and other "upper-layer" protocols instead of IPv6. A lower-layer protocol, for example, IEEE 802.15.4 may be able to confirm reachability to a neighboring node. However, it is not aware of its upper layer addresses, that is, IPv6 address. Because of that, such a lower-layer protocol cannot be a place for `uip_ds6_nbr_refresh_reachable_state()`, either.

# Related PRs
- https://github.com/contiki-os/contiki/pull/1394
- https://github.com/contiki-os/contiki/pull/1400


simonduq:
Is `nbr_table_update_lladdr` still needed then?

laurentderu:
The NULL entry in the neighbor table and the fact that there could be only a single instance of a MAC address have many unexpected consequences and it's great to see them slowly solved.

Currently I'm testing your code with more complex setup and and still have a few problems (that are also present in current Contiki). They all boils down to the fact that an IPv6 interface has several IP addresses, in most common setup two : the link local address and the global address.

What I have observed is sometimes two nodes first exchange with their link-local addresses, so in the neighbor table the MAC address is associated with the link-local. However, when their gobal address is tentatively used the address resolution fails.

In current Contiki it fails when the the NA is received, nbr_table_update_lladdr() detect a duplicate entry and therefore the NA is dropped and the neighbor entry for the global address stays associated with the NULL MAC address. This ends up with an endless exchange of NS/NA messages.

With the current PR, this still fails but for different reasons : The NS/NA exchange is correctly done for the global address, but the entry in the neighbor table stays in INCOMPLETE state and so can not be used. This is due to the condition of unicity in uip_ds6_nbr_update_lladdr(). In a neighbor with the same MAC address exists, we abort the table update.

As modifying the neighbor table to support more than one entry with the same MAC is not trivial, I tried a simpler workaround : Instead of aborting the update in uip_ds6_nbr_update_lladdr() when a duplicate is detected, I instead removed the duplicated entry before updating the new one. The rationale is that if we can not store all the IP addresses related to a MAC address, at least keep the last used one as it's probably the address the node will use for next messages. This workaround solves the issue and does not seems to have unwanted side effect, but more testing would be welcome.

yatch:
@simonduq That's a good question. `nbr_table_update_lladdr` is not needed any more, I guess. No caller of the function is found in this branch. If I miss something, I would get a linker error anyway. I'll remove the function later.

yatch:
Updated the branch.
- extend the regression test by adding ping tests to global IPv6 addresses
- remove `nbr_table_update_lladdr()`

@laurentderu
Thank you for your detailed feedback! I've updated the regression test. I believe it reproduces the issue you pointed out. Here is an excerpt of a log file generated by the test with enabling debug print.

```
01:28.943       ID:2    Adding neighbor with ip addr fd00::201:1:1:1 link addr (NULL LL addr) state 0
01:28.943       ID:2    Sending NS to ff02::1:ff01:1 from fd00::202:2:2:2 with target address fd00::201:1:1:1
01:28.947       ID:1    Received NS from fd00::202:2:2:2 to ff02::1:ff01:1 with target address fd00::201:1:1:1
01:28.947       ID:1    Adding neighbor with ip addr fd00::202:2:2:2 link addr 00:02:00:02:00:02:00:02 state 2
01:28.947       ID:1    Sending NA to fd00::202:2:2:2 from fd00::201:1:1:1 with target address fd00::201:1:1:1
01:28.947       ID:1    tcpip_ipv6_output: nbr cache entry stale moving to delay
01:28.952       ID:2    Received NA from fd00::201:1:1:1 to fd00::202:2:2:2 with target address fd00::201:1:1:1
01:28.952       ID:2    uip_ds6_nbr_update_lladdr(): new_ll_addr is associated with another IPv6 address

(failed to update the NC for the link-local address)
```

I have another idea to deal with the issue arose when a link-local address is associated more than one IPv6 address. I'll push another commit later which implements the idea.

yatch:
@laurentderu It'd be better to discuss how to manage multiple IPv6 addresses for a single link-layer address at a separate PR. So, I removed the code to test ping to global IPv6 addresses from `cooja-ping6.js`, that is applied in the different branch for the new PR...

laurentderu:
@yatch agreed, this PR otherwise will not converge and get merged.

yatch:
@laurentderu It took longer than I thought. I've created a new PR, https://github.com/contiki-os/contiki/pull/2098, which hopefully solves the issue you pointed out.

Your solution or workaround is reasonable as long as a node doesn't communicate with different IPv6 destinations at the same time which are associated with an identical link-layer address. A good thing is it's simple.

My idea was registering a peer IPv6 address as an IPv6 routing entry if there is another IPv6 address with the same link-layer address in the neighbor cache. The destination address of the routing entry is the peer IPv6 address, and the next-hop address is the address in the neighbor cache. A drawback of this idea is we cannot track the NUD state of peer IPv6 addresses registered in the routing table.

So, I decided to try the "not-trivial" thing. 😄 
> As modifying the neighbor table to support more than one entry with the same MAC is not trivial, I tried a simpler workaround

As I mentioned before, tests with ping to global IPv6 addresses are introduced at the new PR. Your comments are very welcome!!

PR #1987, opened 2016-12-11
Fixed COOJA to avoid using UI based message list when not using UI.

joakimeriksson:
This hopefully fixes a few of the last parts of the COOJA code that still made use of the MessageList with UI that sometimes cause strange class cast exceptions when running in travis. Fixing the parts of the code that cause the problems that is mentioned in PR #1986.

PR #1989, opened 2016-12-13
Refactoring tsch-packet

yatch:
# Summary
The original purpose of this refactoring is to make `tsch_packet_create_eack()` and `tsch_packet_create_eb()` not build a IEEE 802.15.4 frame by themselves, instead, let `NETSTACK_FRAMER.create()` do its job for them. That is, there is duplicate code in `tsch-packet.c` which is similar to what `create_frame()` of  `framer_802154` does.

This PR introduces three new packetbuf attributes for this purpose, currently used only for IEEE 802.15.4 frames:
-  `PACKETBUF_ATTR_MAC_METADATA`: set with 1 if Information Element is present in a frame
-  `PACKETBUF_ATTR_MAC_NO_SRC_ADDR`: set with 1 if source address is omitted
-  `PACKETBUF_ATTR_MAC_NO_DEST_ADDR`: set with 1 if destination address is omitted

# Test
`regression-tests/25-ieee802154/02-tsch-packet-create.csc` provides regression tests for `tsch_packet_create_eack()` and `tsch_packet_create_eb()`.
```shell
$ make 02-tsch-packet-create.testlog
Running test 02-tsch-packet-create with random Seed 1: .......... OK
```

There are three configuration sets used in the tests:
- DEFAULT: use the default configuration parameters
- SECURITY_ON: enable the security feature by the following settings:
```C
#undef LLSEC802154_CONF_ENABLED
#define LLSEC802154_CONF_ENABLED 1
#undef LLSEC802154_CONF_USES_EXPLICIT_KEYS
#define LLSEC802154_CONF_USES_EXPLICIT_KEYS 1
#undef LLSEC802154_CONF_USES_FRAME_COUNTER
#define LLSEC802154_CONF_USES_FRAME_COUNTER 0
```
- ALL_ENABLED: enable all the features other than security, that affects `tsch_packet_create_eack()` and `tsch_packet_create_eb()` behaviors
```C
#undef TSCH_PACKET_CONF_EACK_WITH_DEST_ADDR
#define TSCH_PACKET_CONF_EACK_WITH_DEST_ADDR 1
#undef TSCH_PACKET_CONF_EACK_WITH_SRC_ADDR
#define TSCH_PACKET_CONF_EACK_WITH_SRC_ADDR 1
#undef TSCH_PACKET_CONF_EB_WITH_TIMESLOT_TIMING
#define TSCH_PACKET_CONF_EB_WITH_TIMESLOT_TIMING 1
#undef TSCH_PACKET_CONF_EB_WITH_HOPPING_SEQUENCE
#define TSCH_PACKET_CONF_EB_WITH_HOPPING_SEQUENCE 1
#undef TSCH_PACKET_EB_WITH_SLOTFRAME_AND_LINK
#define TSCH_PACKET_CONF_EB_WITH_SLOTFRAME_AND_LINK 1
```

# Disadvantage
The static variable named `tmp_pktbuf` is introduced in `tsch-packet.c` in order to keep the original contents of `packetbuf` during `tsch_packet_create_eack()` and `tsch_packet_create_eack()`. This could impact on memory constrained platforms.

# By-products
Three issues found during the refactoring has been resolved.

Firstly, it turned out that `create_frame()` of `framer_802154` always set `linkaddr_node_addr` to the source address of an IEEE 802.15.4 frame.

```C
[framer-802154.c]
 172   /*
 173    * Set up the source address using only the long address mode for
 174    * phase 1.
 175    */
 176   linkaddr_copy((linkaddr_t *)&params.src_addr, &linkaddr_node_addr);
```
This bug is fixed to set the address of `PACKETBUF_ADDR_SENDER` instead.

By this bugfix, two regression tests, `11-ipv6/04-exp5438-udp-fragmentation-contikimac` and `13-ipv6-apps/01-sky-servreg-hack1`, got to failure. Looking into frames In these tests, the all zero address was set to the source link-layer address of frames. This prevents a recipient from building the correct source address of a compressed IPv6 packet. This issue is resolved by making `qsend_list()` of `contikimac_driver` set `PACKETBUF_ADDR_SENDER` with `linkaddr_node_addr` before calling `NETSTACK_FRAMER.create()`. Although I'm not sure if setting the source address at the RDC layer is a right thing, I just borrowed the idea that is already there; the same thing has been done in `send_packet()` of `contikimac_driver`. 

The last one is `regression-tests/04-rime/07-sky-collect`. This test failed after the refactoring was done. The interesting thing is this test succeeded on my local PC but failed on Travis-CI.  One packet loss caused the test failure according to the failure log on Travis. I could not find any problem in the Contiki stack code. My conclusion is that the packet was lost by chance.

During the investigation, I found a bug in `examples/rime/example-collect.c`, a source file used in the test. Each node in the simulation is supposed to send a message every 30 seconds. However, it sends messages more frequently because of the bug. This may cause collision and packet loss. After fixing the bug, the packet loss disappeared and the test succeeded.

# Other refactoring things
- Indentation fixes
- Put `tsch_` prefix to some global names
- Introduce `packetbuf_freelen()` which returns the length of free space in packetbuf
- Add missing error checks
- Add the `const` modifier to `dest addr` in `tsch_packet_create_eack()`


joakimeriksson:
Can anyone please "pull" my fixes for Cooja in #1987 so that we know if these things will happen again? (e.g. the fail for this travis test in RPL tests with Cooja. I am eager to know if the fix actually works or not...

simonduq:
Great! Did you check the tsch non-regression tests? (they are now not enabled in travis, because the linker for z1 fails randomly...).
Any chance you could split this in several PRs, one per feature?
@joakimeriksson yes, eager to find out too!

yatch:
@joakimeriksson I'm happy to do what I can do :-) Ah, OK, this branch had the same failure as the issue. I did "pull" your branch. Let's see if the failure is gone.

yatch:
@simonduq Yeah, I know this PR has become eventually big... I'll make some new PRs derived from this later. :-)

yatch:
@joakimeriksson Now, Travis gets green. I bet it's thanks to your changes! Interestingly, the failure always happens on a test for `non-storing`, `BUILD_TYPE=non-storing`. I have the failure on my another branch, https://travis-ci.org/yatch/contiki/jobs/183911836 . That's just for your information.

simonduq:
Yeah I've seen it occur on other simulations just as well, e.g.
https://travis-ci.org/contiki-os/contiki/jobs/181255620

yatch:
@simonduq With regard to tsch non-regression tests, I didn't notice even their existence... Now I ran the following tests; it seems good. Is there any other test I should run?
- `19-z1-rpl-tsch.csc.flaky`
- `20-z1-rpl-tsch-orchestra.csc.flaky`
- `21-z1-rpl-tsch-security.csc.flaky`

simonduq:
these are the 3 tests I had in mind yes :) thanks

yatch:
@simonduq OK; I'll run the tests again after reshaping this branch later.

yatch:
Updated the branch and the PR message. Hope I didn't miss anything... Anyway, I'll rebase the branch after all the dependent PRs are merged and confirm if all the tests are passed.

core/net/mac/tsch/tsch-security.c
```diff
@@ -159,7 +159,7 @@ tsch_security_secure_frame(uint8_t *hdr, uint8_t *outbuf,
 
   if(!frame.fcf.security_enabled) {
     /* Security is not enabled for this frame, we're done */
-    return 1;
+    return 0;
```
simonduq: another good catch :)

core/net/mac/tsch/tsch-packet.c
```diff
@@ -62,64 +62,95 @@
 #endif /* TSCH_LOG_LEVEL */
 #include "net/net-debug.h"
 
+/* temporary buffer used in tsch_packet_create{eack,eb}() */
```
simonduq: missing underscore: should be `tsch_packet_create_{eack,eb}`

simonduq:
All commits but the two last could be merged as is (other PR?)

simonduq:
I meant, couldn't easily have a "misc tsch fixes" PR with only the 6 first commits of this pr?

yatch:
@simonduq Aha, definitely. The last two commits need more time. I agree with you; let me split this PR further.

yatch:
@simonduq Kept the commits on `tsch_packet_create_eack()` and `tsch_packet_create_eb()` and dropped the others from this branch. And, rebased. It's got lean!! Thanks!

yatch:
@simonduq Updated the commits; I realized we don't need to take a backup of `packetbuf` in either `tsch_packet_create_eack()` or `tsch_packet_create_eb()`.

In `tsch_rx_slot()`, a received frame is copied into `current_input->payload` before `tsch_packet_create_eack()` is called. So, `tsch_packet_create_eack()` can use `packetbuf` for its own job.

```C
 719 /*---------------------------------------------------------------------------*/
 720 static
 721 PT_THREAD(tsch_rx_slot(struct pt *pt, struct rtimer *t))
 722 {
(snip)
 746     static struct input_packet *current_input;
(snip)
 759     current_input = &input_array[input_index];
(snip)
 794         current_input->len = NETSTACK_RADIO.read((void *)current_input->payload, TSCH_PACKET_MAX_LEN);
 795         NETSTACK_RADIO.get_value(RADIO_PARAM_LAST_RSSI, &radio_last_rssi);
(snip)
 860               /* Build ACK frame */
 861               ack_len = tsch_packet_create_eack(ack_buf, sizeof(ack_buf),
 862                   &source_address, frame.seq, (int16_t)RTIMERTICKS_TO_US(estimated_drift), do_nack);
```

Regarding `tsch_packet_create_eb()`, this is called from `tsch_send_eb_process()` with the pointer to `packetbuf`. Apparently, `tsch_packet_create_eb()` doesn't need to take a backup of `packetbuf`.
```C
 709 /*---------------------------------------------------------------------------*/
 710 /* A periodic process to send TSCH Enhanced Beacons (EB) */
 711 PROCESS_THREAD(tsch_send_eb_process, ev, data)
 712 {
(snip)
 750         eb_len = tsch_packet_create_eb(packetbuf_dataptr(), PACKETBUF_SIZE,
 751             &hdr_len, &tsch_sync_ie_offset);
```

Please note that the signatures of both functions are changed. They don't need `buf` and `buf_size` any more.

After all the dependent PRs are merged, I'll rebase it and confirm if the tests are passed with this branch.

simonduq:
Yes, but the problem is that packets travel through packetbuf at the interface with upper layers. Specifically, `send_packet` and `tsch_rx_process_pending` might have data in packetbuf when TSCH interrupts to run a slot.

yatch:
@simonduq Aha, OK. I think I understand more about your concerns. I forgot the important thing you pointed out...

In `tsch_packet_create_eack()` case, it's called in `tsch_rx_slot` thread, which runs in the (`rtimer`) interrupt context. Therefore, there can be another process working with `packetbuf` when `tsch_packet_create_eack()` accesses `packetbuf`. That's why the original code uses the separate memory, `ack_buf[]`, allocated by `tsch_rx_slot` instead of `packetbuf`. That is what you mentioned yesterday.

That "another process" may be using `send_packet()` or `tsch_rx_process_pending()`, or accessing `packetbuf` directly.

So, for sure, the latest change for `tsch_packet_create_eack()` still has a chance to ruin `packetbuf` to be processed by others. For `tsch_packet_create_eb()`, `packetbuf` can be used. I'll think it again ;-)

simonduq:
Yep, 100% on the same page now :)

simonduq:
It might be that keeping `tsch_packet_create_eack()` as is (i.e. not FRAMER-based) is the best option (depending on how much changes to FRAMER are required otherwise).

yatch:
@simonduq I extracted the logic setting up `frame802154_t params` from `create_frame()` and created a new function with it. This new function, `framer_802154_setup_params()`, sets values to `params` appropriately based on packetbuf attributes. This function takes a getter function (`get_attr`) as an argument so that it can access packetbuf attributes other than stored in `packetbuf_attrs` defined in `packetbuf.c`.

`tsch_packet_create_eack()` uses a local `packetbuf_attr` array instead of `packetbuf_attrs` that may be used by another process. It gets a ready-to-use `frame802154_t` data with `framer_802154_setup_params()` and builds an EACK in a given buffer with `frame802154_create()`. 

If this version of `tsch_packet_create_eack()` has any critical problem, I'll revert all the changes on the function. This is the last try 😄

simonduq:
Great :) I think this should work. Am ready for a detailed review as soon as Travis is green ;)

yatch:
@simonduq I'll rebase this branch after both of PR/#1998 and PR/#2002 are merged. Then, I'll make sure that Travis gets green as well as non-regression tests are passed.

kkrentz:
BTW, in #880, I also adapted the creation and parsing of EACKs and EBs to reuse code of the framer.

yatch:
@kkrentz Thanks for the info. 😄

PR #1993, opened 2016-12-14
Bugfix on regression-tests/04-rime/07-sky-collect.csc

yatch:
Mainly two fixes:
- `04-rime/07-sky-collect.csc`: while a received `seqno` is supposed to be less than 10 in the test script, it doesn't handle a case where  `seqno` is larger than or equal to 10.
- `example-collect.c`: a node sends messages more frequently than expected, every 30 seconds.

Other changes are trivial: indentation fix and shortening time to wait in case of test failure.


simonduq:
👍

nvt:
:+1:

PR #1997, opened 2016-12-14
Introduce packetbuf_remaininglen()

yatch:
As far as I know, we don't have any API to get the length of remaining data space in `packetbuf`. `packetbuf_remaininglen()` provides this.


simonduq:
👍

alignan:
:+1:

PR #2001, opened 2016-12-15
fixed so that Cooja can run in quick mode when having UI

joakimeriksson:
The fix for travis (#1987) made the regular UI mode a bit boring to use since the "quick" mode did not work anymore so that everybody need to click 10 times to start the simulation. This fix re-enabled quick mode when using UI again and make using Cooja. Now any component that is configured at startup should respect `simulator.isQuickSetup()` if it supports quick mode.

### DIFF
--- a/tools/cooja/apps/mspsim/src/org/contikios/cooja/mspmote/AbstractMspMoteType.java
+++ b/tools/cooja/apps/mspsim/src/org/contikios/cooja/mspmote/AbstractMspMoteType.java
@@ -81,7 +81,7 @@ public abstract class AbstractMspMoteType extends MspMoteType {
         }
 
         /* If visualized, show compile dialog and let user configure */
-        if (visAvailable) {
+        if (visAvailable && !simulation.isQuickSetup()) {
 
             /* Create unique identifier */
             if (getIdentifier() == null) {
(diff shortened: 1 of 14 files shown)


yatch:
Oh, I need this fix! ✨

alignan:
@joakimeriksson this is good to go?

joakimeriksson:
Yes, I would say so!

alignan:
:+1:

yatch:
It seems that Cooja still needs one click to start in the quickmode...

For example, I need to click the "Compile" button on the small window named "Create Mote Type" in Cooja when I start a simulation by followings:
```shell
$ java -Xshare:on -jar ../../tools/cooja/dist/cooja.jar -quickstart=01-panid-handling.csc  -contiki=../..
```

window

Is it an expected behavior? I think Cooja used to start up in the quickstart mode without any click... For example,  back to the commit of ac2a31455e1665d2cee4252ef0fe0e0b0a74572a, it worked as I thought.

Sidenote: I got an error with the nogui mode which was not seen at the commit of ac2a31455e1665d2cee4252ef0fe0e0b0a74572a. I have no idea whether it's relevant or not...
```shell
 INFO [main] (ContikiMoteType.java:404) - Creating core communicator between Java class Lib1 and Contiki library '/home/yatch/work/contiki/regression-tests/25-ieee802154/code/obj_cooja/mtype80.cooja'
javac 1.8.0_111
org/contikios/cooja/corecomm/Lib1.java:40: error: cannot find symbol
public class Lib1 extends CoreComm {
                          ^
  symbol: class CoreComm
1 error
```

joakimeriksson:
Thanks for the feedback - I will check if there is more places where the quick mode needs to be handled!

PR #2011, opened 2016-12-21
Enable llsec on Z1 nodes

jdede:
This PR add the required includes and commands to use link layer security (llsec) with Zolertia Z1 nodes (copied from Sky platform).


simonduq:
👍

alignan:
:+1:

PR #2012, opened 2016-12-22
JN516x: Add support for hardware MAC

laurentderu:
The PR allows the micromac driver of the JN516x platform to use the hardware MAC layer engine provided by NXP.

At first I wanted to have a JN516 framer to create a specific frame for micromac, however there are so many pieces of code that assume that the content of a packetbuf is a real 802.15.4 frame, that the complexity of the change were skyrocketing... Instead, the driver receives the 802.15.4 frame from the RDC layer and dissect it to create the JN516 specific frame and send it to the micromac driver. Similarly, when a frame is received from the microbic layer, the driver recreates a full 802.15.4 frame.

This driver has been tested with NullRDC and ContikiMAC, but without LLSEC, this would require additional changes to support the security header.

By default the hardware mac layer is still disabled, it can be activated using  MICROMAC_CONF_RADIO_MAC macro. However, the default configuration in contiki-conf.h of JN516x reference csma_driver and nullrdc_driver which requires the hardware MAC to be enabled, so in the future either the contiki-conf.h should be modified to activate it by default and modify the TSCH examples to disable it explicitly or make TSCH the default MAC layer for JN516x :)

### DIFF
--- a/platform/jn516x/dev/micromac-radio.c
+++ b/platform/jn516x/dev/micromac-radio.c
@@ -63,6 +63,18 @@
 #define DEBUG DEBUG_NONE
 #include "net/ip/uip-debug.h"
 
+#ifdef MICROMAC_CONF_RADIO_MAC
+#define MICROMAC_RADIO_MAC MICROMAC_CONF_RADIO_MAC
+#else
+#define MICROMAC_RADIO_MAC 0
+#endif
+
+#if MICROMAC_RADIO_MAC
+#define MICROMAC_FRAME tsMacFrame
+#else
+#define MICROMAC_FRAME tsPhyFrame
+#endif
+
 /* Perform CRC check for received packets in SW,
  * since we use PHY mode which does not calculate CRC in HW */
 #define CRC_SW 1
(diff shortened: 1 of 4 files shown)


simonduq:
Hi @laurentderu did you also test this with TSCH?
Please note Travis is not happy, there is a warning on jn516x

laurentderu:
@simonduq The root cause of travis issue is I believe due to the fact that the version of JN516 SDK used by Travis is a bit older and does not include the definition of function vMMAC_SetChannelAndPower(), the latest one does (in Components/MMAC/Include/MMAC.h).

I haven't really tested with TSCH as my idea was to still use the raw radio layer to TSCH.

platform/jn516x/dev/micromac-radio.c
```diff
@@ -57,14 +57,24 @@
 #include "JPT.h"
 #include "PeripheralRegs.h"
 
-void vMMAC_SetChannelAndPower(uint8 u8Channel, int8 i8power);
-
```
simonduq: so, maybe you don't actually want to remove this?

laurentderu: Actually we want as the signature of the function has also changed :)

In the current SDK it is :

    PUBLIC void vMMAC_SetChannelAndPower(uint8 u8Channel, int i8TxPower);

So, int instead of int8 for TxPower.

simonduq: then I would suggest to either fix this signature, or update the sdk that Travis uses

laurentderu: I would suggest to update the SDK so the code is inline with the SDK people would download from NXP. I think you have to do it as the SDK is stored on your github.io.

simonduq: right, I'll do it then!

neophob:
any news on this PR? I've tested it (using the older JN SDK) and it improves connection reliability, especially if you send larger packages (for example via coap)

simonduq:
Sorry for the delay. @laurentderu , rebasing after https://github.com/contiki-os/contiki/pull/2129 is merged should do!

simonduq:
https://github.com/contiki-os/contiki/pull/2129 now merged, can you rebase this?

laurentderu:
Rebased and now travis is green

simonduq:
👍

PR #2023, opened 2016-12-29
Adding support for OpenMote-CC2538 Rev.A1 and OpenUSB

peretuset:
This PR adds support for the OpenMote-CC2538 Rev.A1 board, which uses the CC2538SF23 (256 KB Flash) instead of the CC2538SF53 (512 KB Flash), and the new OpenUSB carrier board, which includes a Texas Instruments CC1200 radio transceiver for 868/915 MHz and has support for automatic bootloading using the cc2538-bsl Python script.

### DIFF
--- a/platform/openmote-cc2538/Makefile.openmote-cc2538
+++ b/platform/openmote-cc2538/Makefile.openmote-cc2538
@@ -36,7 +36,7 @@ MODULES += core/net core/net/mac \
            core/net/llsec core/net/llsec/noncoresec
 
 PYTHON = python
-BSL_FLAGS += -e --bootloader-invert-lines -w -v -b 500000
+BSL_FLAGS += -e --bootloader-invert-lines -w -v -b 450000
 
 ifdef PORT
   BSL_FLAGS += -p $(PORT)
--- a/platform/openmote-cc2538/Makefile.openmote-cc2538
+++ b/platform/openmote-cc2538/Makefile.openmote-cc2538
@@ -31,7 +31,7 @@ MODULES += core/net core/net/mac \
            core/net/llsec core/net/llsec/noncoresec
 
 PYTHON = python
-BSL_FLAGS += -e -w -v -b 450000
+BSL_FLAGS += -e --bootloader-invert-lines -w -v -b 500000
 
 ifdef PORT
   BSL_FLAGS += -p $(PORT)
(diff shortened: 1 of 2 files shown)


marknnd:
Hi Peretuset,
The automatic bootloading mode can work with baudrate 450000 or 115200, cannot work with baudrate 500000. Mean that using below script
BSL_FLAGS += -e --bootloader-invert-lines -w -v -b 450000 or
BSL_FLAGS += -e --bootloader-invert-lines -w -v -b 115200.

Regards,

peretuset:
@marknnd, can you please check that you using the OpenUSB rev.A1 board? We have tested it in various computers and operating systems and it worked well in all at 500000 bauds/s.

marknnd:
I run on OpenUSB Rev.A1, Openmote Rev.E.

peretuset:
What OS?
Pere

marknnd:
Ubuntu 14.04 32bits

peretuset:
Allright, thanks!
Ubuntu is working natively or on a VM?

marknnd:
On a VM, plug directly, no hub.

peretuset:
I see. Could you try on a Ubuntu machine that runs natively? On my side I will try with a Ubuntu VM to see if there is any effect. So far, we only tried with machines natively, but if there are some side effects we can reduce the baudrate to make sure it works well on all platforms.

marknnd:
After test on native Ubuntu PC, it still cannot work at 500000 baudrate. I think we need to reduce a little bit.

peretuset:
I've tried myself within a Ubuntu 16.04 VM (64 bit) and it work just fine at 500000; I tried it several times with various boards and it was reliable. I will try with other operating system configurations and computers and see if this is a general or specific to certain environments to help decide. In the mean time, please use 450000 baudrate to continue working if that works reliably for you.
By the way, what computer are you using (i.e., specs)?

marknnd:
I am using a Dell Desktop PC with CPU core i5.

msloth:
Many times when such issues arise it might be as simple as a poor USB
cable. I'd suggest trying with another cable, and for this pr to use the
slightly slower speed that will give less support issues at the cost of a
neglible slower flash speed.

peretuset:
100% agreed!
By the way, happy new year to you @msloth and to all... :smile:

simonduq:
Is this ready to get in?

peretuset:
I've just reduced the bootloader baudrate to be on the safe side, so it should be okay to merge in.

peretuset:
It seems to me that the build is failing but is not caused by the last commit. Can you please double-check?

simonduq:
Seems unrelated indeed. Restarted the job

peretuset:
Seems good now, feel free to merge in when ready.

simonduq:
👍

peretuset:
Is there a plan on merge this in any time soon?

simonduq:
Let's get this merged, thanks!

PR #2025, opened 2016-12-29
Override HTTP host/path lenght and updated http-socket example

alignan:
This PR allows to override the maximum host and path lenght used by the http-socket library, and adds reconnections attempts to the http-socket example.

### DIFF
--- a/core/net/http-socket/http-socket.h
+++ b/core/net/http-socket/http-socket.h
@@ -33,6 +33,18 @@
 
 #include "tcp-socket.h"
 
+#ifdef MAX_HTTP_SOCKET_CONF_PATHLEN
+#define MAX_PATHLEN MAX_HTTP_SOCKET_CONF_PATHLEN
+#else
+#define MAX_PATHLEN 80
+#endif
+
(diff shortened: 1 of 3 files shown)


core/net/http-socket/http-socket.h
```diff
@@ -33,6 +33,18 @@
...
+#define MAX_PATHLEN MAX_HTTP_SOCKET_CONF_PATHLEN
+#else
+#define MAX_PATHLEN 80
+#endif
+
+#ifdef MAX_HTTP_SOCKET_CONF_HOSTLEN
+#define MAX_HOSTLEN MAX_HTTP_SOCKET_CONF_HOSTLEN
+#else
+#define MAX_HOSTLEN 40
+#endif
```
greg-king5: I wonder about those configuration names.  It seems to me that more appropriate names are `HTTP_SOCKET_CONF_MAX_PATHLEN` and `HTTP_SOCKET_CONF_MAX_HOSTLEN`.

alignan: Agree, I'll make the changes

PR #2028, opened 2016-12-29
added support for 6-bytes link-layer address

joakimeriksson:
This PR adds the ability for the nbr table to use 6-bytes link-layer address. Should improve things for cases where Ethernet is used instead of 802.15.4 (8 or 2 bytes). Also fixed config for minimal-net to use the 6-bytes address as it make use of a TAP (ethernet) to hook into the IP-stack of the host.


simonduq:
👍

alignan:
:+1:

PR #2035, opened 2017-01-05
CSMA bugfix: do not use field on a struct that may have been deallocated

simonduq:
`free_packet` de-allocates a packet and, if this was the last packet in queue, also frees the neighbor. Hence, we need to store `n->transmissions` before `free_packet` in order to return it safely.


nvt:
:+1:

PR #2043, opened 2017-01-11
Fixed rime-tsch sync problem in Cooja

e120guru:
The rime-tsch example does not sync properly in cooja when using cooja motes. Compiling for instance a sky mote works.

As it turns out there seems to be lacking the directive to simulate the turnaround time in this example.

This PR adds a fix for this, after applying this fix cooja motes syncronize as expected even when using cooja motes.


yatch:
You should be right. `COOJA_CONF_SIMULATE_TURNAROUND` needs to be set with 0 when you're using TSCH witt Cooja mote as I mentioned in the Introduction section of https://github.com/contiki-os/contiki/pull/1661.

Sorry for the inconvenience. I should have added something about `COOJA_CONF_SIMULATE_TURNAROUND` into README of TSCH.  I'll send a PR to update the README.

examples/rime-tsch/project-conf.h
```diff
@@ -85,4 +85,10 @@
 #undef TSCH_LOG_CONF_ID_FROM_LINKADDR
 #define TSCH_LOG_CONF_ID_FROM_LINKADDR(addr) ((addr) ? (addr)->u8[LINKADDR_SIZE - 2] : 0)
 
+#if CONTIKI_TARGET_COOJA
+#define COOJA_CONF_SIMULATE_TURNAROUND 0
+#define COOJA_CONF_TRANSMIT_ON_CCA 1
```
yatch: This one is needed as well to run TSCH with Cooja...? As per `platform/cooja/dev/cooja-radio.h`, the default value of `COOJA_TRANSMIT_ON_CCA` is `1`... I may miss something.

simonduq:
Right, would be a good idea to centralize the TSCH-related configs somewhere, at least in the readme! Note that we have some both at the beginning and the end of the project-conf examples..

simonduq:
I agree the def of `COOJA_CONF_TRANSMIT_ON_CCA` does not seem needed. Other than that, thanks for the catch!

e120guru:
PR updated, removed redundant line

simonduq:
thanks. if you could just squash the two commits into one, this is ready to merge!

e120guru:
PR updated

simonduq:
Thanks! 👍

PR #2045, opened 2017-01-11
Improve pic32 spi

francois-berder-imgtec:


francois-berder-imgtec:
SPI##x##BUF is defined as volatile in microchip pic32 headers. So no need for this pull request.

PR #2046, opened 2017-01-11
Bugfix in ringbufindex

yatch:
There are two bugs in `ringbufindex`. They are similar.

While `ringbufindex_peek_put()` is supposed to return the index of the next element to be added, it returns the next index of the adding element. `ringbufindex_peek_get()` returns the next index of the first element.

Travis got green on my account. However, `19-z1-rpl-tsch.csc.flaky` failed, that is one of non-regression tests of TSCH. Since this one fails even on the tip of the master branch, I would say, this PR doesn't cause the failure...

A regression test for `ringbufindex` has been introduced under `regression-tests/03-base` as `04-ringbufindex`. Without the fixes, two of them fails.

```shell
Random seed: 1
549000 node-1 Rime started with address 0.1.0.1.0.1.0.1
549000 node-1 MAC 00:01:00:01:00:01:00:01 sicslowpan/CSMA/nullrdc, channel check rate 1000 Hz
549000 node-1 Tentative link-local IPv6 address fe80:0000:0000:0000:0201:0001:0001:0001
549000 node-1 Tentative global IPv6 address fd00:0000:0000:0000:0201:0001:0001:0001
549000 node-1 Starting 'ringbuf-index.c test'
549000 node-1 =check-me= SUCEEDED - Init
549000 node-1 =check-me= SUCEEDED - Put
549000 node-1 =check-me= FAILED   - PeekPut: exit at L104
549000 node-1 =check-me= SUCEEDED - Get
549000 node-1 =check-me= FAILED   - PeekGet: exit at L168
549000 node-1 =check-me= SUCEEDED - Size
549000 node-1 =check-me= SUCEEDED - Elements
549000 node-1 =check-me= SUCEEDED - Full
549000 node-1 =check-me= SUCEEDED - Empty
TEST FAILED
Test ended at simulation time: 10149000
 FAIL ಠ_ಠ
```

All the tests pass on this branch.
```shell
$ make 04-ringbufindex.testlog
Running test 04-ringbufindex with random Seed 1: ..... OK

$ cat 04-ringbufindex.testlog
Random seed: 1
549000 node-1 Rime started with address 0.1.0.1.0.1.0.1
549000 node-1 MAC 00:01:00:01:00:01:00:01 sicslowpan/CSMA/nullrdc, channel check rate 1000 Hz
549000 node-1 Tentative link-local IPv6 address fe80:0000:0000:0000:0201:0001:0001:0001
549000 node-1 Tentative global IPv6 address fd00:0000:0000:0000:0201:0001:0001:0001
549000 node-1 Starting 'ringbuf-index.c test'
549000 node-1 Run unit-test
549000 node-1 ---
549000 node-1 =check-me= SUCEEDED - Init
549000 node-1 =check-me= SUCEEDED - Put
549000 node-1 =check-me= SUCEEDED - PeekPut
549000 node-1 =check-me= SUCEEDED - Get
549000 node-1 =check-me= SUCEEDED - PeekGet
549000 node-1 =check-me= SUCEEDED - Size
549000 node-1 =check-me= SUCEEDED - Elements
549000 node-1 =check-me= SUCEEDED - Full
549000 node-1 =check-me= SUCEEDED - Empty
549000 node-1 =check-me= DONE
TEST OK
Test ended at simulation time: 549000
```

### DIFF
--- a/regression-tests/03-base/code/test-ringbufindex.c
+++ b/regression-tests/03-base/code/test-ringbufindex.c
@@ -1,0 +1,366 @@
+/*
+ * Copyright (c) 2017, Yasuyuki Tanaka
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <stdio.h>
+
+#include "contiki.h"
+#include "unit-test.h"
+
+#include "lib/ringbufindex.h"
+
+PROCESS(test_process, "ringbuf-index.c test");
+AUTOSTART_PROCESSES(&test_process);
+
+static struct ringbufindex ri;
+static const uint8_t ri_size = 2;
+
+static void
+test_print_report(const unit_test_t *utp)
+{
+  printf("=check-me= ");
+  if(utp->result == unit_test_failure) {
+    printf("FAILED   - %s: exit at L%u\n", utp->descr, utp->exit_line);
+  } else {
+    printf("SUCCEEDED - %s\n", utp->descr);
+  }
+}
+
+UNIT_TEST_REGISTER(test_ringbufindex_init, "Init");
+UNIT_TEST(test_ringbufindex_init)
+{
+  UNIT_TEST_BEGIN();
+
+  ringbufindex_init(&ri, ri_size);
+
+  UNIT_TEST_ASSERT(ri.mask == ri_size - 1 &&
+                   ri.put_ptr == 0 &&
+                   ri.get_ptr == 0);
+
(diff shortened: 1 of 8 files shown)


neophob:
Great catch. I saw that the JN platform uses the ringbufindex code as well, what's the impact there?

yatch:
@neophob I don't think this change has any impact on `platform/jn516x/dev/micromac-radio.c`.

To my understanding, you don't have any problem with this bug as long as you don't use an index returned by `ringbufindex_get()`. The JN platform gets an index with `ringbufindex_peek_get()` and uses `ringbufindex_get()` just to remove the first element in the ringbuf. In fact, `ringbufindex_get()` removes a different element from one returned by `ringbufindex_peek_get()`. But, still, consistency is maintained.

simonduq:
Right, if I'm correct it looks like everything was off-by-one, causing no problem (jn nor tsch), but that is definitely unwanted indeed  :p 
How does ` 19-z1-rpl-tsch.csc.flaky` fail? It should randomly fail to link. But when it links, it normally runs successfully (the run is not flaky).

yatch:
@simonduq Regarding `19-z1-rpl-tsch.csc.flaky`, the number of "routing links" on node-1 doesn't get to 8 within the simulation time. I think I've found a bug, which is not in `ringbufindex`. I'll send a PR later.

simonduq:
Right, looking forward to the PR then :) I guess the bug simply slipped in now with the test disabled :/

yatch:
@simonduq I've created a new PR, https://github.com/contiki-os/contiki/pull/2079.  My conclusion is, `19-z1-rpl-tsch.csc.flaky` fails because some RPL nodes fail to receive broadcast frames including DIOs. This is caused by a [bug](https://github.com/contiki-os/mspsim/pull/8) in the frame filtering feature of the simulated CC2420 device.

At first, I noticed that the test failed on the commit of bfef0b5d6ab49ed347f3087d26938c4ff7b1f5a4, although it succeeded before that. So, I suspected the adaptive scheduling, but I was wrong. I guess, some timing change by bfef0b5d6ab49ed347f3087d26938c4ff7b1f5a4 triggers the test failure by chance.

A TSCH node is initialized with frame filtering disabled. As long as the frame filtering is disabled, the node doesn't suffer from the bug of CC2420. The filter is enabled once it sends a unicast (ACK-required) frame. Therefore, if it receives a broadcasted DIO before sending any unicast frame, it can join a DODAG. Otherwise, it cannot not. This happens in a failed test of  `19-z1-rpl-tsch.csc.flaky`. I'm confused why the frame filtering needs to be turned on and off in TSCH, though.

Anyway, the failure of the test is nothing to do with `ringbufindex`.

simonduq:
Could it be the recent changes to MAC header that result in frames that the emulated cc2420 filters out? 
Maybe it would be wiser to unset `TSCH_HW_FRAME_FILTERING` by default. It's already manually disabled on the cc2538 platforms in some `project-conf.h` files.

yatch:
@simonduq In the first place, why does TSCH needs the frame filtering disabled during scanning and waiting for an ACK...? (it's totally off-topic, by the way... 😅)

simonduq:
On some platforms (at least cc2420, which was our initial development platform), the radio would filter out enhanced ACKs, probably due to the len different from 3.

Scanning should work both with and without filtering.

yatch:
@simonduq Thank you for the information! Now I understand what the following comment means, that is found in `tsch-slot-operation.c`.
```C
/* Entering promiscuous mode so that the radio accepts the enhanced ACK */
```

regression-tests/03-base/code/test-ringbufindex.c
```diff
@@ -0,0 +1,366 @@
...
+static const uint8_t ri_size = 2;
+
+static void
+test_print_report(const unit_test_t *utp)
+{
+  printf("=check-me= ");
+  if(utp->result == unit_test_failure) {
+    printf("FAILED   - %s: exit at L%u\n", utp->descr, utp->exit_line);
+  } else {
+    printf("SUCEEDED - %s\n", utp->descr);
```
greg-king5: Typo: `SUCEEDED` -> `SUCCEEDED`

yatch: Thanks! I've fixed this typo 😄

simonduq:
Good. I've used the fix in multiple testbeds, successfully.
Could you enable in Travis the test added in https://github.com/contiki-os/contiki/pull/2108 ?

yatch:
@simonduq Sure, I can do that today :-)

yatch:
Added one commit enabling TSCH regression testing. Travis ran [the test](https://travis-ci.org/contiki-os/contiki/jobs/209259777) and got green.

simonduq:
👍

simonduq:
Thanks!

PR #2052, opened 2017-01-13
Make RPL route lifetime RFC compliant

laurentderu:
RPL RFC states that if the path lifetime in a DAO transit option is set to 0xFF, it means that the path lifetime is infinite. On the other hand, a lifetime unit of 0xFFFF has not peculiar meaning.

<pre>
   Path Lifetime: 8-bit unsigned integer.  The length of time in
         Lifetime Units (obtained from the Configuration option) that
         the prefix is valid for route determination.  The period starts
         when a new Path Sequence is seen.  A value of all one bits
         (0xFF) represents infinity.
</pre>

### DIFF
--- a/core/net/rpl/rpl-private.h
+++ b/core/net/rpl/rpl-private.h
@@ -145,8 +145,14 @@
 /* Special value indicating immediate removal. */
 #define RPL_ZERO_LIFETIME               0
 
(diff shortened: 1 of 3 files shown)


alignan:
:+1:

PR #2053, opened 2017-01-13
Use RPL_DAG_LIFETIME when computing dag lifetime

laurentderu:
When updating the dag lifetime, the multiplication constant RPL_DAG_LIFETIME must be used (or it has no use at all :) )

### DIFF
--- a/core/net/rpl/rpl-dag.c
+++ b/core/net/rpl/rpl-dag.c
@@ -1542,7 +1542,7 @@ rpl_process_dio(uip_ipaddr_t *from, rpl_dio_t *dio)
   }
 
   /* The DIO comes from a valid DAG, we can refresh its lifetime */
-  dag->lifetime = (1UL << (instance->dio_intmin + instance->dio_intdoubl)) / 1000;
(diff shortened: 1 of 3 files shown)


alignan:
:+1:

laurentderu:
Forgot to update the event timing in the non-regression test related to multi dag, if the lifetime is multiplied by 3 so must be the test...

nvt:
:+1:

arurke:
I'll just add a related comment/question here on the DAG lifetime (which is also related to #912 by @laurentderu ).

I noticed during testing that a node never leaves a DAG due to DAG lifetime. The function which has a chance to do this is rpl_purge_dags(), added in #912, however it requires the "joined" flag to be 0, which can only happen if the node has switched to an alternative DAG (or has become root itself). Hence a node will forever be joined to the single DAG even if its lifetime is expired.

Now, this might not necessarily be unintended or bad behavior. The only effect I can think of is that this blocks the node from sending DIS messages in search for a new DAG, however glancing at the RFC I cannot see any req. for this to actually happen either.

PR #2054, opened 2017-01-13
Reject parents with invalid rank

laurentderu:
A parent with a rank lower than the DODAG Root rank can not exist by definition and therefore should never be selected.

(Sadly I don't remember which condition triggered the bug, but I guess it does not hurt to add this extra check)


nvt:
Looks good, but could you prepend the debug message with "RPL: "?

laurentderu:
Done

alignan:
:+1:

simonduq:
Why checking this in `best_parent` and not when adding / updating a node's rank?

laurentderu:
@simonduq honestly I did this fix a long time ago and I don't remember if there was a real motivation to do it there or if it was just easier. I believe it's the latter though, otherwise the rank should be checked at several other place.

PR #2057, opened 2017-01-13
Do RPL probing on all the known DAG

laurentderu:
The current RPL probing function does probing only in the current dag. With this PR, all the known dodag are probed in a round-robin manner. This improves a lot DODAG switching.

The code could be made conditional using RPL_MAX_DAG_PER_INSTANCE and only used if the number of dag is >= 2

### DIFF
--- a/core/net/rpl/rpl-timers.c
+++ b/core/net/rpl/rpl-timers.c
@@ -457,11 +457,29 @@ get_probing_target(rpl_dag_t *dag)
   return probing_target;
 }
 /*---------------------------------------------------------------------------*/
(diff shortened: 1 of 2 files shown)


simonduq:
Nice feature, but I feel this should be up to the probing target selection function. How about using the instance as parameter and letting the function do the round-robin on dags?

laurentderu:
That's a good remark, and moving it to the probing function allows one to use another probing function that would use something more smart than a simple round robin between the DAGs.

simonduq:
yep!

PR #2058, opened 2017-01-13
Add missing UIP_LLH_LEN

laurentderu:
At a few places, UIP_BUFSIZE is used without removing UIP_LLH_LEN, which could lead to buffer overflow. 
Most of the time the macros used to calculate header or packet size already take into account UIP_LLH_LEN, these are the only places where UIP_BUF_SIZE is used directly and UIP_LLH_LEN is not subtracted.


alignan:
:+1:

PR #2060, opened 2017-01-13
Properly update IPv6 proto in remove_ext_hdr()

laurentderu:
Instead of explicitly update the proto field of the IPV6 packet after calling remove_ext_hdr() which is error prone, this PR updates the field inside remove_ext_hdr() directly.


g-oikonomou:
This seems to be something we need to look at. @simonduq @nvt any thoughts?

PR #2062, opened 2017-01-13
Adding uip_ds6_defrt_list_head()

laurentderu:
Trivial PR adding the missing uip_ds6_defrt_list_head() function to retrieve the first element of the default router list.


nvt:
The other function in this module that returns a list head is called uip_ds6_route_head, so it would be good for consistency if both functions either have "list" in the name or not.

laurentderu:
I removed the _list part of the function name for default router.

nvt:
:+1:

PR #2064, opened 2017-01-13
Add basic support for static routes

laurentderu:
This PR add support for static routes. The difference between static routes and current routes in Contiki is that the static routes do not depend on an entry in the neighbor table or their link layer address resolution.
Static routes can therefore be created at any time, even if the configured next hop is not reachable yet, and will not be deleted if the neighbor table is full or if the next hop is temporarily unreachable.
Static routes will be needed to support Route Information Option (RIO) in Router Advertisement messages (which is part of an upcoming PR).

For this implementation, instead of creating a complete new set of configuration data, I kept the route object and used the content of the neighborroutes field to distinguish between static routes and dynamic routes. If the field is set to NULL, it means that the route entry is a static one.

### DIFF
--- a/core/net/ipv6/uip-ds6-route.c
+++ b/core/net/ipv6/uip-ds6-route.c
@@ -39,6 +39,9 @@
  *    Routing table manipulation
  */
 #include "net/ipv6/uip-ds6.h"
+#if UIP_CONF_IPV6_RPL
+#include "net/rpl/rpl-private.h"
+#endif
 #include "net/ip/uip.h"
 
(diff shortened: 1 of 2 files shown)


laurentderu:
Forgot to squash in the support for RPL

KaneRoot:
Why has this PR not making its way to the master branch yet? It seems important. Please, provide a feedback.

laurentderu:
People have moved to Contiki-NG so I don't think it will ever be merged. I will in the near future make a PR in Contiki-NG with that functionality.

PR #2072, opened 2017-01-16
CC26xx: enable phase lock in ContikiMAC

atiselsts:
Phase lock did not work on CC26xx with the default settings. in particular, the guard time was insufficient as the CC26xx radio chip is quite slow. This pull adds new ContikiMAC phase lock configuration settings and enables it on this platform.

Tested in a star network with 5 nodes. Larger scale testing might be helpful to better find the optimal value of the settings in terms of energy/robustness tradeoff.


atiselsts:
With this,  the `cc26xx-web-demo` fails to compile because of running out of RAM. Not sure if there's an easy way to fix it, and we might need to reduce the web demo memory usage in any case: see pull request #2089.

atiselsts:
Updated the `cc26xx-web-demo` to explicitly disable phase locking; now it can be linked and Travis is green.

examples/cc26xx/cc26xx-web-demo/project-conf.h
```diff
@@ -53,6 +53,9 @@
 #define NBR_TABLE_CONF_MAX_NEIGHBORS          5
 #define UIP_CONF_MAX_ROUTES                   5
 #define UIP_CONF_TCP_MSS                    128
+/* Additionally disable phase optimization to save RAM */
+#undef CONTIKIMAC_CONF_WITH_PHASE_OPTIMIZATION
```
g-oikonomou: The undef is not required for this platform

platform/srf06-cc26xx/contiki-conf.h
```diff
@@ -78,7 +78,12 @@
 
 /* Configure ContikiMAC for when it's selected */
 #define CONTIKIMAC_CONF_WITH_CONTIKIMAC_HEADER  0
-#define CONTIKIMAC_CONF_WITH_PHASE_OPTIMIZATION 0
+/* Phase optimization: enable, but increase the guard and max strobe time 2 times */
+#ifndef CONTIKIMAC_CONF_WITH_PHASE_OPTIMIZATION
+#define CONTIKIMAC_CONF_WITH_PHASE_OPTIMIZATION 1
```
g-oikonomou: Are you 100% certain that enabling phase optimisation won't lead to a barrage of issues? :) I'm willing to take your word for it here.

g-oikonomou:
I'm keen to merge this. Just waiting for minor adjustments/responses as per my comments.

atiselsts:
I consider this an experimental feature, tested on the table for a few minutes with a few motes. Would be more happy if someone reported results from a real system - as I have done with the code from other pull requests recently submitted. (We're not using ContikiMAC in our production systems, so this PR is not a priority...)

examples/cc26xx/cc26xx-web-demo/project-conf.h
atiselsts: Seemy previous commit - I think it ran out of RAM when trying to link the web demo with phase lock enabled.

g-oikonomou:
Yes, I understand you are considering this experimental. Which is why I think that `contiki-conf.h` should default to disabling phase optimisation and then the web demo's `project-conf.h` doesn't need to do anything. Also, to change the configuration in `project-conf.h`, you only need to `#define`, you don't need to `#undef` first.

Perhaps we can demonstrate that it can be enabled at own risk in the simple demo.

Issue #1858, opened 2016-09-13
cc26xx-web-demo and very-sleepy-demo stop working after 600 seconds


magicm76:
Hello, 
I compiled examples (as they are) in contiki 3.x (DriverLib v0.46593) on different boards (SmartRF06 with cc2650em and SensorTag2.0) and: 
- cc26xx-demo is working well
- cc26xx-web-demo is rebooting after about 600 sec.
- very-sleepy-demo hangs after about 600 sec. 
I compiled example in Instant Contiki, with commands: 
- make TARGET=srf06-cc26xx name.bin CPU_FAMILY=cc26xx (for SmartRF06 with cc2650em)
- make TARGET=srf06-cc26xx BOARD=sensortag/cc2650 name.bin CPU_FAMILY=cc26xx (for SensorTAG)
Except rebooting and hanging applications seems working well. I looked for same problem in internet but I didn't find any solution

Thank You

schuschu:
I cannot tell you why it is happening but it I can observe the same behavior with a CC2650 running the web-demo and one as slip device (with 6LBR).

arurke:
Just for reference: I saw same issue reported on TI forums
http://e2e.ti.com/support/wireless_connectivity/zigbee_6lowpan_802-15-4_mac/f/158/p/544209/1991513

Seems they narrowed it down to COAP and/or MQTT app.

magicm76:
also the post in e2e.ti forum is mine, but is without answers

g-oikonomou:
Hi

I'll assume you are using latest version from master. Could you experiment with CoAP only and also making sure that the net-uart is disabled and report back?

P.S. (You don't really need to specify `CPU_FAMILY`)

schuschu:
Ok i tried on a launchapd to rule out some sensor process or similar: with udp, 6lbr, coap and mqtt it does also reset (as do my sensortags). when only enabling coap it does not occur. I'll try to do some more tests today.

magicm76:
Hi, 
I done more tests and I had a lot of problem but I found 1 configuration stable: 
Now, with master version of contiki, with SmartRF06 and cc2650EM as router with rpl-border-router, with motes as web-demo with only COAP active, with all sensor disabled in sensortag, system is stable. both SmartRF06 and sensortag motes seems stable. 
If I use sensortag web-demo as it is, with sensors active, motes reboot after near 600 seconds (SmartRF works well as it is in contiki example)
[...]

PR #2073, opened 2017-01-16
Fix semantic of UIP_ND6_SEND_NA and add UIP_ND6_SEND_NS

laurentderu:
The compilation flag UIP_ND6_SEND_NA is used to disable NS and NA messages handling. However the name is misleading as actually we mean to disable NS sending and so disable NUD and DAD processes. Also it is not possible currently to disable them yet to still respond to incoming NS messages from full NDP peers.

This FR renames UIP_ND6_SEND_NA into UIP_ND6_SEND_NS in most places and introduces UIP_ND6_SEND_NA to still support incoming NS handling. By default on a RPL host, SEND_NS is deactivated and SEND_NA is still activated.

There was also a potential bug in rpl-dag.c where the compilation flag definition was tested instead of its actual value, and so the code was always included.

This PR should also fix one of the issues reported in https://github.com/contiki-os/contiki/issues/1858


simonduq:
👍

nvt:
:+1:

PR #2076, opened 2017-01-17
Native monotonic timer

laurentderu:
Currently on native platform we use gettimeofday() to create the time of all the timers, however gettimeofday() is not an absolute monotonic time source. When the system clock is updated (either because of drift, DST or user configuration) the time returned is shifted. This breaks completely the Contiki timer, resulting in timers already expired before even starting or times scheduled billions of seconds later in time...

With this PR, I switched to clock_gettime(CLOCK_MONOTONIC) on linux and mach_absolute_time() on MacOS-X. These two functions are guaranteed to always return a monotonic time, independent of any clock change on the platform.

### DIFF
--- a/platform/native/Makefile.native
+++ b/platform/native/Makefile.native
@@ -42,4 +42,8 @@ CURSES_LIBS ?= -lncurses
 
 TARGET_LIBFILES += $(CURSES_LIBS)
 
+ifeq ($(HOST_OS),Linux)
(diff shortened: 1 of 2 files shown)


pengi:
For the record, clock_gettime(CLOCK_MONOTONIC, ...) is available on mac too. No extra libraries is needed though, but it would make it possible to use the same implementation on both linux and mac.

PR #2089, opened 2017-01-24
Add stack usage checker feature on CC26xx, Z1, Sky, Exp5438 and Wismote platforms

atiselsts:
This patch adds library code for estimation on run-time stack usage with the aim to make software development for Contiki more robust.

Using the estimator is as simple as:
  printf("stack usage: %u permitted: %u\n",
         stack_checker_get_usage(), stack_checker_get_reserved_size());
For more, see `examples/stack/main.c`.

The idea behind the implementation is to pre-fill the stack region at the system's initialization with a known patter and then check the boundaries of that pattern during run-time.

When evaluating the checker on some of the example applications in Contiki I was not able to find any that overflows the stack region immediately after starting up. However, `cc26xx-web-demo` on TI SensorTag likely does that as soon it has to do any serious job such as handling some requests. Immediately after startup there's less than 100 bytes free, as stack usage is 560 bytes from the maximal 658 bytes. Meanwhile, functions such as `coap_notify_observers_sub()` in the CoAP implementation use literally hundreds of bytes of stack space on their own.

The amount of stack "reserved" on CC26xx by default by the linker script (`cpu/cc26xx-cc13xx/cc26xx.ld`) is just 0x200 or 512 bytes, which is woefully inadequate in my opinion. (This "reservation" does not ensure the stack does not overflow during runtime, it just creates a compile-time error if there are less-than-this bytes free after the end of the `.data` segment). For our projects in University of Bristol we have changed that to 0x500 or 0x700 depending on the project. In particular, in a real deployment that uses the CoAP / 6LoWPAN / TSCH  / IEEE 802.15.4 network stack I've observed stack usage up to 1750 bytes using this checker library.


laurentderu:
👍 from me, this is really an interesting feature !

Why are you using bytes to fill and check the stack however, a cpu word (16/32bits) would be more efficient.

Could you also add a simpler test to test the bottom of the stack so it could be added into the main loop of Contiki (with a compilation option of course) and so detect a soon as possible a stack overflow without having to scan the whole stack.

atiselsts:
Changed the logic to use a 16-bit word when filling and checking the stack, thanks for the feedback. 16-bit as I'd prefer a single version of the code for both 16-bit and 32-bit CPUs.

Including this in the main loop is possible, are platform maintainers ok with that? @g-oikonomou @alignan

oliverschmidt:
I personally would prefer "stack check" over "stack checker"...

alignan:
I would be more on-board if this was also extended to the CC2538-based platforms :smile:

atiselsts:
Hi Antonio,

deciding the RAM size and address on cc2538 looks tricky, as they are dependent on many defines in  `dev/cc2538-dev.h` and the linker script `cc2538.lds`.

When I now think about this, maybe it's better to make this less accurate but also less platform dependent and just assume that the stack bottom is an address of a local variable in the `main()` function, rather than requiring a platform-specific `#define`.

alignan:
Probably yes, I think this is a great feature and somehow everyone ends up doing their own way, that's why I was hopping for a broader approach.

core/sys/stack-checker.c
```diff
@@ -0,0 +1,104 @@
...
+  while(*p != STACK_FILL) {
+    p++;
+    /* Make sure WD is not trigerred */
+    watchdog_periodic();
+  }
+
+  while(*p == STACK_FILL) {
+    p++;
+    /* Make sure WD is not trigerred */
+    watchdog_periodic();
```
greg-king5: It would be faster to "pet" the watchdog _before_ the first loop, and _after_ the second loop.

greg-king5: >It would be faster to "pet" the watchdog _before_ the first loop, and _after_ the second loop.

Hmpf!!  Maybe not -- some watchdogs are set for only one second -- code must placate them very often!

atiselsts:
Added CC2658DK and Zoul to the platforms. Now `#define STACK_ORIGIN` is optional; if not defined, the stack origin is assumed. So it's very easy to add more platforms, at the minimum just one `#include` and a call to `stack_check_init()` from the `main()`.

Also renamed the library to "stack_check" and took the watchdog pinging out of the `for` loops.

I did not add the periodic checks in the main loops yet and would prefer to leave that to maintainers (or at least get explicit ok) as that might be quite intrusive I think.

arurke:
I'll just add that I tested it on a cc1310 custom board, unsurprisingly it worked fine. Regarding the 500 byte reservation you mention, is this it? 😕 
```
_Min_Stack_Size = 0x100;
```

atiselsts:
Thanks and yes, that's what I meant. So in mainline Contiki that's actually only 256 bytes reserved for the stack, not 512 (0x200) as was in our fork.

bthebaudeau:
- `watchdog_periodic()` should not be useful in `stack_check_init()`.
- `memset()` would be more efficient than your loop in `stack_check_init()`, provided that it is available on all the targeted platforms.
- `stack_check_init()` should just fill the stack from bottom to top, without this `&local` hack, which could fail since `local` and `p` may be placed in any order in the stack relatively to each other. Just define symbols for bottom and top of stack, and make sure that `stack_check_init()` does not use the stack (which could be an issue on some platforms). Another, simpler way, would be to just drop `local` and to instead use the address before that of `p`.
- The stack is already in use in `main()`, so calling `stack_check_init()` from `main()` is a destructive operation (though not as much as it would be with a true stack fill, thanks to your `&local` hack), and it does not allow to trace the first operations using the stack. For instance, on CC2538, it should be called at the beginning of `reset_handler()` in `cpu/cc2538/startup-gcc.c`, which would also avoid duplicating this call in all the CC2538 `contiki-main.c` files.
- `stack_check_init()` slightly slows down boot, so it should actually do something only if some dedicated configuration option is enabled (debug vs. release). And your other stack functions should be adapted accordingly to do nothing and return neutral results (i.e. that would not be interpreted as a stack overflow).
- In the linker scripts, `_end` can be used to mean various things. It would be better to use a more meaningful naming. `_ebss` is already defined to the same value in the linker scripts where you have defined `_end`, but you could also use names such as `_stack` or `_edatabss`.
- What is the point of the empty `examples/stack/project-conf.h`?
- `trigerred` -> `triggered`.

atiselsts:
Hello, let me address your comments  @bthebaudeau.

>    watchdog_periodic() should not be useful in stack_check_init().

Because at the point where it's called watchdog is not started yet? Not sure if that's guaranteed. And any, someone might want to reinitialize this library for some reason in the future, so better to be safe.

>    memset() would be more efficient than your loop in stack_check_init(), provided that it is available on all the targeted platforms.

Not sure how to call it in a safe way, as a function call requires the stack to remain intact in order to jump back to the calling function.

>  stack_check_init() should just fill the stack from bottom to top, without this &local hack, which could fail since local and p may be placed in any order in the stack relatively to each other. Just define symbols for bottom and top of stack, and make sure that stack_check_init() does not use the stack (which could be an issue on some platforms). Another, simpler way, would be to just drop local and to instead use the address before that of p.

True, don't need the `local` variable at all. Updated.

>    The stack is already in use in main(), so calling stack_check_init() from main() is a destructive operation (though not as much as it would be with a true stack fill, thanks to your &local hack), and it does not allow to trace the first operations using the stack. For instance, on CC2538, it should be called at the beginning of reset_handler() in cpu/cc2538/startup-gcc.c, which would also avoid duplicating this call in all the CC2538 contiki-main.c files.

The idea was to call this after serial port is already initialized, so the user can do some printing in the init function(that's why there's some printf code left there, only commented out). It's not overwritting the stack used by `main()`.

>  stack_check_init() slightly slows down boot, so it should actually do something only if some dedicated configuration option is enabled (debug vs. release). And your other stack functions should be adapted accordingly to do nothing and return neutral results (i.e. that would not be interpreted as a stack overflow).

Good point, done. `STACK_CHECK_CONF_ENABLED` controls this now. It's left enabled by default, as I think defensive programming is the way to go.

>  In the linker scripts, _end can be used to mean various things. It would be better to use a more meaningful naming. _ebss is already defined to the same value in the linker scripts where you have defined _end, but you could also use names such as _stack or _edatabss.

That's a standard name on msp430. The difference is that msp430 linker scripts are no included in Contiki itself, while CC2538 and CC26xx are. So it's either using `_end` or different names for different platforms, which is ugly for other reasons...

Additionally I reverted back to filling the stack by a single byte. As the checker can now be easily disabled, I prefer to go with prettier code rather than small improvement in boot speed.

g-oikonomou:
Where are we here? Overall, this is a desirable feature but keen on reading more thoughts in terms of its actual implementation. @bthebaudeau @simonduq @laurentderu

bthebaudeau:
@atiselsts

> >    watchdog_periodic() should not be useful in stack_check_init().
> 
> Because at the point where it's called watchdog is not started yet? Not sure if that's guaranteed. And any, someone might want to reinitialize this library for some reason in the future, so better to be safe.

If `stack_check_init()` may be called from several places, then it may set `stack_origin` to different values, so `stack_check_get_usage()` may then report inaccurate values.

In `platform/zoul/contiki-main.c`, you call `stack_check_init()` after `watchdog_init()`. My comment was assuming that this call is moved before `main()` as suggested in one of my next comments. So yes, one reason is that the watchdog has not yet been started at this point. But my comment is still valid without this assumption, because another reason is that the `.data` and `.bss` sections can be safely initialized without taking care of the watchdog. I mean: the stack is never gigantic on MCUs, so a `memset()` over it should be pretty fast, and the watchdog timeouts are usually very long. Otherwise, no buffer could be initialized without triggering the watchdog.

Moreover, `stack_check_init()` resets the watchdog only before and after the actual stack initialization, which is assuming that there is a long operation that could take place either before or after this function without resetting the watchdog, rather than assuming that this function could take too much time by itself.

Lastly, the call to `stack_check_init()` cannot be moved before `main()` if `stack_check_init()` resets the watchdog, because `watchdog_init()` is called only in `main()`.

> >    memset() would be more efficient than your loop in stack_check_init(), provided that it is available on all the targeted platforms.
> 
> Not sure how to call it in a safe way, as a function call requires the stack to remain intact in order to jump back to the calling function.

This actually depends on the MCU architecture and on the ABI. This is indeed true in some cases. But your loop does not guarantee anything either, as the compiler may generate code using the stack below `p` for it, even if this is unlikely. A reasonably safe way would be to consider that some stack area below `p` should not be touched. The only strictly safe way is to use assembly code, preferably at the beginning of the reset handler.

> >  stack_check_init() should just fill the stack from bottom to top, without this &local hack, which could fail since local and p may be placed in any order in the stack relatively to each other. Just define symbols for bottom and top of stack, and make sure that stack_check_init() does not use the stack (which could be an issue on some platforms). Another, simpler way, would be to just drop local and to instead use the address before that of p.
> 
> True, don't need the `local` variable at all. Updated.

You're still filling the stack from top to bottom, which is usually slower.

The result of the 1st iteration of your loop is undefined because the `--` and the `=` operations target the same address (that of `p`). This loop should exclude `&p`.

> >    The stack is already in use in main(), so calling stack_check_init() from main() is a destructive operation (though not as much as it would be with a true stack fill, thanks to your &local hack), and it does not allow to trace the first operations using the stack. For instance, on CC2538, it should be called at the beginning of reset_handler() in cpu/cc2538/startup-gcc.c, which would also avoid duplicating this call in all the CC2538 contiki-main.c files.
> 
> The idea was to call this after serial port is already initialized, so the user can do some printing in the init function(that's why there's some printf code left there, only commented out).

This is not required: it is only a debug `printf()` to debug the stack debug code. :wink:

> It's not overwritting the stack used by `main()`.

No, but it's destroying the stack usage information regarding what occurred before `main()`.

> >  stack_check_init() slightly slows down boot, so it should actually do something only if some dedicated configuration option is enabled (debug vs. release). And your other stack functions should be adapted accordingly to do nothing and return neutral results (i.e. that would not be interpreted as a stack overflow).
> 
> Good point, done. `STACK_CHECK_CONF_ENABLED` controls this now.

Good. However, it would be better to move the empty definitions into the header file as macros, so that they can just be optimized away.

> It's left enabled by default, as I think defensive programming is the way to go.

This code does nothing more than waste time if the `stack_check_*()` functions are not called, which is the case everywhere except in this example, so it should be disabled by default.

> >  In the linker scripts, _end can be used to mean various things. It would be better to use a more meaningful naming. _ebss is already defined to the same value in the linker scripts where you have defined _end, but you could also use names such as _stack or _edatabss.
> 
> That's a standard name on msp430. The difference is that msp430 linker scripts are no included in Contiki itself, while CC2538 and CC26xx are. So it's either using `_end` or different names for different platforms, which is ugly for other reasons...

At least on ARM, `_end` is usually the end of the stuff before the heap, i.e. the beginning of the heap (if the heap is used). It would be easy to use the same meaningful name for all platforms. E.g. for MSP430, you could use `--defsym=_stack=_end` with `ld` (if the `_stack` symbol name is not already used).

> Additionally I reverted back to filling the stack by a single byte. As the checker can now be easily disabled, I prefer to go with prettier code rather than small improvement in boot speed.

Agreed.

atiselsts:
Hi, thanks for the in-depth discussion - appreciated even if i don't agree with all.

Let me clarify the intention. The goal of patch is to allow easier debugging and fault detection of user applications. Therefore checking what happens with the stack at the start of and before `main()` is not relevant from my perspective. If a platform has code that overflows the stack in the first few lines of `main()` or before that, that platform should not even be part of Contiki anyway, no? The stack usage there is not measured for the same reason, say, `energest` energy usage is not measured right from the very start. Can you think of any counterexamples for this?

> At least on ARM, _end is usually the end of the stuff before the heap, i.e. the beginning of the heap (if the heap is used). It would be easy to use the same meaningful name for all platforms. E.g. for MSP430, you could use --defsym=_stack=_end with ld (if the _stack symbol name is not already used).

Agree about what `_end` means and added a comment in the loader scripts to clarify that. Still, I think that `_end` symbol should be added for cc2650 and cc2538, even if separately from this patch (but platform maintainers can override me of course). No one at the moment knows what the symbol `_stack` means, but many programmers do known [what `_end` means](http://man7.org/linux/man-pages/man3/end.3.html).
And it's not just msp430 that has the `_end` symbol - for example, the AVR platforms and x86 (the "native" platform) do have it too.

> The result of the 1st iteration of your loop is undefined 

Ok, that's fair point, this looks like UB and now is changed.

> A reasonably safe way would be to consider that some stack area below p should not be touched.

Is there a strong argument for a specific value of `N` rather than `N+1` bytes? If not, then `N==0` is an good as any other value, provided it's empirically known to work on all tested platforms, compilers and optimization settings - which is the case so far.

> If stack_check_init() may be called from several places, then it may set stack_origin to different values, so stack_check_get_usage() may then report inaccurate values.

I can easily imagine an application where the the use re-init the stack once a minute or so, to keep track of the dynamics of stack usage. The API should allow this and be robust enough to work from any state (even almost expired WDT) and clean after itself. Please don't ask for removal of robustness features.

> the .data and .bss sections can be safely initialized without taking care of the watchdog. I mean: the stack is never gigantic on MCUs, so a memset() over it should be pretty fast, and the watchdog timeouts are usually very long. Otherwise, no buffer could be initialized without triggering the watchdog.

It it indeed a known problem on msp430 that applications may fail to enter `main()` if there are too large .data and .bss buffers that need to be initialized.  (See [here](https://e2e.ti.com/support/development_tools/compiler/f/343/p/496916/1797071) for an example). As far as I remember, GCC on msp430 does disable the WDT by default as the very first thing, but not all compilers do that.That's why my patch initially had watchdog_periodic() inside the loops - until testing convinced me that it's probably an overkill.

> Good. However, it would be better to move the empty definitions into the header file as macros, so that they can just be optimized away.

I agree from efficiency's point of view, however, here I was following the typical Contiki API code style, which typically does have function definitions in `.c` files - probably for reasons keeping the `.h` files cleaner - for example, `energest_init()` is always defined.

bthebaudeau:
Hi @atiselsts,

> Let me clarify the intention. The goal of patch is to allow easier debugging and fault detection of user applications. Therefore checking what happens with the stack at the start of and before `main()` is not relevant from my perspective. If a platform has code that overflows the stack in the first few lines of `main()` or before that, that platform should not even be part of Contiki anyway, no? The stack usage there is not measured for the same reason, say, `energest` energy usage is not measured right from the very start. Can you think of any counterexamples for this?

For `energest`, it does not matter much, because the devices do not reboot very often. There is however the specific case of MCUs that have to go through the boot process in order to get out of their most deep sleep states. In that case, it would be interesting to have the `energest` info for the boot too. If that's not what's implemented, it's probably more because it's complicated than because it's not wanted.

Early stack overflows or not less important than later ones. Bugs should be detected and fixed as much as possible. If a platform has a bug, it should not be rejected from Contiki, but fixed. It is a common practice to use the linker scripts and/or the startup code to automatically fill the stack with a debug pattern such as `0xcd`, only once, and at the very beginning of the execution, before `main()`, or even before the execution (think debugger). Then, it's easy to determine the maximum stack usage at any time. Some IDEs also provide this feature.

However, if it makes things much simpler here not to handle early stack overflows, that's good enough for me.

> > At least on ARM, _end is usually the end of the stuff before the heap, i.e. the beginning of the heap (if the heap is used). It would be easy to use the same meaningful name for all platforms. E.g. for MSP430, you could use --defsym=_stack=_end with ld (if the _stack symbol name is not already used).
> 
> Agree about what `_end` means and added a comment in the loader scripts to clarify that. Still, I think that `_end` symbol should be added for cc2650 and cc2538, even if separately from this patch (but platform maintainers can override me of course). No one at the moment knows what the symbol `_stack` means, but many programmers do known [what `_end` means](http://man7.org/linux/man-pages/man3/end.3.html).
And it's not just msp430 that has the `_end` symbol - for example, the AVR platforms and x86 (the "native" platform) do have it too.

I agree, and my point was based on this argument. What I meant is that the usual definition of `_end` is the end of BSS as you pointed out, which often coincides with the beginning of the heap on MCUs. So defining `_end` as you suggest for CC2538 and CC2650 would be fine. But my point was that:
 - custom linker scripts may follow other conventions,
 - `_end` is usually the end of BSS, not necessarily the beginning of the stack.

Consequently, using `_end` to fill the stack is wrong because it may touch other things than the stack, e.g. the heap, especially if you do this after `main()`. That's why introducing a custom symbol pointing to the beginning of the stack would be better.

> > The result of the 1st iteration of your loop is undefined 
> 
> Ok, that's fair point, this looks like UB and now is changed.

You're doing pointer arithmetic on `void *`. This is non-standard. It is a GCC extension, and not all the compilers used by the Contiki platforms are GCC-based. You're still filling the stack from top to bottom. The `trigerred` typo has not been fixed everywhere. Why not just do the following (a bit like what you do in `stack_check_get_usage()`?
```
static uint8_t *stack_origin;
/* [...] */
void
stack_check_init(void)
{
  uint8_t *p;

  stack_origin = &p;
  p = &_stack; /* Assuming my previous suggestion */

  /* Make sure WDT is not triggered */
  watchdog_periodic();

  while(p < stack_origin) {
    *p++ = STACK_FILL;
  }

  /* Make sure WDT is not triggered */
  watchdog_periodic();
}
```

> > A reasonably safe way would be to consider that some stack area below p should not be touched.
> 
> Is there a strong argument for a specific value of `N` rather than `N+1` bytes? If not, then `N==0` is an good as any other value, provided it's empirically known to work on all tested platforms, compilers and optimization settings - which is the case so far.

I'm fine with `N==0` if it has been thoroughly tested as you say.

> > If stack_check_init() may be called from several places, then it may set stack_origin to different values, so stack_check_get_usage() may then report inaccurate values.
> 
> I can easily imagine an application where the the use re-init the stack once a minute or so, to keep track of the dynamics of stack usage. The API should allow this and be robust enough to work from any state (even almost expired WDT) and clean after itself. Please don't ask for removal of robustness features.

Isn't the goal here to detect stack overflows? What would be the point of tracking the stack depth over time? The stack may do whatever it wants as long as there is a margin large enough to make sure that it may never overflow. Only tracking the stack maximum depth over time is useful to detect when a stack overflow occurred, which requires a single stack fill. Then, it's again about the early stack fill discussion, so as I said above, it's just good enough as you did.

> > the .data and .bss sections can be safely initialized without taking care of the watchdog. I mean: the stack is never gigantic on MCUs, so a memset() over it should be pretty fast, and the watchdog timeouts are usually very long. Otherwise, no buffer could be initialized without triggering the watchdog.
> 
> It it indeed a known problem on msp430 that applications may fail to enter `main()` if there are too large .data and .bss buffers that need to be initialized.  (See [here](https://e2e.ti.com/support/development_tools/compiler/f/343/p/496916/1797071) for an example). As far as I remember, GCC on msp430 does disable the WDT by default as the very first thing, but not all compilers do that.That's why my patch initially had watchdog_periodic() inside the loops - until testing convinced me that it's probably an overkill.

In that case, let's keep your watchdog resets as a precaution.

> > Good. However, it would be better to move the empty definitions into the header file as macros, so that they can just be optimized away.
> 
> I agree from efficiency's point of view, however, here I was following the typical Contiki API code style, which typically does have function definitions in `.c` files - probably for reasons keeping the `.h` files cleaner - for example, `energest_init()` is always defined.

With what I suggest, there would be functions defined in the `.c` with the option enabled, and macros defined in the `.h` otherwise, not functions, which is fine with the rules in Contiki. What has been done for `energest_init()` is a choice, not a rule. See `ENERGEST_ON()` for an example of a macro defined in a header file.

Travis failed, but it does not want to display the job log right now, so I can't analyze why.

atiselsts:
Updated the pull request. I believe we're towards reaching a consensus here. Anyone else wants to have a say?

Updates:
- define `_stack` symbol on CC2538 and CC2650 in the linker script files, and an as alias to _end in `Makefile.msp430`
- also keep the `_end` symbol in CC2538 and CC2650 linker scripts as it's a fairly standard thing to have
- stack filling logic is now as suggested

bthebaudeau:
Very good! Thanks. The only remaining issues that I see are:
 - the default value of `STACK_CHECK_ENABLED`, which has no reason to be `1`, because the check functions are called nowhere except in this example,
 - the build failure of Travis on 6502 because of a RAM overflow of 73 bytes, but it does not seem to be related to this PR.

atiselsts:
Yes, Travis is failing on all recent PRs, so it's unrelated.

About being enabled by default - I know you had this complaint in the past and I did not respond to it. I would go for enabled by default as:
1) The longer-term idea is (I hope) to actually make more use of this in the main events loops of some platforms. Adding this code to a platform should be as simple as possible instead of requiring to put stuff in `platform-conf.h` files which already are quite hard to manage
2) Beta testing of this feature (with odd compiler versions, settings etc.) is not really going to happen unless it's enabled by default
3) Just the fact itself that the stack is filled with `0xcd` sometimes does make a difference - it can be helpful when looking at memory dumps in `mspsim` or by using gdb a big region of `0xcd` will clearly indicate untouched stack

The downside is slightly higher flash and RAM usage, and a difference in startup time that no-one will notice unless looking for it.

bthebaudeau:
1. This is a debugging feature. Users should not have to know that they need to disable it to optimize their binaries. On the other hand, they should know that this features exists whenever they need it, and it should be easy to use. If it is disabled by default, well documented in the Wiki, and integrated in the main event loops of most platforms, then it is easy to enable on a per-project basis thanks to `project-conf.h`, as you did in your example. There would be nothing about this feature in `platform-conf.h` in that case. We could even go further and introduce a global DEBUG make switch (maybe it already exists?) that would enable most debug features if set, and optimize performance and size for release if unset, like the default debug and release project configurations found in most IDEs. What do you think? One more thing: if enabled by default and integrated in the main event loops of most platforms, the associated periodic `printf()` would pollute the text output of the examples, unless it prints something only in case of stack overflow, but that could be too late to be able to print anything.
2. It's Travis' job, using your example. Actually, I see that this example has not been added to the test suite for the supported platforms. Please do so. You can do it like for the `cfs-coffee` example. E.g., see https://github.com/contiki-os/contiki/blob/master/regression-tests/18-compile-arm-ports/Makefile .
3. Yes, that would be useful. Perhaps it could be enabled by default if `mspsim` is used. As to the `gdb` case, it's a typical debugging context that could use `project-conf.h` when needed.

cpu/cc26xx-cc13xx/cc26xx.ld
```diff
@@ -90,11 +90,15 @@ SECTIONS
...
     } > SRAM
 
+    _end = .;  /* End of the .bss segment. */
+
     .ccfg :
     {
         KEEP(*(.ccfg))
     } > FLASH_CCFG
 
+    _stack = .;  /* This symbol is used by the stack check library. */
```
arurke: I revisited this PR now on our custom cc13xx board, but got problem booting with _stack defined here. I moved it before .ccfg (as I noticed you do in contiki-ng) and it worked fine. Not versed in linker scripts but looking at the values it yields it seems strange to me that it should work at all as is?

bthebaudeau: You're right: here, `_stack` has its VMA set to flash memory, so there is no way it can work.

PR #2090, opened 2017-01-26
CC26xx IEEE mode radio: fix LQI reading

atiselsts:
LQI reading was taken with an incorrect offset in the received packet buffer.

Also, the mask for the LQI in the status word should be 0x3f, not 0x7f, as the bit 0x40 refers to the frame filtering status of the packet.


g-oikonomou:
Nice catch Atis, 👍

PR #2091, opened 2017-01-26
Revert to static Ethernet driver for the ATARI.

oliverschmidt:
After all there's no momentum for a W5100-based Ethernet solution on the ATARI. So I revert to static linkage of the CS8900A driver as this improves program load times.

### DIFF
--- a/cpu/6502/lib/config.c
+++ b/cpu/6502/lib/config.c
@@ -90,15 +90,15 @@ config_read(char *filename)
   log_message("Def. Router: ",  ipaddrtoa(&config.draddr, uip_buf));
   log_message("DNS Server:  ",  ipaddrtoa(&config.resolvaddr, uip_buf));
 
-#ifndef ETHERNET
+#ifndef STATIC_DRIVER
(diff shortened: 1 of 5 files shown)


cpu/6502/lib/config.c
```diff
@@ -90,15 +90,15 @@ config_read(char *filename)
...
   log_message("DNS Server:  ",  ipaddrtoa(&config.resolvaddr, uip_buf));
 
-#ifndef ETHERNET
+#ifndef STATIC_DRIVER
   log_message("Eth. Driver: ",  config.ethernetcfg.name);
-#else /* !ETHERNET */
+#else /* !STATIC_DRIVER */
   #define _stringize(arg) #arg
   #define  stringize(arg) _stringize(arg)
   log_message("Eth. Driver: ",  stringize(ETHERNET));
```
greg-king5: That `ETHERNET` name, too, needs to be changed.

oliverschmidt:
Thanks for the review and pointing out :-)

I'll fix it with an upcoming pull request - the one I prepared for with the renaming...

PR #2093, opened 2017-01-30
Fix Bug in MQTT App: Refused Connections should not be pursued.

thomas-blank:
If the MQTT connection is acked by the broker with an error, like when the given client ID is not accepted by the server, the MQTT app recognizes that error, sends a corresponding event to the app user process, but then continues to use the connection as if the connection had been acked without an error, including sending another event to the app user process.

### DIFF
--- a/apps/mqtt/mqtt.c
+++ b/apps/mqtt/mqtt.c
@@ -751,6 +751,8 @@ handle_connack(struct mqtt_connection *conn)
     call_event(conn,
                MQTT_EVENT_CONNECTION_REFUSED_ERROR,
                &conn->in_packet.payload[1]);
+    abort_connection(conn);
+    return;
   }
 
   conn->out_packet.qos_state = MQTT_QOS_STATE_GOT_ACK;


g-oikonomou:
This is a good catch. You are not explicitly resetting the connection's state to `MQTT_CONN_STATE_NOT_CONNECTED`. I only had a quick look and it seems to me it is not required, but can you confirm that this is indeed the case and you are doing this intentionally?

g-oikonomou:
In fact I'm wrong, we do need to reset the MQTT connection's state. There are also question marks regarding the state of the TCP connection. Perhaps a better approach to fix this would be to call `abort_connection()`.

thomas-blank:
I had a close look at network traffic while I presented the IBM MQTT broker with a few intentionally broken IDs. While on the MQTT level giving a wrong ID gives only that Connect Ack with the error bits set, on a TCP level the whole exchange stays civilized: The server sends the Connect Ack and then closes the connection with a TCP FIN. We handle that TCP event in lines 1104ff with all the necessary resets and cleanups. That is why the code works as was, and was correct in a certain way: The MQTT connection does not advance out of the CONNECTING state. The TCP connection underneath stays untouched by this.

On the other hand, relying on the server to close the connection sounds stupid once I type it out loud, you're right.

I used the TCP event handling code as a reference to see what needs to be cleaned up. Calling abort_connection() is pretty much it. We do not need to stop the keep_alive_timer because it's not started yet.

What I do not understand about the TCP handler once I look at it is why it also posts a mqtt_abort_now_event - That should cause the mqtt_process to call abort_connection(), too, but in practice I don't see that happening: When I do both in handle_connack(), aborting the connection and posting the event, abort_connection is still only called once. Why does the process not run/get the event/ call abort_connection?

But this is off topic and possibly I'm just not understanding correctly what's going on. I tested this version both with liberal printf use and by checking wireshark and believe this way it's correct.

Should I do anything about the TCP event handler once this fix is through? Should I try to get to the root of the smell, just to make sure there's nothing wrong with the code?

g-oikonomou:
I cannot say I can immediately understand what you are asking there, but it seems as though we are both on the same page in terms of having to abort the connection attempt client-side. Thus, merging this in good faith! :) 👍

thomas-blank:
Yes, we're on the same page there

The question was whether anything should be done about The TCP Event Handler in line 1109ff: https://github.com/contiki-os/contiki/blob/master/apps/mqtt/mqtt.c#L1106 - That code does a lot more, sets the status of the connection, calls abort_connection(), and dispatches a mqtt_abort_now_event. With the mqtt_abort_now_event, the main mqtt process should call abort_connection() again. But with a printf()-statement in abort_connection(), I could see that abort_connection() is called only once. I see that as a code smell with something potentially being wrong.

The question was: Can you tell if that code is correct and I just misunderstood something, or do you agree that this lost double call to abort_connection looks weird, then I'll take the time to have another look at that and maybe prepare another pull request?

PR #2097, opened 2017-02-05
Removed artifacts of former uip_appdata handling.

oliverschmidt:
Adam Dunkels writes on 2/5/2017 on the Contiki mailing list:
[...] the original idea was that the application could just point the uip_appdata pointer to wherever the data was, but we then changed it so that the data actually had to be copied into the uip_aligned_buf buffer. So, yes, the network device driver should only need to read from this buffer. [...]

This change removes comments on the possibility of uip_appdata pointing somewhere outside the uip_aligned_buf. And it removes code in the SLIP drivers not necessary anymore.

Additionally it makes code in a SLIP driver optional that takes care of the Microsoft-specific CLIENT / SERVER / CLIENTSERVER chat.

### DIFF
--- a/core/dev/slip.c
+++ b/core/dev/slip.c
@@ -107,9 +107,6 @@ slip_send(void)
 
   ptr = &uip_buf[UIP_LLH_LEN];
   for(i = 0; i < uip_len; ++i) {
-    if(i == UIP_TCPIP_HLEN) {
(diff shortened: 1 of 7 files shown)


oliverschmidt:
If noone vetos against it I'll merge this soon...

Issue #2031, opened 2017-01-04
RPL-ICMP6 uninitialized pointer


tozon:
dao_input_storing() function uses "rep" local variable.
It is initialized in L761.
It is used in multiple places (also in L854).
If RPL_WITH_MULTICAST=1 goto jump may happen from L757 to L841. In such case rep variable is used without initialization. This can lead to random behaviour or hard fault.

fprwi6labs:
Same bug encountered on branch release-3-1.
The multicast is not operational anymore.

yatch:
@tozon @fprwi6labs Could you try the patch of #2101 with your code to see the issue is resolved?

tozon:
@yatch I have verified the change. If DAO with multicast prefix is received, it is processed without any trouble now.
I see different problem now: DAO containing unicast prefix is resent in case of no ACK received. This mechanism is not implemented for DAO with multicast prefix and they often get lost. But this is other issue.

yatch:
@tozon Thank you for your test. I guess you're right. I'll send a PR as I mentioned in https://github.com/contiki-os/contiki/pull/2101#issuecomment-285236893 and we can discuss there :-)

PR #2101, opened 2017-02-07
RPL: fix a bug accessing an uninitialized pointer

yatch:
This PR aims to resolve https://github.com/contiki-os/contiki/issues/2031.

A new regression test, `11-rpl-multicast.csc`, is introduced under `regression-tests/12-rpl/`.

Without the change, the test fails like this:
```shell
$ cd regression-tests/12-rpl
$ make 11-rpl-multicast.testlog
Running test 11-rpl-multicast with random Seed 1: ............../../regression-tests/simexec.sh: line 70: 83680 Aborted                 (core dumped) java -Xshare:on -j
(snip)
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007fb713def1f3, pid=83680, tid=0x00007fb712844700
#
# JRE version: OpenJDK Runtime Environment (8.0_121-b13) (build 1.8.0_121-8u121-b13-0ubuntu1.16.04.2-b13)
# Java VM: OpenJDK 64-Bit Server VM (25.121-b13 mixed mode, sharing linux-amd64 compressed oops)
```

On the tip of the branch, the test succeeds:
```shell
$ cd regression-tests/12-rpl
$ make 11-rpl-multicast.testlog
Running test 11-rpl-multicast with random Seed 1: ................. OK
```

### DIFF
--- a/core/net/rpl/rpl-icmp6.c
+++ b/core/net/rpl/rpl-icmp6.c
@@ -79,7 +79,7 @@ static void dao_input(void);
 static void dao_ack_input(void);
 
 static void dao_output_target_seq(rpl_parent_t *parent, uip_ipaddr_t *prefix,
-				  uint8_t lifetime, uint8_t seq_no);
+                                  uint8_t lifetime, uint8_t seq_no);
 
 /* some debug callbacks useful when debugging RPL networks */
 #ifdef RPL_DEBUG_DIO_INPUT
@@ -158,8 +158,8 @@ get_global_addr(uip_ipaddr_t *addr)
 static uint32_t
 get32(uint8_t *buffer, int pos)
 {
-  return (uint32_t)buffer[pos] << 24 | (uint32_t)buffer[pos + 1] << 16 |
-         (uint32_t)buffer[pos + 2] << 8 | buffer[pos + 3];
+  return ((uint32_t)buffer[pos] << 24 | (uint32_t)buffer[pos + 1] << 16 |
+          (uint32_t)buffer[pos + 2] << 8 | buffer[pos + 3]);
 }
 /*---------------------------------------------------------------------------*/
 static void
@@ -202,7 +202,7 @@ rpl_icmp6_update_nbr_table(uip_ipaddr_t *from, nbr_table_reason_t reason, void *
   }
 
   return nbr;
- }
+}
 /*---------------------------------------------------------------------------*/
 static void
 dis_input(void)
@@ -238,7 +238,7 @@ dis_input(void)
           PRINTF("RPL: Unicast DIS, reply to sender\n");
           dio_output(instance, &UIP_IP_BUF->srcipaddr);
         }
-	/* } */
+        /* } */
       }
     }
   }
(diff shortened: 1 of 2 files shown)


yatch:
Oops, something wrong with the sequence number of a DAO with a multicast address for the target... I feel I hit another bug. I'm looking into it.

yatch:
I've pushed another commit which hopefully resolves the issue found in the DAO sequence number management. It might be better to discuss the commit in a separate PR...

`RPL_LOLLIPOP_INCREMENT(dao_sequence);` has been moved from `dao_output_target()` to `dao_output()` by https://github.com/contiki-os/contiki/pull/1634 so that a correct sequence number is stored in `my_dao_seqno`. However, to my understanding, this causes a DAO for multicast routes to have the same sequence number as the last a DAO for unicast routes which is sent by `dao_output()`.

I reverted the change with taking into account `my_dao_seqno`.

lcharpenwi6labs:
Hello
Thanks for the correction.

I tested it but I am not sure it's the best solution.
According to me, it's good to keep the route state even if it's a multicast route.

Here is my proposal:

In uip-macst6-route.h:

```
/** \brief An entry in the multicast routing table */
typedef struct uip_mcast6_route {
  struct uip_mcast6_route *next; /**< Routes are arranged in a linked list */
  uip_ipaddr_t group; /**< The multicast group */
  uint32_t lifetime; /**< Entry lifetime seconds */
  void *dag; /**< Pointer to an rpl_dag_t struct */
+#ifdef UIP_DS6_ROUTE_STATE_TYPE
+ UIP_DS6_ROUTE_STATE_TYPE state;
+#endif
} uip_mcast6_route_t;
```

Then, in rpl_icmpv6.c

```
/* prepare for forwarding of DAO */
static uint8_t
prepare_for_dao_multicast_fwd(uint8_t sequence, uip_mcast6_route_t *rep)
{

  /* not pending - or pending but not a retransmission */
  RPL_LOLLIPOP_INCREMENT(dao_sequence);

  /* set DAO pending and sequence numbers */
  rep->state.dao_seqno_in = sequence;
  rep->state.dao_seqno_out = dao_sequence;
  RPL_ROUTE_SET_DAO_PENDING(rep);

  return dao_sequence;
}

static void
dao_input_storing(void)
{
....
      if(RPL_ROUTE_IS_DAO_PENDING(rep) &&
         rep->state.dao_seqno_in == sequence) {
        /* keep the same seq-no as before for parent also */
        out_seq = rep->state.dao_seqno_out;
      } else if((uip_is_addr_mcast_global(&prefix)) && (mcast_group != NULL)) {
        out_seq = prepare_for_dao_multicast_fwd(sequence, mcast_group);
      } else if(rep != NULL) {
        out_seq = prepare_for_dao_fwd(sequence, rep);
      }
....
```

yatch:
@lcharpenwi6labs Thank you for your comment! I was going to do a similar thing as your idea adding the `UIP_DS6_ROUTE_STATE_TYPE state` member to `uip_mcast6_route_t`. I think there are two issues with the idea. One is major and the other is minor.

- [major issue] `rpl_route_entry_t` assumes an associated route is registered by a single neighbor. This is the case for a unicast route in most cases. Not for a multicast route. And, I don't think it's trivial to handle the multicast route case in the current code base...
- [minor issue] Since `uip_mcast6_route` has already `lifetime` and `dag`,  simply adding `rpl_route_entry_t` to `uip_mcast6_route` would waste memory. `rpl_route_entry_t` has also `lifetime` and `dag` members. 

For the issues mentioned above, I decided to give up the End-to-End DAO-ACK for multicast for now.

lcharpenwi6labs:
Thanks for your feedback
I understand your point and I agree with you.

We have tested it and it works fine.

yatch:
@lcharpenwi6labs Thank you for your testing. I'll update the body message of this PR to close the issue automatically once it's merged.

simonduq:
Doesn't the PR contain multiple different fixes? Shouldn't be split in mupltiple PRs?
There seems to be the NULL pointer fix and then some other changes in logic with DAO sequence and ACK.

yatch:
@simonduq You're right; this contains changes on DAO handling for multicast. I found two changes in 1d67b94 are irrelevant to the NULL pointer fix. I'll a separate PR having the two and f3126dc.

simonduq:
thanks!

yatch:
I've removed irrelevant changes and performed rebase with solving conflicts. Once this PR gets merged, I'll send a PR on DAO-ACK handling for multicast routes.

examples/ipv6/multicast/project-conf.h
```diff
@@ -44,7 +44,16 @@
 #include "net/ipv6/multicast/uip-mcast6-engines.h"
 
 /* Change this to switch engines. Engine codes in uip-mcast6-engines.h */
+#if WITH_SMRF
+#define UIP_MCAST6_CONF_ENGINE UIP_MCAST6_ENGINE_SMRF
+#define RPL_CONF_MOP 3
```
simonduq: why `3` and not `RPL_MOP_STORING_NO_MULTICAST`?

simonduq:
thanks for catching and fixing! 👍

examples/ipv6/multicast/project-conf.h
yatch: Good catch...!

I think I hesitated to use `RPL_MOP_STORING_MULTICAST` because it's a *private* declaration in `rpl-private.h`, which is the file for "private declarations for ContikiRPL." Of course, it's by far better to use the macro instead of the immediate value.

Now I found `rpl-private.h` is included from outside of the `rpl` module. So, I guess I can do the same thing as others did, although it seems we don't need to include `rpl-private.h` within `project-conf.h`.

yatch:
Oops, something wrong. I'm looking into it. :sweat:

yatch:
It turned out that including "rpl-private.h" within `project-conf.h` is not a good idea... I've updated the branch and Travis got green.

g-oikonomou:
I _strongly_ prefer leaving the example's Makefile untouched and controlling the contents of your build directly through `project-conf.h`. Any particular reason why you need to use make variables to manipulate your build's configuration?

g-oikonomou:
> Any particular reason why you need to use make variables to manipulate your build's configuration?

It's purely to get the correct build params for the travis test, right?

yatch:
@g-oikonomou Thank you for your comment :-)

Yes, it is; I need firmwares supporting SMRF and DAO-ACK in order for the test for this particular fix. :blush:

I found `examples/ipv6/multicast` was a nice set of code for my purpose, but they didn't meet my requirements, SMRF and DAO-ACK support.

However I didn't want to change the default configuration of them since, as you know, this example is used in other regression tests, `17-cooja-multicast-11-hops.csc` and `18-cooja-multicast-31-hops.csc`, in `11-ipv6`, and may be used somewhere else with assumption that it's using `UIP_MCAST6_ENGINE_ROLL_TM`. So, I added a couple of variables into `Makefile` and an #ifdef block in `project-conf.h` so that we can switch configuration settings easily from the command line interface, like `make WITH_SMRF=1`, with keeping the default configuration. The command with `make WITH_SMRF=1 WITH_DAO_ACK=1` is used in the newly created .csc file, `11-rpl-multicast.csc`, to create test firmwares.

Any other, nicer way to do this without touching the `Makefile`...?

g-oikonomou:
I would recommending using the `DEFINES` make variable. Have you tried this:

`$ make DEFINES=RPL_CONF_WITH_DAO_ACK=1,UIP_MCAST6_CONF_ENGINE=1,PROJECT_CONF_H="\\\""project-conf.h"\\\""`

For this to work, you will have to change `project-conf.h` to only define `UIP_MCAST6_CONF_ENGINE` if not already defined.

You do not need to set `RPL_CONF_MOP` to `RPL_MOP_STORING_MULTICAST`, enabling SMRF will do this automatically.

yatch:
@g-oikonomou I got the idea, overriding `DEFINES` whose value will be used in `CFLAGS`. I've updated the branch following your recommendation.. 😄 Let's see what color Travis will be. 

Just out of curiosity; why changing Makefile is a bad idea in this case? It would be acceptable if it were not an example project? I'd appreciate it if you could share your thoughts.

In addition, I'm going to leave some comments here:
- How can we escape double quotes in .csc file...? `PROJECT_CONF_H="\\\""project-conf.h"\\\""` doesn't work in <commands>...</commands> of a .csc file. 
- `RPL_CONF_MOP` is not set with `RPL_MOP_STORING_MULTICAST` by  `UIP_MCAST6_CONF_ENGINE=1` or `UIP_MCAST6_CONF_ENGINE=UIP_MCAST6_ENGINE_SMRF`.

Regarding the second comment, you'll get an error as follows unless `RPL_CONF_MOP` is set explicitly.
```console
../../../core/net/rpl/rpl-private.h:258:2: error: #error "RPL Multicast requires RPL_MOP_DEFAULT==3. Check contiki-conf.h"
 #error "RPL Multicast requires RPL_MOP_DEFAULT==3. Check contiki-conf.h"
  ^
../../../Makefile.include:240: recipe for target 'obj_native/uip6.o' failed
make: *** [obj_native/uip6.o] Error 1
```
When `UIP_MCAST6_ENGINE_SMRF` is specified as `UIP_MCAST6_ENGINE`, `RPL_WITH_MULTICAST` is set with `1` in `core/net/ipv6/multicast/uip-mcast6.h` and this is supposed to make `RPL_MOP_DEFAULT` become `RPL_MOP_STORING_MULTICAST`. The following is an excerpt from `rpl-private.h`:

```c
    208 /* RPL Mode of operation */
    209 #ifdef  RPL_CONF_MOP
    210 #define RPL_MOP_DEFAULT                 RPL_CONF_MOP
    211 #else /* RPL_CONF_MOP */
    212 #if RPL_WITH_MULTICAST
    213 #define RPL_MOP_DEFAULT                 RPL_MOP_STORING_MULTICAST
    214 #else
    215 #define RPL_MOP_DEFAULT                 RPL_MOP_STORING_NO_MULTICAST
    216 #endif /* RPL_WITH_MULTICAST */
    217 #endif /* RPL_CONF_MOP */
```

However, `RPL_CONF_MOP` could be *defined* at the time reading `rpl-private.h`. `contiki-default-conf.h` sets `RPL_MOP_STORING_NO_MULTICAST` to `RPL_CONF_MOP` unless `RPL_CONF_MOP` is defined. To my understanding,  `contiki-default-conf.h` is included before `rpl-private.h` in the error case shown above.
```c
    151 /* RPL_CONF_MOP specifies the RPL mode of operation that will be
    152  * advertised by the RPL root. Possible values: RPL_MOP_NO_DOWNWARD_ROUTES,
    153  * RPL_MOP_NON_STORING, RPL_MOP_STORING_NO_MULTICAST, RPL_MOP_STORING_MULTICAST */
    154 #ifndef RPL_CONF_MOP
    155 #define RPL_CONF_MOP RPL_MOP_STORING_NO_MULTICAST
    156 #endif /* RPL_CONF_MOP */
```

Thank you!!

yatch:
Oh, I clicked the wrong button by mistake... Re-opened this PR. Sorry for the noise.

g-oikonomou:
> Just out of curiosity; why changing Makefile is a bad idea in this case? It would be acceptable if it were not an example project? I'd appreciate it if you could share your thoughts.

Because to change the configuration of an example the user would have to make changes in possibly two different places. Also because changing the value of a make variable sometimes requires a make clean (which some users don't realise and some other users simply forget).

> How can we escape double quotes in .csc file...? PROJECT_CONF_H="\\\""project-conf.h"\\\"" doesn't work in ... of a .csc file.

Sorry, I don't have a clue about this. This is how one needs to escape from the command line, perhaps if you typed it directly in the respective cooja window the csc file would get updated correctly?

> Regarding the second comment, you'll get an error as follows unless RPL_CONF_MOP is set explicitly.

This is not intended :( I'm looking at right now.

g-oikonomou:
So basically the example was broken at some point in the past by some change that I cannot dig out right at this moment. The example is now unable to switch between engines the way it is meant to (i.e. simply by changing the engine in project-conf.h). We need to fix this in a separate pull.

g-oikonomou:
#2145 should fix the problem with the configuration of RPL MOP. When it gets merged, it should be possible to switch between engines simply by changing `UIP_MCAST6_CONF_ENGINE`. Once this has been done, then this pull can be updated to leave the makefile untouched as per my request. 

Other than that 👍 for the pointer fix and the respective travis test.

Thank you for your contribution here!

greg-king5:
> - How can we escape double quotes in a .csc file?
> ```sh
> PROJECT_CONF_H="\\\""project-conf.h"\\\""
> ```
> doesn't work in a `<command>` section of a .csc file.

```sh
$ echo PROJECT_H="\\\""project.h"\\\""
PROJECT_H=\"project.h\"
$ echo PROJECT_H=\\\"project.h\\\"
PROJECT_H=\"project.h\"
$ echo PROJECT_H=\\\\\\\\\\\"project.h\\\\\\\\\\\"
PROJECT_H=\\\"project.h\\\"
```
I don't know if the CSC scanner is as crazy as the shell's scanner; but, you should try eleven (!) back-slashes on each side.

g-oikonomou:
Far from ideal... There are two acceptable workarounds:

Either we change the `Makefile` to specify `PROJECT_CONF_H` like so:
`CFLAGS += -DPROJECT_CONF_H=\"project-conf.h\"`

or we change the Makefile to use the `override` directive for `DEFINES`, like so:
`override DEFINES += PROJECT_CONF_H=\"project-conf.h\"`

In both cases, we no longer have to specify `PROJECT_CONF_H` in the command line. It should be possible to simply run:
`make DEFINES=RPL_CONF_WITH_DAO_ACK=1,UIP_MCAST6_CONF_ENGINE=1`

greg-king5:
I assume that the `DEFINES+=` method is used to encourage people to modify "project-conf.h" instead of doing
```sh
make savedefines DEFINES=...
```
In that case, it should be `override DEFINES+=`.

yatch:
@g-oikonomou @greg-king5 Thank you for your comments! Regarding the escape thing, I cannot still specify `PROJECT_CONF_H` within a .csc file nor in the corresponding cooja window named "Create Mote Type", although I tried the "eleven baskslash" method by @greg-king5 😆 Cooja always takes away double quotes out of "compile commands"...

By the way, the idea by @g-oikonomou is nice. I'd like it. :smile:

> In both cases, we no longer have to specify PROJECT_CONF_H in the command line. It should be possible to simply run:
make DEFINES=RPL_CONF_WITH_DAO_ACK=1,UIP_MCAST6_CONF_ENGINE=1

g-oikonomou:
With #2145 merged, it should be possible to now simply switch engines through `project-conf.h`. This means we can also do `make DEFINES=RPL_CONF_WITH_DAO_ACK=1,UIP_MCAST6_CONF_ENGINE=1` to build as per previous posts.

We also need to change `project-conf.h` like so:

```
$ git diff -- project-conf.h
diff --git a/examples/ipv6/multicast/project-conf.h b/examples/ipv6/multicast/project-conf.h
index 0c47ab5..7ce29af 100644
--- a/examples/ipv6/multicast/project-conf.h
+++ b/examples/ipv6/multicast/project-conf.h
@@ -44,7 +44,9 @@
 #include "net/ipv6/multicast/uip-mcast6-engines.h"
 
 /* Change this to switch engines. Engine codes in uip-mcast6-engines.h */
+#ifndef UIP_MCAST6_CONF_ENGINE
 #define UIP_MCAST6_CONF_ENGINE UIP_MCAST6_ENGINE_ROLL_TM
+#endif
 
 /* For Imin: Use 16 over NullRDC, 64 over Contiki MAC */
 #define ROLL_TM_CONF_IMIN_1         64
```

g-oikonomou:
@yatch it appears as though you have addressed all my comments here about not changing the example's `Makefile`. Are you ready for this to be considered?

yatch:
@g-oikonomou Thanks! I've updated the branch with removing `RPL_CONF_MOP=RPL_MOP_STORING_MULTICAST` from `make` command lines as follows:

```diff
index c4981de..cd985d9 100644
--- a/regression-tests/12-rpl/11-rpl-multicast.csc
+++ b/regression-tests/12-rpl/11-rpl-multicast.csc
@@ -27,7 +27,7 @@
       <description>Cooja Mote Type #1</description>
       <source>[CONTIKI_DIR]/examples/ipv6/multicast/root.c</source>
       <commands>make clean TARGET=cooja
-      make DEFINES=RPL_CONF_WITH_DAO_ACK=1,UIP_MCAST6_CONF_ENGINE=UIP_MCAST6_ENGINE_SMRF,RPL_CONF_MOP=RPL_MOP_STORING_MULTICAST root.cooja TARGET=cooja</commands>
+      make DEFINES=RPL_CONF_WITH_DAO_ACK=1,UIP_MCAST6_CONF_ENGINE=UIP_MCAST6_ENGINE_SMRF root.cooja TARGET=cooja</commands>
       <moteinterface>org.contikios.cooja.interfaces.Position</moteinterface>
       <moteinterface>org.contikios.cooja.interfaces.Battery</moteinterface>
       <moteinterface>org.contikios.cooja.contikimote.interfaces.ContikiVib</moteinterface>
@@ -53,7 +53,7 @@
       <description>Cooja Mote Type #2</description>
       <source>[CONTIKI_DIR]/examples/ipv6/multicast/sink.c</source>
       <commands>make clean TARGET=cooja
-      make DEFINES=RPL_CONF_WITH_DAO_ACK=1,UIP_MCAST6_CONF_ENGINE=UIP_MCAST6_ENGINE_SMRF,RPL_CONF_MOP=RPL_MOP_STORING_MULTICAST sink.cooja TARGET=cooja</commands>
+      make DEFINES=RPL_CONF_WITH_DAO_ACK=1,UIP_MCAST6_CONF_ENGINE=UIP_MCAST6_ENGINE_SMRF sink.cooja TARGET=cooja</commands>
       <moteinterface>org.contikios.cooja.interfaces.Position</moteinterface>
       <moteinterface>org.contikios.cooja.interfaces.Battery</moteinterface>
       <moteinterface>org.contikios.cooja.contikimote.interfaces.ContikiVib</moteinterface>
@@ -79,7 +79,7 @@
       <description>Cooja Mote Type #3</description>
       <source>[CONTIKI_DIR]/examples/ipv6/multicast/intermediate.c</source>
       <commands>make clean TARGET=cooja
-      make DEFINES=RPL_CONF_WITH_DAO_ACK=1,UIP_MCAST6_CONF_ENGINE=UIP_MCAST6_ENGINE_SMRF,RPL_CONF_MOP=RPL_MOP_STORING_MULTICAST intermediate.cooja TARGET=cooja</commands>
+      make DEFINES=RPL_CONF_WITH_DAO_ACK=1,UIP_MCAST6_CONF_ENGINE=UIP_MCAST6_ENGINE_SMRF intermediate.cooja TARGET=cooja</commands>
       <moteinterface>org.contikios.cooja.interfaces.Position</moteinterface>
       <moteinterface>org.contikios.cooja.interfaces.Battery</moteinterface>
       <moteinterface>org.contikios.cooja.contikimote.interfaces.ContikiVib</moteinterface>
```

Let's see if Travis gets green...

g-oikonomou:
The way things are now, this regression test ignores the example's `project-conf.h`. Can you confirm this is what you intend?

yatch:
@g-oikonomou Yes; it should be fine.

I used `project-conf.h` in order to define `RPL_CONF_WITH_DAO_ACK` and `UIP_MCAST6_CONF_ENGINE` as per arguments of  the `make` command. Now, I learned how to define those macros directly from a command line, using `DEFINES`,  which appears to work well. :-)

g-oikonomou:
👍 from me here, just need someone to quickly double check the RPL-related code, just in case

nvt:
:+1:

PR #2103, opened 2017-02-08
Don't drop packets bound for outside the RPL mesh in non-storing mode

stringr9:
When running a network in non-storing mode, the RPL root will drop any packets bound for outside the network  because `rpl_update_header()` returns `0` when root in non-storing mode if the destination ip address in the IP buffer isn't in a known DAG.

This adds an additional check to not drop the packet if we're in non-storing mode and the packet destination isn't in a known DAG.  Though this seems to work for me, suggestions for a better way to handle this are welcome.

### DIFF
--- a/core/net/rpl/rpl-ext-header.c
+++ b/core/net/rpl/rpl-ext-header.c
@@ -642,12 +642,18 @@ rpl_update_header(void)
 
   if(default_instance->current_dag->rank == ROOT_RANK(default_instance)) {
     /* At the root, remove headers if any, and insert SRH or HBH
-     * (SRH is inserted only if the destination is in the DODAG) */
+    * (SRH is inserted only if the destination is in the DODAG) */
     rpl_remove_header();
-    if(RPL_IS_NON_STORING(default_instance)) {
-      return insert_srh_header();
+    if(rpl_get_dag(&UIP_IP_BUF->destipaddr) != NULL) {
+      /* dest is in a DODAG; the packet is going down. */
+      if(RPL_IS_NON_STORING(default_instance)) {
+        return insert_srh_header();
+      } else {
+        return insert_hbh_header(default_instance);
+      }
     } else {
-      return insert_hbh_header(default_instance);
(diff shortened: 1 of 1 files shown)


yatch:
How about updating `rpl_update_header()` itself? Here is my idea, although I don't perform any test with this change...
```diff
$ git diff
diff --git a/core/net/rpl/rpl-ext-header.c b/core/net/rpl/rpl-ext-header.c
index 80de287..b0563af 100644
--- a/core/net/rpl/rpl-ext-header.c
+++ b/core/net/rpl/rpl-ext-header.c
@@ -644,10 +644,16 @@ rpl_update_header(void)
     /* At the root, remove headers if any, and insert SRH or HBH
      * (SRH is inserted only if the destination is in the DODAG) */
     rpl_remove_header();
-    if(RPL_IS_NON_STORING(default_instance)) {
-      return insert_srh_header();
+    if(rpl_get_dag(&UIP_IP_BUF->destipaddr) != NULL) {
+      /* dest is in a DODAG; the packet is going down. */
+      if(RPL_IS_NON_STORING(default_instance)) {
+        return insert_srh_header();
+      } else {
+        return insert_hbh_header(default_instance);
+      }
     } else {
-      return insert_hbh_header(default_instance);
+      /* dest is outside of DODAGs; no ext header is needed. */
+      return 1;
     }
   } else {
     if(uip_ds6_is_my_addr(&UIP_IP_BUF->srcipaddr)
```

simonduq:
Thanks @stringr9 for pointing this out and suggesting a fix.
I find @yatch 's fix easier to read though. @yatch , could you PR?

stringr9:
@simonduq I incorporated @yatch 's  fix into this PR, hope that's ok.  Have done a quick test here and it seems to work well.

simonduq:
great! if you could just rebase and squash the two commits I'll be happy to support this PR :)

stringr9:
done!

simonduq:
👍

nvt:
:+1:

PR #2116, opened 2017-02-22
cpu/pic32: Prepare for upcoming pic32 platform and cleanup

Avinash95:
Contiki currently has support for PIC32MX795F512L(PIC32 model), which is used by SEED EYE board.
We have extended this(cpu/pic32) module to support PIC32MX470F512H in order to support Mikro-e 6LoWPAN Clicker board.


evilaliv3:
@Avinash95 i'm sorry for not having reviewed this till now.

The pull request seems legit to me and more than correct.

do you confirm that your work is still functional?

@g-oikonomou @nvt maybe this addition could be of value for other users?

PR #2132, opened 2017-03-08
add CPU energest measurements that were lacking in this platform

JoaoS:
This platform did not have the Energest functions to measure CPU rticks, so they were added.

### DIFF
--- a/platform/avr-rss2/contiki-main.c
+++ b/platform/avr-rss2/contiki-main.c
@@ -258,6 +258,8 @@ initialize(void)
 /* rtimers needed for radio cycling */
   rtimer_init();
 
+/* we can initialize the energest arrays here */
+  energest_init();
 /* after the timer intitialisation we start the cpu measurement */
   ENERGEST_ON(ENERGEST_TYPE_CPU);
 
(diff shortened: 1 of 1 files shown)


g-oikonomou:
@herjulf Robert any thoughts here?

herjulf:
Hello. Thanks. 

Yes It seems to missing. energest_init() could have a better place in the initialize() function where we keep initializations.

Joao is testing a new rtimer arch. I think? 

Cheers

JoaoS:
The place where energest_init() is located could be inside the initialize because it is just initializing arrays to zero.
No i am not testing a new rtimer, i am however testing placements to update the Energest CPU values when sleep mode is inactive

herjulf:
Please put energest_init() in initialize() directly, 
Well I got the impression that you had problems with the timers on your Atmel board.
Cheers

herjulf:
Thanks. I'm fine that.

JoaoS:
I thought my issue was with the rtimer, however i found the problem is rather with energest, i have ran it 25 times during 1 hour tests and get a drift of 3.7% in the expected number of rticks

g-oikonomou:
Are we happy to merge this one folks?

herjulf:
Yes. Thanks.

PR #2138, opened 2017-03-14
CC26xx IEEE mode driver fixes and robustness improvements

atiselsts:
This patch add a fix for the radio driver that limits the time it spends in busywaiting operations. At the moment it can lock up in case turning off the radio Rx mode fails, leading to a reboot of the system.

The patch also makes the radio timer (RAT) handling more robust.


g-oikonomou:
👍

PR #2139, opened 2017-03-14
CC26xx SensorTag I2C: add and use LIMITED_BUSYWAIT macro

atiselsts:
This limits the busywaiting time of the SensorTag I2C driver as a robustness improvements. We have not experienced any problems with SensorTag as such; however, on a custom platform that does not have all the SensorTag sensors attached the code would occasionally lock up when trying to read a non-attached sensor, leading to a system reboot.


g-oikonomou:
👍

PR #2145, opened 2017-03-17
Don't force RPL MOP configuration in contiki-default-conf.h

### LINKED ISSUE #2143: Multicast example cannot select a different multicast engine.
The multicast example should be capable of switching between multicast engines merely by changing `UIP_MCAST6_CONF_ENGINE`. The multicast code under core should be able to automatically set RPL to operate in MOP 3 when required. This used to work but not any more. It appears to be related to the order of header file inclusion.

As a result of that, attempting to change the default multicast engine will lead to this error:

```
  CC        ../../../core/net/ipv6/uip6.c
In file included from ../../../core/net/ipv6/uip6.c:85:0:
../../../core/net/rpl/rpl-private.h:258:2: error: #error "RPL Multicast requires RPL_MOP_DEFAULT==3. Check contiki-conf.h"
 #error "RPL Multicast requires RPL_MOP_DEFAULT==3. Check contiki-conf.h"
  ^
make: *** [obj_srf06-cc26xx/uip6.o] Error 1
```

This was discovered during the discussion under #2101.

g-oikonomou:
We are trying to set the default MOP in two places:

* `rpl-private.h`: Considers multicast engine selection and correctly allows MOP to be set to 3 when required.
* `contiki-default-conf.h` is not only redundant but also fails to handle situations where RPL multicast is required.

When selecting a multicast engine that requires RPL to be in MOP 3, compilation fails with the error shown in #2143.

This pull simply removes RPL MOP-related configuration from `contiki-default-conf.h` and allows user-provided configuration to take care of things, with `rpl-private.h` setting correct defaults if required.

Fixes #2143

### DIFF
--- a/core/contiki-default-conf.h
+++ b/core/contiki-default-conf.h
@@ -148,13 +148,6 @@
 #endif /* NBR_TABLE_FIND_REMOVABLE */
 #endif /* UIP_CONF_IPV6_RPL */
 
-/* RPL_CONF_MOP specifies the RPL mode of operation that will be
- * advertised by the RPL root. Possible values: RPL_MOP_NO_DOWNWARD_ROUTES,
- * RPL_MOP_NON_STORING, RPL_MOP_STORING_NO_MULTICAST, RPL_MOP_STORING_MULTICAST */
-#ifndef RPL_CONF_MOP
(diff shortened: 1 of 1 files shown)


nvt:
:+1:

PR #2148, opened 2017-03-21
native-border-router: fix for disappearing timer 'uip_ds6_timer_perio…

niziak:
…dic'

When UIP_ND6_SEND_NS is enabled, I've noticed that unreachable
neighbours still remains in REACHABLE state even if lifetime
(nbr->reachable) expired.

During network init 'tcpip_process' is scheduling
'uip_ds6_timer_periodic' is to tick every 100ms and make necessary
expiration.

When MAC address is received from slip-radio (from 'etimer_process'
context), network is "reinitialized" and timer 'uip_ds6_timer_periodic'
is set again with wrong process.


nvt:
:+1:

PR #2150, opened 2017-03-21
Several changes to fix and harden mqtt code.

posjodin:
1. The PT_MQTT_WAIT_SEND() macro has several issues:

- It does not check the return value from process_post(), which
  sometimes returns an error code. See next issue.

- Each time the macro is called, is posts an event to itself. The idea
seems to be that the event should be absorbed by the macro itself, so
when the macro terminates there is NOT a net growth of the event
queue. This does not work.  The reason is that the
PROCESS_WAIT_EVENT() sometimes absorbs a broadcast event instead of
its own event, and then the number of events in the event queue
increases. This leads to event explosions and overflow in the event
queue.

- The macro cannot call PT_EXIT(). This will expand to a return
statement, causing a return from the function calling the macro
(mqtt_process), rather then exiting the protothread (which was
probably the intention). Protothreads have lexical scope...

Fixes: 1) Check return value from process_post() 2) Loop until the
event posted to itself is absorbed (rather than just absorbing the
next event) 3) Replace PT_EXIT() with PT_INIT() (doesn't really make a
difference, could probably just be removed).

2. Change order of the checks in the protothread-calling loops in
mqtt_process().  Reason: When a protothread has been cleared by
PT_MQTT_WAIT_SEND(), it will not return a value, so checking against
PT_EXITED does not make sense.

3. PT_MQTT_WRITE_BYTES() should initialize conn->out_write_pos to 0.
When PT_MQTT_WRITE_BYTES() does not finish (due to TCP disconnect for
instance), it may leave conn->out_write_pos with a non-zero
value. Next time PT_MQTT_WRITE_BYTES() is called, it will take data
from the wrong place.

4. Put MQTT_CONN_STATE_ABORT_IMMEDIATE before
MQTT_CONN_STATE_NOT_CONNECTED in the enum list, so that the check
if(conn->state > MQTT_CONN_STATE_NOT_CONNECTED) in mqtt_connect()
fails when in state MQTT_CONN_STATE_ABORT_IMMEDIATE. Otherwise, it
will deadlock and not reattempt connections while in this state.


herjulf:
Peter maybe you should also say that is patch has been used in the GreenIoT pilot for quite some time.

alexandruioanp:
This looks pretty reasonable to me.

g-oikonomou:
Thanks for the input @herjulf  and @alexandruioanp 

@posjodin, I see you are indenting using tabs. Please put the modified files through uncrustify to fix the indentation (we use 2 spaces) and push again.

Let me know when it's done, I'm otherwise happy to merge this.

posjodin:
The tabs are gone now. Uncrustify changed other things as well, so I used emacs untabify instead.

g-oikonomou:
👍

PR #2155, opened 2017-03-23
er-coap: Fix style and add missing `addr` argument

cladmi:
Er-coap style cleanup (`uncrustify-fix-style.sh`).

I wanted to clean up the style before contributing new code (https://github.com/contiki-os/contiki/pull/1995).

I also added a missing attribute name in the header.

PR #1995, opened 2016-12-14
Er-coap-dtls: integration of tinydtls and using it in er-coap

cladmi:
Hello,

this set of patches adds support for er-coap-dtls using 'tinydtls' as a submodule and code modified from cetic/6lbr er-coap adaptations.

They still need to be cleaned-up, wait for tinydtls patch integration, fix contribution informations.
But they are a starting point for discussing what you think about this and test them.

More information on the changes:

## Tinydtls ##

The given version is currently my maintained version where I fixed several issues with contiki, which include :

* a regression in `dtls_retransmit_process` where a node should be popped from the queue before calling transmit
* a bug where structs are defined differently because a `#define` which is not imported in all files.
* Make tinydtls work without requiring `autoreconf; ./configure --with-contiti`

These have been reported upstream and should soon be reviewed by the maintainer @obgm .

You may also want to take a look at what was done in this repository https://github.com/iot-lab/armour-tinydtls so I can give other feedbacks upstream.

## er-coap ##

I copied the changes from https://github.com/cetic/6lbr @laurentderu which are under the same license as Contiki: https://github.com/cetic/6lbr/wiki#licence
Currently, there is no contribution information from them but I will gladly add it when I know what to put. I slightly adapted them but the fixes have been removed by the following patches.

After using them, I removed the new "context" argument which was not necessary.
This removed the need for the additional headers and made the connection management only to the "communication" layer implementation. It's still done in separate commits to see the difference.

## er-rest-dtls ##

Example copied from er-rest-example with features removed and a specific project-conf.h.
I removed the netstack configuration and set a big UIP_CONF_BUFFER_SIZE, which should be the same on the border router.
It works on my motes (cortex-m3 based with lot of ram) but did not tested it on other platforms.

The only requirement to make an er-rest example work with dtls is to set "WITH_DTLS_COAP=1" in the makefile and have a big enough UIP_CONF_BUFFER_SIZE (not tested the minimal value).

What do you think ?

Regards,
cladmi

PR #2156, opened 2017-03-23
er-coap: separate communication layer

cladmi:
Separates the communication layer in separate source file.
This will allow changing the communication layer, for example to dtls, by replacing `er-coap-udp.c`.

This code was copied and modified from https://github.com/cetic/6lbr written by Laurent Deru @laurentderu

This commit is based on https://github.com/contiki-os/contiki/pull/2155.
This commit is a part of https://github.com/contiki-os/contiki/pull/1995.

### DIFF
--- a/apps/er-coap/Makefile.er-coap
+++ b/apps/er-coap/Makefile.er-coap
@@ -4,3 +4,5 @@ er-coap_src = er-coap.c er-coap-engine.c er-coap-transactions.c      \
 
 # Erbium will implement the REST Engine
 CFLAGS += -DREST=coap_rest_implementation
+
+er-coap_src += er-coap-udp.c
--- a/apps/er-coap/er-coap-communication.h
+++ b/apps/er-coap/er-coap-communication.h
@@ -1,0 +1,15 @@
+#ifndef _ER_COAP_COMMUNICATION_H_
+#define _ER_COAP_COMMUNICATION_H_
+
+#include "contiki.h"
+
+void
+coap_init_communication_layer(uint16_t port);
+
+void
+coap_send_message(uip_ipaddr_t *addr, uint16_t port, uint8_t *data, uint16_t length);
+
+void
+coap_handle_receive(void);
+
+#endif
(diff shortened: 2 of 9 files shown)


cladmi:
I think travis build broke for unrelated reasons.

nvt:
I restarted the Travis tests in case it was a transient problem.

cladmi:
I currently added no license/author/doxygen informations.

If you like the changes this way I will add them. I preferred getting a review before taking time on it.

joakimeriksson:
Hi,

This is a bit similar to what we have been doing with CoAP in our LWM2M standalone implementation where we have made CoAP independent of transport layer and also independent of addressing model. This was in order to enable things like using LWM2M over serial links or SMS (as specified in the LWM2M specification). Examples of implementations on top of the regular linux stack (or posix) and also over HEX serial line is available here:

https://github.com/sics-iot/lwm2m-contiki/tree/lwm2m-standalone/examples/oma-lwm2m/standalone

(where you also can find the implementation and some brief description on the wiki pages).

We defined something we called coap_transport that is similar to your communication layer abstraction. We also defined a coap_endpoint that takes care of addressing (IPv4/6 or whatever is the addressing model). The we finally removed any other specifics such as processes and other things so that it is fully independent of anything else also - to make it possible to only use CoAP + LWM2M in any environment (just by implementing a very basic timer and the transport/endpoint abstractions).

cladmi:
I looked at your implementation. I see the two different parts

* Allow overwriting the whole network layer and using "endpoints" instead of address and port
    * This requires also moving "uip" management in a separate file as you may not depend on it anymore to receive messages
* Moving parts of the contiki "process" management out.


My PR, addresses using the same transport layer as before, but adding the possibility to modify packets content on the fly (session/presentation layer). Still relying on uip below. 


So we are doing a similar thing but for a different goal.
And I think, none on us currently handles the other one problem.

If I would try to add DTLS on top of your stack, I would need to copy all the `er-coap-uip` code to create dtls+uip layer. (I think it would only require prefixing your functions names so I could re-use the uip layer and add #define configuration).

And saying you could replace the "communication layer" with my PR is wrong as it still relies on uip only.


How where these problems addressed in other parts on the code ? Because it looks like a problem that can happen in other places.

cladmi:
It is the last day of my current job so I will not be able to work on this anymore.
Feel free to fork/update/change what I have done.
Regards, cladmi

PR #2164, opened 2017-03-31
Zero out httpd_state before deallocating

alexstanoev:
When a connection is aborted by the HTTP server while it's still being processed it is possible to hit a null pointer dereference issue by jumping back to a protothread (outputpt) after its httpd_state has been freed. This can be triggered by sending a POST to any form in the CC26xx web demo server using Firefox.
This patch prevents that by zeroing out httpd_state structs before freeing them, thus also clearing the httpd_state->outputpt field.
Tested using Firefox 55.0a1 on a CC2650 LaunchPad.


alexstanoev:
This also fixes the issue where headers sometimes show twice up after a POST request:

headers

g-oikonomou:
👍

PR #2165, opened 2017-03-31
Check broker IP conversion. Adjust state machine accordingly

alexandruioanp:
The result of converting the IP address of the broker wasn't checked. As a result, if an invalid IP address was entered, the pointer was left uninitialised and the IPv6 address used for connecting was some random data. The function now returns an error. Before connect_to_broker is called, mqtt_register is executed, which memsets conn to 0, making its state 0 (MQTT_CONN_STATE_ERROR). In order to recover from this error state, the extra check was added in the MQTT_CLIENT_STATE_NEWCONFIG state.

This was discovered by @jackmcbride using [CodeSonar](https://www.grammatech.com/products/codesonar).


g-oikonomou:
👍

PR #2167, opened 2017-03-31
Adding sensniff support for Atmel via avr-rss2 platform

herjulf:
Title says it. Gives a very useful app with wireshark. Use 500 kbaud (default) or 1 Mbaud which goes well with the UART XTAL at 16Mhz. Tested on AtMega256RFR2.


g-oikonomou:
Splendid! 👍

PR #2168, opened 2017-04-02
Correct SPI pins

tarakanov:
I make some changes to SPI pin definition for Sensortag so it will be similar to launchpad. I successfully use this to connect ENC28J60 module and run 6lbr (eth-module and external flash works).


g-oikonomou:
I'm not so sure that using the same pins for an external flash and for the ENC28J60 is a good idea, but this pull request is valid.

tarakanov:
Indeed, when I move to dedicated spi pins for ENC, 6lbr become more stable.

PR #2170, opened 2017-04-03
'Step instruction' button in 'Instructions' tab of 'Msp Code Watcher'

fmag7:
This is a small but helpful enhancement of Cooja. When debugging with the "Msp Code Watcher", stepping through instructions was only possible in the "Source code" tab but not in the "Instructions" tab. This commit adds this feature.


fabricio-godoi:
Greetings,
It's a nice modification, it's annoying to have to change view all the time to debug some assembly. I made the changes and notice some bug, that I think isn't wanted with this change.
When the Source code tab set a source file, the Code Watcher view is changed to the "Source code" tab with every "Step Instruction" done in the Instructions tab.

I added this lines to prevent updating the view, check if it's useful.

In the MspCodeWatcher class (arround line 93)
private static final int INSTRUCTIONS = 1;

In function displaySourceFile (arround line 295)
if(mainPane.getSelectedIndex() != INSTRUCTIONS){
    mainPane.setSelectedIndex(SOURCECODE); /* code */
}

Thanks,
Best regards!

fmag7:
Thanks for your fix, indeed this tab switch wasn't intended.

PR #2171, opened 2017-04-03
Z1x

onurrrrrr:


nvt:
Please look at https://github.com/contiki-os/contiki/wiki/Code-Contributions for a list of some issues regarding new platforms that should be clarified in the description of the PR.

The code also seems to have been developed by @atiselsts, but he is not the originator of the PR. Any comments on this?

atiselsts:
@onurrrrrr , please contact me privately if you're interested in contributing to this. My contact details are easily available online. I recommend closing this pull request.

About the platform - this is just a simple clone of the Z1 platform, but with more flash and RAM memory. It's not based on a real sensor node, but intended to be used by Cooja simulations. We're finding it very useful to use it for TSCH testing in particular, and also in general, as a mock up of our CC2650 platform in the SPHERE project (no Cooja support for ARM based MCUs, as you know), to run regression tests on the complete system etc.

The main reason why I haven't made a pull request of this personally is that it also requires updating the `mspsim` code. Perhaps if there's interest we can move forward?

nvt:
@atiselsts Would the exp5438 or Wismote platforms be sufficient for your experiments on MSP430-based systems? I'm not so sure that building a "virtual" platform based on the old Z1 is the right way to go for the official repository.

atiselsts:
At least at the point when this Z1x code was developed, TSCH support on Wismote's radio (CC2520) was problematic. Don't know about the state now. I'd still rather use CC2420 than CC2520 any time, as CC2420 is so well know and supported.

Another reason was that we were interested in having the same sensors as on TI SensorTag, not whatever Wismote / exp5438 has.

I will not push for it you don't think it's appropriate to the mainline, sure.

nvt:
Closing this PR as @atiselsts recommended above. We'll see if there is interest from other mergers to include this platform in Contiki, in which case a new PR could perhaps be opened with support from the original authors.

PR #2176, opened 2017-04-05
Fix HDC sensor reading conversion

atiselsts:
HDC sensor on SensorTag is currently having invalid values when temperature >+42.x degrees Celsius.

The cast to int16_t in this line seems to be at fault:
  `*temp = ((double)(int16_t)raw_temp / 65536) * 165 - 40;`

Casting to signed is just not necessary here, as `raw_temp` can never be negative. `raw_temp` of zero corresponds to -40 degrees Celsius, which is the minimal value the sensor can read according to datasheet.


g-oikonomou:
👍 just restarted the failed travis task.

PR #2180, opened 2017-04-08
Keep CoAP 'observe' option length <= 3 bytes

atiselsts:
Currently Contiki generates Observe notifications with the observe counter going up to 4 bytes. Yet RFC 7641(https://tools.ietf.org/html/rfc7641) clearly shows that the Observe option may have length from 0 to 3 bytes.

This causes problems for out application because with 4 byte observe option some packets will get too large to be sent over the air.


nvt:
Looks reasonable, but would it make sense to do the masking just once in coap_set_header_observe?

atiselsts:
I went for the current approach to keep the internal state consistent with what is sent over-the-air. Otherwise the `obs_counter` variable sometimes will have different values from what's inside the CoAP packets, which would be a little confusing, to me at least.

nvt:
:+1:

PR #2188, opened 2017-04-11
Fix several uninitialized variables

tonnenpinguin:
This fixes several potentially unitinitialized variables found by codesonar.

### DIFF
--- a/core/net/mac/frame802154.c
+++ b/core/net/mac/frame802154.c
@@ -205,7 +205,7 @@ frame802154_has_panid(frame802154_fcf_t *fcf, int *has_src_pan_id, int *has_dest
 int
 frame802154_check_dest_panid(frame802154_t *frame)
 {
-  int has_dest_panid;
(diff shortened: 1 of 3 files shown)


nvt:
SDCC complains about the way that you initialize a variable in rpl-icmp6.: <code>uip_ipaddr_t prefix = {0};</code>. A possible fix is to always do <code>memset(&prefix, 0, sizeof(prefix));</code> before the option processing, instead just within the processing of the RPL target option.

tonnenpinguin:
That's weird. Initializing to zero like that should be standard compliant, or not?
I'll update my pr.

greg-king5:
Look in "core/net/ip/uip.h".  `uip_ipaddr_t` is a union of arrays.  Therefore, its initializer needs _two_ pairs of braces.

tonnenpinguin:
Touché!

nvt:
:+1:

PR #2189, opened 2017-04-11
Fix potentially unterminated strings

### LINKED ISSUE #2185: Possibly unterminated strings in resolv.c
My static code analysis tool shows several possibly unterminated string warnings for resolv.c, due to the use of [strncpy ](http://www.cplusplus.com/reference/cstring/strncpy/)- which does not implicitly add a null terminator if the original string is longer than len.

This issue occurs in lines [1092](https://github.com/contiki-os/contiki/blob/master/core/net/ip/resolv.c#L1092) and [1252](https://github.com/contiki-os/contiki/blob/master/core/net/ip/resolv.c#L1252), the latter of which causes several other functions to have possibly undefined  behaviour.

~~A relatively easy fix would be to switch to BSD's [strlcpy](http://man.openbsd.org/strlcpy), however, I'm not sure if this is available for all platforms.~~
Enforcing string termination using `resolv_hostname[sizeof(resolv_hostname) - 1] = '\0';` should work.

tonnenpinguin:
This fixes some potential issues with strings being copied using strncpy, which causes char arrays that are used as strings later on to be non-terminated.

I have also changed the typecasts in the tolower function calls to unsigned char.

Fixes #2185

### DIFF
--- a/core/net/ip/resolv.c
+++ b/core/net/ip/resolv.c
@@ -1094,7 +1094,7 @@ resolv_set_hostname(const char *hostname)
   /* Add the .local suffix if it isn't already there */
   if(strlen(resolv_hostname) < 7 ||
      strcasecmp(resolv_hostname + strlen(resolv_hostname) - 6, ".local") != 0) {
-    strncat(resolv_hostname, ".local", RESOLV_CONF_MAX_DOMAIN_NAME_SIZE);
+    strncat(resolv_hostname, ".local", RESOLV_CONF_MAX_DOMAIN_NAME_SIZE - strlen(resolv_hostname));
   }
 
   PRINTF("resolver: hostname changed to \"%s\"\n", resolv_hostname);
@@ -1248,8 +1248,8 @@ remove_trailing_dots(const char *name) {
   static char dns_name_without_dots[RESOLV_CONF_MAX_DOMAIN_NAME_SIZE + 1];
   size_t len = strlen(name);
 
-  if(name[len - 1] == '.') {
-    strncpy(dns_name_without_dots, name, sizeof(dns_name_without_dots));
+  if(len && name[len - 1] == '.') {
+    strncpy(dns_name_without_dots, name, RESOLV_CONF_MAX_DOMAIN_NAME_SIZE);
     while(len && (dns_name_without_dots[len - 1] == '.')) {
       dns_name_without_dots[--len] = 0;
     }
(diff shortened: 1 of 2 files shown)


core/net/ip/resolv.c
```diff
@@ -397,7 +397,7 @@ dns_name_isequal(const unsigned char *queryptr, const char *name,
         return 0;
       }
 
-      if(tolower((unsigned int)*name++) != tolower((unsigned int)*queryptr++)) {
+      if(tolower((unsigned char)*name++) != tolower((unsigned char)*queryptr++)) {
```
darconeous: Unrelated change?

darconeous:
I would remove the unrelated change to the casts. `tolower()` is supposed to take an `int`. I'm not sure what's up with the `unsigned` part, but that's not really important at this point. I would rather the fix for the strncat issue to be clean.

greg-king5:
`name` points to type `char`.  The original programmer wanted to guarantee that the promotion to `int` doesn't extend the sign bit (if a compiler implements `char` as signed).

I don't see any advantage in changing those casts.  The result is the same: `char` to ` signed int` without sign extension.

core/net/ip/resolv.c
```diff
@@ -1094,7 +1094,7 @@ resolv_set_hostname(const char *hostname)
   /* Add the .local suffix if it isn't already there */
   if(strlen(resolv_hostname) < 7 ||
      strcasecmp(resolv_hostname + strlen(resolv_hostname) - 6, ".local") != 0) {
-    strncat(resolv_hostname, ".local", RESOLV_CONF_MAX_DOMAIN_NAME_SIZE);
+    strncat(resolv_hostname, ".local", RESOLV_CONF_MAX_DOMAIN_NAME_SIZE - strlen(resolv_hostname));
```
greg-king5: There is another  strncat()  call like that one down at line 1482.

tonnenpinguin:
@darconeous i removed the unrelated changes.
@greg-king5 I handled the second strncat issue as well!

core/net/ip/resolv.c
nvt: The multiple calls to strlen(resolv_hostname) could be something to consider to make more elegant later.

nvt:
Good to see security fixes, but we could at a later point also consider to provide an indication to the user somehow when buffers are insufficiently large for copying, instead of silent truncation.

:+1:

PR #2194, opened 2017-04-12
Added serial-raw driver for unbuffered and unprocessed serial input.

posjodin:
Intended for devices that attach via serial interface, such as
Plantower dust sensors.

### DIFF
--- a/core/dev/serial-raw.c
+++ b/core/dev/serial-raw.c
@@ -1,0 +1,99 @@
+/*
+ * Copyright (c) 2017, Peter Sjödin, KTH Royal Institute of Technology
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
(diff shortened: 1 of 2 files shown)


posjodin:
We use this for the PMS5003 dust sensor from Plantower. The sensor driver comes in a separate pull request.

posjodin:
The differences between this and the regular serial-line driver aren't very large. An alternative design could be to have just a single driver, and use a #define to select buffer mode at compile time.

core/dev/serial-raw.c
```diff
@@ -0,0 +1,99 @@
...
+#if (BUFSIZE & (BUFSIZE - 1)) != 0
+#error SERIAL_RAW_CONF_BUFSIZE must be a power of two (i.e., 1, 2, 4, 8, 16, 32, 64, ...).
+#error Change SERIAL_RAW_CONF_BUFSIZE in contiki-conf.h.
+#endif
+
+static struct ringbuf rxbuf;
+static uint8_t rxbuf_data[BUFSIZE];
+
+PROCESS(serial_raw_process, "Serial driver for raw unbuffered input");
+AUTOSTART_PROCESSES(&serial_raw_process);
```
greg-king5: `serial_raw_process` must not be auto-started.  It is started by `serial_raw_init()`.

core/dev/serial-raw.c
```diff
@@ -0,0 +1,99 @@
...
+  return 1;
+}
+/*---------------------------------------------------------------------------*/
+PROCESS_THREAD(serial_raw_process, ev, data)
+{
+  static uint8_t buf[1];
+
+  PROCESS_BEGIN();
+  
+  while(1) {
```
greg-king5: You need to get an event number for the `serial_raw_event_message` variable.  Look at [line 94](https://github.com/contiki-os/contiki/blob/master/core/dev/serial-line.c#L94) in "core/dev/serial-line.c".

nvt:
@posjodin Yes, I think that it would be more elegant to have a single driver with configurable buffering like setvbuf() in stdio or as compile-time option. There might be a backward-compatibility issue, however, if the API names should become more generic rather than line-oriented, as is currently the case with the serial-line driver.

herjulf:
@nvt 
You mean just a simple compile-time option as  SERIAL_LINE_CONF_RAW to serial_line.c?

nvt:
A couple of issues make this a bit tricky to solve without requiring a number of code changes among the modules that depend on the serial-line module.

1. The module and function names including "_LINE_" are no longer so suitable if more buffer modes are to be supported.

2. Rather than incurring the overhead of sending a broadcast event for each byte, we may prefer to have a common callback-based API for all buffer modes, where a buffer and a length are delivered,

A quicker option could be to circumvent these API discussions and build something directly on top of the UART API in your driver.

herjulf:
Right. 
From what I understand Peter was considering something based on the UART API.
We'll see.

posjodin:
Yes, I put the sensor driver directly on top of the UART API.

After considering this some more, I feel that Integrating the two buffer modes into the same code may not be a very elegant solution after all. There would be too many conditionals cluttering up the code...

In terms of functionality, the main difference between serial-raw and the UART API is really the broadcast. With only one consumer, one may as well use the UART API.

nvt:
Sounds good. Shall we close this one then?

herjulf:
Just talked to Peter. Yes I think so. He just forgot to close.

posjodin:
Yes, sorry. Closing now.

PR #2206, opened 2017-04-28
Load platform before apps

pengi:
Some code can be reused between platform, so makeing platforms being
able to load apps helps.

For example, different platforms might still use the same segger-rtt
interface, similar handling of debug io

This patch makes it possible to add "APPS += my_app" to the platform
makefile


pengi:
bump?

This isn't anything that can be added to contiki? Would really help us if it was, and I can't really see that it will break or affect anything within contiki at the moment.

fleutot:
I'd appreciate that too.

PR #2213, opened 2017-05-05
Validate MQTT broker IP and escape quotes when displaying

alexstanoev:
This commit implements address validation for the broker address in the MQTT configuration page of the CC26XX web demo example. Additionally, the Type ID, Org ID, Auth Token, Command Type and Event Type ID fields have quotes escaped (" -> &quot;) to prevent XSS/broken page issues when displaying user-sourced input in HTML input fields.

### DIFF
--- a/examples/cc26xx/cc26xx-web-demo/httpd-simple.c
+++ b/examples/cc26xx/cc26xx-web-demo/httpd-simple.c
@@ -137,6 +137,10 @@ PROCESS(httpd_simple_process, "CC26XX Web Server");
 #define REQUEST_TYPE_GET  1
 #define REQUEST_TYPE_POST 2
 /*---------------------------------------------------------------------------*/
+/* Temporary buffer for holding escaped HTML used by html_escape_quotes */
+#define HTML_ESCAPED_BUFFER_SIZE 128
+static char html_escaped_buf[HTML_ESCAPED_BUFFER_SIZE];
+/*---------------------------------------------------------------------------*/
 static const char *NOT_FOUND = "<html><body bgcolor=\"white\">"
                                "<center>"
                                "<h1>404 - file not found</h1>"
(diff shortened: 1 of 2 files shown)


examples/cc26xx/cc26xx-web-demo/httpd-simple.c
```diff
@@ -305,6 +309,30 @@ url_unescape(const char *src, size_t srclen, char *dst, size_t dstlen)
   return i == srclen;
 }
 /*---------------------------------------------------------------------------*/
+static char*
+html_escape_quotes(const char *src)
+{
+  size_t dstpos = 0;
+  memset(html_escaped_buf, 0, HTML_ESCAPED_BUFFER_SIZE);
+  for(size_t i = 0; i < HTML_ESCAPED_BUFFER_SIZE; i++) {
```
greg-king5: The variable `i` is the _source_ index; but, you are comparing it to the _destination_ length.  You need to compare `dstpos` to that length.  (If there are quotation marks in the source string, then `dstpos` will be larger than `i`.  Then, a very long source string could overflow the buffer.)

alexstanoev: Thanks for spotting this!

g-oikonomou:
Alex, I think we need a little bit more work here.

Correct me if I'm wrong, we are relying on the source string to be null-terminated. If it is not, we shall overrun it and keep reading up to a possible `HTML_ESCAPED_BUFFER_SIZE` bytes, right?

By looking at `mqtt-client.c` (those `_post_handler` functions), it appears as though some of the source strings here may under circumstances not be null-terminated (`mqtt_config.type_id` and `.mqtt_config.org_id` are the ones I looked at). This needs fixed in a separate pull.

For the purposes of this pull, I think we should be able to handle bad inputs (not null-terminated). One way of doing that is by having this function accept source length as an additional argument. Then the if statement at L318 should also check for source overrun. This should handle situations where the input is not null-terminated, while it will still truncate the destination correctly.

Can I also ask to remove variable declarations from inside for statements? This is C99, `dstpos` should be defined near `srcpos`.

alexstanoev:
Yes, this would expect the input to be properly null terminated. I'll add another argument and pass it the source buffer lengths defined in `mqtt-client.h`.

I'll have a go at sorting out `mqtt-client.c` too, it does indeed seem that there's no space for a null terminator if the input string is exactly the length of the buffer.

I'll move out the variable declaration and squash all these commits after everything seems fine, thanks.

PR #2215, opened 2017-05-07
Fix uninitialised variables

alexandruioanp:
If fcf (the Frame Control Field) is NULL, frame802154_has_panid will return before initialising has_dest_pan_id and has_src_pan_id. A consequence of this was caught and fixed in 848f801, but this fixes the root cause by initialising the two variables before the function returns.

This was discovered by @jackmcbride using [CodeSonar](https://www.grammatech.com/products/codesonar).


greg-king5:
A more efficient way to do it is to change the test to
```C
if(fcf != NULL) {
```
And, put all of the code that accesses `fcf` into the body of that `if()` statement.

The code at the end of the function always will be executed; the return-variables always will be given proper values.

g-oikonomou:
👍

PR #2223, opened 2017-05-12
Driver for Plantower PMS 5003 dust sensors.

posjodin:
Platform-specific files for avr-rss2.

### DIFF
--- a/dev/pms5003/README.pms5003
+++ b/dev/pms5003/README.pms5003
@@ -1,0 +1,21 @@
+Contiki implementation for Plantower PMS 5003 series
+dust sensors.
+
+The PMS sensors support serial communication through UART and
+I2C. This implementation supports both modes, and they can be
+individually turned on/off through config options. Both can be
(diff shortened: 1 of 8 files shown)


herjulf:
And after reading up about different air-quality projects I see the Plantower PMSX003 dust sensors are getting widely used. So having a driver is an advantage. The driver should work for most models.

herjulf:
You have not added the latest update of the duty cycling patches?

And the review of the PMS5003 sensor seems very interesting. So driver comes in handy...

posjodin:
I made an update with revised logic for duty cycling with runtime-configurable timers. It also adds support for reading dust size counters for sensors that have that (PMS5003, at least).

posjodin:
It seems that a Travis check fails with a Java error in Cooja. Not much I can do about that I guess.

posjodin:
Rebased and resubmitted as PR #2347.

PR #2226, opened 2017-05-16
nativer: fix 100% CPU usage when started from systemd

niziak:
systemd starts process with stdin redirected to /dev/null.
In this case we shouldn't read from stdin.


nvt:
A concern here is that isatty() will exclude more than just /dev/null. For example, some may wish to use pipes or files as standard input. Perhaps we need a more precise check on the file descriptor type; e.g., involving a call to fstat() and checking the st_mode field of the stat structure.

PR #2232, opened 2017-05-23
rename "srf06-cc26xx" to "cc26xx"

casaroli:
This simply renames the srf06-cc26xx platform by renaming files and replacing all occurrences of "srf06-cc26xx" with just "cc26xx".

Using "srf06" in the platform name is misleading when used in other platforms such as the launchpad.

This has already been discussed in the past http://contiki-developers.narkive.com/ASbH7YQr/rename-platform-srf06-cc26xx

So, here is the PR. I hope it does gets accepted. :)

Thanks


arurke:
Would it not make more sense to rename this to _cc26xx-cc13xx_ or something similar to indicate that it actually holds both platforms? With this PR you would get strange conterintuitive situations like (?)
```
make PLATFORM=cc26xx BOARD=launchpad/cc1310
```

PR #2236, opened 2017-05-31
Fix to rtimer

kkrentz:
I noticed that `rtimer_arch_schedule` gets called twice if `rtimer_set` is called from within the rtimer ISR, which is often the case. This PR fixes this issue.


awolokita:
This patch helps quite a bit with avoiding missed slots when scheduling slots in TSCH.

PR #2237, opened 2017-05-31
Fix UIP buffer overrun

alexandruioanp:
When copying data to the uip buffer, the size of the MTU was checked,
but not the size of the actual buffer. Each platform can set this size
to a different value (often times smaller than the MTU). This would
results in an overrun when creating an ICMPv6 error packet.


g-oikonomou:
👍

Issue #2240, opened 2017-06-01
er-coap - memory corruption in coap_parse_message (SIGSEGV)


stze:
Parsing the following coap message results in a crash of contiki:

coap message hexdump:
```
00000000  42 43 42 42 42 42 42 9e  80 42 42 28 01 e1 e1 e1  |BCBBBBB..BB(....|                          
00000010  e1 e1 e1 e1 e1 e1 e1 e1  e1 e1 e1 bf e1 00 00 10  |................|
00000020  00 43 42 53 42 ff 49                              |.CBSB.I|
00000027
```

How to reproduce:
```c
static coap_packet_t message[1];
uint8_t teststr[] = { 0x42 ,0x43 ,0x42 ,0x42 ,0x42 ,0x42 ,0x42 ,0x9e ,0x80 ,0x42 ,0x42 ,0x28 ,0x01 ,0xe1 ,0xe1 ,0xe1 ,0xe1 ,0xe1 ,0xe1 ,0xe1 ,0xe1 ,0xe1 ,0xe1 ,0xe1 ,0xe1 ,0xe1 ,0xe1 ,0xbf ,0xe1 ,0x00 ,0x00 ,0x10 ,0x00 ,0x43 ,0x42 ,0x53 ,0x42 ,0xff ,0x49 };
coap_parse_message(message, (unsigned char *)teststr, sizeof(teststr));
```

gdb:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000000000041464e in coap_parse_message (packet=0x63a240 <message>, data=0x7fff7270aba0 "BCBBBBB\236\200BB(/", '\341' <repeats 14 times>, "\277", <incomplete sequence \341>, data_len=39)
    at ../../apps/er-coap/er-coap.c:537
537	    SET_OPTION(coap_pkt, option_number);
(gdb) bt
#0  0x000000000041464e in coap_parse_message (packet=0x63a240 <message>, data=0x7fff7270aba0 "BCBBBBB\236\200BB(/", '\341' <repeats 14 times>, "\277", <incomplete sequence \341>, data_len=39)
    at ../../apps/er-coap/er-coap.c:537
#1  0x0000000000407828 in process_thread_er_example_client (process_pt=0x639198 <er_example_client+24>, ev=<optimized out>, data=<optimized out>) at er-example-client.c:116
[...]

Issue #2239, opened 2017-06-01
er-coap - memory corruption in coap_merge_multi_option (SIGSEGV)


stze:
Parsing the following coap message results in a crash of contiki:

coap message hexdump:
```
00000000  53 53 53 53 53 53 53 53  84 53 53 53 53 53 53 53  |SSSSSSSS.SSSSSSS|                          
00000010  5d 00 00 0e 53 53 53 73  32                       |]...SSSs2|
00000019
```
How to reproduce:
```c
static coap_packet_t message[1];
uint8_t teststr[] = { 0x53 ,0x53 ,0x53 ,0x53 ,0x53 ,0x53 ,0x53 ,0x53 ,0x84 ,0x53 ,0x53 ,0x53 ,0x53 ,0x53 ,0x53 ,0x53 ,0x5d ,0x00 ,0x00 ,0x0e ,0x53 ,0x53 ,0x53 ,0x73 ,0x32 };
coap_parse_message(message, (unsigned char *)teststr, sizeof(teststr));
```

gdb:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f7b1837f49f in __memmove_avx_unaligned_erms () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007f7b1837f49f in __memmove_avx_unaligned_erms () from /usr/lib/libc.so.6
#1  0x000000000041502c in coap_merge_multi_option (dst=0x63a318 <message+216>, dst_len=0x63a310 <message+208>, option=0x7ffdf02a77c6 "2\377\377\377{\177", option_len=21600, separator=38 '&')
    at ../../apps/er-coap/er-coap.c:231
#2  0x0000000000414ad0 in coap_parse_message (packet=0x63a240 <message>, data=0x7ffdf02a77b0 "SSSSSSSS\204SSSSSSS]", data_len=25) at ../../apps/er-coap/er-coap.c:621
#3  0x00000000004077e2 in process_thread_er_example_client (process_pt=0x639198 <er_example_client+24>, ev=<optimized out>, data=<optimized out>) at er-example-client.c:115
#4  0x7830000000000041 in ?? ()
#5  0x000000007ffdf02a in ?? ()
#6  0x0008000000000000 in ?? ()
#7  0x0000000000000000 in ?? ()
```
valgrind:
```
==18107== Invalid read of size 2
[...]

Issue #2238, opened 2017-06-01
er-coap - memory corruption in coap_parse_int_option (SIGSEGV)


stze:
Parsing the following coap message results in a crash of contiki:

coap message hexdump:
```
00000000  51 51 51 00 80 51 51 51  51 4e 51 51 51 51 51 51  |QQQ..QQQQNQQQQQQ|
00000010  51 f5 06                                          |Q..|
00000013
```

How to reproduce:
```c
static coap_packet_t message[1];
uint8_t teststr[] = {  0x51, 0x51, 0x51, 0x00, 0x80, 0x51, 0x51, 0x51, 0x51, 0x4e, 0x51, 0x51, 0x51, 0x51, 0x51, 0x51 };
coap_parse_message(message, (unsigned char *)teststr, sizeof(teststr));
```

gdb:
```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000414f43 in coap_parse_int_option (bytes=0x7fffffffc27c "QQQQQ\365\006", length=21086) at ../../apps/er-coap/er-coap.c:92
92	    var |= bytes[i++];
(gdb) bt
#0  0x0000000000414f43 in coap_parse_int_option (bytes=0x7fffffffc27c "QQQQQ\365\006", length=21086) at ../../apps/er-coap/er-coap.c:92
#1  0x000000000041462b in coap_parse_message (packet=0x63a240 <message>, data=0x7fffffffc270 "QQQ", data_len=19) at ../../apps/er-coap/er-coap.c:546
#2  0x0000000000407788 in process_thread_er_example_client (process_pt=0x639198 <er_example_client+24>, ev=<optimized out>, data=<optimized out>) at er-example-client.c:115
#3  0x000000000041bde8 in call_process (p=0x639180 <er_example_client>, ev=129 '\201', data=0x0) at ../../core/sys/process.c:190
#4  0x000000000041b7f5 in process_post_synch (p=0x639180 <er_example_client>, ev=129 '\201', data=0x0) at ../../core/sys/process.c:366
#5  0x000000000041b7b3 in process_start (p=0x639180 <er_example_client>, data=0x0) at ../../core/sys/process.c:120
#6  0x000000000041aaeb in autostart_start (processes=0x430500 <autostart_processes>) at ../../core/sys/autostart.c:57
#7  0x000000000041a161 in main (argc=1, argv=0x7fffffffe568) at ../../platform/native/./contiki-main.c:247
```
valgrind:
```
[...]

PR #2249, opened 2017-06-14
Added boundary checks when parsing CoAP packets.

nfi:
This should address the issues reported in #2238, #2239 and #2240. Thanks to Stephan Zeisberg for reporting the issues and for the excellent test cases.


nvt:
:+1:

nvt:
With this fix, there is no longer a segmentation fault for any of the three mentioned issues.

PR #2254, opened 2017-06-20
fix overflow in RPL timer

bschwand:
RPL_CONF_DIO_INTERVAL_MIN + RPL_CONF_DIO_INTERVAL_DOUBLINGS
are large, leading to DIO spamming because the calculated timer
interval would be too small

### DIFF
--- a/core/net/rpl/rpl-timers.c
+++ b/core/net/rpl/rpl-timers.c
@@ -115,7 +115,7 @@ new_dio_interval(rpl_instance_t *instance)
   time = 1UL << instance->dio_intcurrent;
 
   /* Convert from milliseconds to CLOCK_TICKS. */
-  ticks = (time * CLOCK_SECOND) / 1000;
+  ticks = time / ( 1000 / CLOCK_SECOND);
   instance->dio_next_delay = ticks;
 
   /* random number between I/2 and I */


core/net/rpl/rpl-timers.c
```diff
@@ -115,7 +115,7 @@ new_dio_interval(rpl_instance_t *instance)
   time = 1UL << instance->dio_intcurrent;
 
   /* Convert from milliseconds to CLOCK_TICKS. */
-  ticks = (time * CLOCK_SECOND) / 1000;
+  ticks = time / ( 1000 / CLOCK_SECOND);
```
atiselsts: Some platforms have CLOCK_CONF_SECOND set to 1024, leading to division by zero after this change. Also the change makes the timer a little bit inaccurate for small values of `time`. BTW, it seems that with the default values of `RPL_CONF_DIO_INTERVAL_MIN` and `RPL_CONF_DIO_INTERVAL_DOUBLINGS` the existing code should work fine on all platforms, so the problem should only appear when those are increased?

bschwand: good point. as CLOCK_CONF_SECOND is a user setting, I could not foresee such, I just saw values 32, 125, 128, 1000.

The alternative I see is to typecast the intermediary results to a wider size (64bit), which might not be available on certain platforms. What do you suggest ?

> _so the problem should only appear when those are increased?_

as mentioned in the comment 

_when RPL_CONF_DIO_INTERVAL_MIN + RPL_CONF_DIO_INTERVAL_DOUBLINGS
is large,_

In addition there is no guard agains too large values, but I do not think such should be done at runtime in a platform like contiki. A judicious cassert() would be best.

PR #2263, opened 2017-07-11
Fixed issues with MQTT keep alive timer

posjodin:


alexandruioanp:
This looks okay. The missing ```ctimer_restart``` is a good catch. What issues were you encountering that required the changes in ```disconnect_tcp```?

posjodin:
The issue with `disconnect_tcp()` is that after a keepalive timeout, MQTT ends up in a stale state where nothing happens.

In the original code, `disconnect_tcp()` clears the socket structure without waiting for TCP to close.
Then there will be no more events to keep the state machine going. By just calling tcp_socket_close() instead, without wiping the socket, there will be a callback event from TCP when the socket has been closed, which will trigger MQTT to reconnect.

It's only the keepalive timer that uses `disconnect_tcp()`, by the way.

PR #2265, opened 2017-07-13
sync code

lanwan:


alexrayne:
IMHO this check should optional, and selectable for release and debug stages.
possibly this checks could print failure, since `c->u == NULL`  not normal for most cases.
possibly introduce CONF macro with behaviour of check. some like:

greg-king5:
"`c->u == NULL`" can be normal.  If a "connection" only transmits, and one doesn't need to know when the packets actually are sent, then the callbacks aren't needed.  Therefore, the connection would be openned with a `NULL` as the callback argument.

alexrayne:
agree, but is it general situation? on examples a saw - all connections have callbacks.
thus my mention can be reclasified to:  selectable for generic and full defined applications. Where full defined - have all connections with callbacks, and threfore a bit more optimised.
this economy is small - a few bytes of code, a few tacts for cpu. for avr, msp430 mcu it can be sufficient.
but maybe it negligible, to have more work on this patch.

alexrayne:
lanwan, you post it patch from "unknown repository", can you publish it on you public repos, where i can take it?

PR #2271, opened 2017-07-31
make for сс26-13 generates hex only if  srec exist

alexrayne:
hex generate only if srec_cat exist. this patch helps avoid make errors if srecord absent.

*when build on windows host, there are hard to achieve srecord. and now this is not a problem - elf images, and binaries are enough. So, with this patch hex generation just ignored, and make - success

### DIFF
--- a/cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
+++ b/cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
@@ -6,7 +6,15 @@ OBJCOPY = arm-none-eabi-objcopy
 OBJDUMP = arm-none-eabi-objdump
 NM      = arm-none-eabi-nm
 SIZE    = arm-none-eabi-size
+SREC_TEST = $(shell srec_cat --version > null && echo 1)
+ifeq ([${SREC_TEST}],[1])
 SREC_CAT = srec_cat
+CAN_HEX  = %.hex
+else
+$(waring "install srecord to allow hex image generation")
+SREC_CAT =
+CAN_HEX  =
+endif
 
 CPU_ABS_PATH       = cpu/cc26xx-cc13xx
 TI_XXWARE = $(CONTIKI_CPU)/$(TI_XXWARE_PATH)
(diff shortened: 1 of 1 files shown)


cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
```diff
@@ -6,7 +6,15 @@ OBJCOPY = arm-none-eabi-objcopy
 OBJDUMP = arm-none-eabi-objdump
 NM      = arm-none-eabi-nm
 SIZE    = arm-none-eabi-size
+SREC_TEST = $(shell srec_cat --version > null && echo 1)
+ifeq ([${SREC_TEST}],[1])
 SREC_CAT = srec_cat
+CAN_HEX  = %.hex
+else
+$(waring "install srecord to allow hex image generation")
+SREC_CAT =
```
greg-king5: `SREC_CAT` doesn't need to be defined conditionally.  It's make-rule simply won't be used if `CAN_HEX` is empty.  Therefore, it always can be defined as `srec_cat`; and, it can be used in the `$(shell)` test.

cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
```diff
@@ -6,7 +6,15 @@ OBJCOPY = arm-none-eabi-objcopy
 OBJDUMP = arm-none-eabi-objdump
 NM      = arm-none-eabi-nm
 SIZE    = arm-none-eabi-size
+SREC_TEST = $(shell srec_cat --version > null && echo 1)
+ifeq ([${SREC_TEST}],[1])
 SREC_CAT = srec_cat
+CAN_HEX  = %.hex
+else
+$(waring "install srecord to allow hex image generation")
```
alexrayne: ok

cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
alexrayne: ok

cpu/cc26xx-cc13xx/Makefile.cc26xx-cc13xx
```diff
@@ -8,6 +8,14 @@ NM      = arm-none-eabi-nm
 SIZE    = arm-none-eabi-size
 SREC_CAT = srec_cat
 
+SREC_TEST = $(shell $(SREC_CAT) --version > null && echo 1)
```
greg-king5: On POSIX systems (including Cygwin), that I/O device is named `/dev/null` ; on pure Windows development platforms (e.g., MSYS2), that device is named `NUL:` .

alexrayne: thank you. but is there any method to distinguish wich dev to use?

PR #2273, opened 2017-07-31
enhance cc26 prop-mode rf-core to support tsch

alexrayne:
Here is an set of patches, for cc26xx prop-mode rf-core driver, that provides minimal functionality enough for TSCH.
1) for aside work - was deployed RAT timer syncronise-monitor code from ieee-mode driver into stanalone object rf-rat-monitor.
2) optimised set_value operation for speed - now it not on/off rf-core if one powerdowned. this gives speed effort this method. this affects radio-chanel and tx-power setup parameters.
3) appends agressive RAT sync checking mode - this will checks that retrieved rx-operation timestamp is valid, and if not - checks sincronisation of RAT vs RTtimer.
4) a bit optimised code for size

### DIFF
--- a/cpu/cc26xx-cc13xx/rf-core/rf-rat-monitor.c
+++ b/cpu/cc26xx-cc13xx/rf-core/rf-rat-monitor.c
@@ -1,0 +1,208 @@
+/*
+ * rfcore-rat-moninor.c
+ *
+ *  Created on: 21 èþë. 2017 ã.
+ *      Author: alexrayne
+ *
+ * here is RAT timer overflow monitoring moved from ieee-mode
+ * to populate it for prop-mode.
+ */
+/*---------------------------------------------------------------------------*/
+/**
+ * \addtogroup rf-core
+ * @{
+ *
+ * \defgroup rf-core-ieee CC13xx/CC26xx IEEE mode driver
+ *
+ * @{
+ *
+ * \file
+ * Implementation of the CC13xx/CC26xx IEEE mode NETSTACK_RADIO driver
+ */
+/*---------------------------------------------------------------------------*/
+#include "contiki.h"
+#include "dev/radio.h"
+#include "sys/clock.h"
+#include "sys/rtimer.h"
+#include "ti-lib.h"
+#include "rf-core/rf-core.h"
+#include "rf-core/rf-switch.h"
+#include "rf-core/rf-rat.h"
+/*---------------------------------------------------------------------------*/
+#include <stdint.h>
+#include <string.h>
+#include <stdio.h>
+#include <stdbool.h>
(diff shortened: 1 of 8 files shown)


cpu/cc26xx-cc13xx/rf-core/rf-rat-monitor.c
```diff
@@ -136,9 +138,9 @@ static void
 handle_rat_overflow(void *unused)
 {
   if(rf_core_is_accessible()) {
-  uint_fast8_t success = 0;
       //* if RF is powerdown, nothing to check
-  success = rf_rat_check_overflow(false);
+      WHEN_ANNOTATE(uint_fast8_t success = )
```
tonnenpinguin: Does the warning stop when you just add this: `(void) success;` ?
IMHO this macro is calling for trouble and is not exactly straight forward to read :)

alexrayne: thank you, accept it.

greg-king5: You should remove the (now unused) `WHEN_ANNOTATE()` macro definitions.

cpu/cc26xx-cc13xx/rf-core/rf-rat-monitor.c
```diff
@@ -0,0 +1,213 @@
...
+  return 1;
+}
+
+/*---------------------------------------------------------------------------*/
+static const clock_time_t RAT_CHECK_PERIOD = RAT_OVERFLOW_PERIOD_SECONDS * CLOCK_SECOND / 6;
+static void
+handle_rat_overflow(void *unused)
+{
+  if(rf_core_is_accessible()) {
+      //* if RF is powerdown, nothing to check
```
greg-king5: That comment should be above the `if() {` line.

alexrayne:
added hotfix for  ightweight TX power change.
also add conf RF_CORE_CONF_PENDING - denotes behaviour, how pending_packets are calculated. old behaviour of pending_packet was not the same as for read packet. that is why possibility to have some buffered packets, but read cant get anything. this situation leads strange errors.

alexrayne:
More work on reqcv que handling - sometimes my cc1310 looks leave curents slot in PENDING state, and goes to next. read_frame() stalls on this slot. so RF_CORE_PENDING_READSANY mode provided for this - read_frame scans all que, skips PENDING slots until FINISHED one.

alexrayne:
fixed bad sync RAT - RTC when current time is large - more then 18hours.
 there was incorrect RAT overloads calculation, since used contiki RTIMER time insteard RTC 64 bit value. RF core uses 64bit value for sync.

PR #2285, opened 2017-08-10
TSCH: rtimer us_to_rtimerticks/rtimerticks_to_us - routines prefer macros

alexrayne:
Here is rtimer.h provided convertion routines us_to_rtimerticks/rtimerticks_to_us - static code wraps version of RTIMERTICKS_TO_US/US_TO_RTIMERTICKS macro, that intends to common code optimise

*Note: rtime-us convertion is not trivial, often use mul/div. so better to provide routines
        for variable conversion. vs macro - uses onluh for constansts evaluation


alexrayne:
migrates some more tsch code to rtimerticks_to_us/us_to_rtimerticks

PR #2288, opened 2017-08-10
supress warns on DEBUG define to compiler

alexrayne:
now undef DEBUG everywhere it redefines, to supress compiler warning about redefine

* usual way to note compiler about release/debug target is provide DEBUG definition.
  contiki code usualy localy override it, so build with DEBUG generates a lot of warnings

### DIFF
--- a/core/net/mac/tsch/tsch-log.c
+++ b/core/net/mac/tsch/tsch-log.c
@@ -52,8 +52,10 @@
 #include "lib/ringbufindex.h"
 
 #if TSCH_LOG_LEVEL >= 1
+#undef DEBUG
 #define DEBUG DEBUG_PRINT
 #else /* TSCH_LOG_LEVEL */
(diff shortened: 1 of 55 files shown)


core/net/mac/tsch/tsch-log.c
```diff
@@ -52,8 +52,10 @@
 #include "lib/ringbufindex.h"
 
 #if TSCH_LOG_LEVEL >= 1
+#undef DEBUG
 #define DEBUG DEBUG_PRINT
 #else /* TSCH_LOG_LEVEL */
+#undef DEBUG
```
greg-king5: `DEBUG` should be undefined in only one place (above the `#if` line).

alexrayne: thought about it. but prefer place undefine right before define.

PR #2293, opened 2017-08-11
TSCH: log tsch-ques messages instead print

alexrayne:
tsch-ques used PRINTF, and thererfore introduced large delays in timeslot processing. migrating to log makes timeslot more robust.

also provided more kinds of log messages:
+ tsch_log_text  -add printer of constant text string
+ tsch_log_fmt - formated string with 3 args
+ tsch_log_packet[_verbose] - prints formated messages with packet info
+ tsch_log_change_timesrc - prints chnging time source message

this messages intends for more light log

### DIFF
--- a/core/net/mac/tsch/tsch-log.c
+++ b/core/net/mac/tsch/tsch-log.c
@@ -70,6 +70,8 @@ static struct ringbufindex log_ringbuf;
 static struct tsch_log_t log_array[TSCH_LOG_QUEUE_LEN];
 static int log_dropped = 0;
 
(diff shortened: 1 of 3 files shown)


alexrayne:
change TSCH log prints nodes adress as hex. imho ti si handy

alexrayne:
change TSCH log prints nodes adress as hex, and fix a bit compiler warnings with TSCH_LOGF

alexrayne:
more TSCH log prints nodes adress as hex

alexrayne:
fixed TACH_LOGF( ) - macro args - that one should be in commas, to avoid errors wth complex expressions

PR #2315, opened 2017-09-01
+cc26xx uart tx buffer - UART_TXBUFSIZE enables buffer for uart output

alexrayne:
Hallow! here provided:

buffering uart output, that can be enabled by UART_TXBUFSIZE configuration macro.
enhanced cc26xx_uart api :
- cc26xx_uart_write_bytes - nonbocking write to uart fifo optimised to buffering driver
-cc26xx_uart_space_avail - status of tx fifo
putchar:puts, dbg_send_bytes - migrates to cc26xx_uart_write_bytes for optimisation.
a bit optimised UART ISR - more light check for power and clock.
*Note - using buffering (via UART_TXBUFSIZE) skips waiting for send completion in
routines puts, dbg_send_bytes - that was used for to protect vs poweroff
during transfer.

### DIFF
--- a/core/lib/ringbuf16index.c
+++ b/core/lib/ringbuf16index.c
@@ -1,0 +1,168 @@
+/*
+ * Copyright (c) 2015, SICS Swedish ICT.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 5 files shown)


alexrayne:
fixes` :
1) TX IRQ enabling - when TX_BUFFER defined, UART IRQ enables now even if RX receive handle disabled. UART IRQ always need in buffered mode
2) LPM request has loose chars that are in send registers, so should wait for UART transmiter empty before DEEP_SLEEP

alexrayne:
fixed work USRT in LPM SLEEP mode - now USRT leaves work in sleep state. Now UART can flush fifo values during cpu sleep.

alexrayne:
fixes LPM mode evaluation when TX buffer empty - only way to detect that UART transmiter is empty, is poll UART flags. so, keem AWAKE LPM mode while polling.

PR #2326, opened 2017-09-14
lib/list.h more inlines - optimise code a bit

alexrayne:
here is a trivial code optimisation via inlines of some lib/list.h routines

### DIFF
--- a/core/lib/list.c
+++ b/core/lib/list.c
@@ -53,6 +53,7 @@ struct list {
   struct list *next;
 };
 
+#if !LIB_INLINES
 /*---------------------------------------------------------------------------*/
 /**
  * Initialize a list.
@@ -102,6 +103,7 @@ list_copy(list_t dest, list_t src)
 {
   *dest = *src;
 }
+#endif // #if !LIB_INLINES
 /*---------------------------------------------------------------------------*/
 /**
  * Get the tail of a list.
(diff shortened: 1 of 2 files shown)


msloth:
What is the gain of the optimization (ie is it useful), do you have any measurement? The cost is more complexity, duplicated code, and another configuration option in a mature lib, so imo the gain must be >> the cost.

alexrayne:
1) there is no "another configuration options" used. it uses alredy commited behaviour from packetbuf.h
2) the same questions you can ask AdamDunkels about packetbuf.h:PACKETBUF_CONF_ATTRS_INLINE usage and reson.
3) about profit. use arm with O1 optimisation gives:
list_init - inline uses 1 extra instruction vs routine call
list_head - routine-call use 1-2 extra instructions, vs inline. 
list_copy not used in any code, so i catn inspect it

in both cases i not counts cpu cycles use - surely inline wins about 2 or more cycles

So, looks better to get off list_init optimisation.
and i will remove _GNUC_ define from inline usage condition. it will uses only with PACKETBUF_CONF_ATTRS_INLINE

nvt:
Thanks for the contribution, but I agree with @msloth here. There are a few issues with the PR: 1) it duplicates code and makes the implementation harder to read, 2) it introduces a dependency on a packetbuf configuration variable in list.h, and 3) Contiki coding style and naming conventions are not followed entirely. Hence, I am closing this PR unfortunately.

PR #2334, opened 2017-09-22
*TSCH:tsch_XXX_process_pending  smarter events pending

alexrayne:
make pending TSCH events thread more granular:
- yields on every pending step
- and log processing with less prioritete - prefer rx/tx pendings

* this patch should make tsch events fater response, and more flexibe for pt scheduling

### DIFF
--- a/core/net/mac/tsch/tsch-log.c
+++ b/core/net/mac/tsch/tsch-log.c
@@ -72,8 +72,7 @@ static int log_dropped = 0;
 
 /*---------------------------------------------------------------------------*/
 /* Process pending log messages */
-void
-tsch_log_process_pending(void)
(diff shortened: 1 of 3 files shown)


alexrayne:
provided trivial fix for robustness over proto-thread

alexrayne:
now provided choose of tsch_polling _event thread code by config TSCH_CONF_POLLING_STYLE 
provided style TSCH_POLLING_STRONG and TSCH_POLLING_RELAXED. relaxed style use PROCESS_PAUSE for yielding. 
also in relaxed style fixed - now TX events process surely before RECV events. This is must provide order that event for packet sent invokes before received answer packet.

PR #2338, opened 2017-09-26
cc26xx/sensortag: Use correct calibration data structure size

tim-ist:
The CALIB_DATA_SIZE was set to 24 while the actual structure size is 28.
Because of this, some other variables in the module were overwritten after the sensor read operation. This, for example, prevented reading the temperature immediately after the pressure (or vice versa), because the "enabled" static variable contained garbage.


msloth:
👍 looks good, and is correct

PR #2346, opened 2017-10-09
Set Orchestra's default debug output level to none

fleutot:
Debug printf is meant for development. An application using Orchestra
does not want output on by default.


fleutot:
Regression test for tsch-orchestra fails, since it expects the output to check if Orchestra started.

PR #2349, opened 2017-10-11
cc26xx:rtc isr ch2 handler support

alexrayne:
Here provided patch adding handler soc_rtc_on_isr_ch2 - for RTC irq on CH2 compare


alexrayne:
accasionaly here is PR#2351

PR #2351, opened 2017-10-11
+cc26xx:prop-mode:rf on timeout - prevents code hung when OSC not start

alexrayne:
turn on radio checks HF OSC for timeout, ant fails if OSC not ready for RF_CORE_CONF_HFOSC_STARTUP_TOUS
+cc26xx:oscillators:oscillators_wait_ready_hf_xosc - waiting HF OSC ready until timeout

*Note - if OSC is bad, and can`t start, therefore mcu hung forever. when RF_CORE_CONF_HFOSC_STARTUP_TOUS defined, it checks with this timeout and reports failure

### DIFF
--- a/cpu/cc26xx-cc13xx/dev/oscillators.c
+++ b/cpu/cc26xx-cc13xx/dev/oscillators.c
@@ -111,6 +111,27 @@ oscillators_request_hf_xosc(void)
   /* Release the OSC AUX consumer */
   aux_ctrl_unregister_consumer(&osc);
 }
+
+/*---------------------------------------------------------------------------*/
+bool oscillators_wait_ready_hf_xosc(rtimer_clock_t timeout)
+{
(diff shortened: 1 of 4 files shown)


amitbhanja:
Thanks for the commit. What should be the value of RF_CORE_HFOSC_STARTUP_TOUS ?

alexrayne:
> Thanks for the commit. What should be the value of RF_CORE_HFOSC_STARTUP_TOUS ?

RF_CORE_HFOSC_STARTUP_TOUS  - is a timeout in [us]. for cc1350 OSC should wakeup less then  300-500us. so, i guess, 700us enough

richarddonkin:
> Thanks for the commit. What should be the value of RF_CORE_HFOSC_STARTUP_TOUS ?

I use 1000 (as in the commented #define). It seems to work well.

PR #2372, opened 2017-11-02
TSCH boolean flasg optimisations

alexrayne:
here provided an trivial optimisations:
- for some null pointer chesks
- for boolean values allocation. uses bool type from "stdbool.h", unstead int. this could give a bit less data mamory usage . 
- also provide optimisations via inline checks of bool flags


alexrayne:
append boolean migration for tsch_que_is_empty

PR #2373, opened 2017-11-03
!TSCH:seq:ffff - avoid this value in counter

alexrayne:
framer 802154 treats value of seq=ffff as invalid, so avoid one in seq counter of TSCH stack


alexrayne:
look like TSCH use 1byte seq, so check value ffff is not actual

PR #2388, opened 2017-11-10
+TSCH:tsch_packet_parse_my_eb - deploed EB parse and validity test.

alexrayne:
this patch exposes  for app usage:
1) tsch_temp_packet  - temporary input packet used by tsch_scan.
2)  tsch_packet_parse_my_eb  - tsch_scan EB parse and valifation procedure.
3) move struct input_packet to tsch-packet.h  - looks there is more native place
this resources allow provides application own air scan schemes


alexrayne:
fix strange gcc6-arm optimisation bug - static function optimises stack-frame, and it make  loose local variables

PR #2389, opened 2017-11-10
+process:process_abort - silent exit process

alexrayne:
process_abort provides extit process silently, without any events broadcast. 
ordinal process END/EXIT generates broadcasts events.

### DIFF
--- a/core/sys/process.c
+++ b/core/sys/process.c
@@ -203,6 +203,11 @@ process_exit(struct process *p)
 {
   exit_process(p, PROCESS_CURRENT());
 }
+
+void process_abort(void){
+    PROCESS_CURRENT()->state = PROCESS_STATE_NONE;
+}
+
 /*---------------------------------------------------------------------------*/
 void
 process_init(void)
--- a/core/sys/process.h
+++ b/core/sys/process.h
@@ -389,6 +389,17 @@ CCIF void process_post_synch(struct process *p,
  */
 CCIF void process_exit(struct process *p);
 
+/**
+ * \brief Makes status of current process not running, for silent process exiting
+ *
+ *       exiting of process generates events PROCESS_EVENT_EXITED for all
+ *       active processes, and PROCESS_EVENT_EXIT for current process
+ *       such notification can be useless and wasteful. so process_abort can
+ *       mark process alredy stops, to block notification in process_exit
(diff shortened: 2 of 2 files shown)


msloth:
What's the use case with this PR? What benefit does it add?

You have about 45 open PRs (of the 150 open PRs). I hope you don't mind, I had a look at a couple, and have some general feedback that I hope you are open to.

You rarely state the benefit of the change, ie why it should be accepted. That users are given more choice (as this PR) is not really enough - things can be done in many ways, and adding more ways (and more code) is not necessarily a good thing. You need to provide a _reason_ why something should be changed. Some PR messages claim performance gains, or code size reductions. Then, please add an example or two.

You generally need to work on code style. Naming of variables is mixed, extra whitespace added, and so on. I don't really follow your commit message style either, you seem to use chars like `+` and `*` to mean some various things. Spelling errors and similar sloppiness don't add confidence.

You often push more commits to old PRs long (months) after opening them, without explaining what changed or why, making it harder to review. Suppose someone is in the middle of testing or reviewing your PR and you then push more commits? Please do at least write a comment when you do this. Otherwise, how can someone know when the PR is ready to be reviewed? Ideally, it should be ready _before_ the PR is opened.

I for one appreciate efforts in contributing, but I hope you understand the efforts required by mergers (of which I'm not) is most often much, much more. Finally, I hope you accept this as constructive feedback and an opportunity to improve.

alexrayne:
msloth:
1) well, i provide explanation and reason as much as i can for this PR. is that enoungh? For other PR i take revision for some time. What example you mention for? is example in comment well?
i don`t made examples cause i not understand examples folder structure, and how work tests system.

2) i make many smapp PR, insteard one big sequence. to allow maintainer choose any one. is it bad?

3) for mixed variables names - looks i need to point it. normaly i try to use same naming style as other code. 

4) what spaces are problem? can you point it?

5)about + and * - do you mean that i should not use it in commit message? why?

6)about my PR evolution - since PR is not accepted\rejected, i have no any sence that some one work with it (Travis always do 8) ). i places hotfix, or minor changes that not affects PR sentence. As saw other PRs, they includes fixes and changes during some conversation. So, i thought, this is normal practice. isn`t it?

PR #2393, opened 2017-11-13
Stm32nucleo st7580

stebosisio:
The port supports the following boards from ST:

    - NUCLEO-L152RE board, based on the STM32L152RET6 ultra-low power microcontroller
    - X-NUCLEO-PLM01A1 based on the ST7580 FSK, PSK multi-mode power line networking
    system-on-chip expansion board

The port follows the same idea/structure of our already integrated stm32nucleo-spirit1 platform and we have kept in mind issues already discussed in the past for it (e.g. use a git submodule for CMSIS and for our existing headers/sources files that are kept unchanged, put long term maintainers' names in the readme file...) 

ST CLAB team

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,3 +17,6 @@
 [submodule "tools/sensniff"]
 	path = tools/sensniff
 	url = https://github.com/g-oikonomou/sensniff.git
+[submodule "platform/stm32nucleo-st7580/stm32cube-lib"]
+	path = platform/stm32nucleo-st7580/stm32cube-lib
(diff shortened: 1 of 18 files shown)


stebosisio:
I see that only doxygen regression test is failing, due to warnings, but these are also on contiki core's files. In fact I see that this test is failing on all the newest pull requests.

stebosisio:
Hello,

we didn't receive any feedback on this PR... Can we have, at least, an initial feedback and/or some instructions on the next steps needed to propose a new platform for evaluation?

PR #2399, opened 2017-12-01
fix TSCH - stop slots  operations on disassociate and some rtimer issue

alexrayne:
provide explisit slot operations stop on disassciaton. this helpful when need to sutdown after TSCH off.
also fix slot rtimer override - cancels slots rtimer before start new operation,
                cause some implementation of rtimer can`t override current running
                rtimer, and therefore slot not starts on expired rtimer. this situation
                leads to block whole tsch activity. i`ve got this issue wih PR#1290


alexrayne:
closed for rearrange patch contents

PR #2400, opened 2017-12-02
fix TSCH - stop slots operations on disassociate with some rtimer issue

alexrayne:
here fixed some issues of rtimer use vs disassociation TSCH.
1) provide explisit slot operations stop on disassciaton. this helpful when need to sutdown after TSCH off.
2) also fix slot rtimer override - cancels slots rtimer before start new operation,
rtimer doesnot override current running rtimer. But some implementation even block running rtimer set again. Therefore slot not starts on rinning rtimer. this situation leads to block whole tsch activity. ive got this issue wih PR#1290


alexrayne:
fixed stop during slot operation. there was leaved tsch lock in captured state, now stop forced TSCH to unlock

PR #2408, opened 2017-12-09
Small code style change

cpsw:
Making the operator precedence explicit.


alexrayne:
why you not make this change everywhere?
you got same code below:
`size = (prec || flags & ALT)`

cpsw:
Hi @alexrayne, I have a tool that can capture all occurrences of this pattern as well as other similar ones. Should I do change the other occurrences also?

alexrayne:
imho, better to fix other occurencec too.

PR #2427, opened 2018-01-16
Magnetometer Sensor Added in cc26xx-web-demo 

nicholasinatel:
Here provided modifications in the cc26xx-web-demo example, 
magnetometer from the MPU_9250 for the sensortag (TI) board in the srf06-cc26xx platform is active.

Observation: Could only compile the example with some of its features deactivated, when fully activated got error of memory SRAM overflow.

at **examples/cc26xx/cc26xx-web-demo/project-conf.h**

`/* Enable/Disable Components of this Demo */

#define CC26XX_WEB_DEMO_CONF_MQTT_CLIENT      0

#define CC26XX_WEB_DEMO_CONF_6LBR_CLIENT      0

#define CC26XX_WEB_DEMO_CONF_COAP_SERVER      1

#define CC26XX_WEB_DEMO_CONF_NET_UART         0`


platform/srf06-cc26xx/sensortag/mpu-9250-sensor.c
```diff
@@ -377,6 +432,121 @@ enable_sensor(uint16_t axes)
...
+  val = 0;
+  sensor_common_read_reg(INT_PIN_CFG, &val, 1);
+  SENSOR_DESELECT();
+  if(val == 0x22) {
+    success = true;
+  }
+  else {
+    success = false;
+  }
+  return success;
```
greg-king5: The success/failure of the function should be returned directly:
```c
  return val == 0x22;
```

platform/srf06-cc26xx/sensortag/mpu-9250-sensor.c
```diff
@@ -377,6 +432,121 @@ enable_sensor(uint16_t axes)
...
+ * \return magnetometer status, 0 if no error
+ */
+uint8_t 
+mag_read(int16_t *data)
+{
+  uint8_t val;
+  uint8_t rawData[7];
+  magStatus = MAG_NO_POWER;
+  if(power_test() != 0)
+    magStatus = MAG_STATUS_OK;
```
greg-king5: I think that `magStatus` should be initiated to `MAG_STATUS_OK`.  And, if `power_test()` is zero, then it should be changed to `MAG_NO_POWER`.  That sequence is a better match to how `magStatus` is used in the rest of this function.

platform/srf06-cc26xx/sensortag/mpu-9250-sensor.c
```diff
@@ -511,16 +752,27 @@ power_up(void)
...
 /*---------------------------------------------------------------------------*/
+/**
+* \brief Test if MPU power is ON or OFF
+* \return state of MPU power, TRUE if it is ON
+*/
+bool power_test(void)
+{
+  if(ti_lib_gpio_read_dio(BOARD_IOID_MPU_POWER) == 1)
+    return true;
+}
```
greg-king5: The compiler might give a warning on this function because there is no `return` statement in the "else" path.

It's more efficient to return the test result directly:
```c
  return ti_lib_gpio_read_dio(BOARD_IOID_MPU_POWER) == 1;
```

platform/srf06-cc26xx/sensortag/mpu-9250-sensor.c
nicholasinatel: Thank you, i have implemented that for the next commit

platform/srf06-cc26xx/sensortag/mpu-9250-sensor.c
nicholasinatel: Thinking about doing something more sophisticated here, your suggestion did not work,  but gave some ideas.

nicholasinatel:
Hi @g-oikonomou, greetings.

Please i would like to understand how does the **continuous-integration/travis-ci/pr** works and what it is.
Also if possible, could you explain why the commit was not approved?

I checked the **Details** link in the ( **continuous-integration/travis-ci/pr** — The Travis CI build could not complete due to an error ) and didn't understand what happened.

In my Fedora27, compiling and testing in the Sensortag/CC1350 with an RPL Bourder Router, the CoAP works fine.

nicholasinatel:
Hi @casaroli, greetings

I ask you please to take a review in my code and why it was refused by **continuous-integration/travis-ci/pr** when you are able to.

Best Regards

platform/srf06-cc26xx/sensortag/mpu-9250-sensor.c
nicholasinatel: i agree, thank you

ql1179:
So there has to be a 10ms delay each time we read magnetometer?

PR #2435, opened 2018-02-01
Add function to setup UART dynamically on cc13xx/cc26xx devices

keyfour:
There is useful to have ability to change UART parameters on working device with running program. I ran into this when connects different devices with serial interfaces on remote working device. I added function
to cc26xx UART driver which re-configure device with custom parameters.


alexrayne:
Hallow Keyfour!
IMHO, this new feature better be optional, since your requrements to change setup is rare. Most projects no need this, therefore your new code useless for them. 
You may introduce config definition like UART_SETUP_ENABLE to enable new function. and without that - uses constants only.

PR #2437, opened 2018-02-01
Lpm cc26xx cc13xx config

cquijano:
Add Config for lpm on cc26xx and cc13xx devices


alexrayne:
why you need it? What pupose of such conf you intends?

cquijano:
I do not always want to set the CPU ON LPM_MODE_SHUTDOWN , some times I want set the LPM on the cc26xx  to LPM_MODE_DEEP_SLEEP, for example, When on the application  is no external interrupt to wake up the CPU.

PR #2457, opened 2018-03-28
Provide exclude source list for development flexibility

Vinggui:
This new parameter is a list capable of ignoring specific source files during compilation. which will allow more flexibility for developers.
One example is that "nullradio.c" and "nullradio.h" (inside core/dev) was not wanted, since I'm building a new network architecture.
To ignore this files, I added the following definition into my makefile:
```
# For the .h
CFLAGS += -DRADIO_H_=1
#for the .c
IGNORE_SOURCE_FILE_LIST += nullradio.c
```


alexrayne:
You would rewrite PR name, for better readability? some about  - "provide exclude source list" ?

PR #2587, opened 2019-05-02
Spellchecked root level files and apps

taigrr:
Manually ran spellcheck across root level text files and all files in the /apps folder.


taigrr:
Almost completely comments/docstring updates. Only changed one misspelled variable name and one debug print line, in case that info makes accepting the PR faster.

taotailiang:
你好

PR #2660, opened 2020-09-25
 openmote-websense.c

rana509:


rana509 (commented):
This program is similar to sky-websense.c . To flash it in the openmote-cc2538 platform you just have to download it in the contiki/examples/ipv6/sky-websense folder with the proposed Makefile

